PR #32, opened 2023-03-14
Stripe integration

hult:
- [x] Backend for initiating a pledge
  - [x] Create Reward
  - [x] Create payment intent
  - [x] Also allow modifying reward amount (and the attached payment intent)
  - [ ] Shouldn't require org and repo membership
  - [x] Track email if pledging without being logged in
  - [ ] Track account id if pledging while being logged in
  - [ ] Allow for pledge id to be tied to account when creating it
  - [ ] Allow for credit card token to be saved on account when creating it
- [ ] Webhook for when payment succeeds
  - [x] Happy path
  - [ ] What if amount or something else doesn't match?
  - [ ] What about if it fails?


server/polar/integrations/stripe/tasks.py
```diff
@@ -0,0 +1,18 @@
...
+from polar.postgres import AsyncSessionLocal
+from polar.models import Reward
+from polar.reward.schemas import State
+
+
+@task("stripe.webhook.payment_intent.succeeded")
+async def payment_intent_succeeded(ctx: JobContext, event: stripe.Event) -> None:
+    async with AsyncSessionLocal() as session:
+        print(event)
+        reward = await Reward.find_by(
```
hult: Service method

clients/apps/web/src/components/Website/Checkout/index.tsx
```diff
@@ -0,0 +1,258 @@
...
+          </Elements>
+        )}
+      </div>
+    </>
+  )
+}
+
+/*
+ * To discuss with Magnus:
+ * - From decimal to int (cents)?  Consistent with Stripe
```
hult: I think it makes sense to do cents, yeah. I didn't touch the existing columns in `Reward`.

clients/apps/web/src/components/Website/Checkout/index.tsx
```diff
@@ -0,0 +1,258 @@
...
+ * - Rename rewards to checkout?
```
hult: I was thinking perhaps even payment?

hult: Renamed to pledges

PR #36, opened 2023-03-16
feat(github): refresh user tokens

zegl:
- [x] Refresh user tokens (and refresh the refresh tokens) if expired
- [x] Use GitHub API to track which organisations the user can access
- [x] Use GitHub AIP to track which repositories the user can access
- [x] Cache access (currently for 1 day)
- [x] Use `UserRepository` to grant access to repo (`UserOrganization` is not enough)
- [ ] On login, refresh access?
- [ ] How to deal with invites to the organization and/or repo? New users should get access to Polar. Same applies for users that have their access revoked.

Screenshot 2023-03-17 at 16 58 19

☝️ Looks like it's working! Two users can now have different sets of repositories in the same org, depending on their access levels on GitHub.

### DIFF
--- a/server/polar/auth/dependencies.py
+++ b/server/polar/auth/dependencies.py
@@ -87,7 +87,6 @@ class Auth:
             session,
             platform=platform,
             name=org_name,
-            # user_id=user.id,
         )
         if not organization:
             raise HTTPException(status_code=404, detail="Organization not found")
--- a/server/polar/integrations/github/service/organization.py
+++ b/server/polar/integrations/github/service/organization.py
@@ -75,7 +75,7 @@ class GithubOrganizationService(OrganizationService):
 
         return organization
 
-    async def ensure_installed(
+    async def save_from_github(
         self, session: AsyncSession, installation: github.rest.Installation
     ) -> Organization | None:
         to_create = OrganizationCreate.from_github_installation(installation)
--- a/server/polar/integrations/github/service/repository.py
+++ b/server/polar/integrations/github/service/repository.py
@@ -223,7 +223,7 @@ class GithubRepositoryService(RepositoryService):
 
         return instances
 
-    async def ensure_installed(
+    async def upsert_from_github(
         self,
         session: AsyncSession,
         organization: Organization,
--- a/server/polar/integrations/github/service/user.py
+++ b/server/polar/integrations/github/service/user.py
@@ -84,15 +84,20 @@ class GithubUserService(UserService):
         # Fetch which installations the user can access
         installations = await self.fetch_user_accessible_installations(session, user)
 
+        # Build list of GitHub-organizations that the user can access
+        # This will be used to remove access from Polar-organizations that the user
+        # no longer should have access to
         access_to_orgs_ids: List[UUID] = []
 
         # Write to cache
         for i in installations:
-            org = await github_organization.ensure_installed(session, i)
+
+            # Save the github organization as a Polar organization (if not exists)
+            org = await github_organization.save_from_github(session, i)
 
             if not org:
                 log.error(
-                    "failed to ensure installation",
+                    "failed to save organization/installation",
                     installation_id=i.id,
                     login=i.account.login,
                 )
@@ -100,27 +105,29 @@ class GithubUserService(UserService):
 
             access_to_orgs_ids.append(org.id)
 
+            # Create/Update UserOrganization entry
             await github_organization.add_user(session, org, user)
 
-            log.info(
-                "ensured installation", installation=i.account.login, user=user.email
-            )
-
-            # install for org
-            await github_repository.install_for_organization(session, org, i.id)
-
+            # For this user, fetch which GitHub-repositories in the GitHub-organization
+            # that the user can access.
             github_repos = await self.fetch_user_accessible_installation_repositories(
                 session, user, i.id
             )
 
+            # Build list of repositories in this org that the user can access.
+            # Will be used to remove access from repositories that the user no longer
+            # can access.
             access_to_repo_ids: List[UUID] = []
 
             for github_repo in github_repos:
-                repo = await github_repository.ensure_installed(
+                # Create a polar-Repository from the GitHub-repository if it doesn't exist
+                repo = await github_repository.upsert_from_github(
                     session, org, github_repo
                 )
                 if not repo:
                     continue
+
+                # Create/Update UserRepository entry
                 await github_repository.add_user(session, org, repo, user)
                 log.info(
                     "ensured repository",
@@ -130,17 +137,27 @@ class GithubUserService(UserService):
                 )
                 access_to_repo_ids.append(repo.id)
 
-            log.info(
-                "cleanup access", access_to_repo_ids=access_to_repo_ids, user=user.id
-            )
-
             # Remove access entries for repositories that the user doesn't have access to anymore
             await github_repository.cleanup_repositories_access(
                 session, org.id, access_to_repo_ids, user
             )
 
+            log.info(
+                "synced repository access",
+                org_id=org.id,
+                access_to_repo_ids=access_to_repo_ids,
+                user=user.id,
+            )
+
         # Remove access entries for organizations that the user doesn't have access to anymore
         await github_organization.cleanup_access(session, access_to_orgs_ids, user)
+
+        log.info(
+            "synced organization access",
+            org_id=org.id,
+            access_to_orgs_ids=access_to_orgs_ids,
+            user=user.id,
+        )
 
     async def fetch_user_accessible_installations(
         self, session: AsyncSession, user: User
@@ -284,19 +301,19 @@ class GithubUserService(UserService):
 
         user_repos = await lookup()
 
-        outOfDate = False
+        out_of_date = False
 
         for r in user_repos:
             if not r.validated_at:
-                outOfDate = True
+                out_of_date = True
                 break
             cutoff = datetime.now(r.validated_at.tzinfo) - timedelta(days=1)
             if r.validated_at < cutoff:
-                outOfDate = True
+                out_of_date = True
                 break
 
         # No stale data
-        if not outOfDate:
+        if not out_of_date:
             return user_repos
 
         # Unknown status, refresh from API


server/polar/auth/dependencies.py
```diff
@@ -78,16 +83,19 @@ async def user_with_org_access(
...
+            name=org_name,
+            # user_id=user.id,
         )
         if not organization:
-            raise HTTPException(
-                status_code=404, detail="Organization not found for user"
-            )
+            raise HTTPException(status_code=404, detail="Organization not found")
+
+        if not await github_user.user_can_access_org(session, user, organization):
```
birkjernstrom, line 94: Once we move syncing permissions to a task and allow "runtime code" (Auth) to treat it as updated, we should only use `Organization.get_for_user` here as before since it joins against the `user_organization` mapping

server/polar/auth/dependencies.py
```diff
@@ -100,18 +108,30 @@ async def user_with_org_and_repo_access(
...
+            session, organization_id=organization.id, name=repo_name
         )
-        if not org_and_repo:
-            raise HTTPException(
-                status_code=404,
-                detail="Organization/repository combination not found for user",
+        if not repository:
+            raise HTTPException(status_code=404, detail="Repository not found")
+
+        if not await github_user.user_can_access_repo(
```
birkjernstrom, line 124: Same here that `Organization.get_with_repo_for_user` should suffice with one database roundtrip to get an organization and repository that matches the ones belonging to the given user and that our tasks sync that continously.

server/polar/integrations/github/client.py
```diff
@@ -104,6 +109,61 @@ def ensure_expected_response(
...
+    )  # A new refres token (is only set if the app is using expiring refresh tokens)
+    refresh_token_expires_in: Union[int, None] = Field(default=...)
+    scope: str = Field(default=...)  # Always an empty string
+    token_type: str = Field(default=...)  # Always "bearer"
+
+
+async def get_user_client(
+    session: AsyncSession, user: User
+) -> GitHub[TokenAuthStrategy]:
+    oauth = user.get_platform_oauth_account(Platforms.github)
```
birkjernstrom, line 128: Very nice that we refresh the user token here upon needing/using it! Let's discuss how to schedule permission syncing via tasks, but potentially we do it here if we see the TTL be within a certain threshold.

server/polar/integrations/github/client.py
```diff
@@ -104,6 +109,61 @@ def ensure_expected_response(
...
+        raise Exception("no github oauth account found")
+
+    # if token expires within 30 minutes, refresh it
+    if (
+        oauth.expires_at
+        and oauth.refresh_token
+        and oauth.expires_at <= (time.time() + 60 * 30)
+    ):
+        refresh = await GitHub().arequest(
+            method="POST",
```
birkjernstrom, line 139: You can use `githubkit.oauth.refresh_token` here instead which performs this API request

server/polar/integrations/github/service/repository.py
```diff
@@ -225,8 +214,94 @@ async def install_for_organization(
             RepositoryCreate.from_github(organization, repo)
             for repo in response.parsed_data.repositories
         ]
+
         instances = await self.upsert_many(session, repos)
+
+        if trigger_sync_job:
```
birkjernstrom, line 220: Nice that you moved this here vs. subclass of `upsert_many`to be more deliberate and clear

server/polar/integrations/github/service/repository.py
```diff
@@ -225,8 +214,94 @@ async def install_for_organization(
...
+        repository: Repository,
+        user: User,
+    ) -> None:
+        nested = await session.begin_nested()
+        try:
+            relation = UserRepository(
+                user_id=user.id,
+                organization_id=organization.id,
+                repository_id=repository.id,
+                validated_at=datetime.now(),
```
birkjernstrom: Use `polar.kit.utils.utc_now` to get a datetime with timezone (UTC)

server/polar/integrations/github/service/repository.py
```diff
@@ -225,8 +214,94 @@ async def install_for_organization(
...
+        # Update validated_at
+        stmt = (
+            sql.Update(UserRepository)
+            .where(
+                UserRepository.user_id == user.id,
+                UserRepository.organization_id == organization.id,
+                UserRepository.repository_id == repository.id,
+            )
+            .values(
+                validated_at=datetime.now(),
```
birkjernstrom: Same here with `polar.kit.utils.utc_now`. Worth quickly looking into for us for better DX: I think Postgres without a specified timezone would assume UTC and if so we can use `datetime.utcnow()` to always provide the UTC timestamp even without the timezone specified.

server/polar/integrations/github/service/repository.py
```diff
@@ -225,8 +214,94 @@ async def install_for_organization(
...
+        user: User,
+    ) -> None:
+        """
+        Remove entries for repositories that the user doesn't have access to anymore
+        """
+        stmt = sql.Delete(UserRepository).where(
+            UserRepository.user_id == user.id,
+            UserRepository.organization_id == organization_id,
+            UserRepository.repository_id.not_in(repository_ids),
+        )
```
birkjernstrom, line 303: Would be nice with some logging that permissions were dropped for the given user/organization and list of repository ids.

server/polar/organization/service.py
```diff
@@ -131,10 +131,12 @@ async def add_user(
...
         try:
             relation = UserOrganization(
-                user_id=user.id, organization_id=organization.id
+                user_id=user.id,
+                organization_id=organization.id,
+                validated_at=datetime.now(),
             )
             session.add(relation)
-            await session.commit()
+            await nested.commit()
```
birkjernstrom, line 194: You should return after this to avoid execution of the SQL update below which is unnecessary.

Also the `polar.kit.utils.utc_now` :-) Yeah, we should really look into if we can shorten that to datetime.utcnow() instead.

server/polar/organization/service.py
```diff
@@ -171,5 +203,15 @@ async def update_settings(
         )
         return updated
 
+    async def get_user_organizations(
```
birkjernstrom: This is not used anywhere? Let's drop it

server/polar/integrations/github/client.py
```diff
@@ -104,6 +109,61 @@ def ensure_expected_response(
 ###############################################################################
 
 
+class RefreshAccessToken(rest.GitHubRestModel):
```
birkjernstrom, line 112: We can get rid of this when using `githubkit.oauth.refresh_token` which returns this model

server/polar/integrations/github/service/organization.py
```diff
@@ -67,13 +66,24 @@ async def install(
...
+            session, organization, installation_id, trigger_sync_job=True
         )
+
         # TODO: Use SQLAlchemy features here vs. hard setter
-        organization.repositories = repositories
+        # organization.repositories = repositories
+
         return organization
 
+    async def ensure_installed(
```
birkjernstrom: Stylistic feedback: I typically want/expect `ensure_*` methods to raise upon error to ensure things either crash if not handled appropriately, or that invocations explicitly capture and manage it accordingly. Should we call this simply `install` since it fails gracefully by returning None?

server/polar/integrations/github/service/repository.py
```diff
@@ -225,8 +214,94 @@ async def install_for_organization(
...
+                validated_at=datetime.now(),
+            )
+            session.add(relation)
+            await nested.commit()
+            log.info(
+                "repository.add_user",
+                user_id=user.id,
+                organization_id=organization.id,
+                repository_id=repository.id,
+            )
```
birkjernstrom, line 257: Need to `return` here to avoid the duplicate & unnecessary SQL update below for existing records.

server/polar/integrations/github/service/user.py
```diff
@@ -63,5 +78,265 @@ async def update_profile(
...
+
+            # Remove access entries for repositories that the user doesn't have access to anymore
+            await github_repository.cleanup_repositories_access(
+                session, org.id, access_to_repo_ids, user
+            )
+
+        # Remove access entries for organizations that the user doesn't have access to anymore
+        await github_organization.cleanup_access(session, access_to_orgs_ids, user)
+
+    async def fetch_user_accessable_installations(
```
birkjernstrom: Typo of `accessable` vs. `accessible`

server/polar/integrations/github/service/user.py
```diff
@@ -63,5 +78,265 @@ async def update_profile(
...
+            map_func=lambda r: r.parsed_data.repositories,
+            installation_id=installation_id,
+        ):
+            res.append(repo)
+
+        return res
+
+    async def user_can_access_org(
+        self, session: AsyncSession, user: User, org: Organization
+    ) -> bool:
```
birkjernstrom, line 207: We can remove this in favour of the previous model in `Auth` of getting the organizations joined by `user_organization` and that our tasks calling `populate_user_accessable_orgs_and_repositories` keep them up-to-date along with future webhooks from Github

server/polar/integrations/github/service/user.py
```diff
@@ -63,5 +78,265 @@ async def update_profile(
...
+
+        # Unknown status, refresh from API
+        await self.populate_user_accessable_orgs_and_repositories(session, user)
+
+        # TODO: somehow cache false access requests as well
+
+        # Fetch again
+        return await lookup()
+
+    async def user_can_access_repo(
```
birkjernstrom, line 245: We can remove this in favour of the previous model in `Auth` of getting the organizations joined by `user_organization` and that our tasks calling `populate_user_accessable_orgs_and_repositories` keep them up-to-date along with future webhooks from Github

server/polar/integrations/github/service/user.py
```diff
@@ -63,5 +78,265 @@ async def update_profile(
...
+
+        # Unknown status, refresh from API
+        await self.populate_user_accessable_orgs_and_repositories(session, user)
+
+        # TODO: somehow cache false access requests as well
+
+        # Fetch again
+        return await lookup()
+
+    async def user_accessable_repos(
```
birkjernstrom: With tasks ensuring the data is up-to-date and we can simplify this. Essentially, the nested function (lookup) becomes the method itself. Finally, we can skip the `user_` prefix since it's within the user service, i. e `get_accessible_repositories`

server/polar/integrations/github/service/user.py
```diff
@@ -63,5 +78,265 @@ async def update_profile(
...
+        # No stale data
+        if not outOfDate:
+            return user_repos
+
+        # Unknown status, refresh from API
+        await self.populate_user_accessable_orgs_and_repositories(session, user)
+
+        return await lookup()
+
+    async def user_accessable_orgs(
```
birkjernstrom: With tasks ensuring the data is up-to-date and we can simplify this. Essentially, the nested function (lookup) becomes the method itself.

server/polar/user/endpoints/user_organization.py
```diff
@@ -17,17 +19,21 @@ async def get_user_organizations(
...
+
+    # fetch orgs
+    orgs = [await organization.get(session, o.organization_id) for o in user_orgs]
+
+    async def expand_children(org: Organization) -> OrganizationRead:
+        # get the repositories in this org that the user can access
+        user_repos = await github_user.user_accessable_repos(session, user, org)
+
+        # fetch as repo
+        repos = [await repository.get(session, r.repository_id) for r in user_repos]
```
birkjernstrom, line 33: Lot of nested database calls here in loops. However, we should be able to get rid of them all and revert back to `get_all_org_repos_by_user_id` and using the query with joins against user_organizations and user_repositories instead if we move permission syncing to a separate task?

server/polar/integrations/github/service/user.py
```diff
@@ -63,5 +78,265 @@ async def update_profile(
...
+            access_to_orgs_ids.append(org.id)
+
+            await github_organization.add_user(session, org, user)
+
+            log.info(
+                "ensured installation", installation=i.account.login, user=user.email
+            )
+
+            # install for org
+            await github_repository.install_for_organization(session, org, i.id)
```
birkjernstrom: We're installing repositories twice with this invocation and the code below? Shouldn't the code below related to fetching repositories from Github and then installing them simply replace `github_repository.install_for_organization`? Nice to have them broken up into two methods for clearer separation, but I think we're doing it twice now?

server/polar/integrations/github/service/user.py
```diff
@@ -63,5 +78,265 @@ async def update_profile(
...
+
+        access_to_orgs_ids: List[UUID] = []
+
+        # Write to cache
+        for i in installations:
+            org = await github_organization.ensure_installed(session, i)
+
+            if not org:
+                log.error(
+                    "failed to ensure installation",
```
birkjernstrom: Small feedback, but I'd drop "ensured" in the logs and go with "failed to install" or "installed" etc.

server/polar/integrations/github/service/user.py
```diff
@@ -63,5 +78,265 @@ async def update_profile(
...
+        async def lookup() -> Sequence[UserOrganization]:
+            stmt = sql.select(UserOrganization).where(
+                UserOrganization.user_id == user.id,
+            )
+            res = await session.execute(stmt)
+            return res.scalars().unique().all()
+
+        user_repos = await lookup()
+
+        outOfDate = False
```
birkjernstrom, line 338: Since we can drop this method later it doesn't matter, but more for the future as a heads-up: In Python the norm is snake case (out_of_date) except for classes

server/polar/integrations/github/service/user.py
zegl: Nicely spotted, there was some overlap here! I cleaned it up. 👍

PR #44, opened 2023-03-21
Set up Stripe account

hult:
- [x] Create account `POST /{platform}/{org_name}/accounts`
- [x] Create onboarding link `POST /{platform}/{org_name}/accounts/{stripe_id}/links`
- [x] Redirect back to api goes to http://127.0.0.1:8000/api/v1#/integrations/stripe/return?org_name=polarsource, where did that # come from?
- [ ] Only user who created Stripe account should be able to finish

### DIFF
--- a/server/migrations/versions/2023-03-22_merge_8db2ec8b80c7_and_9630755cf256.py
+++ b/server/migrations/versions/2023-03-22_merge_8db2ec8b80c7_and_9630755cf256.py
@@ -1,0 +1,27 @@
+"""Merge 8db2ec8b80c7 and 9630755cf256
+
+Revision ID: 125627402691
+Revises: 8db2ec8b80c7, 9630755cf256
+Create Date: 2023-03-22 14:01:13.650522
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# Polar Custom Imports
+from polar.kit.extensions.sqlalchemy import PostgresUUID
+
+# revision identifiers, used by Alembic.
+revision = '125627402691'
+down_revision = ('8db2ec8b80c7', '9630755cf256')
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    pass
+
+
+def downgrade() -> None:
+    pass


server/migrations/versions/2023-03-21_accounts_columns_nullable.py
```diff
@@ -0,0 +1,50 @@
...
+        "accounts", sa.Column("account_type", sa.String(length=10), nullable=False)
+    )
+    op.alter_column(
+        "accounts",
+        "type",
+        new_column_name="business_type",
+        nullable=True,
+    )
+    op.alter_column(
+        "accounts", "email", existing_type=sa.VARCHAR(length=254), nullable=True
```
birkjernstrom, line 33: Interesting, I haven't seen SQLAlchemy/Alembic prefer `VARCHAR` before over `String`. Strange that autogenerate yields different ones, but might have to do with the intended length of the strings.

server/polar/account/endpoints.py
```diff
@@ -0,0 +1,48 @@
...
+    org_name: str,
+    stripe_id: str,
+    auth: Auth = Depends(Auth.user_with_org_access),
+    session: AsyncSession = Depends(get_db_session),
+) -> AccountLink:
+    created = await account_service.create_link(
+        session,
+        auth.organization.id,
+        stripe_id,
+        f"?platform={platform.value}&org_name={auth.organization.name}",
```
birkjernstrom, line 44: Do we need to pass the query parameters? Only Stripe would consume them and have no usage for them, right?

server/polar/account/service.py
```diff
@@ -0,0 +1,75 @@
...
+class AccountService(ResourceService[Account, AccountCreate, AccountUpdate]):
+    async def create_stripe_account(
+        self,
+        session: AsyncSession,
+        organization_id: UUID,
+        admin_id: UUID,
+        account: AccountCreate,
+    ) -> Account | None:
+        if (
+            await self.get_by(session=session, organization_id=organization_id)
```
birkjernstrom: Personal preference, but I'm not a fan of await calls in an if-statement. Would rather capture the fetch/getter into a variable and then check the variable. Benefit being debugging with breakpoints and less code-diff in case we need to use the data.

server/polar/account/service.py
```diff
@@ -0,0 +1,75 @@
...
+            is not None
+        ):
+            # TODO: Error?
```
birkjernstrom: I think it's fine at this stage of rapidly getting to an alpha to simply return None vs. have custom error exceptions for each case, handle them etc (I don't think we will or want to either in all these cases). So I think we can remove the `# TODO: Error?`, but replace them with logging (info/warning) so that we can see them in the logs and then know what to address/improve in the future.

server/polar/enums.py
```diff
@@ -1,9 +1,12 @@
...
 # We will likely want to support Gitlab users in the near future so this
 # abstraction is here to hopefully support a drop-in – albeit custom – Gitlab client.
-
-
 class Platforms(str, Enum):
     github = "github"
+
+
+class AccountType(str, Enum):
+    stripe = "stripe"
```
birkjernstrom, line 12: Nice that `enums.Platforms` is getting a friend to warrant the module ;-)

server/polar/integrations/stripe/endpoints.py
```diff
@@ -45,6 +53,40 @@ async def enqueue(event: stripe.Event) -> WebhookResponse:
     return WebhookResponse(success=True, job_id=enqueued.job_id)
 
 
+@router.get("/return")
```
birkjernstrom, line 56: Very nice! I'm most likely going to move the Github Auth callback to a handler on the backend too vs. having frontend be involved (current stage). 

Do we want this to be auto-documented in the OpenAPI schema and /docs playground? Otherwise, we can add `include_in_schema=False`, but at least we should remove it from the automatic API TypeScript client generation. But we haven't done removals there before. Just modification of names. So unless it's a quick fix, I think we can add a TODO comment and move on for now.

server/polar/integrations/stripe/endpoints.py
```diff
@@ -45,6 +53,40 @@ async def enqueue(event: stripe.Event) -> WebhookResponse:
...
+async def stripe_connect_return(
+    auth: Auth = Depends(Auth.user_with_org_access),
+    session: AsyncSession = Depends(get_db_session),
+) -> RedirectResponse:
+    account = await account_service.get_by(
+        session, organization_id=auth.organization.id
+    )
+    if not account:
+        raise HTTPException(status_code=400, detail="Error while getting account")
+    stripe_account = stripe_lib.retrieve_account(account.stripe_id)
```
birkjernstrom: Imported as `stripe`, but called as `stripe_lib` here?

server/polar/kit/db/models/base.py
```diff
@@ -2,18 +2,30 @@
...
-from sqlalchemy import TIMESTAMP
+from sqlalchemy import TIMESTAMP, MetaData
 from sqlalchemy.orm import DeclarativeBase, Mapped, MappedColumn, mapped_column
 
 from polar.kit.extensions.sqlalchemy import PostgresUUID, IntEnum
 from polar.kit.utils import utc_now, generate_uuid
 
 from .mixins import ActiveRecordMixin, SerializeMixin
 
+my_metadata = MetaData(
```
birkjernstrom, line 13: Great work on this yesterday!

PR #45, opened 2023-03-21
Issue References

zegl:


server/migrations/versions/2023-03-21_repository_repository_issues_references_.py
```diff
@@ -0,0 +1,31 @@
...
+# revision identifiers, used by Alembic.
+revision = '7f3d2182a4fb'
+down_revision = '79c64025f7cf'
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('repositories', sa.Column('repository_issues_references_synced_at', sa.TIMESTAMP(timezone=True), nullable=True))
```
birkjernstrom, line 24: Let's shorten this to `issue_references_synced_at`

server/polar/dashboard/endpoints.py
```diff
@@ -142,11 +142,9 @@ async def get_dashboard(
...
     # get linked pull requests
     for i in issues:
-        if not i.repository_id:
-            continue
-
-        prs = await pull_request.list_by_repository_for_issue(
-            session, i.repository_id, i.number
+        prs = await pull_request.list_referencing_issue(
+            session,
+            i,
```
birkjernstrom, line 147: Let's add an issue to review the dashboard endpoint later to streamline database roundtrips a bit. I'm allergic to database calls in a for-loop (some rare exceptions). I think perhaps we should add a `get_many` method in the base ResourceService class that we can leverage. So we can pass in a list of IDs and get everything with one call and roundtrip. I know we're doing this in some places, so would be nice to standardize and make easy so we don't have to do nested database calls in loops.

server/polar/integrations/github/service/reference.py
```diff
@@ -0,0 +1,209 @@
...
+        """
+        sync_repo_references lists repository events to find issues that have been mentioned.
+        When we know which issues that have been mentioned, a job to fetch and parse timeline
+        events will be triggered.
+        """
+
+        if (
+            repo.repository_issues_references_synced_at
+            and (utils.utc_now() - repo.repository_issues_references_synced_at).seconds
+            < 60
```
birkjernstrom: Let's add `60` as a config variable. Worth testing, but I think all of these API calls will consume quite a lot of rate limit so we're likely only able to do it a few times per day. Would be nice to easily be able to customize and change it. Low (60) is probably nice now during development and testing.

server/polar/integrations/github/service/reference.py
```diff
@@ -0,0 +1,209 @@
...
+        events = res.parsed_data
+
+        synced_issues: Set[UUID] = set()
+
+        for event in events:
+            if event.event == "referenced" and event.issue:
+                issue = await github_issue.get_by_external_id(session, event.issue.id)
+                if not issue:
+                    log.warn(
+                        "github.sync_repo_references.issue-not-found",
```
birkjernstrom, line 78: Up for discussion, but I think we should have some convention on logging event names. Personally, I prefer only having `.` and `_` vs. dashes. Combined with trying to have them be quite short. This length is fine, but probably at the ceiling

server/polar/integrations/github/service/reference.py
```diff
@@ -0,0 +1,209 @@
...
+
+                # sync has already been triggered for this issue
+                if issue.id in synced_issues:
+                    continue
+                synced_issues.add(issue.id)
+
+                # Trigger issue references sync job
+                await enqueue_job("github.issue.sync.issue_references", issue.id)
+
+        return None
```
birkjernstrom, line 92: This can be removed. None is the default "return"

server/polar/integrations/github/service/reference.py
```diff
@@ -0,0 +1,209 @@
...
+        async for event in client.paginate(
+            client.rest.issues.async_list_events_for_timeline,
+            owner=org.name,
+            repo=repo.name,
+            issue_number=issue.number,
+        ):
+            if isinstance(event, github.rest.TimelineCrossReferencedEvent):
+                await self.parse_issue_pull_request_reference(session, event, issue)
+
+        return
```
birkjernstrom, line 123: We can remove this return at the end too

server/polar/integrations/github/service/reference.py
```diff
@@ -0,0 +1,209 @@
...
+
+    async def parse_issue_pull_request_reference(
+        self,
+        session: AsyncSession,
+        event: github.rest.TimelineCrossReferencedEvent,
+        issue: Issue,
+    ) -> None:
+        if not event.source.issue:
```
birkjernstrom, line 131: Can we have this case and what does it mean? I'd add a comment for readability. Potentially check for this at invocation and in context

server/polar/integrations/github/service/reference.py
```diff
@@ -0,0 +1,209 @@
...
+        self, session: AsyncSession, issue_id: UUID, pr_id: UUID
+    ) -> None:
+        nested = await session.begin_nested()
+        try:
+            relation = IssueReference(
+                issue_id=issue_id,
+                pull_request_id=pr_id,
+            )
+            session.add(relation)
+            await nested.commit()
```
birkjernstrom, line 191: Why do we need a nested transaction here? It can be called numerous times within the same session and sync task? Since we're committing the main session below + not doing another database call in case of an integrity error (except for the standard rollback), I wonder if we need the nested transaction at all?

server/scripts/github.py
```diff
@@ -75,6 +87,17 @@ async def resync_issues(org_name: str) -> None:
         await trigger_issue_sync(session, org)
 
 
+@cli.command()
```
birkjernstrom, line 90: Very nice with the added CLI options to try this out

PR #65, opened 2023-03-30
Issue dependencies on dashboard

hult:
- [x] Dashboard API still not conforming to JSONAPI (every key in `relationships` should be a dict with a data member that is either `{ type: ..., id: ...}` or a list of the same)
- [x] Return issue dependencies in the issue list, rather than as relations of the issues depending on them


server/polar/dashboard/endpoints.py
```diff
@@ -111,9 +112,12 @@ async def get_dashboard(
     pledges = await pledge.get_by_issue_ids(session, issue_ids)
 
     # start building issue relationships with pledges
-    issue_relationships: Dict[UUID, List[Relationship]] = {}
-    for i in issues:
-        issue_relationships[i.id] = []
+    issue_relationships: Dict[UUID, IssueRelationship] = {}
+    # for i in issues:
```
hult: Remove

PR #253, opened 2023-04-26
Fix API development environment 

birkjernstrom:
Issue: Couldn't start and/or stop the API development environment without `kill -9` quite frequently.

Two things changed here.

**Start: Github App verification script**
Interesting one, but before starting the dev environment we always run the script to verify the accurate Github App configuration. I improved logging a bit to make that more clear.

The problem was the introduction of `RedisCache` which automatically keeps a cached instance of the app JWT. This is of course fantastic for production and development, but...

Github's `/app` endpoint only seems to return HTTP 200 and no error codes. So we can have a cached JWT, but it could be the wrong one for the new process/app/config. Regardless, what ends up happening is Github or Githubkit doesn't return and is waiting for an API response. So the process hangs and the API is never started.

Solved this by overriding the usage of `RedisCache` and turning it off for the verification script. Starts like a 'beaut again.

**Stop: SSE infinite loop**
Stopping the API development process also didn't work and we experienced SSE events happening after Ctrl-C. I re-introduced the sleep (which is good to have, I think), but dramatically lowered the time vs. before. This seems to play more nicely with uvicorn which handles the signals and aborts the process with the sleeping loop.

### DIFF
--- a/server/polar/eventstream/endpoints.py
+++ b/server/polar/eventstream/endpoints.py
@@ -29,6 +29,8 @@ async def subscribe(redis: Redis, channels: list[str]):
                 log.info("redis.pubsub", message=message["data"])
                 yield message["data"]
 
+            await asyncio.sleep(0.01)
+
 
 @router.get("/user/stream")
 async def user_stream(


server/polar/eventstream/endpoints.py
```diff
@@ -29,6 +29,8 @@ async def subscribe(redis: Redis, channels: list[str]):
                 log.info("redis.pubsub", message=message["data"])
                 yield message["data"]
 
+            await asyncio.sleep(0.01)
```
hult, line 32: 😮

PR #427, opened 2023-05-15
Remove duplicate and make str enums in config

sultaniman:
Hey,

This small PR is to adjust enums to explicitly say they are string based and removed duplicate when initializing at the bottom.

### DIFF
--- a/server/polar/config.py
+++ b/server/polar/config.py
@@ -4,14 +4,14 @@ from enum import Enum
 from pydantic import AnyHttpUrl, BaseSettings, PostgresDsn, validator
 
 
-class Environment(Enum):
+class Environment(str, Enum):
     development = "development"
     testing = "testing"
     staging = "staging"
     production = "production"
 
 
-class EmailSender(Enum):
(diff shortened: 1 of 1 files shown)


birkjernstrom:
How can you not be romantic about open source? ([Moneyball reference](https://youtu.be/9rrAbLNePxU?t=88)) 

This is truly special. We open sourced Polar and only a few hours later you committed the first ever contributor PR. It's been a special day, but this takes the prize! @sultaniman please drop me a note at birk@polar.sh – I'd love to ship you something nice to remember and celebrate the moment :) 

In terms of the PR: Looks great – merging now. Fun fact with the duplicate line: Don't run two linters at once in VSCode. Fully switched to the amazing Ruff now, but there was a day when competing linters caused duplicate lines on each save 😅 Thanks for cleaning that one up.

sultaniman:
Woohoo, what a great honor! I love the tech stack used and this comes very close to what I am used to develop with FastAPI especially in my own recent OSS project https://github.com/lbrty/observer

PR #489, opened 2023-05-17
Fix pledge overlay disappearing when mouse leaves window

### LINKED ISSUE #452: Dashboard pledge overlay having some display issues
- [ ] Toggles visibility when "Pledge"-button is hidden since it's a child component
- [ ] Issue activities can appear over the overlay (higher z-index)

hult:
Fixes #452

In this solution, the button still disappears if the mouse leaves the window, but the modal stays


birkjernstrom:
Nice! Tested and it works great. Merging.

PR #537, opened 2023-05-23
WIP Introduce OpenAPI helper to add more response statuses

sultaniman:
This PR introduces enhancements to OpenAPI generation, adds additional responses to swagger schema  https://fastapi.tiangolo.com/advanced/additional-responses/?h=responses

### DIFF
--- a/server/polar/kit/api/responses.py
+++ b/server/polar/kit/api/responses.py
@@ -1,0 +1,29 @@
+from typing import Any, Dict, Union
+
+from starlette.exceptions import HTTPException
+
+
+def get_api_errors(*additional_responses) -> Dict[Union[int, str], Dict[str, Any]]:
+    """Adds additional responses to swagger schema
+    https://fastapi.tiangolo.com/advanced/additional-responses/?h=responses
+
+    ```
+    get_api_errors(
+        status.HTTP_401_UNAUTHORIZED,
+        status.HTTP_403_FORBIDDEN,
+        (status.HTTP_404_NOT_FOUND, "Project or member not found"),
+    )
+    ```
+    """
+    responses = {}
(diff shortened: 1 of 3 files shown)


server/polar/kit/api/responses.py
```diff
@@ -0,0 +1,29 @@
+from typing import Any, Dict, Union
+
+from starlette.exceptions import HTTPException
+
+
+def get_api_errors(*additional_responses) -> Dict[Union[int, str], Dict[str, Any]]:
```
Kludex, line 6: This is not needed... 👀

sultaniman: @Kludex thanks for taking a look 🤝 , how can we change this what do you suggest?

Kludex: Just create the dictionary on each endpoint

birkjernstrom: Thanks @Kludex for the great feedback ♥️ I agree with it that we can simply add the responses directly to the router decorator, i.e no need for the `get_api_errors` function. @sultaniman Thanks for helping out with this and improving the documentation/Swagger of our endpoints – much appreciated 🙏🏼

sultaniman: @Kludex @birkjernstrom then I will close this PR and will instead open a separate one.

birkjernstrom: @sultaniman Awesome 🙏🏼 You can convert it to a draft too and ping me at any point if you want an earlier review too – whatever you prefer 🙂

birkjernstrom: Oops. Saw now that you closed it. I was too slow :)

Issue #451, opened 2023-05-16
Redirect to dashboard w/ toast for authenticated – public – pledges


birkjernstrom:
In case an authenticated backer makes a pledge via Github Badge, i.e public flow, we should redirect them to their dashboard with a toast confirming the pledge vs. public thank you page.

**Depends on**
#447

PR #564, opened 2023-05-25
Removing CC on file + inline dashboard pledges

### LINKED ISSUE #531: Dashboard pledge should link to public page + remove CC on file (for now)
We've decided to remove this feature for now to avoid confusion and risky outcomes. It might not be entirely clear that you save a CC on file for the entire organization and potentially not the right approach vs. individual cards (expenses). 

Instead, we'll re-introduce this more properly through a combination of later efforts, e.g:
1) Save personal CC on file
2) Team management / budgets (more explicit and intentional)

**Todo now**
- [x] Remove CC on file from settings and our API
- [x] "Pledge" in dashboard view should link to public pledge page
- [x] No animation for the "Pledge" button in the dashboard (don't hide it)
- [x] Redirect to dashboard dependency view
- [x] Show success toast in dashboard if pledge was successful
- [x] Switch to redirect from `/status` page + Zustand store. Cleaner and simpler logic.
- [x] Test Stripe "0" HTTP calls in `main` and compare against this branch. Seems excessive. Wondering if it's always been like this

birkjernstrom:
Fixes #531 #451


birkjernstrom:
Accidentally created a PR vs. draft (intent). Still have one more thing todo before it's ready.

PR #729, opened 2023-06-13
Fix SSE

birkjernstrom:
Root cause was using & awaiting request.is_disconnected().

It uses anyio.CancelScope to await/receive cancellations. It crashed
silently when it ran inside of our infinite SSE loop.

As I dug deeper into redis-py, sse-starlette and starlette investigating this,
I saw that all examples had a short sleep to spare CPU even with get_message() being
non-blocking. Since it polls constantly using select. Nice to have a tiny breather in-between.

- server/sse: Remove usage of Starlette request.is_disconnected
- server/sse: Add minimal sleep

### DIFF
--- a/server/polar/config.py
+++ b/server/polar/config.py
@@ -81,6 +81,8 @@ class Settings(BaseSettings):
 
     POSTHOG_PROJECT_API_KEY: str = ""
 
+    SSE_SLEEP_INTERVAL: float = 0.001
+
     class Config:
         env_prefix = "polar_"
         env_file_encoding = "utf-8"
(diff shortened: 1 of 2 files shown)


birkjernstrom:
Further clarification. `sse-starlette` will trigger a `asyncio.CancelError` exception upon `http.disconnect` client-side which we're already capturing. Using that as a signal to when we should abort vs. polling `request.is_disconnected` which doesn't work here.

birkjernstrom:
Converting to draft. Needs further investigation. Works in production. Just for me it seems that this is a bug...

PR #754, opened 2023-06-15
apps/dashboard: allow signup by anyone

### LINKED ISSUE #726: Toggle invite gatekeeping via config
For launch we need to remove gatekeeping by invite codes. Rather than delete the code directly we could have a configuration flag, e.g `polar.config.settings.PUBLIC_SIGNUP`, and in case it's `TRUE` we skip the invite code requirement.

**Todo**
- [ ] Allow signup/login from anyone and skip invite code requirement if config is `TRUE`
- [ ] Only show the ToS + Privacy Policy agreement in `Gatekeeper` client-side if invites are required

zegl:
Fixes #726 

It's not a config, but I believe that this is easier to maintain. :-)

### DIFF
--- a/clients/apps/web/src/components/Dashboard/Gatekeeper/Gatekeeper.tsx
+++ b/clients/apps/web/src/components/Dashboard/Gatekeeper/Gatekeeper.tsx
@@ -9,7 +9,8 @@ const Gatekeeper = (props: { children: React.ReactElement }) => {
     return <></>
   }
 
-  if (!currentUser.invite_only_approved) {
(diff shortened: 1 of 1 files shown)


birkjernstrom (commented):
Ah ok, I thought we had some gatekeeping on the backend too. Only problem with this is that users signing up would have `invite_only_approved` as false, and so if we re-introduced `requireInvite` we would shut them out. Bit theoretical though, I think once we go live, we're live and no point of return. So I think this is good for now and we either 1) remove the codepaths entirely later OR 2) solve it based on the needs then.

birkjernstrom:
Absolutely agree 👌

PR #782, opened 2023-06-21
Setup mkdocs-material for docs+blog

birkjernstrom:


birkjernstrom:
Going to deploy this skeleton of `mkdocs-material` to setup DNS etc for `docs.polar.sh` via Vercel and get it all up & running before continuing on customizations + content.

PR #801, opened 2023-06-26
Enable 'pip install .' for docs' dependencies

fredrikaverpil:
### Why this change?

I saw @birkjernstrom 's tweet: https://twitter.com/birk/status/1673337242486558721?s=20

It's possible to install a poetry-managed pyproject.toml with pip, since poetry-core and pip has [PEP 517](https://peps.python.org/pep-0517) support:

```bash
# cd into the polar repo's docs folder (I guess this is how you want this to work)
cd polar/docs

# install
pip install .

# install as "editable" (noice for local development)
pip install -e .
```

But you've got a typo in the `packages` spec (doesn't line up with your docs folder name), and you must add an `__init__.py` into the package folder in order to make it into a Python package.

### What was changed?

- Add `__init__.py`
- Fix typo in `pyproject.toml`

### Notes

- You can check this branch out and run `pip install .` to install the mkdocs-material dep + the docs as a package itself. The versions specified in poetry.lock is supposed to be respected.
- The latter part of installing the docs package is likely redundant to you, but it is something poetry requires. Might be possible to somehow configure this to not happen 🤷  (a rabbit hole I haven't had to dive into)
- Even if you are installing the insiders version of mkdocs 🤟:100: from vercel now, you might want to consider fixing the typo and the need for this `__init__.py` if you expect users to be able to install mkdocs via this pyproject.toml file, with or without poetry. There's also an option to define e.g. [poetry dependency groups](https://python-poetry.org/docs/master/managing-dependencies/) for server, client, docs, tests etc in one central pyproject.toml file 😄 (will also remove overlap of dependencies dependabot/renovate will bug you about).

### DIFF
--- a/docs/pyproject.toml
+++ b/docs/pyproject.toml
@@ -5,7 +5,9 @@ description = "Documentation, FAQ & Blog for Polar"
 authors = ["Polar Team <support@polar.sh>"]
 license = "Apache-2.0"
 readme = "README.md"
-packages = [{include = "polar_docs"}]
+packages = [
+  { include = "docs" }
+]
 
 [tool.poetry.dependencies]
 python = "^3.9"


birkjernstrom:
Thank you so much @fredrikaverpil. So awesome with open source how I can share a question over a tweet and get a PR answer ♥️

However, reading `mkdocs-material` documentation on how to best deploy Insiders to Vercel, it turns out we only need the pip/poetry dependencies for local development. So not an issue anymore in production since it's installed via pip against remote & official mkdocs-material packages.

So unfortunately, no need to merge this one. However, the typo would be nice to fix (in case of usage in the future + clarity) so if you remove the `__init__.py`, I'll merge this :-)

birkjernstrom:
Truly sorry for the delay on this one @fredrikaverpil ❤️

This sent me into a bit of a rabbit hole of refreshing and better understanding the current state of Python package and dependency management. Honestly, I started using Poetry because I saw others use it (apps vs. packages) and preferred the `pyproject.toml` format including the centralized configuration it provided (mypy etc).

You're absolutely right though. Reading this [excellent write-up](https://caremad.io/posts/2013/07/setup-vs-requirement/) it's intended for packages vs. applications (our case).  Seems like there are a few others surprised by this here: https://github.com/pypa/pip/issues/11584

It's a shame honestly. Feels like `pyproject.toml` would be an excellent standard for both use cases provided it could support some form of flag, e.g `package = false` versus having it, requirements.txt, Pipfile etc. So I'm a bit torn here. The right choice is using `requirements.txt`, but it simultaneously feels like a step backward in DX. Having to have multiple configuration files and running pip + virtualenv. I don't remember those days with particular fondness since last I used Python substantially :-) You have pipenv, but I saw a lot of discouraging posts about that when quickly investigating this earlier this year. Will look into it again.

However, for now, we'll keep `pyproject.toml` with Poetry even though it's wrong. We should do it right eventually, but it's not a high priority now in exchange for – imho – a worse DX. Therefore, I think we can close this PR and I'll create an issue for this. Is that fair @fredrikaverpil ?

fredrikaverpil:
> Therefore, I think we can close this PR and I'll create an issue for this. Is that fair @fredrikaverpil ?

Absolutely. 😄 

Unfortunately, there is no npm/packages.json simple approach in the Python packaging ecosystem today. Instead there is a myriad of ways to achieve different things. If you don't mind, I'll just share my thoughts here; my personal rule of thumb is to use Poetry for apps and Hatch/hatchling for libraries. This comes from having used both for a long time and interacted with their maintainers.

I would personally advise against using Pipenv or PDM, but we can take that offline perhaps 😄 as I don't want this to drag out for too long but also not come out as dunking on other people's hard work.

I'm sure this will be a lot of stuff you already know, but maybe some golden nuggets that could help you figure this out to cater to your own needs.

### Poetry for apps

When developing apps (like deploying to AWS with a Dockerfile) this requires the ability to maintain a lockfile (like package.json). Poetry handsdown provides the simplest DX among the choices you have out there and kind of "just works" (at least to the extent we can talk about in the python eco system). An honorable mention goes to pip-tools but the DX is not nearly as streamlined.

Even if you do not intend to send anyone a wheel or a source distribution of your app, it still makes sense to build the package internally, as you'd catch beginner mistakes early on how you've structured your application. You can see the build step as an invisible part of the installation phase.

### Hatchling for libraries

When you intend to distribute a library, you don't want a lockfile. Then you want to be a lot more lenient about which deps versions the end user is on. Here it would be better to (still have a package definition) and define the _ranges_ or _minimum_ dependency versions supported. Hatchling provides some great convenience/DX functionalities, which makes this a lot more useful than just using a "vanilla" `pyproject.toml` IMHO.

Hatchling has surpassed tools like setuptools in popularity and even the pypa refers to hatchling in the [official packaging guide](https://packaging.python.org/en/latest/tutorials/packaging-projects/) these days.

### One project, one `pyproject.toml`

Obviously, the `pyproject.toml` specifies _one_ project. There is no story for a monorepo of multiple projects here. The convention AFAIK is to add one `pyproject.toml` for each repo's root. This assures that all the end-user has to do is set up one virtual environment and run just one command (e.g. `pip install -e .` or `poetry install`) to be able to consume your main app/library. And that would be all you'd have to do. Anything else means "you're on your own" kind of, figuring out your own take on a monorepo, and you'd have to start your main README to outline how you intend the end user to navigate your monorepo (since this is rarely using any conventions, at least not in the Python world).

Taking the above into account, for both an app and a library, the dependencies for your docs would most often be something you add into this main/root `pyproject.toml`. Most usually in a "docs" dependency group to make it clear you don't need these dependencies to run your app or use your library, but you can optionally install them. With Poetry you have [dependency groups](https://python-poetry.org/docs/master/managing-dependencies/#dependency-groups) and for hatchling/vanilla `pyproject.toml` you have [optional dependencies](https://hatch.pypa.io/latest/config/metadata/#optional).

I personally use such dependency groups for linting, formattting, testing etc - maybe overusing them a bit here, but it can in some situations make CI faster and has a smaller cache footprint... Example [here](https://github.com/fredrikaverpil/creosote/blob/main/pyproject.toml#L36-L41).

### Keeping docs separated from your projects

In your case, it seems (at least right now) you want to write docs completely detached from e.g. running [doctests](https://github.com/koaning/mktestdocs) or similar (which would require your app/lib dependencies). It also looks like your docs Vercel deployment story does not require a distribution (such as a wheel). Because of this, I would've personally just used a `requirements.txt` file, at least short-term, to indicate that all you need here is that one mkdocs dependency and no further setup.

You might also want to look at Pydantic's docs story for inspiration, as they consume docs from other projects like pydantic-settings, making it possible to encompass all aspects of Pydantic eventhough it is spread over mutiple projects.

### Closing notes

The Python ecosystem for a long time relied on requirements.txt files and projects such as pip-tools emerged to provide the capability of coupling that with a lockfile mechanism. Over time, PEPs were driven in a direction to allow to tap into pip behavior by providing your own build frontend/backend, supplying your own code to e.g. resolve dependencies or read from your own lockfile mechanism. This has resulted in fragmentation, and I recommend you [listen to this talkpython podcast episode](https://talkpython.fm/episodes/show/406/reimagining-pythons-packaging-workflows) if you want to hear more about how the PyPa are in the process of reevaluating the situation, maybe focussing more on providing a "sane default" as to how to deal with these things.

Comparing this whole situation with the rust/go ecosystem, we still have a long way to go in terms of DX simplicity though... (according to rambling old me, that is).

birkjernstrom:
Wow thank you @fredrikaverpil for this excellent write-up! 🙏 Creating an issue out of it to have in the backlog so this doesn't get lost.

I agree with the recommendation outlined here in regards to using `poetry` still and that our projects are treated as packages, but internal ones. Just doing it for `server` for now since that's our only true Python "repo" in development within our monorepo and docs are separate. Likely moving docs to a simple `requirements.txt`.

PR #831, opened 2023-07-04
Add support for OpenCollective

### LINKED ISSUE #712: Support Open Collective for Maintainers
**Background**
Several maintainers have requested to use Open Collective instead of Stripe, e.g collective of maintainers using OC as their fiscal host. I chatted with OC and am glad to share that they're excited to support Polar through the same setup/flow they have with Github today. 
* Maintainers can link their OC account (URL) to Polar vs. Stripe
* Inbound pledges work as normal, i.e goes to the Polar Stripe platform account
* Once completed we don't transfer it using Stripe Transfer API (via backoffice).
* Instead, we:
   - Highlight it as "Open Collective" in our backoffice instead of Stripe actions
   - Run a monthly report to export all such outstanding balances to be paid to Open Collective (outside the scope of the initial task here – we'll do that ad-hoc in partnership with OC)

### Considerations & limitations
* We should expand our `accounts` model to support + introduce types, e.g `stripe` & `open_collective`
* We should only support one `active` account per organization, i.e it's either or, but multiple records can exist with one active. _Note: Supporting changing accounts via UI is outside the scope for this issue. Will require support outreach + help_
* Required fields across both will be `country`, `currency` & `business_type`. For OC, we'll prompt the maintainer to specify theirs as with Stripe. For `currency` we'll hardcode it to `USD` and `business_type` as `fiscal_host`
* `is_details_submitted`, `is_charges_enabled` & `is_payouts_enabled` are quite nice to leverage for OC as well to standardize the model/usage in the future. However, they will default to `true` for valid OC submissions.

#### Setup
`https://polar.sh/finance/:org_slug:`

**Current**
image

**Changes to be made**
* Bonus: Fix "Setup up Stripe" to "Setup Stripe" 😅
* Change copy to "You need to setup **Stripe** or **Open Collective** for payouts"
* Change button to "Setup" (onClick opens the next module as it does today) 

image

**Changes to be made**
* Add dropdown to choose between `Stripe` & `Open Collective`. This should be before current country dropdown. Stripe as the default.
* Move country explanation, e.g "If this is a personal account", below the country dropdown. Make it `text-xs`
* If Stripe: That's it. "Setup account" links to Stripe as is today.
* If OC: Add Input for Open Collective URL.
  - Label: "Open Collective Profile URL"
  - Input
  - Explanation below (text-rs): "Example: https://opencollective.com/curl" 
  - On "Setup account". Parse, validate & save. Close modal to return to financial view (next).

#### Review finances

image

**Changes to be made**
* If Stripe: Keep as is
* If OC:
  - Change icon to OC icon
  - Copy: "Payouts will be sent in bulk once per month to the connected Open Collective account"
  - "Go to Open Collective". Link to the OC profile URL.

### Expected Backoffice UX

URL: `https://polar.sh/backoffice/pledges`

image

**Changes to be made**
We'll start simple here. As we do our first payouts with OC and have the operational flows setup in collaboration with them, we'll learn and refine this. For now, let's first make sure we clearly see it's OC vs. Stripe.
* Add column to table. "Account" with `Stripe` vs. `Open Collective` for these pledges in terms of the receiving account
* Keep actions (buttons) for Stripe. Replace them for `Open Collective` with "Manual Open Collective Payouts"

## Funding

<a href="https://polar.sh/polarsource/polar/issues/712">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/polarsource/polar/issues/712/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

frankie567:
Fix #712


server/polar/models/account.py
```diff
@@ -24,14 +25,15 @@ class Status(str, Enum):
         UniqueConstraint("stripe_id"),
     )
 
-    account_type: Mapped[str] = mapped_column(String(10), nullable=False)
+    account_type: Mapped[AccountType] = mapped_column(String(255), nullable=False)
```
frankie567, line 28: I had to stretch the size of several fields so it can fit other type of values

server/polar/models/account.py
```diff
@@ -24,14 +25,15 @@ class Status(str, Enum):
...
 
     organization_id: Mapped[UUID] = mapped_column(
         PostgresUUID, ForeignKey("organizations.id"), unique=True
     )
     admin_id: Mapped[UUID] = mapped_column(PostgresUUID, ForeignKey("users.id"))
 
-    stripe_id: Mapped[str] = mapped_column(String(100), nullable=False)
+    stripe_id: Mapped[str | None] = mapped_column(String(100), nullable=True)
+    open_collective_slug: Mapped[str | None] = mapped_column(String(255), nullable=True)
```
frankie567, line 36: I added a dedicated field to handle Open Collective. Not a super-big fan of this (probably can be more generic), but on the other hand, Stripe ID and Open Collective Slug are not semantically the same thing.

`stripe_id` is now nullable.

server/polar/account/endpoints.py
```diff
@@ -18,30 +19,36 @@ async def create_account(
...
-        session, auth.organization.id, auth.user.id, account
-    )
-    if not created:
-        raise HTTPException(status_code=400, detail="Error while creating account")
+    try:
+        created = await account_service.create_account(
+            session, auth.organization.id, auth.user.id, account
+        )
+    except AccountServiceError as e:
+        raise HTTPException(status_code=400, detail=e.message)
```
frankie567, line 27: I've revamped error handling there using exception classes. It helps us to give more meaningful error messages to the user.

server/polar/account/endpoints.py
```diff
@@ -18,30 +19,36 @@ async def create_account(
...
+
     return AccountRead.from_orm(created)
 
 
 @router.get(
-    "/{platform}/{org_name}/accounts/{stripe_id}/onboarding_link",
+    "/{platform}/{org_name}/accounts/{account_id}/onboarding_link",
```
frankie567, line 33: I changed the endpoints so they now directly consider the Account ID instead of Stripe ID. This way, it's generic no matter the type of account.

server/polar/account/endpoints.py
```diff
@@ -50,28 +57,30 @@ async def onboarding_link(
...
+
+    link = await account_service.dashboard_link(account)
     if not link:
         raise HTTPException(status_code=400, detail="Error while creating link")
     return link
 
 
 @router.get("/{platform}/{org_name}/accounts", response_model=list[AccountRead | None])
-async def get_account(
+async def get_accounts(
```
frankie567, line 83: Fixed a typo in the route name.

That said, now that [we agreed](https://github.com/polarsource/polar/issues/712#issuecomment-1625083182) an organization can have only one `Account`, it feels a bit weird to return a list.

server/polar/account/schemas.py
```diff
@@ -13,9 +14,18 @@ class AccountLinkTypes(str, Enum):
 
 
 class AccountCreate(Schema):
-    account_type: AccountType | None
+    account_type: AccountType
+    open_collective_slug: str | None = Field(None, min_length=1)
     country: str
 
+    @root_validator(skip_on_failure=True)
+    def validate_open_collective(cls, values: dict[str, Any]) -> dict[str, Any]:
```
frankie567, line 22: Validate that `open_collective_slug` is provided if `AccountType` is `open_collective`

server/polar/account/schemas.py
```diff
@@ -28,8 +38,10 @@ class AccountUpdate(Schema):
 
 
 class AccountRead(AccountCreate):
-    account_type: AccountType | None
-    stripe_id: str
+    id: UUID
```
frankie567, line 41: We missed the `id` in the output schema.

server/polar/account/service.py
```diff
@@ -14,36 +14,67 @@
...
+    open_collective,
+    OpenCollectiveAPIError,
+    CollectiveNotFoundError,
+)
 
 
 log = structlog.get_logger()
 
 
+class AccountServiceError(Exception):
```
frankie567, line 27: As mentioned above, I've implemented an approach based on exceptions to handle errors and return more precise error messages to the user.

server/polar/account/service.py
```diff
@@ -55,53 +86,89 @@ async def create_account(
...
-        )
-        if account is None or account.admin_id != user.id:
-            # TODO: Error?
-            return None
+        admin_id: UUID,
+        account: AccountCreate,
+    ) -> Account:
+        assert account.open_collective_slug is not None
+        try:
+            collective = await open_collective.get_collective(
```
frankie567, line 98: We have a service to query the Open Collective API and check if the collective exists, is valid and has the right financial host

server/polar/backoffice/schemas.py
```diff
@@ -30,6 +33,10 @@ def from_db(cls, o: Pledge) -> Self:
             pledger_name = o.by_organization.name
             pledger_avatar = o.by_organization.avatar_url
 
+        to_organization_account_type = None
+        if o.to_organization.account is not None:
+            to_organization_account_type = o.to_organization.account.account_type
```
frankie567, line 38: Surface the account type when querying a `Pledge` from the backoffice.

server/polar/config.py
```diff
@@ -66,6 +66,9 @@ class Settings(BaseSettings):
     # Stripe webhook secret
     STRIPE_WEBHOOK_SECRET: str = ""
 
+    # Open Collective
+    OPEN_COLLECTIVE_PERSONAL_TOKEN: str | None = None
```
frankie567, line 70: Open Collective API can be queried without a token. However, it's limited to 10 req/min. Thus, it's probably better to set up a token of our own to reach 100 req/min.

Ref: https://graphql-docs-v2.opencollective.com/access#authentication

server/polar/integrations/open_collective/service.py
```diff
@@ -0,0 +1,103 @@
...
+
+class OpenCollectiveService:
+    def __init__(self, personal_token: str | None = None) -> None:
+        self.personal_token = personal_token
+
+    def create_dashboard_link(self, slug: str) -> str:
+        return f"https://opencollective.com/{slug}"
+
+    async def get_collective(self, slug: str) -> OpenCollectiveCollective:
+        async with self._get_graphql_client() as client:
```
frankie567, line 53: This performs a GraphQL query to retrieve a collective from its slug. We have different exceptions to handle different error cases.

clients/packages/polarkit/src/api/errors.ts
```diff
@@ -0,0 +1,29 @@
...
+        ...map,
+        [loc]: [...map[loc], error.msg],
+      }
+    }
+    return {
+      ...map,
+      [loc]: [error.msg],
+    }
+  }, {})
+}
```
frankie567, line 29: I had to add more muscles to our error handling in the UI. Basically, this helps us to work with FastAPI `ValidationError` coming from Pydantic.

frankie567 (commented):
So, *finally*, here it is! I've highlighted the most important aspects and opinionated choices in comments.

It also comes with the service to query Open Collective API and check if the collective exists, is valid and has the right fiscal host.

clients/apps/web/src/components/Dashboard/SetupAccount.tsx
```diff
@@ -63,30 +111,71 @@ const SetupAccount = ({ onClose }: { onClose: () => void }) => {
...
+                  )} 
+                </select>
+                {validationErrors.account_type?.map((error) => <p key={error} className="text-xs text-red-500 mt-2">{error}</p>)}
+              </div>
+
+              {accountType === AccountType.OPEN_COLLECTIVE &&
+                <div>
+                  <div className="relative mt-2">
+                    <div className="pointer-events-none absolute inset-y-0 left-0 flex items-center pl-3">
+                      <span className="font-display text-gray-500 dark:text-gray-40">https://opencollective.com/</span>
```
birkjernstrom, line 138: This was a lovely UX enhancement @frankie567 😍

clients/packages/polarkit/src/api/errors.ts
birkjernstrom: This is such a fantastic enhancement! It's been on the back of my mind for a bit, but never got around to it (a bit too happy path programming now during the early days). Love how simple this is and that with Pydantic we can easily add more user-friendly error messages for the future

clients/packages/polarkit/src/api/client/services/AccountsService.ts
```diff
@@ -14,11 +14,11 @@ export class AccountsService {
   constructor(public readonly httpRequest: BaseHttpRequest) {}
 
   /**
-   * Get Account
+   * Get Accounts
    * @returns AccountRead Successful Response
    * @throws ApiError
    */
-  public getAccount({
+  public getAccounts({
```
birkjernstrom, line 21: @frankie567 Did you find and run `cd clients/packages/polarkit && pnpm generate` yourself to generate these changes or did you do them by hand? 

Just realized it's something we haven't documented in our `clients/README.md`. I feel awful if you had to change these by hand 😬

server/.env.template
```diff
@@ -19,6 +19,7 @@ POLAR_GITHUB_CLIENT_SECRET="*** REPLACE ***"
 
 POLAR_GITHUB_BADGE_EMBED=1
 
+POLAR_GITHUB_APP_WEBHOOK_SECRET="*** REPLACE ***"
```
birkjernstrom, line 22: ♥️ Thanks!

server/polar/account/endpoints.py
birkjernstrom: As mentioned in the comment about `getValidationErrorsMap` this is so nice. Absolutely lovely to see how user-friendly error messages can be added with such nice DX on our end between API<>clients.

server/polar/account/endpoints.py
birkjernstrom: Makes total sense. Thanks Francois!

server/polar/account/endpoints.py
```diff
@@ -18,30 +19,36 @@ async def create_account(
...
 @router.get(
-    "/{platform}/{org_name}/accounts/{stripe_id}/onboarding_link",
+    "/{platform}/{org_name}/accounts/{account_id}/onboarding_link",
     response_model=AccountLink,
 )
 async def onboarding_link(
     platform: Platforms,
     org_name: str,
-    stripe_id: str,
+    account_id: UUID4,
```
birkjernstrom, line 39: I can't believe I didn't know about this built-in pydantic type. TIL. So nice. I think we have quite a few places where we're using `str` today. Will update & enhance as we progress throughout the codebase now knowing this.

server/polar/account/endpoints.py
birkjernstrom: Agreed.

We've been moving fast towards our alpha and our API has been exclusively designed for our own clients. So not RESTful or standardized to a high degree.

We're going to change that into an API-first approach with a public API as far as we can and some internal endpoints where needed. I think as part of our process of making a public API we'll address this (https://github.com/polarsource/polar/issues/834).

server/polar/account/service.py
```diff
@@ -14,36 +14,67 @@
...
         session: AsyncSession,
         organization_id: UUID,
         admin_id: UUID,
         account: AccountCreate,
-    ) -> Account | None:
+    ) -> Account:
         existing = await self.get_by(session=session, organization_id=organization_id)
         if existing is not None:
-            return None
+            raise AccountAlreadyExistsError()
```
birkjernstrom, line 48: I know I've said it 2 times already, but it's beautiful to see the enhanced error handling and especially how it leverages pydantic + early exceptions.

server/polar/config.py
birkjernstrom: +1. I'll fix this before deploying.

server/polar/models/account.py
birkjernstrom: Great and thanks for expanding them sufficiently for more flexible options in the future 👍🏼 In hindsight, I regret these limits 🤓

server/polar/models/account.py
birkjernstrom: I would have gone through the same thought process of internally debating something more generic 🤓 However, I'm fully supportive (+ glad) of this pragmatic route. It's not a model that will expand with a lot of types over time so the added abstraction to make it generic would likely only cause overhead/confusion vs. be clear to the developer. Good that we're starting simple 🙂

birkjernstrom:
I'm quite 🤯 of this PR @frankie567! You've absolutely addressed the issue (https://github.com/polarsource/polar/issues/712) to completion and through a high-quality contribution. It's ready to ship [*].

I've added comments inline and in response to your own (which I highly appreciated for the additional context btw), but I want to specifically callout and applaud a few things:

- Understanding & adoption of code base: It's super impressive to see this PR and how you've mastered the setup, structure and idioms of our code base without any documentation or help. Case in point: `polar.integrations.open_collective`. It's incredible to see and a bliss to review ♥️🙏🏼
- Error handling: Worthy of praise as a standalone contribution. Super nice addition to our foundations and a lovely surprise to see here – much appreciated!
- UX choices. Not mentioned in the issue, but was really nice to see how you clarified the input to already include the open collective URL so that users only need to provide the slug. Great decision and additional delight.

In short: Amazing work @frankie567 & thank you so much!

[*] One test is failing, but unrelated to this. I'll fix it + rebase (if needed) after a few meetings so we can launch this 💪🏼

clients/packages/polarkit/src/api/client/services/AccountsService.ts
frankie567: Nah, don't worry 😄 I suspected this was automatically generated code, so I looked for the command!

server/polar/account/endpoints.py
frankie567: Thanks! A small precision though: in this case, this is a "manual" error that needs to be handled specifically in the UI. Like here: https://github.com/polarsource/polar/pull/831/files#diff-37e2aa3c84916b4c0d96182c53db993dcc266f18d4792d20223c6a51abdc29caR167-R169

`getValidationErrorsMap` and friends are specifically to handle built-in 422 Pydantic validation errors.

---

However, it's entirely possible to implement a generic/global error handling at FastAPI level for our own exceptions. That's what exception handlers are for: https://fastapi.tiangolo.com/tutorial/handling-errors/#install-custom-exception-handlers

It could be a nice approach as we wouldn't even need to `except` and raise an `HTTPException` in the route handler. It would directly look like this:

```py
@router.post("/{platform}/{org_name}/accounts", response_model=AccountRead)
async def create_account(
    ...
) -> AccountRead:
    created = await account_service.create_account(
        session, auth.organization.id, auth.user.id, account
    )
    return AccountRead.from_orm(created)
```

If all our exceptions classes properly inherit from a base one (e.g. `PolarError`), this would enable a generic/global error handling at very little cost.

server/polar/account/endpoints.py
frankie567: Yes, very convenient! It automatically converts the parameter to a proper UUID object (or throws a validation error if not possible).

frankie567:
Thank you, @birkjernstrom, I really appreciate all those positive feedbacks 🙏

Regarding the failing tests, yes, I saw it but I suspected it was not because of my changes, so I preferred to let you handle it 🙂

Feel free to ping me if you need me to do the rebase!

clients/packages/polarkit/src/api/client/services/AccountsService.ts
birkjernstrom: Phew :-)

server/polar/account/endpoints.py
birkjernstrom: Yes, completely with you re: precision so no worries there :-)

Love the suggestion. It would be incredible DX, clean up the code and make sure we have standardized errors more broadly. Added a separate issue for it to tackle. https://github.com/polarsource/polar/issues/849

birkjernstrom:
Fixed the failing tests: https://github.com/polarsource/polar/pull/831/commits/de70533c454041543ecf3a36e63ad9be329464f8

I originally thought a rebase was all that was necessary, but turned out to be a bit more hairy 😅 Our `issue_service.list_by_repository_type_and_status` sorts based on `issues.created_at`. No sorting of joined pledges. Yet, our test checked against indexes where sorting of them mattered.

I diffed the query before/after and only the additional `accounts.open_collective_slug` was added. So it's surprising that this caused Postgres to sort differently during the join. But of course, it doesn't care since we're not asking it to explicitly sort by them.  Since it doesn't matter in the product today in terms of how this data is being used, I decided to not enforce sorting in the method itself, but do it within the test.

frankie567:
Oh it *was* my fault, then, sorry about that 😞 I was sure I saw this test failing even before I started to make changes. Thank you for tackling this 🙂

birkjernstrom:
Not at all @frankie567 🙂 It was a side effect that could have hit anyone. In fact, because of that it was a bit tricky to debug vs. track towards by reviewing changes. So definitely not on you. Super thankful for the contribution here – thank you again ♥️

PR #867, opened 2023-07-19
remove: hight from the drawing

### LINKED ISSUE #853: Auth page in Phone is not aligned and small
When I tried to log in with my phone the button is not aligned and small

## Funding

<a href="https://polar.sh/polarsource/polar/issues/853">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/polarsource/polar/issues/853/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

Mahmoudgalalz:
this should close #853 
@birkjernstrom


birkjernstrom:
Thanks @Mahmoudgalalz. I've now reviewed this PR and appreciate you finding this + helping out to fix it.

Unfortunately this removes the background gradient entirely which we want to keep. We should leverage CSS media queries to distinguish mobile devices and customize the gradients height/width accordingly. Solving for alignment of the login while still keeping the visual effect of the gradient. Is this something you want to fix? Otherwise, let me know and I can take a stab at it 😊

Thank you again 🙏

Mahmoudgalalz:
It seems true, but the color wasn't shown in the dark mode, so I would like to add, I will push an update rn

@birkjernstrom thanks for noticing

Mahmoudgalalz:
I don't know what happened with my local repo, but it seems that smth wrong happen I tried to remove all the commits that already exist in the main branch here if you could help and check the update that I did

birkjernstrom:
Thanks @Mahmoudgalalz! Yes I was honestly also surprised by that for a second as I was running dark mode on my phone :) Thank you for fixing it, I'll review tonight. Re: Branch state I think it's only that a rebase is needed, but I can fix that as I review no problem

birkjernstrom:
Hey @Mahmoudgalalz I was reviewing this more closely now and saw you had made some accidental reverts in this branch. I think the easiest might be to start a new and fresh branch from main and simply cherry pick the correct commits over. Is that something you could do?

Mahmoudgalalz:
@birkjernstrom hello, sorry for the late reply, I will do it and raise another PR, I wasn't need to raise another one but I think it should be done like that

Issue #849, opened 2023-07-12
Custom generic FastAPI exception handler


birkjernstrom:
From @frankie567 in https://github.com/polarsource/polar/pull/831
--------------------
Would be really nice DX. Nice suggestion & enhancement. 
--------------------

It's entirely possible to implement a generic/global error handling at FastAPI level for our own exceptions. That's what exception handlers are for: https://fastapi.tiangolo.com/tutorial/handling-errors/#install-custom-exception-handlers

It could be a nice approach as we wouldn't even need to `except` and raise an `HTTPException` in the route handler. It would directly look like this:

```py
@router.post("/{platform}/{org_name}/accounts", response_model=AccountRead)
async def create_account(
    ...
) -> AccountRead:
    created = await account_service.create_account(
        session, auth.organization.id, auth.user.id, account
    )
    return AccountRead.from_orm(created)
```

If all our exceptions classes properly inherit from a base one (e.g. `PolarError`), this would enable a generic/global error handling at very little cost.

PR #1057, opened 2023-09-04
Base exception and generic exception handler

frankie567:
Following what we discussed in #849.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/polarsource) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog :^)

<a href="https://polar.sh/polarsource/polar/issues/1057">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/polarsource/polar/issues/1057/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

### DIFF
--- a/server/polar/exception_handlers.py
+++ b/server/polar/exception_handlers.py
@@ -1,0 +1,11 @@
+from fastapi import Request
+from fastapi.responses import JSONResponse
+
+from polar.exceptions import PolarError
+
+
+async def polar_exception_handler(request: Request, exc: PolarError) -> JSONResponse:
+    return JSONResponse(
+        status_code=exc.status_code,
+        content={"type": type(exc).__name__, "detail": exc.message},
+    )
--- a/server/polar/exceptions.py
+++ b/server/polar/exceptions.py
@@ -3,29 +3,33 @@ from sqlalchemy import exc as sqlalchemy
 IntegrityError = sqlalchemy.IntegrityError
 
 
-class AuthenticationRequired(Exception):
-    """Attempted action requires authentication."""
+class PolarError(Exception):
+    """
+    Base exception class for all errors raised by Polar.
+
+    A custom exception handler for FastAPI takes care
+    of catching and returning a proper HTTP error from them.
+
+    Args:
+        message: The error message that'll be displayed to the user.
+        status_code: The status code of the HTTP response. Defaults to 400.
+    """
+
+    def __init__(self, message: str, status_code: int = 400) -> None:
+        super().__init__(message)
+        self.message = message
+        self.status_code = status_code
 
 
-class NotPermitted(Exception):
+class NotPermitted(PolarError):
+    def __init__(self, message: str = "Not permitted", status_code: int = 403) -> None:
+        super().__init__(message, status_code)
+
+
+class StripeError(PolarError):
     ...
 
 
-class InvalidPlatform(Exception):
-    ...
-
-
-class InvalidRequest(Exception):
-    ...
-
-
-class StripeError(Exception):
-    ...
-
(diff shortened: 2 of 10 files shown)


server/polar/exception_handlers.py
```diff
@@ -0,0 +1,11 @@
+from fastapi import Request
+from fastapi.responses import JSONResponse
+
+from polar.exceptions import PolarError
+
+
+async def polar_exception_handler(request: Request, exc: PolarError) -> JSONResponse:
```
frankie567, line 7: Magic happens here! The great thing is that we can implement lot of different things there to track errors!

server/polar/exceptions.py
```diff
@@ -3,29 +3,33 @@
 IntegrityError = sqlalchemy.IntegrityError
 
 
-class AuthenticationRequired(Exception):
-    """Attempted action requires authentication."""
+class PolarError(Exception):
```
frankie567, line 6: *Every* exception classes should inherit from this one.

server/polar/exceptions.py
```diff
@@ -3,29 +3,33 @@
...
-    ...
+    Args:
+        message: The error message that'll be displayed to the user.
+        status_code: The status code of the HTTP response. Defaults to 400.
+    """
 
-
-class InvalidRequest(Exception):
-    ...
+    def __init__(self, message: str, status_code: int = 400) -> None:
```
frankie567, line 18: The base exception defines two things: an error message and a status code, that'll be eventually used by the exception handler.

server/polar/exceptions.py
```diff
@@ -3,29 +3,33 @@
...
+        super().__init__(message)
+        self.message = message
+        self.status_code = status_code
 
 
-class StripeError(Exception):
-    ...
+class NotPermitted(PolarError):
+    def __init__(self, message: str = "Not permitted", status_code: int = 403) -> None:
```
frankie567, line 25: For convenience, those exceptions define a default message and status code, but this can be overridden when raising it if needed.

server/polar/exceptions.py
```diff
@@ -3,29 +3,33 @@
 IntegrityError = sqlalchemy.IntegrityError
 
 
-class AuthenticationRequired(Exception):
```
frankie567, line 6: I took the liberty to remove the ones that were unused: `AuthenticationRequired`, `InvalidPlatform`, `InvalidRequest` and `DatabaseRecordExists`.

server/polar/issue/endpoints.py
```diff
@@ -182,37 +182,23 @@ async def lookup(
 
         client = get_polar_client()
 
-        try:
```
frankie567, line 182: Here is a case where we do not need to manually catch the exception anymore.

I made the change in some places, but other places probably need to move to this new "paradigm".

server/polar/issue/endpoints.py
```diff
@@ -182,37 +182,23 @@ async def lookup(
...
+        org, repo, tmp_issue = res
 
         # get for return
         issue = await issue_service.get_loaded(session, tmp_issue.id)
         if not issue:
-            raise HTTPException(
-                status_code=404,
-                detail="Issue not found",
-            )
+            raise ResourceNotFound("Issue not found")
```
frankie567, line 194: Similarly, for logic directly in controllers where we manually raise `HTTPException`, it could be interesting to move to this common pattern.

zegl (approved):
Nice stuff!

PR #1098, opened 2023-09-07
GitHub org/repo/issue sync improvements

### LINKED ISSUE #877: Race condition of repository installation & issue synchronization
We have a known race condition for installing & synchronizing repositories. It's triggered twice:
1. GitHub webhook event: `installation.created`
2. By user during initialization (UI) upon being redirected back to Polar post-installation.

We need to support both since `#2` is not guaranteed. Simultaneously, `#1` is required to give instant feedback and real-time feedback to end-user, and sometimes `#2` is delayed.

This has worked fine: 1) No issues & 2) Average number of issues per repository is not too high so despite this we're still only using a fraction of our rate limit.

However, recently we ran into an issue for the first time (Sentry Issue ID: 4324443250) of an `IntegrityError` exception during task `github.issue.sync.issue_dependencies`. Mostly, we use upserts, but for dependencies we do need to do individual API calls so we fallback on the typical get or create pattern.

Unfortunately, the UI installation `#2` attempted to synchronize the exact same dependency at the same time as `#1`. Causing the `IntegrityError`.

Simultaneously, it was for a repository with 1.5k issues so since we ran it twice it almost consumed all of our rate limit for the installation.

/End of debugging context :-)

**Solution**
We still need to support `#1` and `#2`, however we should introduce a lock mechanism so only one can initiate it.

## Funding

<a href="https://polar.sh/polarsource/polar/issues/877">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/polarsource/polar/issues/877/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

frankie567:
Should eventually fix #877

This PR grow quite big (sorry), but I think it makes sense as it's all related around the external issue synchronization issue. A few highlights:

## 1. Fix typing for `Organization.installation_id`

The property `installation_id` was typed as `int`, but it should be `int | None` as it's nullable (in the case of external organization).

Making this change was causing typing issues everywhere we used it assuming it was non-null. Some places however handled it with a local `Exception`.

To improve this, I introduce a property `safe_installation_id`, which always return an `int` or raise a `NotInstalledOrganization`. This satisfies the type checker and avoid to have manual logic everywhere we use this value.

## 2. Harmonize `create_or_update` methods for Organization, Repository and Issue

We had several flavors to create organization and repositories from GitHub data. I tried to harmonize this so there is one only clear path to perform this action.

It also systematically updates it, so we can better account for cases where an repository/issue has been transferred.

## 3. Implement a shared lock mechanism with Redis

I implemented a simple lock mechanism using built-in Redis tools. This may be handy in a few cases.

Right now, I've used it around the external issue synchronization, so we avoid the race conditions we met in #877.

> Side note: from my testing, the race condition can actually happen also on the web app. Next.js makes two successive calls to `/lookup` on the issue page. That's why I was interested in switching our client from `axios` to `fetch`.

---

That's about it! I really would like your feedback on this one, as it involves trickier changes.

### DIFF
--- a/server/polar/integrations/github/service/issue.py
+++ b/server/polar/integrations/github/service/issue.py
@@ -279,11 +279,8 @@ class GithubIssueService(IssueService):
         installation_id = (
             crawl_with_installation_id
             if crawl_with_installation_id
-            else org.installation_id
+            else org.safe_installation_id
         )
-
-        if not installation_id:
-            raise Exception("no github installation id found")
 
         client = github.get_app_installation_client(installation_id)
 
@@ -497,7 +494,7 @@ class GithubIssueService(IssueService):
         repository: Repository,
         issue: Issue,
     ) -> Issue:
-        client = github.get_app_installation_client(organization.installation_id)
+        client = github.get_app_installation_client(organization.safe_installation_id)
 
         labels = await client.rest.issues.async_add_labels(
             organization.name, repository.name, issue.number, data=["polar"]
@@ -514,7 +511,7 @@ class GithubIssueService(IssueService):
         repository: Repository,
         issue: Issue,
     ) -> Issue:
-        client = github.get_app_installation_client(organization.installation_id)
+        client = github.get_app_installation_client(organization.safe_installation_id)
 
         labels = await client.rest.issues.async_remove_label(
             organization.name, repository.name, issue.number, "polar"
@@ -631,11 +628,8 @@ class GithubIssueService(IssueService):
         installation_id = (
             crawl_with_installation_id
             if crawl_with_installation_id
-            else organization.installation_id
+            else organization.safe_installation_id
         )
-
-        if not installation_id:
-            raise Exception("no github installation id found")
 
         client = github.get_app_installation_client(installation_id)
 
--- a/server/polar/integrations/github/service/issue.py
+++ b/server/polar/integrations/github/service/issue.py
@@ -841,7 +841,7 @@ class GithubIssueService(IssueService):
             if r.private:
                 continue
 
-            org = await github_organization.update_or_create_org_from_github(
+            org = await github_organization.create_or_update_from_github(
                 session, r.owner
             )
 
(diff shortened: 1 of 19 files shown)


server/polar/integrations/github/service/issue.py
```diff
@@ -553,154 +549,55 @@ async def sync_external_org_with_repo_and_issue(
...
-                if e.response.status_code == 401:
-                    raise ResourceNotFound()
-                # re-raise other status codes
-                raise e
-
-            # check if we have org with same external_id
-            organization = await github_organization.get_by_external_id(
-                session, owner.id
+        try:
+            github_repository_response = await client.rest.repos.async_get(
```
zegl, line 559: Previously this function would use locally cached results by default, and would only fallback to syncing from GitHub if it was missing. Maybe the method name is to blame here for the confusion, but that would be a nice property to keep. :)

That's especially important when called with `client=get_polar_client()` as we don't have too many API credits for that user.

server/polar/organization/schemas.py
```diff
@@ -86,34 +86,31 @@ class OrganizationCreate(OrganizationPrivateBase):
...
+        if installation is not None:
+            data = {
+                **data,
+                "installation_id": installation.id,
+                "installation_created_at": installation.created_at,
+                "installation_updated_at": installation.updated_at,
+                "installation_suspended_at": installation.suspended_at,
+            }
+
+        return cls(platform=Platforms.github, **data)
```
zegl, line 113: 🤔 Instead of adding `Any` and `**` can we split this into two classes? One with installation data and one without. That way we can still be type-safe and maybe even remove `exclude_unset`?

zegl (commented):
This PR does three very different things (installation_id / locking / cleanup), the installation_id and locking changes look great, but I have some comments on the other cleanup.

Wanna split out the installation_id and locking changes to their own PRs so that we can merge those separately?

frankie567:
I'll split the PR!

server/polar/integrations/github/service/issue.py
frankie567: Will improve that!

server/polar/organization/schemas.py
frankie567: I kinda like to have a single method that can handle both, but definitely doable to do so with more type-safety.

The `exclude_unset` trick is actually not needed given how we structured our schema. I can remove this.

PR #1115, opened 2023-09-07
Refactor sync_external_org_with_repo_and_issue

frankie567:


frankie567:
_Originally posted by @zegl in https://github.com/polarsource/polar/pull/1098#discussion_r1318370255_

> Previously this function would use locally cached results by default, and would only fallback to syncing from GitHub if it was missing. Maybe the method name is to blame here for the confusion, but that would be a nice property to keep. :)
> 
> That's especially important when called with `client=get_polar_client()` as we don't have too many API credits for that user.

Totally getting the point!

However, the previous approach makes the code quite intricated. Besides, it would have needed further tweaks to account for renamed org/repo and moved issues. The good thing with the naive approach we have here is that it automatically handles this.

Here is what I was thinking to solve this: what about having an "index" linking an `(org_name, repo_name, issue_number)` tuple to an issue `external_id` (this could be in Redis or in a PostgreSQL table)?

This way, we could directly return our local copy if we already saw it. If there is a rename/move, we won't have it in index and thus perform a full-sync, allowing us to update our local copy.

Let me know what you think about this approach :)

server/polar/organization/schemas.py
```diff
@@ -86,33 +86,24 @@ class OrganizationCreate(OrganizationPrivateBase):
...
+            name=user.login,
+            external_id=user.id,
+            avatar_url=user.avatar_url,
+            is_personal=user.type.lower() == "user",
+            installation_id=installation.id if installation else None,
+            installation_created_at=installation.created_at if installation else None,
+            installation_updated_at=installation.updated_at if installation else None,
+            installation_suspended_at=installation.suspended_at
+            if installation
+            else None,
```
zegl: Is there a risk that we're now unsetting the `installation_` and other fields that default to null in the database?

Could we have a separate model for this other use case, and remove the fields from this model?

Like this maybe? **(incomplete example)**

```python
class OrganizationCreate(OrganizationPrivateBase):
    pledge_minimum_amount: int = settings.MINIMUM_ORG_PLEDGE_AMOUNT

    @classmethod
    def from_github(
        cls,
        user: github.webhooks.User | github.rest.SimpleUser,
    ) -> Self:
        return cls(
            platform=Platforms.github,
            name=user.login,
            external_id=user.id,
            avatar_url=user.avatar_url,
            is_personal=user.type.lower() == "user",
        )


class OrganizationInstallationUpdate(OrganizationCreate):
    installation_id: int
    installation_created_at: datetime
    installation_updated_at: datetime | None = None
    installation_suspended_at: datetime | None = None

    @classmethod
    def from_github(
        cls,
        user: github.webhooks.User | github.rest.SimpleUser,
        installation: github.rest.Installation,
    ) -> Self:
        return cls(
            platform=Platforms.github,
            name=user.login,
            external_id=user.id,
            avatar_url=user.avatar_url,
            is_personal=user.type.lower() == "user",
            installation_id=installation.id,
            installation_created_at=installation.created_at,
            installation_updated_at=installation.updated_at,
            installation_suspended_at=installation.suspended_at
            if installation
            else None,
        )
````

Comment (zegl) [reply server/polar/organization/schemas.py]: In general, my strategy has been to avoid schemas with nullable fields, and instead create multiple schemas. :)

Comment (frankie567) [reply server/polar/organization/schemas.py]: You're right, this would have reset `installation_` fields.

I've taken back the approach with `exclude_unset`. I really think it's a good way of managing this kind of cases. Having several separate schemas would have been hard to handle with the `.update` method.

**Comment by frankie567** [comment]
So, I introduced the "index" mechanism I described above. Instead of relying on Redis, it made more sense after second thought to actually store it in the DB under the `external_lookup_key` column. This way, we can just make good-old SQL requests.

This field is filled as part of the `from_github` methods, which is why I needed to pass full `organization` and `repository` objects instead of just their IDs.

The DB migration includes an `UPDATE` query so we can fill this column for our existing issues and PR.

File: server/migrations/versions/2023-09-08_add_external_lookup_key_column_to_issue.py [hunk 2]
```diff
@@ -0,0 +1,70 @@
...
+    )
+    op.create_index(
+        op.f("ix_issues_external_lookup_key"),
+        "issues",
+        ["external_lookup_key"],
+        unique=False,
+    )
+    op.add_column(
+        "pull_requests", sa.Column("external_lookup_key", sa.String(), nullable=True)
+    )
```
zegl: Do we really need to have external_lookup_key on the pull requests table? I don't think we ever look up PRs in this way.

Add `external_lookup_key` to `Issue` instead of `IssueFields` to only add it for issues.

server/polar/integrations/github/service/issue.py
```diff
@@ -550,154 +545,70 @@ async def sync_external_org_with_repo_and_issue(
...
-            log.info(
-                "repository not found by name",
-                organization_id=organization.id,
                 repo_name=repo_name,
+                issue_number=issue_number,
+                external_id=issue.external_id,
             )
+            issue = await self.get_loaded(session, issue.id)
+            assert issue is not None
+            return (issue.repository.organization, issue.repository, issue)
```
zegl, line 567: I just noticed that while this function returns `Tuple[Organization, Repository, Issue]`, the `Issue` is the only field that ever is used. A good idea for some cleanup  😄  (no need to do it in this PR)

server/migrations/versions/2023-09-08_add_external_lookup_key_column_to_issue.py
frankie567: Yeah, probably not needed, will remove it 👍

server/polar/integrations/github/service/issue.py
frankie567: Yes, I saw they were not used, but unsure if it was on purpose or not. Will do a follow-up PR to remove this.

server/polar/kit/services.py
```diff
@@ -160,12 +160,13 @@ async def update(
         update_schema: UpdateSchemaType,
         include: set[str] | None = None,
         exclude: set[str] | None = None,
+        exclude_unset: bool = False,
         autocommit: bool = True,
     ) -> ModelType:
         return await source.update(
             session,
-            include=include,
-            exclude=exclude,
```
zegl, line 168: This was the only usage of these parameters in `ActiveRecordMixin.update()`, it might be worth deleting them there as well.

zegl:
Ok, looks good, but `exclude_unset` still scares me. 😅 

Like how can a value be "unset" if it has a default value of `None`? But setting it explicitly to `None` is different from the default `None`?

frankie567:
> Ok, looks good, but `exclude_unset` still scares me. 😅
> 
> Like how can a value be "unset" if it has a default value of `None`? But setting it explicitly to `None` is different from the default `None`?

Yes! Pydantic makes the difference between a value that has been set manually and a value coming from default (whether it's `None` or any other value). It's very handy for partial updates, e.g.:

```py
class MyUpdateSchema(BaseModel):
    label: str | None = None
    number: int | None = None

update_schema = MyModelUpdate(label="foo")
update_dict = update_schema.dict(exclude_unset=True)
print(update_dict)  # {"label": "foo"} -> "number" is not there
```

In FastAPI endpoints this enables user to make update requests with this kind of payloads:

```json
{"label": "foo"}
```

So they can just update the only field they care about.

PR #1141, opened 2023-09-11
Implement a dedicated email service

frankie567:
* Move out the sender service to its own module
* Move out the Jinja renderer service to its own module
* Add a base template (taken from https://github.com/ActiveCampaign/postmark-templates) for email so we have a bit more polished experience (it supports dark mode!):

Capture d’écran 2023-09-11 à 11 42 10

Capture d’écran 2023-09-11 à 11 42 12


zegl:
LGTM! Did you discuss with @birkjernstrom about redesigning all emails? I don't remember how intentional the previous aesthetic was.

frankie567:
I did not, but yes, we could use @birkjernstrom's opinion. It's just a small wrapper to the existing emails, their content is untouched.

frankie567:
Now, with the logo! It'll be served by our Next.js app on Vercel, so we'll benefit from their CDN.

Capture d’écran 2023-09-12 à 09 08 51

Capture d’écran 2023-09-12 à 09 08 53

Issue #1052, opened 2023-09-03
Backer: Magic link sign in w. upgrade path to GitHub


birkjernstrom:
**Problem**
Upon funding an issue today, backers provide their email and can only upgrade to an account if they do so directly after the pledge on the "Thank you"-page via GitHub login. We should improve this to support backers to login with an account associated with their email using magic links. From there they can easily upgrade their account with GitHub in their settings and that we encourage them to do so via the dashboard.

Full design of the flow is in [Figma](https://www.figma.com/file/1p2oOjsRBgSrr3SC4zySJG/Polar-Latest-Design?type=design&node-id=222%3A6136&mode=design&t=2dP5iMGSceiaWzBp-1). Core flow illustrated below. 

![Image](https://github.com/polarsource/polar/assets/281715/f40eb29c-b827-4fdc-b455-860fc713cf46)
![Image](https://github.com/polarsource/polar/assets/281715/eda7d67b-67fd-40cb-bd0e-ef8e3308866e)
![Image](https://github.com/polarsource/polar/assets/281715/294a87a9-5006-418b-8ccf-8988e6638f89)
![Image](https://github.com/polarsource/polar/assets/281715/ea724d57-596c-47c5-8ec3-74206118085a)

**Considerations**
- We need to create users for each successful pledge. Basic user without GitHub connection.
- We need a migration script to generate such users for historical, successful, pledges (no abandoned checkouts)
- We need to make sure pledges for the same email are connected to a single user

**Todo**
_Not final. To be flushed out as we build _
- [x] Get || Create a user once a pledge is made tied to the given email (case sensitive/insensitive?)
- [x] Connect the pledge to the user
- [x] Ability to sign in using magic links and a verification code we send in the email too for additional security (Vercel-style)
   - [x] Link should expire within X minutes (30?)
   - [x] Link should only be able to be used once
- [x] Ability to connect GitHub to the account via settings
- [x] Automatically connect accounts if backer has an email account and later signs in with GitHub which has the same email address

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/polarsource) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog :^)

<a href="https://polar.sh/polarsource/polar/issues/1052">
<picture>
[...]

PR #1153, opened 2023-09-12
Magic link authentication backend logic

frankie567:
Backend logic for magic link authentication (see #1052)


server/polar/magic_link/endpoints.py
```diff
@@ -0,0 +1,33 @@
...
+
+from polar.tags.api import Tags
+from polar.kit.db.postgres import AsyncSession
+from polar.postgres import get_db_session
+from polar.auth.service import AuthService, LoginResponse
+
+from .schemas import MagicLinkRequest
+from .service import magic_link as magic_link_service
+
+router = APIRouter(prefix="/magic-link", tags=[Tags.INTERNAL])
```
zegl: 1) It would be nice to change the prefix to "magic_link" to be consistent with other endpoints.
2) The tag here should be something like "magic_link". That way a "MagicLinkService" will be automatically created when you're generating the API client. :-)

server/polar/magic_link/tasks.py
```diff
@@ -0,0 +1,35 @@
...
+    ctx: JobContext, magic_link_id: UUID, token: str, polar_context: PolarWorkerContext
+) -> None:
+    with polar_context.to_execution_context():
+        async with AsyncSessionLocal() as session:
+            magic_link = await magic_link_service.get(session, magic_link_id)
+
+            if magic_link is None:
+                raise MagicLinkNotFoundError(magic_link_id)
+
+            await magic_link_service.send(magic_link, token)
```
zegl: I think it would be a good idea to send the emails directly without a queue, that way users can still log in even if the worker is slow/down/whatever.

server/polar/magic_link/endpoints.py
frankie567: 1. Ok, changed. Even though I tend to prefer kebab-case for URL 🙃
2. Added!

server/polar/magic_link/tasks.py
frankie567: Ok, let's do this.

PR #1156, opened 2023-09-13
server/reward: set an order_by clause to RewardService.list

frankie567:
Prevents the test `test_list_rewards_to_user` to be flaky (the DB was not always returningt the Pledges in the same order)


frankie567:
@zegl A small fix for a flaky test in `test_reward`!

Issue #1137, opened 2023-09-11
Propagate log Correlation ID from server to worker task


frankie567:
Currently, we have a log Correlation ID that's generated when we start to process an API request or a worker task.

Frequently, worker tasks are triggered during an API request. Thus, it would be handy to propagate the Correlation ID generated for the API request to the corresponding triggered tasks. This way, we could easily track all the logs for a whole scenario.

To keep things consistent, I suggest that we still generate a new Correlation ID for each worker task execution. The ID coming from the API request should be added to the log as another field, like `request_correlation_id`.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/polarsource) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog :^)

<a href="https://polar.sh/polarsource/polar/issues/1137">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/polarsource/polar/issues/1137/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/polarsource/polar/issues/1137/pledge.svg">
</picture>
</a>

PR #1192, opened 2023-09-18
server/dx: implement `request_correlation_id` in tasks logs

frankie567:
We had to move out the `on_job_start`/`on_job_end` approach.

Unfortunately, we don't have access to task arguments in this hook.

This prevents us to implement things like common arguments handling,
as we do for `request_correlation_id`.

To circumvent this limitation, we implement this behavior
through the `task_hooks` decorator.


frankie567:
Fix #1137

PR #1198, opened 2023-09-18
server/dx: improve typing of Auth by using Annotated approach

frankie567:
I was a bit annoyed by the fact that `Auth.current_user` would return us a class instance typed with an optional user (`User | None`), while we *know* by using this constructor that the user is there and not `None`.

So, I've played a bit with typing and the new fancy [`Annotated` approach](https://fastapi.tiangolo.com/tutorial/dependencies/#share-annotated-dependencies) to help us improve this.

I've implemented this in a single endpoint for now. As we see, it helps us get rid of unneeded `if not auth.user` checks.

I would very like your feedback on this, @zegl, before generalizing it.

### DIFF
--- a/server/polar/account/endpoints.py
+++ b/server/polar/account/endpoints.py
@@ -2,7 +2,7 @@ from uuid import UUID
 
 from fastapi import APIRouter, Depends, HTTPException, Query
 
-from polar.auth.dependencies import Auth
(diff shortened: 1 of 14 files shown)


zegl:
This is really convenient, I like it. :-)

PR #1199, opened 2023-09-18
Badge: show avatars of pledgers

zegl:
### Summary

* Changes `Pledges` to be **public**. Some fields (refunded_at, scheduled_payout_at, hosted_invoice_url) are only shown to users with write access to the pledge.
* Added `pledge_badge_show_amount` to the public `Organization` object
* Show avatars of pledgers (max 4 shown) in the SVG badge

### DIFF
--- a/clients/packages/polarkit/src/components/badge/Badge.tsx
+++ b/clients/packages/polarkit/src/components/badge/Badge.tsx
@@ -19,11 +19,13 @@ export const Badge = ({
   // amountRaised = undefined,
   darkmode = false,
   funding = undefined,
+  avatarsUrls = [],
 }: {
   showAmountRaised?: boolean
   // amountRaised?: string
   darkmode: boolean
   funding?: Funding
+  avatarsUrls: string[]
 }) => {
   const showFundingGoal =
     funding &&
@@ -42,6 +44,10 @@ export const Badge = ({
           1, // Min 1
         )
       : 0
+
+  const showAvatars =
+    avatarsUrls.length > 4 ? avatarsUrls.slice(0, 3) : avatarsUrls
+  const extraAvatarsCount = avatarsUrls.length - showAvatars.length
 
   return (
     <>
@@ -85,7 +91,7 @@ export const Badge = ({
               flexShrink: '0',
             }}
           >
-            Fund this issue
+            Fund
           </div>
 
           {showAmount && funding?.pledges_sum?.amount !== undefined && (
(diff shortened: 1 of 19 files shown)


clients/packages/polarkit/src/components/badge/Badge.tsx
```diff
@@ -211,6 +220,56 @@ export const Badge = ({
...
+                    backgroundColor: darkmode ? '#2e4070' : '#C9DBF4',
+                    color: darkmode ? '#a6c7ea' : '#4667CA',
+                    height: 22,
+                    width: 22,
+                    borderRadius: 22,
+                    marginLeft: '-6px',
+                    textAlign: 'center',
+                    fontSize: '8px',
+                    lineHeight: '20px',
+                    border: darkmode ? '1px solid #374e96' : '1px solid white',
```
birkjernstrom: Let's try to change the dark mode border to "#2e4070" same as the background. So it's more of a carve-out vs. border as is the case in the light mode.

birkjernstrom:
Great work! Super excited to get this in-place (in hindsight, I can't believe we didn't have it from the get go).

Bit late so I focused the review now on the functionality & design vs. code implementation. I assume that's what was mainly requested too :-)

Just two things on that regard:
1. Let's change the border in dark mode to be the same gray as the background. Similar to how we're using white in the light version. So it looks more like a carve-out vs. actual border with the blue.
2. Dumb question, but for peace of mind: We're not exposing the email of the backer anywhere in this public API, right? Except potentially in other contexts if the request is from someone with write access?

In conversation today we discussed making the amount hidden. I've changed my mind. Since pledges are made public/social the amount is pseudo public by extension. Since an issue with one backer will automatically disclose the amount and for followers of it, it will be fairly easy to track the first 2-3 as well. So it's better that we have strict modes of public and anonymous later (if requested, e.g companies might be reasonable at certain stages/amounts). Instead of having subtle in-between nuances that will likely only serve to create confusion and mishaps.

zegl:
> 2. Dumb question, but for peace of mind: We're not exposing the email of the backer anywhere in this public API, right? Except potentially in other contexts if the request is from someone with write access?

Nope, no emails are exposed. The `Pledger` object looks like this:

```python
class Pledger(Schema):
    name: str
    github_username: str | None
    avatar_url: str | None
```

clients/packages/polarkit/src/components/badge/Badge.tsx
zegl: Nice idea, I guess you mean #3E3F43 in that case :)

Issue #1092, opened 2023-09-06
Handle GitHub renamed and moved orgs/repos/issues


zegl:
This will be an ongoing effort, but wherever possible, start to use lookups on external_id instead of names (orgs/repos) and numbers (issues) to support moved resources.

If we encounter a resource that has been moved on GitHub's side of things, also move it on our side.

- [x] Renamed organizations
- [x] Renamed repositories
- [x] Moved repositories (new organization as owner). It's possible for an existing repository (in our DB) to be moved away from an installation to an org that doesn't have the Polar app installed, and vice versa.
- [x] Moved issues (new repo/organization as owner). When an issue is moved between repositories, the issues #number changes.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/polarsource) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog :^)

<a href="https://polar.sh/polarsource/polar/issues/1092">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/polarsource/polar/issues/1092/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/polarsource/polar/issues/1092/pledge.svg">
</picture>
</a>

frankie567:
I've started to work on this in the context of `sync_external_org_with_repo_and_issue`. Will be part of the fix for #877.

frankie567:
The problem was encountered by a user who reported it on Discord (https://discord.com/channels/1078611507115470849/1109142435981967360/1153512562537812071).

I'll start to work on this today. Objectives:

* Handle webhook events for Renamed Organization / Transferred Repository / Transferred Issue (Renamed Repository is already handled).
* Create a migration script to fix outdated data

frankie567:
I propose to close this for now. We have the basic bricks to handle this correctly. There are probably lot of corner cases we didn't think about, but I think they can be handled on case-by-case basis with a dedicated bug report.

PR #1207, opened 2023-09-19
Handle rename and transfer events from GitHub

frankie567:
Related to #1092

This PR introduces handlers for GitHub events to better handle when:

## Organizations are renamed

Quite straightforward; we just need to update the name in our database.

## Repositories are transferred

A bit more involved!

It's worth to note that when a repository is transferred, we also receive the events:

* `installation_repositories.removed` from the old organization.
* `installation_repositories.added` from the new organization.

It means that we already had most of the logic to handle it correctly. However, the effect of `installation_repositories.removed` was to *soft-delete* the repository.

Thanks to the `repositories.transferred` event, we are able to "restore" it, and make sure it's correctly linked to the new organization.

### Corner cases

The happy path is when *both organizations* have installed Polar on *all repositories*.

For the other cases, the user will need to go through their GitHub Polar app settings to give access to the new organization/repository. 

## Issues are transferred

That's the boss fight!

It turns out that, when you transfer an issue, GitHub actually **creates** a new issue on the new repository. The `external_id` is **not the same**!

Thus, it triggers the `issues.added` event on the new repository, which we correctly handle. Now, we have the same issue twice in two different repositories.

By handling the `issues.transferred` event, we are able to be notified of the old and new one.

What we do here is to *transfer* all the meaningful properties of Polar (like funding goal, custom badge message...) and associated Pledges, IssueRewards and Notifications.

> **Warning**
> In the future, we shouldn't forget to update this transfer procedure so we don't miss things.

Finally, we soft-delete the old one. Phew! We did it!

> **Note**
> GitHub only allows issues transfer  **within an organization**. It simplifies a lot of things for us since we're sure the organization is installed.


zegl (approved):
Nice, looks nice and well thought out! Good to see some tests on this as well, as it's quite annoying to test manually. :)

PR #1208, opened 2023-09-19
clients/promotion: start adding tabs to funding modal

zegl:


zegl:
The badge is not updated yet, but it's coming very soon in #1211

Issue #1210, opened 2023-09-20
`/pledges/search` public API leaks too much data


frankie567:
The `/pledges/search` API doesn't require authentication anymore and is not limited to the pledges user has access to. Thus, it's way too easy for anybody to read data about any pledge. For example:

https://api.polar.sh/api/v1/pledges/search?organization_name=frankie567&platform=github

---

IMO, we should keep this endpoint authenticated and limited to the authenticated user pledges.

To be able to get data about an issue's pledges, we probably should come up with a dedicated public endpoint, like `/issues/{id}/pledges` which should only return basic data like amount, pledger login and avatar.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/polarsource) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog :^)

<a href="https://polar.sh/polarsource/polar/issues/1210">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/polarsource/polar/issues/1210/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/polarsource/polar/issues/1210/pledge.svg">
</picture>
</a>

Issue #1201, opened 2023-09-18
Implement updated design of issue page


birkjernstrom:
The left-hand side of this updated design:
<img width="865" alt="Screenshot 2023-09-18 at 15 43 17" src="https://github.com/polarsource/polar/assets/281715/cb0de75c-8c6c-46bb-afe9-a7f47af5b959">

This design reflects some features in the works & some added data that we can progressively enhance later.

**Todo**
- [ ] Design empty states, e.g no funding.

**First iteration**
- Issue title & meta data
- Funding & backer avatars
- Org & repo details

**Progressive enhancements (later)**
- #1125 
- https://github.com/polarsource/polar/issues/1200

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/polarsource) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog :^)

<a href="https://polar.sh/polarsource/polar/issues/1201">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/polarsource/polar/issues/1201/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/polarsource/polar/issues/1201/pledge.svg">
</picture>
</a>

frankie567:
We display the number of 👍 under the title:

<img width="500" alt="Capture d’écran 2023-09-20 à 10 52 25" src="https://github.com/polarsource/polar/assets/1144727/9b912024-3af3-4cc5-a952-58b276aa05af">

Is it:

* The number of 👍 *only*?
* The sum of *all reactions*?
* Or, should we show each reactions (👎, 🚀, 🎉...) separately?

emilwidlund:
🥳

Issue #1200, opened 2023-09-18
Show issue description in our issue page


birkjernstrom:
<img width="865" alt="Screenshot 2023-09-18 at 15 43 17" src="https://github.com/polarsource/polar/assets/281715/c93a2916-fce2-452a-ad6c-bf44bd9e0b73">

Great for two purposes:
1. All context in one-place (better conversion)
2. Capture SEO with context of funding
3. Bonus: Looks nice and more professional

How it's supposed to work:
- Cutoff the description after X amount of words. So it's around the length of what is seen in the design
- Mimic the markdown formatting

However, the last part can become incredibly tricky and we should progressively enhance it over time. E.g code highlighting, todos etc. For starters I think we can begin with a whitelist of only a few support types of markdown and filter out everything else. Cutting up until we reach the first point of non-supported markdown, e.g code example.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/polarsource) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog :^)

<a href="https://polar.sh/polarsource/polar/issues/1200">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/polarsource/polar/issues/1200/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/polarsource/polar/issues/1200/pledge.svg">
</picture>
</a>

frankie567:
Idea for the Markdown rendering problem: apparently, GitHub has an API for that: https://docs.github.com/en/rest/markdown/markdown?apiVersion=2022-11-28

Might be worth to check it out and see what's possible to do with it. If it works, we can cache the result on DB/Redis to save our token rate limits.

birkjernstrom:
Oh wow that's awesome! Yes we should definitely leverage that to avoid reinventing the wheel + making it 1:1 with the source.

Yes, by using app installations for it upon issue updates/creation and overwriting a cache with no TTL - I don't think we'll ever get close of running into ratelimit issues (famous last words).

This made me excited ☺️

PR #1212, opened 2023-09-20
Implement new design of issue page's left part

frankie567:
I've implemented the first bits of the new issue presentation on the Pledge page.

Related to #1201 and #1200 

## Screenshots

Capture d’écran 2023-09-21 à 15 04 03

Capture d’écran 2023-09-21 à 15 04 10

## What we have ✅

* General presentation of the issue (title, number, author, number of comments and reactions)
* Current pledge and optional goal
* Rendered issue description ✨
* Repository details with number of stars
* Optional upfront rewards split notice

## What we miss compared to the design 😞

* Pledgers avatars
    * Component is ready, but still unsure about the API to retrieve them (cf our discussions related to #1210)
* List of rewarded people
    * Feature is not yet implemented AFAIK?
* List of linked PR 
    * We should implement an API to retrieve this data for an issue

## Next steps

What I propose is to merge and deploy those changes, even if we have some holes compared to the design, and address them in follow-up improvements.

The page looks already quite good and it can be a good first-step. What do you think @birkjernstrom?

### DIFF
--- a/clients/apps/web/src/app/(pledge)/[organization]/[repo]/issues/[number]/(create)/page.tsx
+++ b/clients/apps/web/src/app/(pledge)/[organization]/[repo]/issues/[number]/(create)/page.tsx
@@ -71,11 +71,14 @@ export default async function Page({
   params: { organization: string; repo: string; number: string }
 }) {
   let issue: Issue | undefined
+  let issueHTMLBody: string | undefined
 
   try {
-    issue = await authedApi().issues.lookup({
+    const api = authedApi()
+    issue = await api.issues.lookup({
       externalUrl: `https://github.com/${params.organization}/${params.repo}/issues/${params.number}`,
     })
+    issueHTMLBody = await api.issues.getBody({ id: issue.id })
   } catch (e) {
     if (e instanceof ApiError && e.status === 404) {
       notFound()
@@ -88,7 +91,7 @@ export default async function Page({
 
   return (
     <>
-      <Pledge issue={issue} gotoURL={undefined} />
+      <Pledge issue={issue} htmlBody={issueHTMLBody} gotoURL={undefined} />
     </>
   )
 }
--- a/clients/apps/web/src/components/Organization/RepositoryPublicPage.tsx
+++ b/clients/apps/web/src/components/Organization/RepositoryPublicPage.tsx
@@ -1,5 +1,6 @@
 import { Issue, Organization, Repository } from 'polarkit/api/client'
-import { abbrStars, prettyURL } from '.'
+import { formatStarsNumber } from 'polarkit/utils'
+import { prettyURL } from '.'
 import HowItWorks from '../Pledge/HowItWorks'
 import Footer from './Footer'
 import Header from './Header'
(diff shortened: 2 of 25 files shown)


birkjernstrom:
Awesome! Going to be really awesome to have this additional data, updated design and now with all the added functionality as well. Overall, I love it and great work! Below are some specific and smaller feedback to make it shippable and :chef-kiss:

1. Can we use our markdown component for the title to ensure backticks are styled and that the backticks themselves don't show? E.g `example`
2. Same for the issue body. They seem to be styled with bold at least (which is sufficient), but can we get rid of the backticks?
3. Why is the block that we show in dark mode in the light theme? Can we only get one mode and that it's then cached I assume? If so, let's cache one for each mode to ensure it's light in the light theme.
4. Margin between issue title and the author is too little. I have 10px in Figma, but that might be to little too. But let's get some space in-between there
5. Funding amount & goal progress bar are too big. I know they are in the design too, but it became more apparent now. Can we try reducing them by 50%. At least the progress bar to 3px in height. Potentially that the funding amount only needs to be reduced down to 18px

Can you also share a screenshot of what this looks like on mobile? I think we might potentially need to get rid of the issue body on mobile to avoid too much scrolling. Potentially just the title, author, funding and then the funding section.

frankie567:
Thanks for the detailed feedback, @birkjernstrom 🙏  I'll tackle those points!

Regarding **2** and **3**, this is actually because of [Tailwind Typography plugin](https://tailwindcss.com/docs/typography-plugin). Its purpose is to put default styling for content you don't own (which is exactly our case here). So, that's why you have bold titles, and code blocks on dark background. Those can be tweaked — to a certain extent. I'll try to work on your specific points.

Regarding mobile, I tried to work on responsiveness so it should look okay, but yeah, the issue body might take lot of spaces. Here it is (iPhone XR mode):

<details>
  <summary>Open screenshot</summary>
  
</details>

birkjernstrom:
This is great and glad we did the small tweaks - looks better.

Just some small things left on mobile and then this is ready to ship.

1. Hide issue body
2. Hide rewards. I think mobile can be more focused on funding vs. contribution and rewards with the added context. At least for now. Want to save as much space as possible so that the funding module is above the fold to make things clear right off the bat. 
3. Let's left align the org details on mobile. We can hide the number of stars as well I think.

Finally, I noticed there is a lot of whitespace on mobile in the payment form. I assume this might be a timing issue of when the screenshot was taken and that it's Stripe being loaded there so that there isn't any whitespace in reality?

frankie567:
Done!

<details>
    <summary>Mobile light mode</summary>
    
</details>

<details>
    <summary>Mobile dark mode</summary>
    
</details>

> Finally, I noticed there is a lot of whitespace on mobile in the payment form. I assume this might be a timing issue of when the screenshot was taken and that it's Stripe being loaded there so that there isn't any whitespace in reality?

Yes, probably, it corresponds to size of the CC form. Doesn't happen in the screenshot I took today

birkjernstrom:
Great! Two final things that I noticed now. Actually outside the scope of this PR, but part of the funding form. Would be great if you could fix them while you're at it:
1. Add margin to the funding amount summary after the country dropdown. They are glued together now, but we should have the same spacing between them as everything else.
2. "Save credit card on file". Moving the checkbox up next to the label. I think we can remove the label underneath. For our audience I don't think it adds much and only takes up space. Same here re: margin to the pledge button underneath too. 

Those are the final things I can see. Then we can ship this which is going to be awesome! :)

PR #1363, opened 2023-10-05
Polar SDK

emilwidlund:
Creates an SDK package which houses the generated API client.


zegl:
@emilwidlund The build is failing 💔

PR #1503, opened 2023-10-18
A first polish pass on subscriptions

emilwidlund:
A first polish pass on the new subscriptions UI

### DIFF
--- a/clients/apps/web/src/components/Subscriptions/SubscriptionGroupIcon.tsx
+++ b/clients/apps/web/src/components/Subscriptions/SubscriptionGroupIcon.tsx
@@ -4,11 +4,13 @@ import { useMemo } from 'react'
 interface SubscriptionGroupIconProps {
   icon: string
   color: string
+  size?: number
 }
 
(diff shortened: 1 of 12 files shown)


clients/apps/web/src/components/Subscriptions/SubscriptionGroupIcon.tsx
```diff
@@ -36,7 +38,7 @@ const SubscriptionGroupIcon: React.FC<SubscriptionGroupIconProps> = ({
       className={`inline-flex items-center text-[--var-icon-color]`}
       style={style}
     >
-      <IconComponent className="!h-4 !w-4" />
+      <IconComponent className={`!h-${size} !w-${size}`} />
```
zegl: Sadly, the dynamic types here will not play well with the Tailwind JIT compiler.

It scans the sources at _compile time_, not runtime. It might work out of luck if another component specified the same height/width, but if it uses a size that isn't used anywhere else the class will not be generated.

emilwidlund: 🥇 for that feedback!

PR #1510, opened 2023-10-19
Adjust Subscription pages to support TierType instead of Group

emilwidlund:
Refactors the Subscription UI to support the new TierType approach instead of the old SubscriptionGroup API.


emilwidlund:
@frankie567 Mind taking a look at this refactor? It seems to be working; but I just want a second opinion on how I approached this here :)

clients/apps/web/src/components/Subscriptions/SubscriptionGroupIcon.tsx
```diff
@@ -17,26 +20,28 @@ const SubscriptionGroupIcon: React.FC<SubscriptionGroupIconProps> = ({
    * so we could set any icon dynamically.
    */
```
frankie567: This comment doesn't make sense anymore since we have a fixed enum :)

frankie567:
LGTM, great job 🙏

clients/apps/web/src/components/Subscriptions/SubscriptionGroupIcon.tsx
emilwidlund: Aye! Good catch!

PR #1513, opened 2023-10-19
server/account: fix exists check for user-connected accounts

### LINKED ISSUE #1512: Accounts can't be setup both for Backer Rewards and Personal Organization profile
Related to #1407 and to the problems @aeworxet encountered (Discord: https://discord.com/channels/1078611507115470849/1151973214562558085/1164488962363117599)

---

Accounts were designed for maintainer's organization. As such, `organization_id` is mandatory and it has a unique constraint on it.

Since Rewards, backers can also now configure an Account. We added `user_id` column for this.

However, it causes issues if the user already configured an account for their personal organization (as maintainer): indeed, because of the `organization_id` unique constraint, we can't create another account for the same one. The Rewards panel don't detect it since `user_id` is not set.

We should probably make `organization_id` optional and drop the unique constraint.

## Sub-problem

`Account` are not deleted when the organization is uninstalled. We should probably remove them also.

zegl:
Fixes #1512

### DIFF
--- a/server/polar/account/service.py
+++ b/server/polar/account/service.py
@@ -74,7 +74,9 @@ class AccountService(ResourceService[Account, AccountCreate, AccountUpdate]):
                 "both organization_id and user_id is set, this is not supported"
             )
 
-        existing = await self.get_by(session=session, organization_id=organization_id)
+        existing = await self.get_by(
+            session=session, organization_id=organization_id, user_id=user_id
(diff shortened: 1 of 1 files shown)


zegl:
What do you think about this as the fix @frankie567?

frankie567:
LGTM! Thanks 🎉

PR #1519, opened 2023-10-19
clients/web: server-side redirection for logged-in/out users

frankie567:
The idea is to leverage Next.js server-side redirects to automatically redirect the user if they are logged-in or not.

The good thing is that it'll save us some "flashing" states when accessing login page while authenticated and vice versa.

Ultimately, this might allow us to remove the `Gatekeeper` component.

As a bonus, when accessing `polar.sh`, the user is automatically redirected to `polar.sh/feed` is they are logged in (not sure if desired, but I personally like the behavior. Can remove it otherwise).

---

To make this work, we simply check for the existence of our session cookie, `polar`, which I made a root domain cookie (see https://github.com/polarsource/polar/commit/11cac1d451cb935a87defe739241b69c3dfbbb65)

### DIFF
--- a/clients/apps/web/next.config.js
+++ b/clients/apps/web/next.config.js
@@ -22,10 +22,16 @@ const nextConfig = {
         source: '/',
         destination: 'https://splendid-help-401117.framer.app/',
         missing: [
+          // Do not rewrite for polar.new, polar.new has another rule below
           {
             type: 'host',
-            value: 'polar.new', // do not rewrite for polar.new, polar.new has another rule below
+            value: 'polar.new',
           },
+          // Do not rewrite if user is logged in, they'll be redirected to the app
+          {
+            type: 'cookie',
+            key: 'polar',
+          }
         ],
       },
       {
@@ -68,6 +74,7 @@ const nextConfig = {
   },
   async redirects() {
     return [
+      // dashboard.polar.sh redirections
       {
         source: '/',
         destination: '/login/init',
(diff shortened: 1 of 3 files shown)


frankie567:
Would gladly have your call on this, guys 😄

clients/apps/web/next.config.js
```diff
@@ -91,6 +98,54 @@ const nextConfig = {
         permanent: false,
       },
 
+      // Logged-out user redirection
+      {
+        // Paths that shall remain public
+        source: '/((?!_next/static|_next/image|favicon.ico|favicon.png|login).*)',
```
zegl: This is great! Really nice idea!

There is a few more pages that can handle logged out users:

* `/signup/maintainer`
* `/github/installation` (supports both logged in and logged out users)
* `/github/session`

and maybe even a few more? What about the magic link pages?

zegl: Also all `/api/`, `.well-known/`, `/embed/`, and `/og/` routes

zegl: `/new/`

zegl: And all public user pages, the pledge pages, etc.

zegl: It might actually be easier to add a list of pages here that are for logged in users only than doing the opposite. That way we don't need to list all pages in the route, as we would handle it client side of we miss to configure a page.

frankie567: Good point 😄 I didn't think we had so many public pages!

zegl: It was more than I thought too! :D

frankie567: Much cleaner like this indeed!

zegl (approved):
This looks very nice

emilwidlund:
> This looks very nice

Indeed! Good job @frankie567 🥳

emilwidlund (approved):

PR #1538, opened 2023-10-22
Update checkout amount appropriately

### LINKED ISSUE #1537: $20 default persists even when the user picks a larger value
I tried to sponsor a couple issues for $50 and it says that I sponsored them for $20 still.

Steps to reproduce:

1. Paste an issue into the top of the feed to fund it: https://polar.sh/feed
2. Click "continue"
3. Change the amount in the input field with "20" to something larger.
4. Select your payment method and click "Fund this issue".
5. Look at your feed again. It will probably say you are funding the issue for $20.

birkjernstrom:
Fixes #1537

### DIFF
--- a/clients/apps/web/src/components/Pledge/PledgeForm.tsx
+++ b/clients/apps/web/src/components/Pledge/PledgeForm.tsx
@@ -556,7 +556,6 @@ const FundToday = ({
     }
     setFormState(n)
     debouncedSync(n)
-    console.log('changed', { ...n })
   }
 
   return (
--- a/clients/apps/web/src/components/Pledge/PledgeForm.tsx
+++ b/clients/apps/web/src/components/Pledge/PledgeForm.tsx
@@ -315,7 +315,7 @@ const FundToday = ({
         },
       })
     },
-    [],
+    [issue.id],
   )
 
   const updatePaymentIntent = useCallback(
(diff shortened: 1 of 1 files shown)


birkjernstrom:
Unfortunate that the dependencies array causes prettier to update the styling of the function signature – causing additional indentation of the entire function body and an unnecessary large diff. Only the dependencies array are changed in reality.

frankie567:
@birkjernstrom Does this mean the bug was caused because of a missing dependency in the hook array?

If so, it could be nice to add the `react-hooks/exhaustive-deps` ESLint rule to avoid this in the future.

zegl (approved):

birkjernstrom:
Thanks for merging & shipping @zegl :)

@frankie567 Yes, that's what caused it. Oh, nice I didn't know about that lint rule. I think that would be nice to have. However, I think it likely would complain in a few areas today even though it doesn't create a bug at all times, e.g certain object attributes being used for values to API calls (`issue.id`) which doesn't change, but the broader object itself might causing the effect to run again. Of course, this can be solved and perhaps it's worth it, but likely would require some review and refactoring to get right upon introducing such a linting flag.

frankie567:
@birkjernstrom I've PR this in #1540. I don't think we need to overthink it, it's a warning by default that helps us to not forget dependencies. If it's not sensible in a specific context, we can line-disable it or simply... Ignore it.

birkjernstrom:
@frankie567 Nice and agreed. Provided we simply get a warning vs. error it feels like a no-brainer and nice addition and sanity check for the future.

PR #1561, opened 2023-10-24
clients/sdk: add anyOf support to OpenAPI generator

frankie567:
So, I digged up a bit and found that we can [actually override templates locally](https://openapi-generator.tech/docs/templating), so it was a good start to try out things.

It turns out that the generator already has all the required metadata to make it work, it's "just" a matter of adapting the templates. I took inspiration from what was done here: https://github.com/OpenAPITools/openapi-generator/pull/14241

And... It works! There are probably edge cases, but it looks to work pretty well for basic union types. Here is what it outputs for a `Union` type:

```ts
/**
 * @type SubscriptionBenefitCreate
 * @export
 */
export type SubscriptionBenefitCreate = SubscriptionBenefitBuiltinCreate | SubscriptionBenefitCustomCreate;
```

### DIFF
--- a/clients/packages/sdk/generate
+++ b/clients/packages/sdk/generate
@@ -16,6 +16,7 @@ docker run \
     openapitools/openapi-generator-cli:v7.0.1 \
     generate \
     -i /sdk/openapi/updated.json \
+    -t /sdk/openapi_templates \
     -g typescript-fetch \
     -o /sdk/src/client \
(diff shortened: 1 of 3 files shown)


zegl:
Very nice! 🥇 I guess we have to be a bit cautions when using unions, and check that the generated code makes sense. The good thing is that it's generating regular TS, so that should be easy to verify!

PR #1565, opened 2023-10-24
clients/reposelection: fix bug with misaligned repo picker popover

emilwidlund:
Makes sure that we relative-position repository picker popover.

image


zegl:
I just noticed this bug as well, merging...

PR #1637, opened 2023-11-01
WIP: SQLAlchemy dataclasses

zegl:
With data classes we get type checking for the model constructors.


zegl:
😬 I need to find a way to do this more gradually. The good thing is that this finds _a lot_ of bugs in our code!

For example bugs like this one:

```python
    user_id: Mapped[UUID] = mapped_column(
        PostgresUUID, ForeignKey("users.id"), nullable=True
```

Where the column is marked as nullable in the database, but not in our types! And since it's not in our types, we have no guarantee that we handle cases where the column in a good way (most likely it will throw).

zegl:
A good way to do it gradually could be to:

1. Migrate relationships to `@declared_attr` class methods (can be done per-model, one by one)
2. Migrate from `ActiveRecordMixin.create()` to `Model(...).create()`. (type safe!). Can also be done one by one.
3. Enable `MappedAsDataclass` (needs to be done everywhere at once?)

PR #1713, opened 2023-11-13
server/badge: remove onboarded_at check

zegl:
The onboarded_at check was effectively also an installation_id check. Adding one to check that the GitHub App is installed in it's place.


frankie567:
Great catch! Let's go 🚀

Issue #1807, opened 2023-11-28
Merge personal maintainer and contributor `Account`


frankie567:
Currently, if you're both a maintainer with a personal account (e.g. @frankie567) *and* a contributor who wants to get rewards, you'll need to configure two accounts (Stripe or whatever).

Following the work on the new Finance page, we decided it would be clearer to actually have one for both usage.

It'll require to:

* [ ] Revisit the model and the several places where we need to fetch the user's account
* [ ] Migrate "dual accounts" user to use only one account

PR #1907, opened 2023-12-07
Make Account API return only one account per user/organization

frankie567:
We used to have a `/search` API for Account which returned a list of accounts for a given user/organization.

Since `Account` have a unique constraint on `user_id`/`organization_id`, it makes more sense to use a `/lookup` which return a single account or 404. It greatly simplifies the reasoning in the UI.

Preliminary work for #1807.


emilwidlund:
Nice work on this @frankie567 🔥

PR #1977, opened 2023-12-14
clients/web: fix flashing error message on GitHub OAuth callback

### LINKED ISSUE #1914: Attempting to connect a GitHub account which already is linked with another account results in a vague error
### What

If you attempt to connect a GitHub account which already is linked with another account, you'll get a vague error which hints at an issue with the access token. We should handle this better.

image

### Proposal

We should hint to the user that the account already is linked to someone else.

frankie567:
Fix #1914


emilwidlund:
Perfect! 🚀

PR #1997, opened 2023-12-15
Revamp GitHub OAuth and magic link authentication

frankie567:
Let's use server-side redirections to handle authentication instead on relying the browser to make the requests


server/polar/magic_link/endpoints.py
```diff
@@ -1,36 +1,46 @@
...
     magic_link, token = await magic_link_service.request(
         session, magic_link_request, source="user_login"
     )
-    await magic_link_service.send(magic_link, token)
+    base_url = str(request.url_for("magic_link.authenticate"))
+    await magic_link_service.send(magic_link, token, base_url)
 
 
-@router.post("/authenticate", tags=[Tags.INTERNAL])
+@router.get("/authenticate", name="magic_link.authenticate", tags=[Tags.INTERNAL])
```
zegl, line 45: Just FYI: This API is now used from `middleware.ts` as a part of pre-authenticated links in emails.

It might be worth splitting the endpoint into two if the behaviour is changing.

frankie567: I just noticed this, I'll have a look.

PR #2005, opened 2023-12-18
clients/web: fix Markdown language detection for code blocks

frankie567:


emilwidlund:
@frankie567 I think you just broke the inline code fix @zegl made last week :D 

image

frankie567:
Ah I see. But syntax highlighting was actually broken 🙃

PR #2178, opened 2024-01-09
server/stripe: cut payment_intent_succeeded event handling if subscri…

frankie567:
…ption invoice


zegl:
Seems like this PR was created during the GH outage yesterday, so the Actions where never triggered properly.

I rebased and force-pushed to trigger actions.

PR #2339, opened 2024-01-26
chore: use semantic HTML for LongFormPost

isaacharrisholt:
This is a minor change that makes the `LongFormPost` component use semantic HTML tags. There's more that could be done (e.g. using the `<footer>` tag), but it's somewhat of a start.

This should make posts easier for both screenreader users and crawlers to parse!

### Notes

I tried to make the `as` property of the `StaggerReveal` component take in any HTML element, but this caused some weird type issues that I'm not familiar enough with React/Framer Motion to fix.

It's also worth noting that I wasn't able to run a local development environment, and codespaces were giving me issues, so it's entirely possible that this is completely borked. Sorry about that 😅

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -128,3 +128,6 @@ playground.py
 
 .DS_Store
 .pnpm-store
+
+# Secrets
+.secrets/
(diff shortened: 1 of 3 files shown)


birkjernstrom:
😍 Thank you so much @isaacharrisholt. This looks great; I unfortunately have a family event now that I need to run to, but will test/review this tonight/weekend to merge asap :) Thank you again!

frankie567:
Tested it locally and can confirm it perfectly works. Chrome reader mode is able to correctly displays the content of the post (which was not the case before!).

Merging now, thanks @isaacharrisholt for spotting this! Great addition 😄

PR #2378, opened 2024-01-30
GitHub Repository benefit

frankie567:


clients/apps/web/src/app/(topbar)/(backer)/subscriptions/ClientPage.tsx
```diff
@@ -231,6 +232,30 @@ interface BenefitContextWidgetProps {
   subscription: SubscriptionSubscriber
 }
 
+const GitHubRepoWidget = ({
```
frankie567, line 235: @birkjernstrom Where is that component supposed to be? Because I already implemented a similar one with a button to login with GitHub if needed: 
https://github.com/polarsource/polar/blob/d17ca20b5266dc1cb7b84954c61e31c197a99b9f/clients/apps/web/src/components/Benefit/BenefitRow.tsx#L80-L86

PR #2431, opened 2024-02-01
Post keyboard shortcuts with meta-key only vs. ctrl

### LINKED ISSUE #2429: Ctrl-e doesn't work in editor
Ctrl-a works to get to the beginning of the line, but Ctrl-e is bound to creating backticks, e.g \` \`. So we're overwriting a OS-wide text input keybinding currently.

Thanks for @eval for sharing this feedback ♥️🙏🏼

birkjernstrom:
Fix #2429


frankie567:
This will likely break shortcuts on Windows, where the meta key is the Windows key.

PR #2436, opened 2024-02-03
fix: fix PostToolbar overlaying ProfileMenu

isaacharrisholt:
Fixes this in the post editor:

image

After: 

image

---

Note: I also ran `pnpm run format`, which changed some unrelated files. The only file with code changes is `PostToolbar.tsx`.


emilwidlund:
Thanks @isaacharrisholt 👍

PR #2461, opened 2024-02-06
clients/posts: clarify post clickability

### LINKED ISSUE #2437: Premium posts are not clickable
Causing a bit of confusion. Instead of allowing the post to load with the more clear paid subscription upsell.

zegl:
Make sure that we're not using "cursor: pointer" on posts when it's not clickable.

The post div requires JS to be clickable, and you would previsouly see the pointer cursor before document.ready had fired, even if you couldn't click the whole block yet.

Switching to controlling the cursor styling with JS to match the expected result.

Fixes #2437


emilwidlund:
👏

PR #2524, opened 2024-02-11
chore: add  to unsubscribe dropdowns

### LINKED ISSUE #2511: Chore: unsubscribe tooltip should probably have `cursor-pointer`
image

isaacharrisholt:
Fixes #2511 

---

Note: I wasn't sure whether to just update the 'Unsubscribe' dropdown items, or whether all dropdown menu items should have `cursor-pointer`, so I did the former. Happy to update so that the `DropdownMenuItem` in `polarkit` uses the class by default.


zegl:
Thanks!

PR #2525, opened 2024-02-11
fix: don't show upsell to self

### LINKED ISSUE #2510: Bug: you can subscribe to yourself
I managed to subscribe to myself from one of my articles

isaacharrisholt:
Fixes #2510 

---

Might also be worth implementing some sort of check on the backend too. I had a look through the API, but I'd like to better understand how you folks would like this implemented


zegl:
Thanks for the PR @isaacharrisholt! I reworked it a bit to better check if the `currentUser` is the author of the article (checking emails is not a reliable method).

isaacharrisholt:
Much obliged! I'll have a look at what you've done - I didn't know the best way to do things

PR #2582, opened 2024-02-18
feat: `@polar-sh/astro`

isaacharrisholt:
An implementation of #2523. Note: the `astro:content` is unavailable in integrations, so an `AstroCollectionEntry` interface was manually created with the properties we need.

> [!IMPORTANT]
> This requires a release of `@polar-sh/sdk` to add the `"type": "module"` key to its `package.json` - the Astro build would fail without it.

## Review suggestions

I'd recommend looking at things in the following order:

- `client.ts`
- `types.ts`
- `upload.ts`

`upload.ts` contains the bulk of the logic, though the other files will give context for a lot of the decisions made within that module.

---

Still TODO:

- [x] Create a demo site for testing
- [x] README
- [x] `package.json` metadata
- [x] `package.json` scripts
- [x] CI pipeline
- [x] Zod schema


zegl:
Hey @isaacharrisholt! I've been checking this out, and the code seems really solid. Well done!

I'm new to Astro, could you share an example of how you would integrate this with Astro? Some examples for how to use it would also be really nice to have in the packages README. :)

isaacharrisholt:
Hey, this is pretty much all good to go except for CI stuff now. Astro packages don't need a `build` step - they can be published to NPM as-is.

isaacharrisholt:
Note: this requires a release of `@polar-sh/sdk` to add the `"type": "module"` key to its `package.json` - the Astro build would fail without it.

isaacharrisholt:
@zegl the reason `pnpm build` was removed was because Astro runs code on build - since we don't want these posts going to Polar, it's best it doesn't get built. We don't have the required environment variables, so the upload will fail, killing the build process and breaking CI.

As for posts being created as drafts, I left it as-is as there's no concept of a 'draft' in Astro. Typically, a user would set a `published_at` or similar field in their frontmatter and display conditionally based on that (much like Polar does!). If they wanted to, they could then use this in the upload builder.

zegl:
> The reason pnpm build was removed was because Astro runs code on build - since we don't want these posts going to Polar, it's best it doesn't get built. We don't have the required environment variables, so the upload will fail, killing the build process and breaking CI.

Yeah, I figured. I've [disabled building the package by default](https://github.com/polarsource/polar/pull/2582/commits/2225e86c567dd437da026ae3c3f5c625afd9d07b).

PR #2752, opened 2024-03-05
clients/og-image: Fix squashed OG image on Publish page

emilwidlund:
Before
image

After
image


zegl:
Instead of this, I think we should allow the preview to change dimensions. It doesn't have to be square...

emilwidlund:
I think it's best to keep it in a known aspect ratio to reduce the likelihood of weird layout shifts when people upload images, as this is a generic image upload component. 

A wide 3:1 could potentially render the OG description box super small next to it. 

However, in addition, I'd like to implement an OG preview along with this, so that you can see the OG in action. As in render the uploaded image in a 1200x630 box with the title & OG description under.

PR #2759, opened 2024-03-05
astro: ensure client doesn't throw

isaacharrisholt:
The `PolarUploadBuilder` class should never throw an error. This PR fixes some undesirable behaviour where it would throw an incorrectly-instantiated error.

Also use `onfulfilled` to ensure the promise always completes and doesn't throw that way.


zegl:
LGTM

PR #2805, opened 2024-03-11
Dx/remove-save-create

frankie567:

PR #2812, opened 2024-03-11
server/repositories: implement profile_settings feature for repositories

emilwidlund:


server/polar/repository/service.py
```diff
@@ -180,6 +180,28 @@ async def update_badge_settings(
...
+        repository: Repository,
+        settings: RepositoryUpdateProfile,
+    ) -> Repository:
+        if settings.profile_settings is not None:
+            repository.profile_settings = {
+                **repository.profile_settings,
+                **settings.profile_settings.model_dump(mode="json", exclude_unset=True),
+            }
+
+        updated = await repository.save(session)
```
frankie567: I'm currently on a effort to remove the `.save` pattern in #2805. I would recommend instead:

```py
session.add(repository)
await session.flush()
```

frankie567 (changes requested):
Just a small thing around how we commit to the DB, but other than that, looks good :)

server/polar/repository/service.py
emilwidlund: Nice! Fixed :)

PR #2838, opened 2024-03-12
server/repo: add cover_image_field

emilwidlund:


server/polar/repository/service.py
```diff
@@ -207,10 +208,22 @@ async def update_settings(
...
+                profile["cover_image_url"] = (
+                    update.get("cover_image_url")
+                    if update.get("cover_image_url")
+                    else None
+                )
+
+            if update.get("featured_organizations") is not None:
+                profile["featured_organizations"] = update.get("featured_organizations")
+
+            repository.profile_settings = profile
```
zegl: There must be a better way to do this... give me a sec and I'll try to find a better way

zegl: Do it like this, much simpler and it's using the Pydantic model:

```python
    async def update_settings(
        self,
        session: AsyncSession,
        repository: Repository,
        update: RepositoryUpdate,
    ) -> Repository:
        profile_settings = RepositoryProfileSettings.model_validate(
            repository.profile_settings
        )

        if update.profile_settings is not None:
            if update.profile_settings.set_cover_image_url:
                profile_settings.cover_image_url = (
                    update.profile_settings.cover_image_url
                )

            if update.profile_settings.featured_organizations is not None:
                profile_settings.featured_organizations = (
                    update.profile_settings.featured_organizations
                )

            repository.profile_settings = profile_settings.model_dump(mode="json")

        session.add(repository)
        await session.flush()
```

PR #2839, opened 2024-03-12
Update `node` to `18.19.1`

patrick91:
Hi there, I was trying to debug and issue and I tried to use the devcontainer setup provider, but unfortunately the installation was failing with this error:

```txt
../src/common.cc:25:22: fatal error: vips/vips8: No such file or directory
 #include <vips/vips8>
```

installing libvips-dev using apt wasn't working, as I'd get this error:

```txt
/usr/include/vips/vips8:35:10: fatal error: glib-object.h: No such file or directory
│    35 | #include <glib-object.h>
```

this error was happening because the installation of `sharp` was trigger a compilation, instead of downloading the binary.

I fixed this by forcing sharp to the latest version, I'm haven't double checked this properly, but I think the version that's being installed now is not supported (doesn't have binaries) for the version of node running on the devcontainer.

I've also bumped the node version in the client folder to match the one in devcontainer (v18.19.1), but I don't know if that's a good idea, happy to revert that 😊

After updating updating the version I recreated the lock file and that seems to have fixed the issue on the devcontainer (see the commit history)


patrick91:
I also get this error when running the backend tests, but I can try to fix that in another PR 😊

```
E           asyncpg.exceptions.InvalidCatalogNameError: database "polar_test" does not exist
```

zegl:
Very nice, thanks!

PR #2888, opened 2024-03-18
clients/creator: reroute creator upsell to overview

emilwidlund:


clients/apps/web/src/components/Dashboard/Upsell.tsx
```diff
@@ -30,11 +30,13 @@ export const GitHubAuthUpsell = () => {
...
 export const MaintainerUpsell = () => {
   const maintainerUpgrade = useMaintainerUpgrade()
+  const { currentUser } = useAuth()
+  const router = useRouter()
 
   const upgrade = async () => {
     await maintainerUpgrade.mutateAsync()
     // TODO: Find a way to refresh the page using router.refresh() without keeping stale data
-    window.location.reload()
+    router.push(`/maintainer/${currentUser?.username}/overview`)
```
zegl, line 39: Does this work? Given the warning/comment above

emilwidlund: @zegl Yes, I tested this locally. I think the previous issues were related to stale data given the same layout. But as we're now pushing to the dashboard layout instead, it seems to work fine.

PR #2892, opened 2024-03-18
clients/posts: add onboarding cards to admin-empty state on public page

emilwidlund:


emilwidlund:
Also removed the wrapping `<Link />` from the `AnimatedIconButton` component, as it was causing hydration issues, as all usages already had a parent `Link` wrapping.

PR #3002, opened 2024-04-03
clients/public: prevent org description from overflowing on public pages

zegl:


emilwidlund:
Mind if we do a `break-word` here instead of `break-all`?
image

Issue #2771, opened 2024-03-06
Remove IssueReference and IssueDependency models and logic


frankie567:
As pointed out by @zegl in #2761, we are putting great efforts in synchronizing issue references and issue dependencies. Lot of code to maintain, lot of computation power spent for a feature that's not really used.

Since we'll likely adjust the vision around issue funding, it might be good to consider removing this; and only focus on synchronizing direct issues and pull requests. I believe it'll save us lot of headaches 😄

Would like to have your inputs on this, @birkjernstrom 😊

birkjernstrom:
Agree. Chrome extension is another nice negative PR around the corner 🙂

So let's get rid of it. However, now as we're adding the ability to highlight initiatives on your page with the potential desire to add any repo there, we might want a slim version of the "synchronizer". One that syncs a given repo vs. crawls for dependencies though - that part we can definitely drop. In our chat on Monday we'll talk about that page together so we're all on the same page. Depending on that outcome we'll know if we remove 100% or 80%+ here.

PR #3031, opened 2024-04-04
Remove IssueReference and IssueDependency logic

frankie567:
As discussed in #2771

The drawback with this change is that we'll no longer synchronize Pull Requests (before, they were synchronized if referenced from an issue).

IMO, this is acceptable; as we may want to introduce a complete synchronization of Pull Requests, enabling us to allow pledges on them.

Thoughts @birkjernstrom, @zegl?


zegl:
I'm in favour! The only real feature where Pull Request data is currently used is in the Split Rewards modal, where authors of related PRs are pre-filled in the modal.

It's a lot of code for such a simple feature, and if we were to build it from scratch, this is definitely not how we'd build it.

PR #3069, opened 2024-04-09
server/donations: implement endpoint for public donation search

emilwidlund:


server/polar/donation/schemas.py
```diff
@@ -117,3 +117,32 @@ class DonationStatisticsPeriod(Schema):
...
 class DonationStatistics(Schema):
     periods: Sequence[DonationStatisticsPeriod]
+
+
+class DonationSummary(Schema):
+    id: UUID4
+    amount: CurrencyAmount
+    message: str | None
+    donor: DonationOrganization | DonationUser | None
+    created_at: datetime.datetime
```
zegl: Should this be public?

emilwidlund: I think the date is fine to keep in the API - but not necessarily show it in the UI.

zegl (approved):
LGTM

PR #3206, opened 2024-04-25
clients/funding: adjust width

emilwidlund:
Extends badge width a bit more.


emilwidlund:
We'll keep it at 500px.

PR #3255, opened 2024-05-04
Add basic issue templates

pktiuk:
I think they may come in handy :smile: 

You can check them here: https://github.com/pktiuk/polar/issues/new/choose

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -1,0 +1,26 @@
+name: "\U0001F41E Bug Report"
+description: Create a bug report to help us fix it
+labels: ["bug"]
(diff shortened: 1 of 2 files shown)


pktiuk:
@birkjernstrom

birkjernstrom:
Sorry I missed this in the notifications @pktiuk. I really appreciate the intention and contribution. However, I don't think we need issue templates at the moment. Currently, we're not facing a problem of people submitting wrongful issues or any duplicates. So I'd prefer to keep things simple and allow people to submit issues freely until it becomes a problem.

We also use GitHub issues for our backlog/roadmap too and constantly create lightweight issues to capture an idea before we flush it out once we start working on it. So having to go through the form first would cause a bit of additional overhead there too.

PR #3256, opened 2024-05-06
Implement new onboarding flow

emilwidlund:


server/polar/organization/service.py
```diff
@@ -243,6 +248,24 @@ async def update_settings(
 
             organization.profile_settings = profile_settings.model_dump(mode="json")
 
+        if settings.feature_settings is not None:
```
frankie567, line 251: Why do we need to go through every single field like this? That looks really painful 😅 I think we could refactor it like this:

```py
if settings.feature_settings is not None:
    organization.feature_settings = {
        **organization_feature_settings,
        **settings.feature_settings.model_dump(mode="json", exclude_unset=True)
    }
```

We don't need neither to manually `OrganizationFeatureSettings.model_validate`, it has already been validated recursively by Pydantic when validating `OrganizationUpdate`.

emilwidlund: Nice! Updated! 🌮

PR #3294, opened 2024-05-14
Ability to block user/orgs

birkjernstrom:
- server: Remove unused Organization.Status enum
- server: Disable authz and public page in case of blocked subject
- server: Disable user & organization API for blocked resources


birkjernstrom:
Would be great with your eyes on this @frankie567 🙏🏼

server/tests/fixtures/random_objects.py
```diff
@@ -56,6 +56,25 @@ async def second_organization(save_fixture: SaveFixture) -> Organization:
     return await create_organization(save_fixture)
 
 
+@pytest_asyncio.fixture(scope="function")
```
frankie567: The `"function"` scope is the default, so you may omit it :)

server/polar/auth/dependencies.py
```diff
@@ -97,6 +97,11 @@ def __call__(self, auth_subject: AuthSubject[Subject]) -> AuthSubject[Subject]:
             else:
                 raise Unauthorized()
 
+        # Blocked subjects
+        blocked_at = getattr(auth_subject.subject, "blocked_at", None)
+        if blocked_at is not None:
+            raise Unauthorized()
```
frankie567: So, that means we are basically preventing blocked users from doing *anything* on Polar, right?

If I were pedantic, I would say it's a 403 (`NotPermitted`) instead of 401 🤓

server/polar/organization/service.py
```diff
@@ -46,26 +46,49 @@ class OrganizationService(ResourceServiceReader[Organization]):
...
+        conditions.append(Organization.blocked_at.is_(None))
+        query = sql.select(Organization).where(*conditions)
+        res = await session.execute(query)
+        return res.scalars().unique().one_or_none()
+
     async def get_by_platform(
         self, session: AsyncSession, platform: Platforms, external_id: int
     ) -> Organization | None:
-        return await self.get_by(session, platform=platform, external_id=external_id)
+        # TODO: Also add deleted_at=None in a separate commit
```
frankie567: I would be cautious with this particular one. I think we may have some places in the GitHub sync logic that use that method where we need to retrieve deleted organizations to handle reinstallations.

frankie567 (approved):

server/tests/fixtures/random_objects.py
birkjernstrom: Ah yes, nice. Copy pasta 😅

server/polar/auth/dependencies.py
birkjernstrom: Correct, but also if subject is an organization for org scoped API calls for instance. 

Re: `403` that's fair. I'll oblige :) Nice to distinguish it in a pure API vs. 404. I'm a fan of the latter since it obfuscates the reason, but in this rare case I think it's OK to have it as `403`. Will update.

server/polar/organization/service.py
birkjernstrom: That's a really good point. You mean the comment of the intended addition of `deleted_at=None` and not the actual change here right?

I'm also wondering whether this filter/check should live entirely on the endpoint layer vs. within the service. I kind of prefer auth/authz be on the endpoint layer with the service being clean of it, e.g allowing such methods to be used within other use cases (internal admin). But for now and here, I think it's fine to have it within the service for this particular use case.

frankie567: > That's a really good point. You mean the comment of the intended addition of `deleted_at=None` and not the actual change here right?

Yes, that's what I was referring to :)

> I'm also wondering whether this filter/check should live entirely on the endpoint layer vs. within the service. I kind of prefer auth/authz be on the endpoint layer with the service being clean of it, e.g allowing such methods to be used within other use cases (internal admin). But for now and here, I think it's fine to have it within the service for this particular use case.

Well, it probably depends on the situation. Sometimes, it's part of a bigger logic chunk where we don't want to deal with those internal details.

Issue #2636, opened 2024-02-26
File downloads


birkjernstrom:
Subscription & product benefit to offer customer digital files to be downloaded.

**In Progress**
https://github.com/polarsource/polar/pull/3296

**Must haves**
- [x] Flatted checksum in schemas, e.g not within `checksum`, but `checksum_` (not a separate resource)
- [x] Endpoint to get accessible files for user
- [x] Feature flag - enabling early deployment while setting up S3 in production etc
- [x] Edit: Sorting of files
- [x] Edit: Disable file
- [x] Edit: Delete file
- [x] Make `files` more standalone (generic S3 solution) vs. connected with `file_permission` only
- [x] Design polish
- [x] Set benefit icon

**Bugs**
- [x] Filename becomes ID - should rename local filename
- [x] Benefit not shown in UI upon creation (requires refresh)
- [x] Benefit not granted immediately or granted as revoked...

**Must have chores**
- [x] API Documentation
- [x] Write some core tests
- [x] Test with more file types

**Should haves (can be separate enhancements, i.e issues)**
- [x] Polar signature to proxy S3
- [x] Edit: Set file version

birkjernstrom:
🚀 Shipped. Still behind a feature flag, but everything is in-place. Just need to enable & communicate it.

PR #3296, opened 2024-05-14
File downloads

birkjernstrom:
See issue https://github.com/polarsource/polar/issues/2636 for outstanding work to be done


clients/packages/sdk/generate
```diff
@@ -13,7 +13,7 @@ docker rm generate-api || true
 docker run \
     --volume "${SCRIPTPATH}":/sdk \
     --name generate-api \
-    openapitools/openapi-generator-cli:v7.5.0 \
+    openapitools/openapi-generator-cli:v7.4.0 \
```
birkjernstrom, line 16: Required since v7.5.0 created some buggy types. Need to investigate later & separately to upgrade once that bug has been resolved.

birkjernstrom:
Still got some tests & polish to do, but core tests & all logic is in-place. So getting ready to merge with main behind a feature flag.

PR #3298, opened 2024-05-16
Sale model

frankie567:


frankie567:
@birkjernstrom For this second wave of changes, here is what might impact you for your accounting queries:

* `Transaction.subscription_id` and `Transaction. product_price_id` are removed in favor of `Transaction.sale_id`.
* A `Sale` is a new object representing a purchase (one-time or a recurring invoice for a subscription). It's linked to `products`, `product_prices` and, if it's a subscription, `subscriptions`.

PR #3521, opened 2024-06-24
server/orders: add order task for discord notifications

emilwidlund:


frankie567 (changes requested):
LGTM! You just should run the linters:

```bash
poetry run task lint
```

PR #3523, opened 2024-06-24
server/tasks: add order taskss

emilwidlund:


frankie567:
I have a fix for the unit test failing in #3524 ;)

emilwidlund:
> I have a fix for the unit test failing in #3524 ;)

Ah nice thanks!

PR #3549, opened 2024-06-27
server/transaction: fix payout computation

frankie567:
We were not considering all the balance reversal when computing the transfer of a balance. This caused a mismatch when some balance were refunded.


birkjernstrom:
Nicely caught and fixed @frankie567 💪🏼

PR #3623, opened 2024-07-09
clients/web: use full bundle for client-side Shiki

### LINKED ISSUE #3622: Missing languages in client-side code highlighter
A markdown code block won't have syntax highlighting when previewing a post. For published articles it's shown correctly.

frankie567:
Attempt to fix #3622

Not sure to merge it though, I fear the bundle size might explode.


frankie567:
The size increase looks reasonable, probably because Shiki is able to async load languages on demand. Will merge it!

PR #3749, opened 2024-08-09
Benefit: License Keys

birkjernstrom:
**API**
- [x] Endpoint to `/disable` activations
- [x] `PATCH /license-keys/{id}`
- [x] `GET /license-keys`
- [x] `GET /license-keys/{id}/activations/{id}`
- [x] Modify benefit grants upon benefit updates
- [x] `organization_id` as required condition during validation
- [x] revoke raises keyerror
- [x] Logging

**UI**
- [x] Show license key in user receipts/orders
- [x] Feature flag on the UI

**Docs: Separate PR**
- [ ] Review and refine API documentation
- [ ] Write documentation


server/polar/license_key/endpoints.py
```diff
@@ -0,0 +1,153 @@
...
+    response_model=ListResource[LicenseKeyRead],
+    responses={
+        401: UnauthorizedResponse,
+        404: NotFoundResponse,
+    },
+)
+async def list(
+    auth_subject: auth.LicenseKeysRead,
+    pagination: PaginationParamsQuery,
+    organization_id: UUID4,
```
frankie567: Lately, I tend to move away from mandatory/single-valued query parameter. I find it more elegant to make everything optional: by default, it'll show you all license keys you have access to (i.e. across all your organizations). That's the purpose of my `_get_readable_xx_statement` methods:

https://github.com/polarsource/polar/blob/ce29527fa5d323871e289f8bf3690dd2dafb3b81/server/polar/order/service.py#L499-L501

Then, to filter on one or several organization, you can use the `MultipleQueryFilter` helper:

https://github.com/polarsource/polar/blob/ce29527fa5d323871e289f8bf3690dd2dafb3b81/server/polar/order/endpoints.py#L34-L36

And then, extend the statement with:

https://github.com/polarsource/polar/blob/ce29527fa5d323871e289f8bf3690dd2dafb3b81/server/polar/order/service.py#L144-L145

---

The benefit of this approach is that:

* Users can finely tune the result of the query (no or multiple organization filter)
* You don't need to pre-check the existence of the organization, nor the access to it. If it doesn't exist or if you don't have permission, then you don't have results.

---

Anyway, we can deploy it like this and change it later: it won't be a breaking change to the API.

server/polar/license_key/endpoints.py
```diff
@@ -0,0 +1,153 @@
...
+    id: UUID4,
+    session: AsyncSession = Depends(get_db_session),
+    authz: Authz = Depends(Authz.authz),
+) -> LicenseKey:
+    """Get a license key."""
+    lk = await license_key_service.get_loaded(session, id)
+    if not lk:
+        raise ResourceNotFound()
+
+    if not await authz.can(auth_subject.subject, AccessType.read, lk):
```
frankie567, line 87: Related to my comment above: if you adopt a `_get_readable_xx_statement` approach, you don't need to check the permission. If you don't have access to it, you'll get a 404.

server/polar/user/endpoints/license_keys.py
```diff
@@ -0,0 +1,175 @@
...
+    response_model=ListResource[LicenseKeyRead],
+    responses={
+        401: UnauthorizedResponse,
+        404: NotFoundResponse,
+    },
+)
+async def list_license_keys(
+    auth_subject: auth.UserLicenseKeysRead,
+    pagination: PaginationParamsQuery,
+    organization_id: OrganizationID,
```
frankie567: Same thing as my first comment: with an optional/multi-valued approach, we allow users to get all their license keys across several organizations. But not a breaking change to do it afterwards.

server/polar/license_key/endpoints.py
```diff
@@ -0,0 +1,153 @@
...
+    LicenseKeyActivationRead,
+    LicenseKeyRead,
+    LicenseKeyUpdate,
+    LicenseKeyWithActivations,
+    NotFoundResponse,
+    UnauthorizedResponse,
+)
+from .service import license_key as license_key_service
+
+router = APIRouter(prefix="/license-keys", tags=[APITag.documented, APITag.featured])
```
frankie567: 🚨 You need to add a dedicated tag first, like `license_keys`:

```
tags=["license_keys", APITag.documented, APITag.featured]
```

Otherwise, generated JS SDK and documentation will name the group "Documented" (see next comment)

clients/packages/sdk/src/client/apis/DocumentedApi.ts
```diff
@@ -0,0 +1,254 @@
...
+
+export interface DocumentedApiUpdateRequest {
+    id: string;
+    body: LicenseKeyUpdate;
+}
+
+/**
+ * 
+ */
+export class DocumentedApi extends runtime.BaseAPI {
```
frankie567: 🚨 The name is not good, it should be `LicenseKeys`. Add a tag on the router to fix this (see previous comment).

frankie567 (dismissed):
Incredible job!

One thing that needs to be fixed before merging: the missing tag on the license keys API router.

I also added a bit of context/rationale behind the API design re: organization_id filter, but that can be done later as it's not a breaking change.

birkjernstrom:
Thank you so much @frankie567 – appreciate the second pair of eyes 🙏🏼 Great catch re: missed tag. Will fix the reported issues and merge quickly.

server/polar/license_key/endpoints.py
birkjernstrom: Great callout & agree that the approach you've moved over to should be the convention and is better. Updating the implementation accordingly.

PR #3917, opened 2024-09-06
server: differentiate email for subscription cancelation

codenoid:
On the existing code, Polar send the subscription created email for when the even is subscription cancelation.

### DIFF
--- a/server/polar/subscription/service.py
+++ b/server/polar/subscription/service.py
@@ -616,7 +616,6 @@ class SubscriptionService(ResourceServiceReader[Subscription]):
             },
         )
 
-        # Send the email to the user
         email_sender.send_to_user(
             to_email_addr=user.email,
             subject=subject,
--- a/server/polar/subscription/service.py
+++ b/server/polar/subscription/service.py
@@ -509,6 +509,8 @@ class SubscriptionService(ResourceServiceReader[Subscription]):
             and not subscription.cancel_at_period_end
         ):
             await self.send_confirmation_email(session, subscription)
+        elif subscription.cancel_at_period_end:
+            await self.send_cancellation_email(session, subscription)
 
         return subscription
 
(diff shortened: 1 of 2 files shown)


codenoid:
Sorry for the gipiti comment, 3AM Pull Request kinda hit different

frankie567:
Hey @codenoid 👋 

Thank you for this! So, you mean that a confirmation email is sent **after cancelling** a subscription plan? Did you experience this? That sounds really unexpected 😅

codenoid:
@frankie567 correct, that was what happen when I testing (in polar.so) to cancel my subscription

frankie567:
Thank you @codenoid, I've checked on my side and indeed experienced the same behavior; really appreciate your report on this!

I'll have a look at the PR ;)

server/polar/subscription/service.py
```diff
@@ -587,6 +588,41 @@ async def send_confirmation_email(
             from_email_addr="noreply@notifications.polar.sh",
         )
 
+    async def send_cancellation_email(
```
frankie567, line 593: Nice, but this isn't called anywhere, right?

Maybe we should add in `update_subscription_from_stripe` above:

```py
elif subscription.cancel_at_period_end:
    await self.send_cancellation_email(session, subscription)
```

server/polar/subscription/service.py
```diff
@@ -506,6 +506,7 @@ async def update_subscription_from_stripe(
             and subscription.started_at is not None
             and subscription.started_at.date()
             == subscription.current_period_start.date()
+            and not subscription.cancel_at_period_end
```
frankie567, line 509: Good catch!

frankie567 (changes requested):

server/polar/subscription/service.py
codenoid: ah, rookie mistake, code updated!

frankie567:
Thank you for this @codenoid, that's really helpful 🙏

frankie567:
`date` is not a builtin filter in Jinja2 🙂 Fixed and added a unit test here: #3948

Issue #3916, opened 2024-09-06
Webhook Deliveries Logs: show user endpoint response


codenoid:
nice to have, make life easier to debug

frankie567:
Implemented recently

PR #3918, opened 2024-09-07
webhook deliveries: show user's endpoint response

codenoid:
fix for #3916

### DIFF
--- a/clients/apps/web/src/components/Settings/Webhook/WebhookDeliveriesTable.tsx
+++ b/clients/apps/web/src/components/Settings/Webhook/WebhookDeliveriesTable.tsx
@@ -268,9 +268,22 @@ const ExpandedRow = (props: CellContext<DeliveryRow, unknown>) => {
         </Button>
       </div>
       <hr />
-      <pre className="whitespace-pre-wrap">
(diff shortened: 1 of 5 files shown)


birkjernstrom:
Wow @codenoid – really impressed how you've jumped into the codespace and provided several PRs already 💪🏼🙏🏼 I'll review them and hope to merge many of them soon.

Have you joined our Discord btw? Would be great to chat 1-1 :) 
[dub.sh/polar-discord ](https://dub.sh/polar-discord)

Regarding this one: I need to think on it a little bit to ensure we avoid some security concerns. I think we'd at least want to make sure we only accept JSON responses and at a certain size. Otherwise, it can easily become problematic. But need to think and look at what others are doing before I give my final say :)

birkjernstrom:
Sorry again @codenoid for the delay. I know we discussed this a bit on Discord and left it open as it was an ongoing discussion. For the time being, I'd prefer to avoid storing the responses. Might change in the future, but for now we haven't heard a lot of requests for it and from a security/storage POV we'd need to take further precautions to do it. Given the low interest from the community, I don't think it's something we need to prioritize at the moment even though I greatly appreciate your initial stab at it here (so it's painful to deny it atm).

PR #3920, opened 2024-09-07
client: Link Creator Dashboard to current organization if user has th…

codenoid:
client: Link Creator Dashboard to current organization if user has the organization

image

## Issue:

Currently, when a user visits their organization page and clicks the "Creator Dashboard" button, it always links to the first organization in their list. This is incorrect if the user owns the visited organization.

## Solution:

Identify Ownership: Determine if the current user is the owner of the visited organization.
Link to Current Organization: If the user is the owner, link the "Creator Dashboard" button directly to the current organization's dashboard page.

## Expected Behavior:

When a user who owns an organization visits their organization page and clicks the "Creator Dashboard" button, it should link to the dashboard for that specific organization.


emilwidlund:
Nice one! Added a few more occurrances of the `organization` param.

PR #3923, opened 2024-09-07
fix: update url in top bar nav to point to sdk

xn1cklas:
The navigation pointed to "https://docs.polar.sh/api/polar-sdk" though this path leads to 404


birkjernstrom:
Thank you @xn1cklas 🙏🏼

PR #3957, opened 2024-09-14
clients: add .env.template

codenoid:
upcoming pull request will be related with package.json (add @dotenvx/dotenvx) and README.md change


frankie567:
We already have a `.env.template` in `clients/apps/web`; so I don't think this is needed. Could you elaborate on this?

codenoid:
oh, I was never developing on clients/app/web, I might just do some readme update

PR #3963, opened 2024-09-15
clients: blur money input on scroll

codenoid:
https://github.com/user-attachments/assets/5879e618-393b-42ff-b752-0bf4107fbc32

The current input scroll behavior is very error-prone, as shown in the attached video

Lemon Squeezy, Tokopedia also disable the scroll on the input[type=number].


codenoid:
Sorry, by 'blur' I don't mean literally blurring the input. I mean removing the focus from the input when scrolling.

frankie567:
Great catch and great fix! Thank you @codenoid 🙏

Issue #4005, opened 2024-09-18
Private note is required when creating Custom benefit


frankie567:
This should clearly be optional 🤔

frankie567:
Currently, we use a single `BenefitCustomProperties` schema for both input and output. Problem is, the output has no default value, making it required.

We should introduce a `BenefitCustomCreateProperties`, like we already do for other benefits, where the `note` has a `None` default.

nareshNishad:
@frankie567 i have raised the PR for the same https://github.com/polarsource/polar/pull/4099

PR #4099, opened 2024-10-05
created BenefitCustomCreateProperties with default note to None

nareshNishad:
Fix issue #4005 
Create a BenefitCustomCreateProperties same as others CreateProperties in schema


frankie567:
Thank you @nareshNishad, but that's not the way it works in our codebase ☺️

The schema is auto-generated from actual Python (Pydantic) models. So the changes should rather happen over here: https://github.com/polarsource/polar/blob/07dfb2b56a1473ec3ad041e7c1893a08bcc30e5b/server/polar/benefit/schemas.py#L60

nareshNishad:
Thanks @frankie567 for guiding, i will make the changes.

nareshNishad:
@frankie567 updated the code, let me know anything else I can update in it.

frankie567 (changes requested):
That's exactly it, thank you @nareshNishad 🙏 

If you can just revert the changes on `sdk/schema.json` (it's auto-updated by a CI workflow), we should be good to go then.

nareshNishad:
@frankie567  Removed schema.json changes.

frankie567:
Took the liberty to fix a small typing issue. Thank you for your help in fixing this @nareshNishad :pray:

nareshNishad:
Sorry for being negligent about the test case, I will make sure about these in future contributions.

PR #4156, opened 2024-10-11
Signup Attribution & Intent

birkjernstrom:
Get better at tracking initial intent for using Polar. Making it easier to customize onboarding and designing the pathways to onboard users depending on intended use. 

Currently, we rely on an organization to be created to determine a creator. However, this can cause sub-optimal flows where a user signups to become a creator, but doesn't create an organization. Upon return they are treated as a buyer and we don't have a way to distinguish or continue their journey.

### DIFF
--- a/server/polar/integrations/google/endpoints.py
+++ b/server/polar/integrations/google/endpoints.py
@@ -1,4 +1,5 @@
 import uuid
+from typing import Any
 
 from fastapi import Depends, Request
 from fastapi.responses import RedirectResponse
@@ -43,7 +44,7 @@ async def google_authorize(
     return_to: ReturnTo,
     signup_attribution: UserSignupAttributionQuery,
 ) -> RedirectResponse:
-    state = {}
+    state: dict[str, Any] = {}
 
     state["return_to"] = return_to
 
--- a/server/polar/integrations/google/endpoints.py
+++ b/server/polar/integrations/google/endpoints.py
@@ -48,9 +48,7 @@ async def google_authorize(
     state["return_to"] = return_to
 
     if signup_attribution:
-        state["signup_attribution"] = signup_attribution.model_dump_json(
-            exclude_unset=True
-        )
+        state["signup_attribution"] = signup_attribution.model_dump(exclude_unset=True)
 
     if is_user(auth_subject):
         state["user_id"] = str(auth_subject.subject.id)
@@ -96,7 +94,7 @@ async def google_callback(
 
     state_signup_attribution = state_data.get("signup_attribution")
     if state_signup_attribution:
-        state_signup_attribution = UserSignupAttribution.model_validate_json(
+        state_signup_attribution = UserSignupAttribution.model_validate(
             state_signup_attribution
         )
 
--- a/server/polar/integrations/google/endpoints.py
+++ b/server/polar/integrations/google/endpoints.py
@@ -9,7 +9,6 @@ from polar.auth.dependencies import WebUserOrAnonymous
 from polar.auth.models import is_user
 from polar.auth.service import AuthService
 from polar.config import settings
-from polar.enums import UserSignupType
 from polar.exceptions import PolarRedirectionError
 from polar.kit import jwt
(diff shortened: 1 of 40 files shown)


server/polar/integrations/google/endpoints.py
```diff
@@ -41,21 +41,22 @@ async def google_authorize(
...
+    signup_attribution: UserSignupAttributionQuery,
 ) -> RedirectResponse:
     state = {}
 
     state["return_to"] = return_to
 
-    if user_signup_type:
-        state["user_signup_type"] = user_signup_type
+    if signup_attribution:
+        state["signup_attribution"] = signup_attribution.model_dump_json()
```
frankie567: You could serialize it as a plain dict since JWT encode handles the serialization anyway. If you need safe JSON output, you can do:

```py
signup_attribution.model_dump(mode="json")
```

I think it'll save extra bytes in the resulting JWT.

server/polar/integrations/google/endpoints.py
```diff
@@ -90,9 +91,12 @@ async def google_callback(
 
     return_to = state_data.get("return_to", None)
     state_user_id = state_data.get("user_id")
-    state_user_type = UserSignupType.backer
-    if state_data.get("user_signup_type") == UserSignupType.maintainer:
-        state_user_type = UserSignupType.maintainer
+
+    state_signup_attribution = state_data.get("signup_attribution")
+    if state_signup_attribution:
+        state_signup_attribution = UserSignupAttribution.model_validate_json(
```
frankie567: If you follow my previous comment, then this would become:

```py
UserSignupAttribution.model_validate(state_signup_attribution)
```

server/polar/integrations/loops/service.py
```diff
@@ -15,12 +14,8 @@ class Loops:
...
         user: User,
-        signup_type: UserSignupType | None = None,
         **properties: Unpack[Properties],
     ) -> None:
-        if signup_type == UserSignupType.imported:
-            return
-
         properties = {
             "isMaintainer": False,
             "isBacker": False,
```
frankie567, line 21: Can't we get rid of those?

server/polar/magic_link/service.py
```diff
@@ -93,10 +98,19 @@ async def authenticate(self, session: AsyncSession, token: str) -> User:
         if magic_link is None:
             raise InvalidMagicLink()
 
+        signup_attribution = None
+        if magic_link.signup_attribution:
+            signup_attribution = await get_signup_attribution(
```
frankie567: That's weird to use the FastAPI dependency in this context. What about this:

```py
signup_attribution = UserSignupAttribution.model_validate(magic_link.signup_attribution)
```

server/polar/order/service.py
```diff
@@ -314,28 +330,20 @@ async def create_order_from_stripe(
             if user is None:
                 assert invoice.customer_email is not None
                 user = await user_service.get_by_email_or_signup(
-                    session, invoice.customer_email, signup_type=UserSignupType.backer
+                    session,
+                    invoice.customer_email,
+                    signup_attribution=UserSignupAttribution(
+                        intent="order",
+                        attribution=f"order:{order.id}",
```
frankie567: I've seen your attribution format with colons. Wondering how easy/hard it'll be to use that in SQL queries and such? If it becomes complex with several items (like for `utm`), I fear we'll have a hard time, no?

server/polar/integrations/loops/service.py
birkjernstrom: Yes, definitely. However, I was thinking of doing that as the next step. Creating an updated Loops sync + scripts to migrate existing for a new format based on the user attributions.

server/polar/integrations/google/endpoints.py
birkjernstrom: Agreed. I'm quite sure I started that way, but ran into issues and therefore changed it to a JSON string. But I'll try and report back once I've confirmed + know the exact issue (and if it's not solvable in another way)

Issue #3577, opened 2024-07-01
Ctrl+ K is catched by Chrome search bar on Windows


frankie567:
Damned!

https://github.com/polarsource/polar/assets/1144727/e6edf187-f3f8-424c-8ffb-91ee4aa26637

PR #4220, opened 2024-10-18
fix: fix the Ctrl+ K that catches Chrome search bar on Windows

apsinghdev:
fixes: #3577 

This PR fixes the bug that catches the Chrome search bar on Windows instead of Polar's by preventing the default behavior of the browser.

### DIFF
--- a/clients/apps/web/src/components/CommandPalette/commands/trigger.tsx
+++ b/clients/apps/web/src/components/CommandPalette/commands/trigger.tsx
@@ -23,7 +23,8 @@ export const useCommandPaletteTrigger = (onTrigger: () => void): void => {
   useEffect(() => {
     const handleKeyPress = (e: KeyboardEvent) => {
       const controlKeyPressed = isMac(navigator) ? e.metaKey : e.ctrlKey
-      if (e.key === KEY && controlKeyPressed) {
+      if (e.key.toLowerCase() === KEY && controlKeyPressed) {
+        e.preventDefault()
         onTrigger()
       }
     }


apsinghdev:
@frankie567 Most probably, it will fix the issue. Unfortunately, I don't have any Windows environment to test it. If possible, Please test it on your side and let me know if it works.

clients/apps/web/src/components/CommandPalette/commands/trigger.tsx
```diff
@@ -23,7 +23,8 @@ export const useCommandPaletteTrigger = (onTrigger: () => void): void => {
   useEffect(() => {
     const handleKeyPress = (e: KeyboardEvent) => {
       const controlKeyPressed = isMac(navigator) ? e.metaKey : e.ctrlKey
-      if (e.key === KEY && controlKeyPressed) {
+      if (e.key.toLowerCase() === KEY && controlKeyPressed) {
```
frankie567, line 26: Sorry, I don't get it 😅 How lowering the case would solve the issue? 😅

apsinghdev: Actually the solution is preventing the default behaviour of the browser as when we click ctrl+k, the browser catches it first so I added `e.preventDefault` to prevent the default behaviour of the browser.

apsinghdev: added lower casing just for the sake of a double check for the command. (It can be removed though)

frankie567: Oh yeah, I see, sorry. Let's merge it — can't do any harm either way — and I'll check on my Windows computer then when I have a minute

frankie567:
Can confirm it works 🎉 Can't believe I didn't think about `preventDefault`! Thank you @apsinghdev 🙏

apsinghdev:
> Can confirm it works 🎉 Can't believe I didn't think about `preventDefault`! Thank you @apsinghdev 🙏

Really happy that it worked! Thank you! 😊

Issue #4238, opened 2024-10-21
Minimum pledge amount is not correctly rendered in badge settings


frankie567:
### Description

The badge settings input always show $20 even though the actual amount is something else.

### Current Behavior

$20 is always shown.

### Expected Behavior

The actual amount should be shown

### Screenshots

**Settings view always showing $20**

<img width="1323" alt="Capture d’écran 2024-10-21 à 16 25 47" src="https://github.com/user-attachments/assets/d8fd411a-09b4-4b8c-8c7a-4e82a5c1a8e8">

**Funding issue view showing the correct minimum**

<img width="662" alt="Capture d’écran 2024-10-21 à 16 26 22" src="https://github.com/user-attachments/assets/92085d5e-254b-49e8-83a7-440f738ba126">

### Environment:

-   Operating System: *
-   Browser (if applicable): Chrome (but likely all)

---

PR #4241, opened 2024-10-21
fix: Minimum pledge amount is not correctly rendered

apsinghdev:
fixes: #4238 

This PR fixes the default value of `Minimum funding amount` not getting updated bug. With these changes, the value of the Minimum pledge amount is not defaulting to 20 across rerenders.

https://github.com/user-attachments/assets/c7320cff-127f-4e5f-863b-d8e52a778410

I think it fixes the bug. @frankie567


reynaldichernando:
I tested this, and it does work.
However, I'm still curious as to why changing the `value` props doesn't re-render the `MoneyInput` component.

apsinghdev:
> I tested this, and it does work. However, I'm still curious as to why changing the `value` props doesn't re-render the `MoneyInput` component.

By default, the `value` is assigned to `internalValue` using `useState`. So the value is set in the initial render but later on when the value of `value` changes `useState` doesn't cause a re-render because of its nature hence we get default value even after updating it. So a simple solution would be to use a `useEffect` to track the changes in `value` and keep it in sync with `internalValue`.

apsinghdev:
@frankie567 I've fixed and updated this one too.

frankie567:
Since we'll remove issue funding features, this fix is no longer required. Thank you for your time here 🙏

PR #4337, opened 2024-10-29
clients/storefront: Optimize mobile storefront navigation

### LINKED ISSUE #4129: Storefront: Navigation is not mobile optimized
Currently is causes a horizontal scroll of the entire page + "cropped navigation" vs. the navigation being a scrollable section itself. 

image

reynaldichernando:
Fixes #4129 
Use `<Select>` component for storefront mobile navigation.

- Uses the tailwind media query break point for choosing which navigation to show (sm)
- For `Select` component (radix), it does not support putting `Link` inside the `SelectItem` component (nothing will happen upon clicking), that is why it is using router for navigation 

https://github.com/user-attachments/assets/ac1e547b-644b-4692-9342-2d26fca1b649


emilwidlund:
Well done! Thanks! 👍

PR #4362, opened 2024-11-01
clients/docs: add tunnel sample

codenoid:
add guide on request inspection and request replay


birkjernstrom:
Sorry for the delay @codenoid. Unfortunately, I won't merge this one. It's awesome that you've built lokal.so and personally, I'm going to try it out, but in our docs we're using Ngrok since it's well-known within the ecosystem and acts well as a placeholder for the type of setup needed. Otherwise, it might cause confusion.

PR #4379, opened 2024-11-05
feat: More contextual promotion embeds + navigation clean-up

birkjernstrom:
**Clean-up old promote page**
- **clients: Remove `Subscribe` embed suggestion**
- **clinets: Refactor embed component & remove tiers**

**Move GitHub FUNDING.yaml promotion to storefront (contextual)**
- **clients: Move GitHub Funding.yaml embed to storefront**
- **clients: Remove emphasis on GitHub embed in case of disabled storefront**

**Move promote page into issues (contextual)**
- **clients: Move issue embeds to issue page**

**Hide funding navigation unless enabled**
- **clients: Settings to enable/disable certain features**
- **clients: Hide funding features in navigation unless enabled**

**Product Card Embed**
- **clients: Product SVG endpoint**
- **clients: Offer SVG embeds in checkout links**
- **clients: Toggle link form & default to first link**
- **clients: Better UI of embeddable selection**
- **clients: Removed unused import**
- **clients: Improve Product SVG based on pricing type**


clients/apps/web/src/app/embed/product.svg/route.tsx
```diff
@@ -0,0 +1,126 @@
...
+} from '@polar-sh/sdk'
+const { default: satori } = require('satori')
+
+export const runtime = 'edge'
+
+const getProduct = async (
+  productId: string,
+  productPriceId: string,
+): Promise<Product | undefined> => {
+  let url = getServerURL(`/v1/products/${productId}`)
```
frankie567: This won't work anymore with the API changes. You should use the Storefront API, similar to here: 

https://github.com/polarsource/polar/blob/a13aee8f95f2354ee764d631bdbf5d490e46d446/clients/apps/web/src/app/embed/tiers.svg/route.tsx#L13-L28

frankie567: Noticing that we only have access to the product ID here.

Could probably be nice then to introduce an `/embed` API with different endpoints to return the data we need to render each type of embed. I could do that fairly quickly.

birkjernstrom: Thank you @frankie567. Nice catch. Didn't discover it in my branch of course given we did these two things in parallel 😅 I'lll fix this today. Agree re: `/embed` (internal) API being a nice solution and would also offer the ability to do some caching in a better/cleaner way over time which would be nice.

PR #4508, opened 2024-11-21
Fix broken webhooks links

elie222:


frankie567:
Thank you @elie222, really helpful 🙏

PR #4509, opened 2024-11-21
Fix broken link in WebhookSettings.tsx

elie222:


frankie567:
Great catch again, thanks 👍

PR #4510, opened 2024-11-21
Fix broken link in WebhookForm.tsx

elie222:


frankie567 (changes requested):
Thank you! For this one, the link is now like this (event is part of the path instead of an anchor):

```ts
const href = `https://docs.polar.sh/developers/webhooks/events/${event}`
```

elie222:
> Thank you! For this one, the link is now like this (event is part of the path instead of an anchor):
> 
> ```ts
> const href = `https://docs.polar.sh/developers/webhooks/events/${event}`
> ```

Sure. Feel free to make that edit / close this PR. I made the PR from the GitHub web interface and not quite sure how to make an additional update using it.
And don't want to spend the 3mins to clone and fix this either 😂

Issue #4237, opened 2024-10-21
Allow to update account email address


frankie567:
Currently, the email can't be changed.

We should have a process to update it, with proper validation.

- [ ] User settings to update email with validation
- [ ] Update contact email in Loops (`polar.integrations.loops.service.loops.user_update`)

PR #4520, opened 2024-11-22
feat: allow users update their account email address with proper validation.

apsinghdev:
fixes: #4237

This PR implements the feature to allow users to update their email. (Draft)

https://github.com/user-attachments/assets/aafd45b9-6696-4771-ab7d-48d48393d5aa


apsinghdev:
@frankie567 I've raised this draft PR that contains backend code. please have a look and let me know if it make sense.

server/polar/models/email_verification.py
```diff
@@ -0,0 +1,29 @@
...
+from datetime import datetime, timedelta
+from sqlalchemy.orm import Mapped, declared_attr, mapped_column, relationship
+from sqlalchemy import String, TIMESTAMP, Uuid, ForeignKey
+from polar.config import settings
+from polar.kit.db.models import RecordModel
+from polar.kit.utils import utc_now
+from polar.models.user import User
+
+def get_expires_at() -> datetime:
+    return utc_now() + timedelta(seconds=settings.MAGIC_LINK_TTL_SECONDS)
```
frankie567: Let's have a dedicated setting for that (keep the same default as magic link)

server/polar/email_update/service.py
```diff
@@ -0,0 +1,114 @@
...
+from polar.kit.services import ResourceServiceReader
+from polar.kit.utils import utc_now
+from polar.models import EmailVerification
+from polar.models.user import User
+from polar.postgres import AsyncSession
+
+from .schemas import EmailUpdateCreate
+
+
+TOKEN_PREFIX = "polar_ml_"
```
frankie567: The prefix needs to be changed. Let's say `polar_ev_`.

server/polar/email_update/service.py
```diff
@@ -0,0 +1,114 @@
...
+            email=email,
+            token_hash=token_hash,
+            expires_at=expire_at,
+            user_id=user_id
+        )
+        
+        email_update_record = EmailVerification(**email_update_create.model_dump(exclude_unset=True)) 
+        
+        session.add(email_update_record)
+        await session.commit()
```
frankie567: Use `await session.flush()` instead of committing here. The difference is here:

* With `flush`, the changes are sent to the DB (so we have integrity checks triggered and default values generated) but the transaction is *not committed*. It means that if something goes wrong after that in the request, the changes won't be saved to the database, ensuring it's consistent.
* With `commit`, all of the above is done but the transaction is committed, meaning the changes are definitely written to the DB. If something goes wrong after that, it's over, there is no going back.

*PS: I see we do that in the magic link service, but that's bad and we should fix it 😜*

server/polar/email_update/service.py
```diff
@@ -0,0 +1,114 @@
...
+        email_update_record = await self._get_email_update_record_by_token_hash(session, token_hash)
+
+        if email_update_record is None:
+            raise InvalidEmailUpdate()
+        
+        user = email_update_record.user
+        user.email = email_update_record.email 
+
+        await session.delete(email_update_record)
+        await session.commit()
```
frankie567: Same comment as above regarding `commit`. In this case, we can even have just `await session.delete(...)` and nothing else: `commit` is called automatically by a middleware after the response is sent to the client.

server/polar/email_update/service.py
```diff
@@ -0,0 +1,114 @@
...
+                "url": f"{base_url}?{urlencode(url_params)}",
+                "current_year": datetime.datetime.now().year,    
+            }
+        )
+        
+        email_sender.send_to_user(
+            to_email_addr=email_update_record.email, subject=subject, html_content=body
+        )
+
+    async def authenticate(
```
frankie567: We should call the method differently, let's say `verify`.

server/polar/email_update/endpoints.py
```diff
@@ -0,0 +1,64 @@
...
+        extra_url_params=(
+            {"return_to": email_update_request.return_to}
+            if email_update_request.return_to
+            else {}            
+        )
+    )
+
+
+@router.post("/authenticate")
+async def authenticate_email_update(
```
frankie567: Let's call that `/verify` and `verify_email_update`

server/polar/email_update/schemas.py
```diff
@@ -0,0 +1,22 @@
...
+class EmailUpdateCreate(Schema):
+    email: EmailStr
+    token_hash: str
+    expires_at: datetime.datetime | None = None
+    user_id: UUID4
+    
+class EmailUpdateRequest(Schema):
+    email: EmailStrDNS
+    return_to: str | None = None
+    attribution: UserSignupAttribution | None = None
```
frankie567: You can remove `attribution`, this won't be useful in this context. It's used to track new users provenance.

server/polar/email_update/endpoints.py
```diff
@@ -0,0 +1,64 @@
...
+from polar.config import settings
+from polar.exceptions import PolarRedirectionError
+from polar.integrations.loops.service import loops as loops_service
+from polar.kit.db.postgres import AsyncSession
+from polar.kit.http import ReturnTo, get_safe_return_url
+from polar.postgres import get_db_session
+from polar.routing import APIRouter
+
+from .schemas import EmailUpdateRequest
+from .service import email_update as email_upate_service
```
frankie567: Typo: `email_update_service`

server/polar/email_update/endpoints.py
```diff
@@ -0,0 +1,64 @@
...
+    email_update_request: EmailUpdateRequest,
+    auth_subject: WebUser,
+    session: AsyncSession = Depends(get_db_session),
+) -> None:
+    user = auth_subject.subject
+    user_id = user.id
+    email_update_record, token = await email_upate_service.request_email_update(
+        email_update_request.email,
+        session,
+        user_id,
```
frankie567: A bit of a nitpick, but let's pass the `auth_subject` directly to the method.

server/polar/email_update/endpoints.py
```diff
@@ -0,0 +1,64 @@
...
+) -> RedirectResponse:
+    
+    try:
+        user = await email_upate_service.authenticate(session, token)
+    except EmailUpdateError as e:
+        raise PolarRedirectionError(
+            e.message, e.status_code, return_to=return_to
+        ) from e       
+
+    await loops_service.user_update(session, user, emailLogin=True)
```
frankie567: `emailLogin` property is not relevant here (it's not a login). We just need to update the user email.

```py
await loops_service.user_update(session, user)
```

server/polar/email_update/email_templates/email_update.html
```diff
@@ -0,0 +1,33 @@
+{% extends "base.html" %}
+
+{% block body %}
+<h1>Hi,</h1>
+<p>Here is the magic link to update your email. Click the button below to complete the update process. <strong>This link
```
frankie567: Let's write:

```
Here is the verification link to update your email. Click the button below to complete the update process.
```

server/polar/email_update/email_templates/email_update.html
```diff
@@ -0,0 +1,33 @@
...
+<!-- Action -->
+<table class="body-action" align="center" width="100%" cellpadding="0" cellspacing="0" role="presentation">
+    <tr>
+        <td align="center">
+            <!-- Border based button
+https://litmus.com/blog/a-guide-to-bulletproof-buttons-in-email-design -->
+            <table width="100%" border="0" cellspacing="0" cellpadding="0" role="presentation">
+                <tr>
+                    <td align="center">
+                        <a href="{{ url | safe }}" class="f-fallback button">Sign in to Polar</a>
```
frankie567: `Update my email`

frankie567 (changes requested):
That's great work @apsinghdev 👍 Let's change a few things to make it really tailored for email verification :)

apsinghdev:
> That's great work @apsinghdev 👍 Let's change a few things to make it really tailored for email verification :)
> 
> Be sure also to run linters:
> 
> ```shell
> uv run task lint && uv run task lint_types
> ```

@frankie567 Thanks for the reviews. I'll fix these very soon.

apsinghdev:
@frankie567 I've made the updates to apply reviews and implemented client-side logic and UI. Please have a look whenever you are free and let me know your thoughts.

server/polar/email_update/schemas.py
```diff
@@ -0,0 +1,23 @@
+import datetime
+
+from pydantic import UUID4, EmailStr, field_validator
+
+from polar.kit.http import get_safe_return_url
+from polar.kit.schemas import EmailStrDNS, Schema
+
+
+class EmailUpdateCreate(Schema):
```
frankie567: Let's remove that intermediary schema, it's not really useful (see next comment)

server/polar/email_update/service.py
```diff
@@ -0,0 +1,130 @@
...
+        user_id = user.id
+        token, token_hash = generate_token_hash_pair(
+            secret=settings.SECRET, prefix=TOKEN_PREFIX
+        )
+        email_update_create = EmailUpdateCreate(
+            email=email, token_hash=token_hash, expires_at=expire_at, user_id=user_id
+        )
+
+        email_update_record = EmailVerification(
+            **email_update_create.model_dump(exclude_unset=True)
```
frankie567: We can remove `EmailUpdateCreate(...` and directly do:

```py
email_update_record = EmailVerification(
    email=email,
    token_hash=token_hash,
    user=auth_subject.subject,
)

File: server/polar/email_update/service.py [hunk 15]
```diff
@@ -0,0 +1,130 @@
...
+        )
+
+
+class EmailUpdateService(ResourceServiceReader[EmailVerification]):
+    async def request_email_update(
+        self,
+        email: str,
+        session: AsyncSession,
+        auth_subject: AuthSubject[User],
+        expire_at: datetime.datetime | None = None,
```
frankie567: This is not useful, the expiration date is always generated by default on the model itself.

clients/apps/web/src/components/Settings/GeneralSettings.tsx
```diff
@@ -61,42 +63,59 @@ const GeneralSettings = () => {
...
+      </ShadowListGroup.Item>
+      <ShadowListGroup.Item>
+        <div className="flex flex-row items-start justify-between">
+          <div className="flex flex-col gap-y-1">
+            <h3>Update email</h3>
+            <p className="dark:text-polar-500 text-sm text-gray-400">
+              Update your account&apos;s current email
+            </p>
+          </div>
+          <Link href="/email-update">
```
frankie567: Ideally, I would just embed the form right here, no need for a separate page. The current email should be profiled on the input form.

Similarly, the confirmation can appear right here instead of redirecting to a new page.

frankie567 (changes requested):
We're getting there, great work @apsinghdev 👍 Just a few adjustments and we should be able to land this.

clients/apps/web/src/components/Settings/GeneralSettings.tsx
apsinghdev: Thanks for the reviews! btw can we profile the current email as the placeholder or a text in the input form similar to this? : 
Screenshot 2024-12-03 at 3 36 31 PM

apsinghdev: Also please confirm if  "Similarly, the confirmation can appear right here instead of redirecting to a new page." means that when the user enters the email and hits `Update`, it replaces the input form with the `/request` msg. When the user hits the URL sent in the email, it shows the `/verify` msg (in place of that input form) and when the user hits the `Update the email` button, the email gets updated and the input form reappears.

apsinghdev:
@frankie567 I have made the changes. Please have a look whenever you get the chance. 

Demo:

https://github.com/user-attachments/assets/447b3859-8408-428f-beb1-6c7e1d37fa83

apsinghdev:
@frankie567 Thank you for the reviews. I'll make the changes very soon. 🙂

apsinghdev:
Hi @frankie567, I've made the necessary changes to check if an email already exists or not. Here is how it looks:


https://github.com/user-attachments/assets/67354c2a-1604-4f0a-a2f3-b6d69616d719

Please take a look where you are free.

frankie567:
Thank you @apsinghdev 👍 Made a few tweaks + a rebase, it's good to go now 😄

Thank you for your hard work 🙇

apsinghdev:
> Thank you @apsinghdev 👍 Made a few tweaks + a rebase, it's good to go now 😄
> 
> Thank you for your hard work 🙇

Thank you so much for your valuable feedbacks. 🙏😊

PR #4534, opened 2024-11-25
clients/web: derive teams instead of effect + state in useListTeams hook

alexandernanberg:
Instead of using an effect + state we can derive the teams directly from the data. This simplifies the code a bit, but also makes the UX a bit better since on the first render the list won't be empty until the effect has run


frankie567:
Thank you @alexandernanberg 👍 To be honest, this specific hook has really few importance nowadays in our UI, but I agree that using `useMemo` is better in this context.

Issue #4396, opened 2024-11-07
When creating orders/transactions from Stripe, use their creation timestamp instead of generating ours


frankie567:
When we receive an `invoice.paid` event from Stripe, we create a corresponding Order in our system. Naturally, we let the `created_at` property to the default, i.e. `NOW()`. Similarly, we create payment transactions when receiving payment success events.

It could be nice to use instead the creation timestamp provided by Stripe (i.e. their invoice or payment creation date). It doesn't make a difference in production, but in local development, it could be really nice when using Test Clocks: we can advance time in Stripe and our orders/transactions would also reflect that simulated time.

magarpratik:
Hi, I’d love to work on this issue if it's available. I can look at this next week.

If that's okay, could you please assign it to me? 

Thanks!

frankie567:
Sure, no problem!

magarpratik:
Hi @frankie567,

I've created a [draft PR](https://github.com/polarsource/polar/pull/4569) to address the issue, where the order creation logic triggered by an `invoice.paid` event has been updated to use the Stripe-generated `created` timestamp for the `created_at` property.

Is this sufficient, or is more work required? I'm asking because the [issue description](https://github.com/polarsource/polar/issues/4396#issue-2640985500) suggests we might want to apply similar changes to payment success events as well.

If that's the case, could you clarify which of the following events should be considered "payment success" events?

```
DIRECT_IMPLEMENTED_WEBHOOKS = {
    "payment_intent.succeeded",
    "payment_intent.payment_failed",
    "charge.succeeded",
    "charge.refunded",
    "charge.dispute.created",
    "charge.dispute.funds_reinstated",
    "customer.subscription.created",
    "customer.subscription.updated",
    "customer.subscription.deleted",
    "invoice.paid",
}
```

Code snippet taken from [here](https://github.com/polarsource/polar/blob/5df36c74a93e600df263138679fd670043048eeb/server/polar/integrations/stripe/endpoints.py#L19).

PR #4569, opened 2024-11-30
server: use created timestamp value from stripe event

magarpratik:
**Closes:** #4396 

This PR updates the order creation logic triggered by the `invoice.paid` event.

Instead of using the default `NOW()` timestamp for the `created_at` field, we now use the `created` field value ([Stripe doc](https://docs.stripe.com/api/events/object#event_object-created) regarding the `created` field) from the event object, which represents the time the event was generated by Stripe.

### DIFF
--- a/server/tests/fixtures/random_objects.py
+++ b/server/tests/fixtures/random_objects.py
@@ -7,6 +7,7 @@ from collections.abc import Sequence
 from datetime import UTC, datetime, timedelta
 from typing import Any, Unpack
 
+import pytest
 import pytest_asyncio
 
 from polar.enums import (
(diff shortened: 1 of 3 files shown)


server/tests/fixtures/random_objects.py
```diff
@@ -1301,3 +1302,10 @@ async def create_advertisement_campaign(
     )
     await save_fixture(advertisement_campaign)
     return advertisement_campaign
+
+
+@pytest.fixture
+def event_creation_time() -> tuple[datetime, int]:
```
frankie567: May we put this fixture directly in `order/test_service.py`? Seems only useful there for now.

frankie567 (changes requested):
Nice @magarpratik, thank you 👍 Just a small nitpick and we should be good to go :)

server/tests/fixtures/random_objects.py
magarpratik: Done ✅

frankie567:
Perfect, thank you 🙏

PR #4590, opened 2024-12-03
Customers architecture

frankie567:


server/polar/order/service.py
```diff
@@ -412,33 +412,16 @@ async def create_order_from_stripe(
...
-                        subscription=order.subscription.id,
-                    )
-
-                user, _ = await user_service.get_by_email_or_create(
-                    session,
-                    invoice.customer_email,
-                    signup_attribution=signup_attribution,
+            if customer is None:
+                customer = await customer_service.create_from_stripe_invoice(
+                    session, invoice, product.organization
```
frankie567: We already see with changes like this how the implementation feels more natural and straightforward.

I usually tend to think that if the implementation looks good, it's a good sign that the underlying design choice and architecture is right 😄

server/polar/customer/service.py
```diff
@@ -0,0 +1,51 @@
...
+            name=stripe_customer.name,
+            billing_address=stripe_customer.address,
+            # TODO: tax_id,
+            organization=organization,
+        )
+
+        session.add(customer)
+        return customer
+
+
```
frankie567, line 50: We may need to allow to list deleted customers to get historical data.

server/polar/customer/service.py
```diff
@@ -0,0 +1,166 @@
...
+            organization=organization,
+        )
+
+        session.add(customer)
+        return customer
+
+    def _get_readable_customer_statement(
+        self, auth_subject: AuthSubject[User | Organization]
+    ) -> Select[tuple[Customer]]:
+        statement = select(Customer).where(Customer.deleted_at.is_(None))
```
frankie567, line 275: We may need to allow to list deleted customers to get historical data.

PR #4624, opened 2024-12-11
Cancel & Modify Subscriptions

### LINKED ISSUE #4595: Subscriptions API: `HTTP DELETE /subscriptions/{id}`
More context: https://github.com/polarsource/polar/issues/4594

birkjernstrom:
Closes #4595

**Remaining (API)**
- [x] Tests for `DELETE /subscriptions/{id}`
- [x] Better documentation for `customer_reason` 
- [x] Fix reliance/usage of `cancel_at_period_end` now with `cancel_at`
- [x] Figure out a better `cancel_at` timestamp to avoid issues on slower networks...
- [x] Allow `DELETE /subscriptions` to set customer comment too
- [x] Tweak customer Cancellation notification email (subscription ending)
- [x] Send customer a Revoked notification email (fully cancelled)
- [x] Set `cancel_at_period_end` optimistically to keep API response up-to-date
- [x] Uncancel in subscription update

**Remaining (UI)**
- [x] Expose `cancellation_reason` & comment flag with the actual comment in details view
- [x] Better copy
- [x] Make it look beautiful 🤓

Follow-up enhancements
- [ ] Customer Portal: Ability to uncancel a subscription
- [ ] Dashboard: Ability to uncancel a subscription about to expire (`cancel_at_period_end`)
- [ ] API: Set proration setting on changing subscriptions
- [ ] Customer Portal: Prorate immediately upon annual subscription changes (better default)
- [ ] Dashboard: Admin controls to change subscription product & price with proration control
- [ ] Webhook & email: On subscription downgrade/upgrade

### DIFF
--- a/clients/apps/web/src/hooks/queries/subscriptions.ts
+++ b/clients/apps/web/src/hooks/queries/subscriptions.ts
@@ -1,7 +1,7 @@
-import { SubscriptionsApiListRequest } from '@polar-sh/sdk'
-import { useQuery } from '@tanstack/react-query'
+import { SubscriptionsApiListRequest, SubscriptionCancel, ListResourceSubscription } from '@polar-sh/sdk'
+import { useQuery, useMutation } from '@tanstack/react-query'
 
-import { api } from '@/utils/api'
+import { api, queryClient } from '@/utils/api'
 import { defaultRetry } from './retry'
 
 export const useListSubscriptions = (
@@ -17,4 +17,39 @@ export const useListSubscriptions = (
       }),
     retry: defaultRetry,
     enabled: !!organizationId,
+  })
+
+export const useCancelSubscription = () =>
+  useMutation({
+    mutationFn: (variables: { id: string, body: SubscriptionCancel }) => {
+      return api.subscriptions.update(variables)
+    },
+    onSuccess: (result, _variables, _ctx) => {
+      queryClient.setQueriesData<ListResourceSubscription>(
+        {
+          queryKey: [
+            'subscriptions',
+            { organizationId: result.product.organization_id }
+          ],
+        },
+        (old) => {
+          if (!old) {
+            return {
+              items: [result],
+              pagination: {
+                total_count: 1,
+                max_page: 1,
+              },
+            }
+          } else {
+            return {
+              items: old.items.map((item) =>
+                item.id === result.id ? result : item,
+              ),
+              pagination: old.pagination,
+            }
+          }
+        },
+      )
+    },
   })
--- a/server/polar/customer_portal/endpoints/subscription.py
+++ b/server/polar/customer_portal/endpoints/subscription.py
@@ -1,5 +1,6 @@
 from typing import Annotated
 
+import structlog
 from fastapi import Depends, Path, Query
 from pydantic import UUID4
 
@@ -14,6 +15,7 @@ from polar.organization.schemas import OrganizationID
 from polar.postgres import get_db_session
 from polar.product.schemas import ProductID
 from polar.routing import APIRouter
+from polar.subscription.service import AlreadyCanceledSubscription
 
 from .. import auth
 from ..schemas.subscription import (
@@ -21,18 +23,19 @@ from ..schemas.subscription import (
     CustomerSubscriptionUpdate,
 )
 from ..service.subscription import (
-    AlreadyCanceledSubscription,
     CustomerSubscriptionSortProperty,
 )
 from ..service.subscription import customer_subscription as user_subscription_service
+
+log = structlog.get_logger()
 
 router = APIRouter(
     prefix="/subscriptions", tags=["subscriptions", APITag.documented, APITag.featured]
 )
 
-SubscriptionID = Annotated[UUID4, Path(description="The subscription ID.")]
+SubscriptionID = Annotated[UUID4, Path(description="Your subscription ID.")]
 SubscriptionNotFound = {
-    "description": "Subscription not found.",
+    "description": "Your subscription was not found.",
     "model": ResourceNotFound.schema(),
 }
 
@@ -108,7 +111,14 @@ async def get(
     summary="Update Subscription",
     response_model=CustomerSubscription,
     responses={
-        200: {"description": "Subscription updated."},
+        200: {"description": "Customer subscription updated."},
+        403: {
+            "description": (
+                "Customer subscription is already canceled "
+                "or will be at the end of the period."
+            ),
+            "model": AlreadyCanceledSubscription.schema(),
+        },
         404: SubscriptionNotFound,
     },
 )
@@ -124,8 +134,14 @@ async def update(
     if subscription is None:
         raise ResourceNotFound()
 
+    log.info(
+        "customer_portal.subscription.cancel",
+        id=id,
+        customer_id=auth_subject.subject.id,
+        updates=subscription_update,
+    )
     return await user_subscription_service.update(
-        session, subscription=subscription, subscription_update=subscription_update
+        session, subscription, updates=subscription_update
     )
 
 
@@ -134,10 +150,10 @@ async def update(
     summary="Cancel Subscription",
     response_model=CustomerSubscription,
     responses={
-        200: {"description": "Subscription canceled."},
+        200: {"description": "Customer subscription is canceled."},
         403: {
             "description": (
-                "This subscription is already canceled "
+                "Customer subscription is already canceled "
                 "or will be at the end of the period."
             ),
             "model": AlreadyCanceledSubscription.schema(),
@@ -156,4 +172,9 @@ async def cancel(
     if subscription is None:
         raise ResourceNotFound()
 
-    return await user_subscription_service.cancel(session, subscription=subscription)
+    log.info(
+        "customer_portal.subscription.cancel",
+        id=id,
+        customer_id=auth_subject.subject.id,
+    )
+    return await user_subscription_service.cancel(session, subscription)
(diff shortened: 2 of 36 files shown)


server/polar/integrations/stripe/service.py
```diff
@@ -468,11 +469,44 @@ async def update_subscription_price(
...
+            details["comment"] = customer_comment
+
+        params: stripe_lib.Subscription.ModifyParams = {
+            "cancellation_details": details,
+        }
+        if immediately:
+            # Need to be an epoch in the future. However, we want the `ended_at`
+            # to ideally have passed upon receiving the webhook. So hoping
+            # one second is a decent balance.
+            params["cancel_at"] = int(time.time()) + 1
```
frankie567: Couldn't we use the cancel API to achieve that?

https://docs.stripe.com/api/subscriptions/cancel

clients/apps/web/src/hooks/queries/subscriptions.ts
```diff
@@ -1,15 +1,23 @@
...
+  })
+}
+
 export const useListSubscriptions = (
   organizationId?: string,
   parameters?: Omit<SubscriptionsApiListRequest, 'organization_id'>,
 ) =>
   useQuery({
-    queryKey: ['subscriptions', { organizationId, ...(parameters || {}) }],
+    queryKey: ['subscriptions', organizationId, { ...(parameters || {}) }],
```
frankie567: I would prefer to keep everything as an object, much more [versatile](https://tanstack.com/query/latest/docs/framework/react/guides/query-invalidation).

clients/apps/web/src/hooks/queries/subscriptions.ts
```diff
@@ -1,15 +1,23 @@
...
 
-import { api } from '@/utils/api'
+import { api, queryClient } from '@/utils/api'
 import { defaultRetry } from './retry'
 
+export const invalidateSubscriptionCache = (
+  organizationId: string,
+) => {
+  queryClient.invalidateQueries({
+    queryKey: ['subscriptions', organizationId],
```
frankie567: You can perfectly do:

```ts
queryClient.invalidateQueries({
    queryKey: ['subscriptions', { organizationId }],
})
```

Even if there are more keys in the object, Tanstack will match them (unless you specify `exact: true`). This allows to very finely invalidate queries.

server/polar/subscription/schemas.py
```diff
@@ -72,3 +78,50 @@ class SubscriptionCreateEmail(Schema):
     product_id: UUID4 = Field(
         description="The ID of the product. **Must be the free subscription tier**."
     )
+
+
+class SubscriptionCancel(Schema):
+    immediately: bool = Field(
```
frankie567: I don't really like that naming. Maybe we could use Stripe naming `cancel_at_period_end`?

frankie567 (commented):
A few things :)

clients/apps/web/src/hooks/queries/subscriptions.ts
birkjernstrom: Awesome, will fix :)

server/polar/integrations/stripe/service.py
birkjernstrom: True 🤦🏼‍♂️ moment. Can't believe I didn't think of that... Fixing.

server/polar/subscription/schemas.py
birkjernstrom: Agree. I think I'll change it to `now=True` for immediate cancellation perhaps. 

**Immediate**
`polar.subscriptions.cancel(id, { now: true })`

**Period end (default)**
`polar.subscriptions.cancel(id)`

Feels like that reads well. `cancel_at_period_end` is intuitive as part of the `subscription` object, but feels a bit verbose and counterintuitive with `cancel_at_period_end=False` to flag immediate as part of the cancel method.

server/polar/subscription/endpoints.py
```diff
@@ -129,3 +138,73 @@ async def create_csv() -> AsyncGenerator[str, None]:
...
+        404: SubscriptionNotFound,
+    },
+)
+async def update(
+    id: SubscriptionID,
+    subscription_update: SubscriptionUpdate,
+    auth_subject: auth.SubscriptionsWrite,
+    session: AsyncSession = Depends(get_db_session),
+) -> Subscription:
+    """Update a subscription of the authenticated customer or user."""
```
frankie567: Smells like a copy/paste 😊 Just `Update a subscription.` in this context.

server/polar/subscription/endpoints.py
```diff
@@ -129,3 +138,73 @@ async def create_csv() -> AsyncGenerator[str, None]:
...
+    responses={
+        200: {"description": "Subscription revoked."},
+        403: {
+            "description": "This subscription is already revoked.",
+            "model": AlreadyCanceledSubscription.schema(),
+        },
+        404: SubscriptionNotFound,
+    },
+)
+async def revoke(
```
frankie567, line 194: Oh nice! I like the fact that `DELETE` is for cancelling immediately vs. `PATCH` is used to plan a cancellation 👍

server/polar/subscription/service.py
```diff
@@ -568,51 +591,427 @@ async def update_subscription_from_stripe(
...
+        if isinstance(updates, SubscriptionUpdatePrice):
+            return await self.update_product_price(
+                session,
+                subscription,
+                product_price_id=updates.product_price_id,
+            )
+
+        cancel = updates.cancel_at_period_end is True
+        uncancel = updates.cancel_at_period_end is False
+        if not (updates.revoke or cancel or uncancel):
```
frankie567: I don't understand this case, could use a comment :)

server/polar/subscription/service.py
```diff
@@ -568,51 +591,427 @@ async def update_subscription_from_stripe(
...
+        session: AsyncSession,
+        subscription: Subscription,
+    ) -> None:
+        await self._send_webhook(
+            session, subscription, WebhookEventType.subscription_active
+        )
+
+        # TODO: Copied this logic over. Why the extra check for customer
+        # confirmation vs. webhook/creator notification? Look into it, but avoid
+        # changing for now.
```
frankie567: This logic was triggered each time we received an update from Stripe. So this check was there to avoid sending a confirmation email at each billing cycle 🙃

But since here it's wrapped in a `_on_subscription_activated` method where we already know the subscription is becoming active for the first time, then it's safe to remove it.

server/polar/webhook/webhooks.py
```diff
@@ -542,8 +578,8 @@ def get_slack_payload(self, target: User | Organization) -> str:
 
 class WebhookSubscriptionCanceledPayload(WebhookSubscriptionUpdatedPayloadBase):
     """
-    Sent when a subscription is canceled by the user.
-    They might still have access until the end of the current period.
+    Sent when a subscription is canceled.
+    Customers might still have access until the end of the current period.
```
frankie567, line 582: Wow, nice catch 😄

server/polar/customer_portal/endpoints/subscription.py
```diff
@@ -14,25 +15,27 @@
...
+
 router = APIRouter(
     prefix="/subscriptions", tags=["subscriptions", APITag.documented, APITag.featured]
 )
 
-SubscriptionID = Annotated[UUID4, Path(description="The subscription ID.")]
+SubscriptionID = Annotated[UUID4, Path(description="Your subscription ID.")]
 SubscriptionNotFound = {
-    "description": "Subscription not found.",
+    "description": "Your subscription was not found.",
```
frankie567: Not sure about the "Your..." style. This is not something shown to the end-user, only in the API docs.

server/tests/subscription/test_endpoints.py
```diff
@@ -44,3 +74,543 @@ async def test_valid(
             assert "user" in item
             assert "customer" in item
             assert item["user"]["id"] == item["customer"]["id"]
+
+
+@pytest.mark.asyncio
+@pytest.mark.skip_db_asserts
```
frankie567: I removed the `skip_db_asserts` marker. You can do a search/replace safely.

server/tests/subscription/test_endpoints.py
```diff
@@ -44,3 +74,543 @@ async def test_valid(
...
+        assert updated_subscription["status"] == SubscriptionStatus.active
+        assert updated_subscription["cancel_at_period_end"] is False
+        assert updated_subscription["ends_at"] is None
+        assert updated_subscription["ended_at"] is None
+        assert updated_subscription["customer_cancellation_reason"] is None
+        assert updated_subscription["customer_cancellation_comment"] is None
+
+
+@pytest.mark.asyncio
+@pytest.mark.http_auto_expunge
```
frankie567: I remove the `http_auto_expunge` marker. You can do a search/replace safely.

frankie567 (changes requested):
Looks really good! The API for update/cancellation/revokation is really well thought. Just a few nitpick and it'll be good to go :rocket:

server/polar/customer_portal/endpoints/subscription.py
birkjernstrom: Fair and I agree. I wanted to distinguish it more as a few developers have historically been confused between core API endpoints vs. `/customer-portal/*`, but can achieve the same with more clarity vs. `your`. Will fix.

server/polar/subscription/endpoints.py
birkjernstrom: Haha guilty 😅 Nice catch

server/tests/subscription/test_endpoints.py
birkjernstrom: Nice catches & heads-up. Will remove.

server/polar/subscription/service.py
birkjernstrom: Nice catch. Commented, but also changed it to an `any()` to better readability. Just exiting early unless we're asked to revoke, cancel or uncancel.

PR #4633, opened 2024-12-14
clients/organisation: navigate to last visited organisation

### LINKED ISSUE #4619: Remember last organization when opening dashboard
### Description

Opening a fresh browser tab with the polar.sh dashboard should show the most relevant organization if there is more than one.

### Current Behavior

Currently it redirects to the first organization in the dropdown list, which is the most recently created one.

### Expected Behavior

Redirect to the organization that was most recently manually selected from the dropdown list, or some other metric like revenue, etc.

### Environment:

-   Operating System: [e.g., Windows, macOS, Linux]
-   Browser (if applicable): [e.g., Chrome, Firefox, Safari]

---

magarpratik:
Closes #4619 

This PR add functionality to remember last visited organization.

When an authenticated user visits the default home page, they will be redirected to the last visited organization's dashboard.

https://github.com/user-attachments/assets/555e3866-71bd-4d57-84b9-2763a8189191


frankie567:
Thank you for this @magarpratik 👍 That's interesting, but I would have liked a server-side approach. Basically, the logic is handled on the following route:

https://github.com/polarsource/polar/blob/07a4c0ad3e65bb2a6d5c290eba79050ed9c15593/clients/apps/web/src/app/(main)/start/page.tsx#L14-L23

Ideally, this would check for the existence of a cookie (instead of local storage) to directly redirect to the last organization at that point.

This cookie could be then be set client-side on the `DashboardLayout` component for example:

https://github.com/polarsource/polar/blob/ca270978bb99e1243d47ce2fbc608d358ac1ab47/clients/apps/web/src/components/Layout/DashboardLayout.tsx#L91-L120

What do you think? :)

magarpratik:
> Thank you for this @magarpratik 👍 That's interesting, but I would have liked a server-side approach. Basically, the logic is handled on the following route:
> 
> https://github.com/polarsource/polar/blob/07a4c0ad3e65bb2a6d5c290eba79050ed9c15593/clients/apps/web/src/app/(main)/start/page.tsx#L14-L23
> 
> Ideally, this would check for the existence of a cookie (instead of local storage) to directly redirect to the last organization at that point.
> 
> This cookie could be then be set client-side on the `DashboardLayout` component for example:
> 
> https://github.com/polarsource/polar/blob/ca270978bb99e1243d47ce2fbc608d358ac1ab47/clients/apps/web/src/components/Layout/DashboardLayout.tsx#L91-L120
> 
> What do you think? :)

Hi @frankie567,

Thank you for the feedback. 🙌 

It makes sense and it would probably be a bit more efficient too. I'll give it a go. 🫡

magarpratik:
Hi @frankie567, I've implemented your feedback.

PR is ready for re-review. 👍

clients/apps/web/src/utils/cookies.ts
```diff
@@ -0,0 +1,16 @@
+import { ReadonlyRequestCookies } from 'next/dist/server/web/spec-extension/adapters/request-cookies'
+
+const lastVisitedOrg = 'last_visited_org'
+
+export const setLastVisitedOrg = (
+  organization: string,
+  maxAge: number = 86400, // Expires in 1 day
```
frankie567: Could we make it longer, e.g. 30 days? One day seems short if you don't visit Polar every day :)

magarpratik: Done ✅

PR #4689, opened 2024-12-18
docs: Do not require authentication to view documentation

### LINKED ISSUE #4637: Docs: Some pages return `404` if user is not logged in
### Description

When a user is not logged in and attempts to access the following pages, they receive a `404` response.

Screenshot 2024-12-15 at 17 03 26

#### 404 response

Screenshot 2024-12-15 at 17 06 12

### Current Behavior

Users who are not logged in are unable to access some documentation pages.

### Expected Behavior

Users should be able to access all documentation pages without the need to log in.

### Screen recording

https://github.com/user-attachments/assets/c15bcc09-7bc8-4b5c-a32d-d52ee1538757

### Environment:

-   Operating System: macOS
-   Browser: Brave browser

magarpratik:
Closes #4637

This PR implements logic to ensure authentication is not required to view documentation.

### DIFF
--- a/clients/apps/web/src/middleware.ts
+++ b/clients/apps/web/src/middleware.ts
@@ -17,6 +17,13 @@ const AUTHENTICATED_ROUTES = [
 ]
 
 const requiresAuthentication = (request: NextRequest): boolean => {
+  if (
+    request.nextUrl.hostname === 'docs.polar.sh' ||
+    request.nextUrl.pathname.startsWith('/docs/')
+  ) {
+    return false
+  }
+
   return AUTHENTICATED_ROUTES.some((route) =>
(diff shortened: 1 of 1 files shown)


frankie567:
Hmm, I've just checked and was able to access to `docs.polar.sh` or `127.0.0.1:3000/docs` unauthenticated without any problem.

Under which circumstances did you hit this problem?

frankie567:
Ok, I get it now 👍 The problem comes from the fact that, on `docs.polar.sh`, we redirect pages like `/docs/finance/balance` to `/finance/balance`, which then falls under the authenticated route.

Not 100% sure the `.startsWith('/docs/')` is necessary, but that'll be good enough for now!

Thank you again @magarpratik 👍

PR #4751, opened 2025-01-02
Fix hardcoded year in email test data

birkjernstrom:
- **chore(tests): Fix hardcoded year in email test data**
- **chore(tests): Fix hardcoded year in magic link test data**


birkjernstrom:
Realized any future updates using `POLAR_TEST_RECORD` will overwrite the templates... The true solution to this is to have a real template and only store the body of the email in testdata vs. surrounding markup and generate new bodies with `POLAR_TEST_RECORD`. I'll create an issue for that and do it as clean-up.

Issue #4466, opened 2024-11-13
Webhooks: Show Endpoint URL in event log table


birkjernstrom:
### Description

Show the endpoint url we send webhooks to in the log. Helpful confirmation & for debugging of which endpoint received a given webhook.

MustaphaHamida:
![scrnli_zdkh6Jof848725](https://github.com/user-attachments/assets/6d2f7ac7-6a1f-4a13-a8ff-15ec8759a0bf)
im on the sandbox environment, and I am still getting the 307 error, i updated the URL to https://app.ranklly.com/api/webhook/polar

birkjernstrom:
@MustaphaHamida Shared this on Discord (let's continue the discussion there), but the URL provided above on your end responds with a `HTTP 307 /signup` redirect. Probably some authentication middleware that protects the route that you need to configure/change to allow the endpoint to be publicly available, but using our webhook secrets & validating the signature for security.

```
$ curl -vvv -X POST https://app.ranklly.com/api/webhook/polar

... redacted... 
< HTTP/2 307
< date: Wed, 13 Nov 2024 20:46:22 GMT
< location: /signup
< server: railway-edge
< x-railway-request-id: xTi5qLIBSj6Q9QY9ragzbQ_3500563535
<
* Connection #0 to host app.ranklly.com left intact
/signup%```

smokyabdulrahman:
@birkjernstrom I can take this up

smokyabdulrahman:
server part should be ready for review: https://github.com/polarsource/polar/pull/4802 🙏

smokyabdulrahman:
I starting to feel that I am heading towards the wrong direction.
I would love some clarification of what is needed?

My understanding is that we need to show for each webhook delivery the endpoint that that webhook attempted to call.
However, in the WebhookDelivery model, we only have a foreign key to WebhookEndpoint.
This means if someone updates a WebhookEndpoint, we wouldn't know old WebhookDeliveries called which endpoint.

Thus, we need to add a column to WebhookDelivery to store the endpoint that was called at that point of time.

Second part, showing that in the UI, how do you want to show that? (Example photo attached)
1. show it as a new column in the deliveries table in the UI
2. show it in the details section of the delivery (after clicking on a row in the table - it expands to show extra info including the webhook endpoint that we are going to add)

[...]

PR #4802, opened 2025-01-07
#4466 Enrich /webhooks/deliveries endpoint to include webhook endpoint schema

smokyabdulrahman:
[WIP]
Addresses #4466 

P.S.
I believe backend changes should be deployed first before deploying frontend ones, so there would be no downtime/errors window as a result of frontend updating first.

### DIFF
--- a/server/polar/webhook/schemas.py
+++ b/server/polar/webhook/schemas.py
@@ -128,3 +128,6 @@ class WebhookDelivery(TimestampedSchema):
     webhook_event: WebhookEvent = Field(
         description="The webhook event sent by this delivery."
     )
+    webhook_endpoint: WebhookEndpoint = Field(
+        description="The webhook endpoint called by this delivery."
+    )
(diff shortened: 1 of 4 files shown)


server/polar/webhook/schemas.py
```diff
@@ -128,3 +128,6 @@ class WebhookDelivery(TimestampedSchema):
     webhook_event: WebhookEvent = Field(
         description="The webhook event sent by this delivery."
     )
+    webhook_endpoint: WebhookEndpoint = Field(
+        description="The webhook endpoint called by this delivery."
+    )
```
smokyabdulrahman, line 133: Is returning the whole schema too much?
Should I be adding a `webhook_endpoint_url: EndpointUrl` instead and load it like:
```
[WebhookDeliverySchema.model_validate({**result, "webhook_endpoint_url": result.webhook_endpoint.url) for result in results],
```

Mizokuiam:
Thank you for raising this issue! I'll look into it and try to help if I can.

smokyabdulrahman:
> Thank you for raising this issue! I'll look into it and try to help if I can.

Thank you so much. 

I've raised some questions on the original issue. So maybe after answering those questions, the solution might change.

https://github.com/polarsource/polar/issues/4466#issuecomment-2575410083

PR #4805, opened 2025-01-07
Add Scalar for API Docs

emilwidlund:


antlio:
@emilwidlund love polar.sh! feel free to hmu if you need anything regarding using scalar!

PR #4837, opened 2025-01-16
Refunds

birkjernstrom:
Fortunately refunds across sellers on Polar has historically been super low (< 1%). Given our fast support, we've been able to help sellers in those cases with refunds within minutes/hours. But of course, a billing platform cannot live without built-in refund support 🙊 So time to solve this table stake once and for all.

**Before merge:**
- [x] Sync order aggregated refund stats on refunds
- [x] Handle `refund.failed` from Stripe (upsert - we don't know if it's possible and cannot trigger...)
- [x] Remove embedding `order.refunds`
- [x] Remove `charge.refunded` handler in favour of `refund.created|updated`
- [x] Webhooks for `refund.created` and `refund.updated`
- [x] `GET /v1/refunds` with filters on `order_id`, `subscription_id`, `customer_id`
- [x] UI to show refunds in order list view
- [x] Setting to revoke benefits on refund, i.e one-time purchases

**UI Follow-up (Quick)**
- [x] UI: Create refund from order dashboard
- [ ] UI: Show orders on cancellation of subscription - prompt to refund latest charge?


server/polar/kit/db/models/base.py
```diff
@@ -43,6 +43,9 @@ class TimestampedModel(Model):
     def set_deleted_at(self) -> None:
         self.deleted_at = utc_now()
 
+    def set_modified(self) -> None:
```
frankie567: You don't need that: `modified_at` is configured with `onupdate=utc_now`, so SQLAlchemy will automatically update the date when `session.add` is called.

server/polar/models/order.py
```diff
@@ -100,6 +114,59 @@ def subscription(cls) -> Mapped["Subscription"]:
...
+        return self.refundable_amount + self.refundable_tax_amount
+
+    def increment_refunds(self, refunded_amount: int, refunded_tax_amount: int) -> None:
+        new_amount = self.refunded_amount + refunded_amount
+        new_tax_amount = self.refunded_tax_amount + refunded_tax_amount
+        exceeds_original_amount = (
+            new_amount > self.amount or new_tax_amount > self.tax_amount
+        )
+        if exceeds_original_amount:
+            raise ValueError(
```
frankie567: Could we have our own exception for this? Might be easier to track in Sentry. Example: https://github.com/polarsource/polar/blob/9870b4b505aeaaf18cb3d55c1239cd428ee9ef0e/server/polar/models/benefit.py#L16-L21

server/polar/models/transaction.py
```diff
@@ -40,15 +42,6 @@ class TransactionType(StrEnum):
     """Money paid to the user's bank account."""
 
 
-class PaymentProcessor(StrEnum):
```
frankie567, line 43: Heads up on what we mentioned about this :) We should keep that one, but name it better to avoid confusion with the other one. Maybe `Processor` is good enough.

server/polar/pledge/service.py
```diff
@@ -764,7 +764,6 @@ async def refund_by_payment_id(
                 transaction_id=transaction_id,
             )
         )
-        await session.commit()
```
frankie567, line 767: 🥳

server/polar/refund/endpoints.py
```diff
@@ -0,0 +1,113 @@
...
+from .schemas import RefundCreate, RefundID
+from .service import RefundAmountTooHigh, RefundedAlready
+from .service import refund as refund_service
+from .sorting import RefundListSorting
+
+log = structlog.get_logger()
+
+router = APIRouter(prefix="/refunds", tags=["refunds", APITag.documented])
+
+OrderNotFound = {
```
frankie567: `RefundNotFound` (+ description)

server/polar/refund/endpoints.py
```diff
@@ -0,0 +1,113 @@
...
+    order_id: MultipleQueryFilter[OrderID] | None = Query(
+        None, title="OrderID Filter", description="Filter by order ID."
+    ),
+    subscription_id: MultipleQueryFilter[SubscriptionID] | None = Query(
+        None, title="SubscriptionID Filter", description="Filter by subscription ID."
+    ),
+    customer_id: MultipleQueryFilter[CustomerID] | None = Query(
+        None, title="CustomerID Filter", description="Filter by customer ID."
+    ),
+    succeeded: bool = Query(
```
frankie567: Personally, I prefer to avoid having implicitly enabled filter. The reason is that a developer might want for various reasons to get _all_ refunds, no matter the success status. That's the strategy I usually apply in other APIs.

So I would define it as `bool | None` with a `None` default.

server/polar/refund/schemas.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def get_amount_display(self) -> str:
+        return f"{format_currency(
+            self.amount / 100,
+            self.currency.upper(),
+            locale="en_US",
+        )}"
+
+
+class RefundUpdate(MetadataInputMixin, Schema):
```
frankie567: Looks unused

server/polar/refund/service.py
```diff
@@ -0,0 +1,581 @@
...
+        if order_id is not None:
+            statement = statement.where(Refund.order_id.in_(order_id))
+
+        if subscription_id is not None:
+            statement = statement.where(Refund.subscription_id.in_(subscription_id))
+
+        if customer_id is not None:
+            statement = statement.where(Refund.customer_id.in_(customer_id))
+
+        if succeeded:
```
frankie567, line 125: See previous comment about default `succeeded` filter :)

server/polar/refund/endpoints.py
```diff
@@ -0,0 +1,113 @@
...
+    auth_subject: auth.RefundsWrite,
+    session: AsyncSession = Depends(get_db_session),
+) -> Refund:
+    """Create a refund."""
+    order = await order_service.user_get_by_id(
+        session,
+        auth_subject,
+        refund_create.order_id,
+    )
+    if not order:
```
frankie567: Kind of a nitpick, but I would move this to the `create` method and raise a `PolarRequestValidationError` (422) instead of a 404.

server/polar/refund/service.py
```diff
@@ -0,0 +1,581 @@
...
+        return await paginate(session, statement, pagination=pagination)
+
+    async def create(
+        self,
+        session: AsyncSession,
+        order: Order,
+        create_schema: RefundCreate,
+    ) -> Refund:
+        if order.refunded:
+            raise RefundedAlready(order)
```
frankie567, line 172: Similar to previous comment: could be a `PolarRequestValidationError`

server/polar/refund/service.py
```diff
@@ -0,0 +1,581 @@
...
+
+        is_subscription = order.subscription_id is not None
+        if create_schema.revoke_benefits and is_subscription:
+            raise RevokeSubscriptionBenefitsProhibited()
```
frankie567, line 176: Similar to previous comment: could be a `PolarRequestValidationError`

server/polar/transaction/service/payment.py
```diff
@@ -32,6 +34,24 @@ def __init__(self, charge_id: str, payment_intent_id: str) -> None:
 
 
 class PaymentTransactionService(BaseTransactionService):
+    async def get_by_charge_id(
```
frankie567, line 36: Duplicate of `_get_by_charge_id`. You can make it non-private if you need it elsewhere.

server/polar/webhook/webhooks.py
```diff
@@ -250,6 +257,84 @@ def get_slack_payload(self, target: User | Organization) -> str:
...
+
+        amount_display = self.data.get_refunded_amount_display()
+
+        fields: list[DiscordEmbedField] = [
+            {"name": "Product", "value": self.data.product.name},
+            {"name": "Refunded", "value": amount_display},
+            {"name": "Customer", "value": self.data.customer.email},
+        ]
+        if self.data.subscription is not None:
+            fields.append({"name": "Subscription", "value": "Yes"})
```
frankie567, line 281: Is it really useful? 😊 They can probably guess it from the product name.

server/polar/webhook/webhooks.py
```diff
@@ -250,6 +257,84 @@ def get_slack_payload(self, target: User | Organization) -> str:
...
+
+        amount_display = self.data.get_refunded_amount_display()
+
+        fields: list[SlackText] = [
+            {"type": "mrkdwn", "text": f"*Product*\n{self.data.product.name}"},
+            {"type": "mrkdwn", "text": f"*Refunded*\n{amount_display}"},
+            {"type": "mrkdwn", "text": f"*Customer*\n{self.data.customer.email}"},
+        ]
+        if self.data.subscription is not None:
+            fields.append({"type": "mrkdwn", "text": "*Subscription*\nYes"})
```
frankie567, line 313: See previous comment

server/polar/webhook/webhooks.py
```diff
@@ -647,6 +732,187 @@ def get_slack_payload(self, target: User | Organization) -> str:
...
+            {"type": "mrkdwn", "text": f"*Status*\n{self.data.status}"},
+        ]
+        return fields
+
+
+class WebhookRefundCreatedPayload(WebhookRefundBase):
+    """
+    Sent when a refund is created regardless of status.
+
+    **Discord & Slack support:** Basic
```
frankie567: Looks like a `Full` support since you defined a specific payload :)

server/polar/webhook/webhooks.py
```diff
@@ -647,6 +732,187 @@ def get_slack_payload(self, target: User | Organization) -> str:
...
+            }
+        )
+        return json.dumps(payload)
+
+
+class WebhookRefundSucceededPayload(WebhookRefundBase):
+    """
+    Sent when a refund succeeds.
+
+    **Discord & Slack support:** Basic
```
frankie567: Looks like a `Full` support since you defined a specific payload :)

server/polar/webhook/webhooks.py
```diff
@@ -647,6 +732,187 @@ def get_slack_payload(self, target: User | Organization) -> str:
...
+            }
+        )
+        return json.dumps(payload)
+
+
+class WebhookRefundUpdatedPayload(WebhookRefundBase):
+    """
+    Sent when a refund is updated.
+
+    **Discord & Slack support:** Basic
```
frankie567: Looks like a `Full` support since you defined a specific payload :)

frankie567 (dismissed):
A few things there and there, the main thing being that `PaymentProcessor` enum thing :)

server/polar/refund/endpoints.py
birkjernstrom: Actually, it's correct with `OrderNotFound` here since it's thrown in case the order is not found - even though it looks like copy pasta :) Centralized it though to share the same definition from the order schema module.

server/polar/refund/endpoints.py
birkjernstrom: Changing now to stay with the convention, but I'm not sure I agree for the long-term on the principle 🙊 Fun discussion to have, but I think I might lean towards optimizing for the smallest implementation footprint for the most common use cases vs. the more correct RESTful approach to get all resources. E.g most developers likely only caring about the succeeded refunds even though it would feel counterintuitive to have to "unfilter" in case you want them all 🤔

server/polar/refund/endpoints.py
birkjernstrom: Fixing + getting rid of 404/OrderNotFound

server/polar/webhook/webhooks.py
birkjernstrom: Fair point. Leaving for now though to mimic `OrderCreated` since it's weird if they're different for the same object in that sense, but might remove on both.

PR #5191, opened 2025-03-06
clients/packages/ui: fix double scrollbar

qwtel:
Fixes double scrollbars. The outer scrollbar is unnecessary. Currently it always overflows by `.5rem` (caused by padding-top of the child). As best as I can tell there's never a reason to overflow the outermost wrapper, so fixed via clipping.

Screenshot 2025-03-06 at 13 46 51-2


frankie567:
Thank you for the fix @qwtel 🙏  I tweaked it a bit so it doesn't break scrolling on mobile, but otherwise, really nice catch 👍

PR #5245, opened 2025-03-11
Customer deletion improvements

### LINKED ISSUE #5142: Customers: Soft deletion does not free up the email
New issue to sync with Linear from [https://github.com/polarsource/polar/issues/4816](https://github.com/polarsource/polar/issues/4816)

### Description

We soft delete customers, but the unique index on `organization_id` and `email` prevents the email from being freed up for a new customer.

### Expected Behavior

We could probably solve this by modifying the index to include the `deleted_at` timestamp. Causing a unique violation in case of creating a customer with the same email as an existing – non-deleted – customer, but not in case others have a unique deleted_at timestamp.

### LINKED ISSUE #5202: Customers: Delete customer from dashboard
Ability to delete a customer from within the dashboard.

Wait on: [https://github.com/polarsource/polar/issues/5142](https://github.com/polarsource/polar/issues/5142)

So we can fix that bug + disallow deletion of customers who have purchased something, i.e has an order/benefit.

birkjernstrom:
Fixes #5142 #5202

**Todo**
- [x] Chase down all queries with `customer` to ensure they always include an `deleted_at IS NULL` constraint
- [x] Fix #5202 to make testing more convenient + close that issue while we're at it


frankie567:
Thought about that the other day and actually, it won't work: in SQL, `NULL` values are not considered equal. So this will pass:

```sql
INSERT INTO customers VALUES ("cus@example.com", NULL);
INSERT INTO customers VALUES ("cus@example.com", NULL); -- This will succeed
```

server/polar/customer/repository.py
```diff
@@ -47,6 +47,14 @@ async def update(
 
     async def soft_delete(self, object: Customer, *, flush: bool = False) -> Customer:
         customer = await super().soft_delete(object, flush=flush)
+        # Clear external_id for future recycling
```
frankie567, line 50: Could do we put this in the repo? So we don't accidentally forget this operation if we implement it somewhere else.

frankie567 (commented):
Looks good! Just one small nitpick :)

server/polar/customer/repository.py
birkjernstrom: Already done 🤓

PR #5311, opened 2025-03-20
docs: escape $ to fix latex styling

flashblaze:


frankie567 (approved):
Thank you, good catch 🙏

PR #5314, opened 2025-03-20
fix: show 0 in average order value when no order is placed

criskell:
Before with with an account that has just been created (show $NaN in average order value):
image
After:
image

This prevents division by zero, which returns NaN.


frankie567:
Good catch, thank you 🙏

PR #5404, opened 2025-04-02
Support custom URI schemes

emilwidlund:


emilwidlund:
@frankie567 I assume something like this? Still prohibits `http://` schemes - but allows `polar://` for example.

frankie567:
Yes, exactly 👍

PR #5433, opened 2025-04-05
docs: only allow authenticated users to checkout

arsenstorm:
https://github.com/polarsource/polar-adapters/pull/177 extends the Better Auth plugin with an **optional** `onlyAuthenticated` parameter to restrict the creation of checkout sessions only to authenticated users.

This pull request updates the documentation to reflect this change.


frankie567:
Outdated with new plugin version

PR #5751, opened 2025-05-22
feat(client): add local compare to the product sort

romadryud:
Hi, thank you for the project!

This pr aims to fix sorting on the product page:

|  Before | After  |
|---|---|
|  image |  image |

The problem with sorting appears on every page where we display a list, so we could extract the local sorting function to a `lib/utils` directory and reuse it wherever items are displayed in a list. In scope of this PR I have covered only the products list.

### DIFF
--- a/clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/products/ClientPage.tsx
+++ b/clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/products/ClientPage.tsx
@@ -54,7 +54,7 @@ import Markdown from 'markdown-to-jsx'
 import Link from 'next/link'
 import { usePathname, useRouter } from 'next/navigation'
 import { useQueryState } from 'nuqs'
-import { useCallback, useState } from 'react'
+import { useCallback, useMemo, useState } from 'react'
 import { twMerge } from 'tailwind-merge'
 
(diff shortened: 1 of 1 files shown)


romadryud:
@birkjernstrom  @frankie567  Could you please review this one?

emilwidlund:
We should preferably sort by creation timestamp, not alphanumerical.

romadryud:
@emilwidlund I believe the alphanumerical sorting is more user-friendly. Since users primarily read this type of data, they tend to rely on alphabetical or numerical order to find information quickly. While sorting by creation timestamp was suggested, the data is already sorted this way on the backend. Therefore, client-side alphanumerical sorting offers a better user experience for the displayed information. For now on the frontend we have standard js sorting with known problem like in the screenshot.

romadryud:
@psincraian @birkjernstrom @frankie567 @emilwidlund Where we can discuss it? My approach described here: https://github.com/polarsource/polar/pull/5751#issuecomment-2972898827

Thank you!

psincraian:
Probably I will open a discussion. I'm not sure if I'm really convinced by the argument. The problem with sorting in alphanumeric order is that if I have two products:

1. A
2. C

If I create another product B it will just be put in the middle of both, where I would expect to be put at the end.

PR #5930, opened 2025-06-17
Removed skeletons for no product description in checkout

### LINKED ISSUE #5897: Product with no description shows skeleton on checkout page
Shouldn't fallback to the skeleton but just show nothing

mshra:
Fixes #5897


frankie567:
Thank you, much appreciated 🙏

PR #5973, opened 2025-06-22
Fix: validate avatar URL to disallow special unicode characters

### LINKED ISSUE #5734: Unicode filenames not being rendered
Customer added a link to their organization avatar which was a file with special characters causing the content security policy derective to deny it.

Changing the filename to `logo.jpg` fixed the issue in this case. 

```
Refused to load the image 'https://<truncated_for_privacy>/%E5%A4%B4%E5%83%8F800.jpg' because it violates the following Content Security Policy directive: "img-src 'self' blob: data:

```

One solution here could be to simply add validation on the field in organization settings that if we detect special – escaped – characters in the URL we request that they change it to A-Za-z0-9 characters instead

ANISH-SR:
This PR adds the client side validation to prevent uploading avatars with special Unicode characters in filenames using simple Regex validation.
Fixes #5734

### DIFF
--- a/clients/apps/web/src/components/Settings/OrganizationProfileSettings.tsx
+++ b/clients/apps/web/src/components/Settings/OrganizationProfileSettings.tsx
@@ -207,6 +207,15 @@ export const OrganizationDetailsForm: React.FC<
         return
       }
       const lastFile = files[files.length - 1]
+      const fileName = lastFile.name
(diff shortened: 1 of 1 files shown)


frankie567:
Thank you for having a look at this @ANISH-SR. Unfortunately, I think the issue report is actually wrong. Our recently updated CSP don't allow external avatar URL, this has not something to do with special chars.

sranish:
@frankie567 Thanks for the follow up  : )

PR #5975, opened 2025-06-22
docs: fix introduction links

lazakrisz:
Hi, noticed some issues with the links:
* better-auth mistyped
* Nuxt listed twice
* Laravel missing from introduction.


docs/introduction.mdx
```diff
@@ -121,8 +121,7 @@ They can be used to:
 We currently have adapters for the most popular JavaScript frameworks:
```
frankie567: If you allow me a small nitpick: if we add Laravel now, we should remove "JavaScript" here 🙂

frankie567 (changes requested):
Thank you @lazakrisz, that's helpful 🙏 If I may ask you to remove the "JavaScript" mention (see comment below), that would be awesome!

docs/introduction.mdx
kriszlazar: Okay, totally right, completely glanced over that part 😅 , just made the changes.

Issue #5208, opened 2025-03-09
Access Token: Sort scopes alphabetically in create modal


birkjernstrom:
<img src="https://uploads.linear.app/5cb99480-3e0b-4cc4-a2f7-c61944b50350/6522b899-a18b-4463-a905-70e4f2281687/60cd04bc-d451-44dd-b9aa-d04866de46cc?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzVjYjk5NDgwLTNlMGItNGNjNC1hMmY3LWM2MTk0NGI1MDM1MC82NTIyYjg5OS1hMThiLTQ0NjMtYTkwNS03MGU0ZjIyODE2ODcvNjBjZDA0YmMtZDQ1MS00NGRkLWI5YWEtZDA0ODY2ZGU0NmNjIiwiaWF0IjoxNzQxNTUzMzYzLCJleHAiOjMzMzEyMTEzMzYzfQ.jhg4h5AvJ0zmGnf_p9Ka3jdlwKr4bnnV8Q5g1JWqaY0 " alt="image.png" width="3456" data-linear-height="1946" />

Sort them on the client-side

kriszlazar:
Hi!
I've added this PR: https://github.com/polarsource/polar/pull/6009, let me know if something else is needed. Thanks!

frankie567:
Fixed by #6009

PR #6009, opened 2025-06-25
Sort scopes alphabetically

lazakrisz:
Hi!
I've added the implementation for https://github.com/polarsource/polar/issues/5208. Let me know if I need to make any other changes!

### DIFF
--- a/clients/apps/web/src/components/Settings/OAuth/OAuthForm.tsx
+++ b/clients/apps/web/src/components/Settings/OAuth/OAuthForm.tsx
@@ -253,6 +253,9 @@ export const FieldRedirectURIs = () => {
 
 export const FieldScopes = () => {
   const form = useFormContext<EnhancedOAuth2ClientConfiguration>()
+  const sortedAvailableScopes = Array.from(enums.availableScopeValues).sort(
+    (a, b) => a.localeCompare(b),
+  )
 
   return (
     <div className="flex flex-col gap-6">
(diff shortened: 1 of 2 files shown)


clients/apps/web/src/components/Settings/OAuth/OAuthForm.tsx
```diff
@@ -253,6 +253,9 @@ export const FieldRedirectURIs = () => {
 
 export const FieldScopes = () => {
   const form = useFormContext<EnhancedOAuth2ClientConfiguration>()
+  const sortedAvailableScopes = Array.from(enums.availableScopeValues).sort(
```
lazakrisz, line 256: I used `Array.from`, instead of casting it.

kriszlazar, line 256: I noticed that there is another use for the scopes in a different component so I also sorted those.

clients/apps/web/src/components/Settings/OrganizationAccessTokensSettings.tsx
```diff
@@ -52,8 +52,9 @@ const AccessTokenForm = ({ update }: { update?: boolean }) => {
     AccessTokenCreate | AccessTokenUpdate
   >()
 
-  const selectableScopes =
-    enums.availableScopeValues as schemas['AvailableScope'][]
+  const sortedScopes = Array.from(enums.availableScopeValues).sort((a, b) =>
```
kriszlazar, line 55: Also used `Array.from` instead of type casting, which seems to remove the `readonly[]` specifier.

frankie567:
Thank you, that's very helpful 🙏

PR #6021, opened 2025-06-26
backoffice: Pledge transfer view

birkjernstrom:


frankie567 (approved):
That's a quick and dirty UI for sure, but yeah, that should do the trick :)

birkjernstrom:
> That's a quick and dirty UI for sure, but yeah, that should do the trick :)

Claude Code one-shot 😉 Definitely not the prettiest of UIs, but will use this feature for maybe 10 pledges before removing all of this code so more than fine for the purpose :)

PR #6037, opened 2025-06-30
server/order: trigger order.updated after order creation if it's immediately paid

frankie567:
- server/order: trigger order.updated after order creation if it's immediately paid


developedbyumair:
Thanks @frankie567

Issue #5470, opened 2025-04-11
Can't create a discount having the same code as a soft deleted one


frankie567:

mrgoonie:
![image](https://github.com/user-attachments/assets/613a6ce2-7cab-459e-86ef-439c41bd2c2c)

are there any updates @frankie567 ? 😓

malthejorgensen:
Hi @mrgoonie – should be fixed now!

PR #6054, opened 2025-07-01
server/discount: Exclude soft-delete Discounts in unique index

malthejorgensen:
This PR changes the unique constraint `("organisation_id", "code")` on the `Discount`-model to match case-insensitively on `code` and to exclude soft-deleted objects/rows in the index. This is using a Postgres partial index.

Fixes: https://github.com/polarsource/polar/issues/5470


server/polar/models/discount.py
```diff
@@ -57,7 +57,24 @@ class DiscountDuration(StrEnum):
 
 class Discount(MetadataMixin, RecordModel):
     __tablename__ = "discounts"
-    __table_args__ = (UniqueConstraint("organization_id", "code"),)
+
+    @declared_attr
+    def __table_args__(cls):
```
frankie567: I think mypy complains because of missing type annotations here. Two things:

* SQLAlchemy recommends to use `@declared_attr.directive` to please type checkers: https://docs.sqlalchemy.org/en/20/orm/mapping_api.html#sqlalchemy.orm.declared_attr.directive
* You should explicitly set a return type; say, `Sequence[Index]` for ease

---

You can run the type checker like this:

```
uv run task lint_types
```

Or all in one:

```
uv run task lint && uv run task lint_types
```

server/polar/models/discount.py
```diff
@@ -57,7 +58,27 @@ class DiscountDuration(StrEnum):
...
+                Column[datetime | None], cls.deleted_at
+            )  # cast to satisfy mypy
+            cls._memoized_indexes = (
+                Index(
+                    "ix_discounts_code_uniqueness",
+                    "organization_id",
+                    func.lower(cls.code),
+                    unique=True,
+                    # partial index
+                    postgresql_where=(_deleted_at_column.is_(None)),
```
frankie567, line 78: Don't really get why `cls.*` properties are not correctly typed as mapped properties but well. We could also do directly `Discount.deleted_at.is_(None)`, if you think that's better :)

malthejorgensen: I think the issue with `Discount.deleted_at.is_(None)` is that we get a runtime error complaining `Discount` isn't defined. Weirdly, `mypy` doesn't catch that 🤷 (but we get an error when running the tests)

frankie567: Okay, then, let's go with that 😅

PR #6065, opened 2025-07-02
Remove `/personal_access_tokens`-API

malthejorgensen:
This API is unused.
We can't remove the models/tables as they might still be used.


frankie567:
One last thing: update our internal JS client to remove those endpoints:

```sh
cd clients
pnpm generate  # Make sure the API is running locally, so it can pull the OpenAPI schema
```

malthejorgensen:
Did that now ✅ 

But I'm now wondering if this is a good idea. Do we need to warn people? – I guess they _could_ be using `/v1/personal_access_tokens` to list their tokens/read them.

frankie567:
> But I'm now wondering if this is a good idea. Do we need to warn people? – I guess they _could_ be using `/v1/personal_access_tokens` to list their tokens/read them.

Actually, it wasn't a public API:

* It wasn't documented (see `APITag.private`)
* It was only accepting web sessions (see `WebUser`)

So I think it's safe in that regard :)

PR #6079, opened 2025-07-03
Sidebar, account dropdown: Show "pointer" cursor to indicate clickable

malthejorgensen:
Previously, the items in the account dropdown were showing with the
neutral cursor. We now use the "pointer" cursor to indicate they're
clickable.

### DIFF
--- a/clients/apps/web/src/components/Layout/Dashboard/DashboardSidebar.tsx
+++ b/clients/apps/web/src/components/Layout/Dashboard/DashboardSidebar.tsx
@@ -145,7 +145,7 @@ export const DashboardSidebar = ({
                   {organizations.map((org) => (
                     <DropdownMenuItem
                       key={org.id}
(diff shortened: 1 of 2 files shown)


malthejorgensen:
Another, more generic way to do this could be to add `cursor-pointer` automatically to the core DropdownMenuItem-component when the `onClick`-prop is passed.

emilwidlund:
> Another, more generic way to do this could be to add `cursor-pointer` automatically to the core DropdownMenuItem-component when the `onClick`-prop is passed.

Yeah, I would prefer to add it to the DropdownMenuItem-component. However - we can't really apply it directly to that component, as those files are automatically created by the shadcn-registry.

That's why we have a few duplicated components in our `atoms` folder. They're used as proxies - using the Shadcn component, but adding a className with custom styling.

I can go ahead and create a proxied version that has cursor-pointer applied to it :)

PR #6086, opened 2025-07-04
Login: Use 6-character code instead of magic link

malthejorgensen:
This switches logins and signups from previously using a magic link in an email to log in, to now instead using a 6-character code in an email that allows the login or signup.


server/polar/login_code/endpoints.py
```diff
@@ -0,0 +1,80 @@
...
+from polar.openapi import APITag
+from polar.postgres import get_db_session
+from polar.posthog import posthog
+from polar.routing import APIRouter
+
+from .schemas import LoginCodeRequest
+from .service import LoginCodeError
+from .service import login_code as login_code_service
+
+router = APIRouter(prefix="/login-code", tags=["login_code", APITag.documented])
```
frankie567: We should mark this as `APITag.private` instead of `APITag.documented`, so they're not in the docs and in the SDK.

server/polar/models/login_code.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class LoginCode(RecordModel):
+    __tablename__ = "login_codes"
+
+    code_hash: Mapped[str] = mapped_column(CHAR(64), nullable=False, index=True)
+    expires_at: Mapped[datetime] = mapped_column(
+        TIMESTAMP(timezone=True), nullable=False, index=True
+    )
+    email: Mapped[str] = mapped_column(String(320), nullable=False, index=True)
```
frankie567, line 23: We might indeed remove the index on `email`

server/polar/models/login_code.py
```diff
@@ -0,0 +1,31 @@
...
+from polar.kit.db.models import RecordModel
+
+if TYPE_CHECKING:
+    from polar.models import User
+
+
+class LoginCode(RecordModel):
+    __tablename__ = "login_codes"
+
+    code_hash: Mapped[str] = mapped_column(CHAR(64), nullable=False, index=True)
```
frankie567: We might indeed set a unique constraint on `code_hash`

frankie567 (changes requested):
Looks great 👍 One small thing regarding the APITag, and a few tweaks to the DB indices and we should be good to go!

server/polar/login_code/endpoints.py
```diff
@@ -0,0 +1,80 @@
...
+        return_to=login_code_request.return_to,
+        signup_attribution=login_code_request.attribution,
+    )
+
+    # Send the code email
+    await login_code_service.send(code_model, code)
+
+
+@router.post("/authenticate")
+async def authenticate_login_code(
```
psincraian, line 46: To make sure that the person who requested the code and authenticates it's the same, maybe we have the email in the request and perform the validation with code + email.

server/polar/login_code/endpoints.py
```diff
@@ -0,0 +1,80 @@
...
+
+from .schemas import LoginCodeRequest
+from .service import LoginCodeError
+from .service import login_code as login_code_service
+
+router = APIRouter(prefix="/login-code", tags=["login_code", APITag.documented])
+
+
+@router.post("/request", status_code=202)
+async def request_login_code(
```
psincraian, line 27: Not blocking - To avoid sending spam to random emails, we should probably add a CAPTCHA

server/polar/login_code/endpoints.py
malthejorgensen: Fixed ✅

server/polar/models/login_code.py
malthejorgensen: Fixed as well ✅

server/polar/login_code/endpoints.py
malthejorgensen: Done ✅

Issue #5627, opened 2025-05-02
Redirect `/purchases/(products|subscriptions)/{id}`


birkjernstrom:
Currently they 404 which is bad since they are linked in old (<=2024) customer emails. So older customers going back to download files might get a bad experience.

Would be great if we could do a lookup on the ID and redirect to the appropriate customer portal.

PR #6087, opened 2025-07-04
feat: add legacy product and subscription purchase pages

psincraian:
# Context
This is related to #5627, where old emails have been sent with `/purchases/(products|subscriptions)/{id}`. This will redirect the user to the customer portal of the organization, so `/${organization.slug}/portal`.

## Open questions
1. Does anyone has any email to make sure I'm redirecting the right url?
1. Should we log the errors to keep track of them?


clients/apps/web/src/app/(main)/purchases/products/[id]/page.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+  try {
+    product = await getProductById(api, params.id)
+    organization = await unwrap(
+      api.GET('/v1/organizations/{id}', {
+        params: {
+          path: { id: product.organization_id },
+        },
+      }),
+    )
```
Comment (frankie567) [inline clients/apps/web/src/app/(main)/purchases/products/[id]/page.tsx]: I'm afraid this won't work 😊

This link will probably be opened by customers that are not logged in. Even if they were, customers don't have permission to call private API. So, calling `/v1/products` and `/v1/organizations/{id}` will likely fail with a 401.

Handling this will I think require that we add a special (temporary) storefront endpoint: https://github.com/polarsource/polar/blob/main/server/polar/storefront/endpoints.py Something that can let us retrieve without a token an organization slug from a product ID or a subscription ID, and nothing more (we don't want to leak data here).

Comment (psincraian) [reply clients/apps/web/src/app/(main)/purchases/products/[id]/page.tsx]: ok, now I see it. I thought that the API was using a special token to have cross organization permissions but it may be only using the cookies to call it. Let me give another look then.

frankie567 (approved):
Perfect, that should do the trick :)

PR #6088, opened 2025-07-04
Fix meter credits empty benefits issue #5972

### LINKED ISSUE #5972: Meters: Not showing benefits in the order details page
Customer tried metered credits and buying them as a customer. Seeing them in the customer portal, but the list of benefits for the order of that subscription showed an empty list of benefits.

amankitsingh:
### Issue
Customers with metered subscriptions see empty benefits in the customer portal, even though the products have metered pricing configured.

### Root Cause
- Products with metered pricing lack associated `meter_credit` benefits
- Customer portal repository doesn't load product benefits relationship

### Changes
- Add migration to create missing meter_credit benefits for metered products
- Update customer portal to load product benefits via selectinload
- Add test coverage

### Testing
Verified locally that metered products now show "1000 API Call Credits" in customer portal.

Fixes #5972

### DIFF
--- a/server/migrations/versions/2025-07-05-0028_fix_meter_credits_empty_benefits_5972.py
+++ b/server/migrations/versions/2025-07-05-0028_fix_meter_credits_empty_benefits_5972.py
@@ -1,0 +1,87 @@
+"""Fix meter credits empty benefits
+
+Revision ID: 5972_fix_meter_credits
+Revises: 4b8976c08210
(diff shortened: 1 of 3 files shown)


frankie567:
Thank you for having a look @amankitsingh, but I'm afraid the solution is not to create a benefit for every metered price product. Some of our users have pure metered pricing that don't grant any credits.

Issue #6018, opened 2025-06-26
Don't document OAuth2 Clients CRUD endpoints


frankie567:
The endpoints to CRUD OAuth2 clients show up in the API reference docs.

Since they're only designed to be used within the dashboard with a web session, we shouldn't expose them. In any case, I don't think I've already seen in the industry a programmatic API to create OAuth2 clients.

amankitsingh:
@frankie567 This has been fixed! 🎉

Removed APITag.documented from all 5 OAuth2 client CRUD endpoints:

These endpoints remain fully functional for dashboard use but are now hidden from the public API documentation.

PR #6104, opened 2025-07-08
OAuth2 client CRUD endpoints have been hidden

amankitsingh:
Removed APITag.documented from the 5 client management endpoints (/oauth2/, /oauth2/register/*) as requested in issue #6018. 

These endpoints remain fully functional for dashboard use but won't clutter the public API documentation.
Standard OAuth2 endpoints (/token, /revoke, /introspect, /userinfo) stay documented as they're part of the public API.
Clean solution - no breaking changes, just cleaner docs that follow industry standards.

### DIFF
--- a/server/polar/oauth2/endpoints/oauth2.py
+++ b/server/polar/oauth2/endpoints/oauth2.py
@@ -45,7 +45,7 @@ router = APIRouter(prefix="/oauth2", tags=["oauth2"])
 @router.get(
     "/",
     summary="List Clients",
-    tags=["clients", APITag.documented],
+    tags=["clients"],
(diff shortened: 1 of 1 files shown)


frankie567:
Thank you @amankitsingh for having a look, that's indeed the idea. Can you actually add `APITag.private` instead on those endpoints? It'll also mark them as private in our schema, which means they won't be generated in our SDK.

amankitsingh:
@frankie567  Done! ✅
Updated all 5 OAuth2 client CRUD endpoints to use APITag.private

frankie567:
Perfect, thanks 🙏

PR #6114, opened 2025-07-09
server: Copy improvement to account review approval

birkjernstrom:


server/polar/notifications/notification.py
```diff
@@ -80,7 +80,7 @@ class MaintainerAccountReviewedNotificationPayload(NotificationPayloadBase):
 
     @classmethod
     def subject(cls) -> str:
-        return "Your Polar account is now completed"
+        return "Your Polar account review is now completed"
```
raunofreiberg: Super nit and this is not wrong grammatically but I think even more correct is

> Your Polar account review is now complete~~d~~

birkjernstrom: Love it. Should have spotted it too. Copy is ♥️ so appreciate you sweating the details 🙏🏼

PR #6116, opened 2025-07-09
server/web_backoffice: able to change name/slug and status of an organization

iranzithierry:
Added new form `Update Organization` to update organization name and slug
Added a way to "Deny" it and change it to "Under Review" too.

### DIFF
--- a/server/polar/account/service.py
+++ b/server/polar/account/service.py
@@ -254,5 +254,19 @@ class AccountService:
         if account.account_type == AccountType.stripe and account.stripe_id:
             await stripe.update_account(account.stripe_id, name)
 
+    async def deny_account(self, session: AsyncSession, account: Account) -> Account:
+        """Sets an account's status to 'denied'."""
+        account.status = Account.Status.DENIED
+        session.add(account)
+        await session.commit()
+        return account
+
+    async def set_account_under_review(self, session: AsyncSession, account: Account) -> Account:
+        """Sets an account's status to 'under_review'."""
+        account.status = Account.Status.UNDER_REVIEW
+        session.add(account)
+        await session.commit()
+        return account
+
 
 account = AccountService()
--- a/server/polar/account/service.py
+++ b/server/polar/account/service.py
@@ -254,15 +254,16 @@ class AccountService:
         if account.account_type == AccountType.stripe and account.stripe_id:
             await stripe.update_account(account.stripe_id, name)
 
-    async def deny_account(self, account: Account) -> Account:
-        """Sets an account's status to 'denied'."""
+    async def deny_account(self, session: AsyncSession, account: Account) -> Account:
         account.status = Account.Status.DENIED
-        enqueue_job("account.denied", account_id=account.id)
+        session.add(account)
         return account
 
-    async def set_account_under_review(self, account: Account) -> Account:
-        """Sets an account's status to 'under_review'."""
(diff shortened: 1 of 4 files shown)


server/polar/account/service.py
```diff
@@ -254,5 +254,19 @@ async def sync_to_upstream(self, session: AsyncSession, account: Account) -> Non
         if account.account_type == AccountType.stripe and account.stripe_id:
             await stripe.update_account(account.stripe_id, name)
 
+    async def deny_account(self, session: AsyncSession, account: Account) -> Account:
+        """Sets an account's status to 'denied'."""
+        account.status = Account.Status.DENIED
+        session.add(account)
+        await session.commit()
```
frankie567: We should never call `session.commit` inside the business logic: this is handled by the application lifecycle. The good thing about this approach is that if an error happens during the process, then the DB is left untouched, without inconsistent data recorded. Long story: https://www.francoisvoron.com/blog/sqlalchemy-stop-calling-session-commit

So you can just remove it, it'll work :)

server/polar/account/service.py
```diff
@@ -254,5 +254,19 @@ async def sync_to_upstream(self, session: AsyncSession, account: Account) -> Non
...
+        return account
+
+    async def set_account_under_review(self, session: AsyncSession, account: Account) -> Account:
+        """Sets an account's status to 'under_review'."""
+        account.status = Account.Status.UNDER_REVIEW
+        session.add(account)
+        await session.commit()
```
frankie567: Same comment as above re: `session.commit()`

server/polar/account/service.py
```diff
@@ -254,5 +254,19 @@ async def sync_to_upstream(self, session: AsyncSession, account: Account) -> Non
...
+        return account
```
frankie567, line 268: We should enqueue the following job:

```py
enqueue_job("account.under_review", account_id=account.id)
```

This will trigger a background task that'll take care of sending an email to the customer and notify us about the review.

server/polar/web_backoffice/organizations/endpoints.py
```diff
@@ -175,6 +175,82 @@ async def list(
...
+    if request.method == "POST":
+        data = await request.form()
+        try:
+            form = form_class.model_validate(data)
+            if form.name:
+                organization.name = form.name
+            if form.slug:
+                organization.slug = form.slug
+            session.add(organization)
+            await session.commit()
```
frankie567: Same comment as above re: `session.commit()`

server/polar/web_backoffice/organizations/endpoints.py
```diff
@@ -175,6 +175,82 @@ async def list(
...
+
+    if request.method == "POST":
+        data = await request.form()
+        try:
+            form = form_class.model_validate(data)
+            if form.name:
+                organization.name = form.name
+            if form.slug:
+                organization.slug = form.slug
+            session.add(organization)
```
frankie567: You could also do something like this:

```py
organization = await org_repo.update(organization, update_dict=form.model_dump(exclude_none=True))
```

iranzithierry:
I've removed the session stuff from the business logic and set up a queue trigger to background task for account status changes.

frankie567:
Thank you, I did a few tweaks; good to go 👍

frankie567 (approved):

Issue #6111, opened 2025-07-09
Customer Portal: Invoice generation does not work on mobile


birkjernstrom:
A customer tried to generate and download an invoice on the go (iOS), but nothing happens when they click on the "Download"-button.

Assuming it's because we're not displaying the modal and it becomes a noop.

pieterbeulque:
I just did this as well and it worked, but it showed this dialog.

![image](https://github.com/user-attachments/assets/3074a438-6a14-40d9-a284-3ea37df318bb)

But after tapping "Allow" it does redirect properly to S3. Maybe if you block this once it's persisted somewhere?

Surely there must be a way to redirect without this dialog showing.

pieterbeulque:
This turned out to be a rabbit hole to debug because there's a timing element to it. Safari will block `window.open` calls (silently or with the dialog shown above) depending on how quickly the async action resolves. You can [verify that with this Codepen](https://codepen.io/pieterbeulque/full/pvjoebr) on your iOS device or on macOS Safari.

It's easy to capture that `window.open` has been blocked is by checking the return value which will be `null`.

The most straight-forward way to ensure the customer receives their invoice is a fallback to updating the current tab: #6122

pieterbeulque:
For housekeeping: can be closed since #6122 was merged

frankie567:
Thanks @pieterbeulque 🙏

PR #6122, opened 2025-07-10
Fallback to same-tab navigation when window.open fails

pieterbeulque:
Quick fix for #6111 

This UX is suboptimal. An improvement could be falling back to an in-app dialog titled "Your invoice is ready to download" and another download button which is a simple `<a download>` to the pre-signed URL.


emilwidlund:
Yeah! Good one!

PR #6154, opened 2025-07-13
devcontainer: rely on devcontainers/features/node:1 to install pnpm

clemg:
The original devcontainers-contrib repo has disappeared, which blocks codespace installation.
The pnpm installation is done anyway by devcontainers/features/node:1.

See: https://github.com/devcontainers/features/tree/main/src/node


frankie567:
Thank you, that's helpful 🙏

PR #6179, opened 2025-07-16
Backoffice[User]: Add delete Organization action

malthejorgensen:


server/polar/organization/service.py
```diff
@@ -162,6 +164,50 @@ async def update(
...
+        session: AsyncSession,
+        organization: Organization,
+    ) -> Organization:
+        """Anonymizes fields on the Organization that can contain PII and then
+        soft-deletes the Organization.
+
+        DOES NOT:
+        - Delete or anonymize Users related Organization
+        - Delete or anonymize Account of the Organization
+        - Delete or anonymize Customers, Products, Discounts, Benefits, Checkouts of the Organization
```
psincraian, line 180: This is more a general question about the deletion of organizations, but what happens if the org has customers and purchases? Can these customers still access their benefits? Should we allow it?

malthejorgensen: It's a good question. As it's implemented here, the delete functionality is intended for empty "leftover" organizations.

I think for an organization that has purchases we should likely "never" delete (at least not for 5 years or so), because we need the information if we get audited or just in general for accounting purposes.

Potentially we could block deletes when the org has purchases.

PR #6182, opened 2025-07-16
feat(Subscriptions): enhance CustomerChangePlanModal with payment method redirection

adwaitgawade:
When the Free plan user wants to upgrade to a higher tier they are shown modal with plan selection but they need to have payment setup already or they couldn't upgrade, the prompt/tip shown is easy to overlook so I added color to the text and also an underline text that takes the user to setting page where they can add payment details

- Added a visual reminder for users to add a payment method before changing their subscription plan.
- The message now includes a link to the Customer Portal Settings for easier navigation.
- Updated imports to include useSearchParams for dynamic settings path generation.

# Before:
<div class="flex gap-4">
  image
  image
</div>

# After
<div class="flex gap-4">
image
image
</div>


adwaitgawade:
The "Before" screenshots and "After" screenshots are taken from different accounts, I apologize if they're causing confusion.

PR #6184, opened 2025-07-16
Issue-6183: check validity of tax address

### LINKED ISSUE #6183: Check validity of tax address even if first month is free
When a customer checkouts, we validate the billing address by trying to compute tax with Stripe. If the address is invalid, we propagate the Stripe error and block the payment.

However, in the case of a setup intent (first month free), it seems we discard the tax calculation payment, which may allow customer to input invalid address. Then, Stripe will complain if we try to create a subscription with automatic tax with that invalid address.

We need to check the validity of the address even in the case of a setup intent.

amankitsingh:
Fixes #6183 
Regular payments (is_payment_required is True): Only require full addresses if require_billing_address is True or is_business_customer is True
Setup intents (is_payment_required is False but is_payment_setup_required is True): Require full addresses for tax calculation
Free products (is_payment_required is False and is_payment_setup_required is False): Don't require full addresses

### DIFF
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -1541,7 +1541,9 @@ class CheckoutService:
     async def _update_checkout_tax(
         self, session: AsyncSession, checkout: Checkout
     ) -> Checkout:
-        if not (checkout.is_payment_form_required and checkout.product.is_tax_applicable):
+        if not (
+            checkout.is_payment_form_required and checkout.product.is_tax_applicable
+        ):
             checkout.tax_amount = 0
             checkout.tax_processor_id = None
             return checkout
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -797,7 +797,11 @@ class CheckoutService:
                     }
                 )
 
-        if checkout.require_billing_address or checkout.is_business_customer:
+        if (
+            checkout.require_billing_address
+            or checkout.is_business_customer
+            or checkout.is_payment_form_required
+        ):
             if (
                 checkout.customer_billing_address is None
                 or not checkout.customer_billing_address.has_address()
@@ -1541,7 +1545,16 @@ class CheckoutService:
     async def _update_checkout_tax(
         self, session: AsyncSession, checkout: Checkout
     ) -> Checkout:
-        if not (checkout.is_payment_required and checkout.product.is_tax_applicable):
+        # For setup intents (first month free), we still need to validate the address
+        # even though no payment is currently required, because the subscription will
+        # need automatic tax enabled for future billing
+        if not (
+            (checkout.is_payment_required and checkout.product.is_tax_applicable)
+            or (
+                checkout.is_payment_setup_required
+                and checkout.product.is_tax_applicable
+            )
+        ):
             checkout.tax_amount = 0
             checkout.tax_processor_id = None
             return checkout
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -800,7 +800,7 @@ class CheckoutService:
         if (
             checkout.require_billing_address
             or checkout.is_business_customer
-            or checkout.is_payment_form_required
+            or (checkout.is_payment_setup_required and not checkout.is_payment_required)
         ):
             if (
                 checkout.customer_billing_address is None
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -797,11 +797,7 @@ class CheckoutService:
                     }
                 )
 
-        if (
-            checkout.require_billing_address
-            or checkout.is_business_customer
-            or (checkout.is_payment_setup_required and not checkout.is_payment_required)
-        ):
+        if checkout.require_billing_address or checkout.is_business_customer:
             if (
                 checkout.customer_billing_address is None
                 or not checkout.customer_billing_address.has_address()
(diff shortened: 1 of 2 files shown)


server/polar/checkout/service.py
```diff
@@ -797,7 +797,11 @@ async def _confirm_inner(
                     }
                 )
 
-        if checkout.require_billing_address or checkout.is_business_customer:
+        if (
+            checkout.require_billing_address
+            or checkout.is_business_customer
+            or (checkout.is_payment_setup_required and not checkout.is_payment_required)
+        ):
```
frankie567: We can revert this change, it's not related to the problem we have at hand here. This logic is here to check if we need a full address (with line, postal code and so forth), but we don't want to always enforce it for payment setup.

server/polar/checkout/service.py
```diff
@@ -1541,7 +1545,16 @@ async def _update_checkout(
     async def _update_checkout_tax(
         self, session: AsyncSession, checkout: Checkout
     ) -> Checkout:
-        if not (checkout.is_payment_required and checkout.product.is_tax_applicable):
+        # For setup intents (first month free), we still need to validate the address
+        # even though no payment is currently required, because the subscription will
+        # need automatic tax enabled for future billing
+        if not (
```
frankie567, line 1544: I think we can simplify the boolean test like this:

```py
if not (checkout.is_payment_form_required and checkout.product.is_tax_applicable):
```

`is_payment_form_required` is a property checking if either payment or setup is required.

---

You can also remove the comment, I think it's sufficiently clear without.

server/tests/checkout/test_service.py
```diff
@@ -2671,9 +2671,23 @@ async def test_valid_stripe(
...
+                {"tax_country": "FR"},
+            ),
+            (
+                {
+                    "line1": "456 Oak Ave",
+                    "postal_code": "H2Y 1C6",
+                    "city": "Quebec",
+                    "state": "CA-QC",
+                    "country": "CA",
+                },
```
frankie567: We can revert this change, it doesn't add much more value to the test.

server/tests/checkout/test_service.py
```diff
@@ -3213,3 +3295,35 @@ async def test_valid_with_redemption(
...
+                product_price_id=price.id,
+                customer_billing_address=Address.model_validate({"country": "FR"}),
+            ),
+            auth_subject,
+        )
+
+        assert checkout.tax_amount == 100
+        assert checkout.tax_processor_id == "TAX_PROCESSOR_ID"
+        assert checkout.customer_billing_address is not None
+        assert checkout.customer_billing_address.country == "FR"
```
frankie567: We can revert this change, it doesn't add much more value to the tests

server/tests/checkout/test_service.py
```diff
@@ -3094,6 +3108,74 @@ async def test_existing_customer_email_changed(
...
+                        "city": "New York",
+                        "state": "US-NY",
+                        "country": "US",
+                    },
+                }
+            ),
+        )
+
+        assert checkout.status == CheckoutStatus.confirmed
+        stripe_service_mock.create_setup_intent.assert_called_once()
```
frankie567: So, the idea of this test is rather to make sure calculate tax is called when we are in a setup scenario. What we would like to see is the following:

```py
async def test_setup_intent_address_validation(
    self,
    calculate_tax_mock: AsyncMock,
    session: AsyncSession,
    locker: Locker,
    auth_subject: AuthSubject[Anonymous],
    checkout_discount_percentage_100: Checkout,
) -> None:
    calculate_tax_mock.side_effect = IncompleteTaxLocation(
        stripe_lib.InvalidRequestError("ERROR", "ERROR")
    )

    # Verify this is a setup intent scenario
    assert checkout_discount_percentage_100.is_payment_required is False
    assert checkout_discount_percentage_100.is_payment_setup_required is True
    assert checkout_discount_percentage_100.is_payment_form_required is True

    with pytest.raises(PolarRequestValidationError) as e:
        await checkout_service.confirm(
            session,
            locker,
            auth_subject,
            checkout_discount_percentage_100,
            CheckoutConfirmStripe.model_validate(
                {
                    "confirmation_token_id": "CONFIRMATION_TOKEN_ID",
                    "customer_name": "Customer Name",
                    "customer_email": "customer@example.com",
                    "customer_billing_address": {
                        "line1": "123 Main St",
                        "postal_code": "12345",
                        "city": "New York",
                        "state": "US-CA",
                        "country": "US",
                }
            ),
        )
```

Basically, we make `calculate_tax` returns us an error; we want to simulate an invalid address. We want to verify we get this error even in the context of a setup intent, which was not the case before.

frankie567 (changes requested):
Thank you for having a look at this. That's the idea, but we'll need some changes to make it a bit more targeted at the problem we have.

server/polar/checkout/service.py
amankitsingh: done

server/polar/checkout/service.py
amankitsingh: done

server/tests/checkout/test_service.py
amankitsingh: done

server/tests/checkout/test_service.py
amankitsingh: done

server/tests/checkout/test_service.py
amankitsingh: done

amankitsingh:
@frankie567 I have made the suggested changes, also I am curious to know something

What specific scenarios do you have in mind where partial addresses should be allowed for setup intents?

frankie567:
Thank you, that's perfect!

> @frankie567 I have made the suggested changes, also I am curious to know something
> 
> What specific scenarios do you have in mind where partial addresses should be allowed for setup intents?

Basically, whether it's payment or setup intents, we try to ask the most *minimal* billing information required to compute the tax/VAT (unless specifically required by the `is_business_customer`/`require_billing_address` options). Typically, for European VAT, the country is enough.

The case we solve here is the following: we're in a Setup Intent (i.e. first month of subscription free, but credit card required for later). Before your changes, we were not validating the billing address against the Stripe Tax API. What we've seen is a customer entering an invalid US address, which was then preventing us to properly compute the tax for the next invoices.

PR #6187, opened 2025-07-17
Add Organization users page

malthejorgensen:
This PR adds a "Members" page under the "Settings" menu item in the left-hand navigation.

Here the members of the organization are listed and it is possible to invite further members to the organization.


server/polar/organization/endpoints.py
```diff
@@ -214,3 +220,64 @@ async def members(
...
+        if hasattr(auth_subject.subject, "email")
+        else "A team member"
+    )
+
+    # Send invitation email
+    body = render_email_template(
+        "organization_invite",
+        {
+            "organization_name": organization.name,
+            "inviter_email": inviter_email,
```
frankie567: `mypy` is failing here because `EmailStr` is not considered as a strict `str`. So we should `str(inviter_email)`.

PR #6193, opened 2025-07-17
feat: add retry subscription functionality

psincraian:
# Context
This PR is preparing for the subscription retry functionality. Mainly:

1. We are adding a new field in the order to mark when the payment retry should happen.
2. We are handling the first failed payment, setting the subscription to past due, removing benefits, and setting the next retry time for the order.

In the next PR we are going to implement the missing functionalities: handle the other dunning retries, the manual payment by the user, etc.


.github/copilot-instructions.md
```diff
@@ -5,6 +5,9 @@ This document provides essential guidance for AI agents contributing to the Pola
 ## General Guidelines
 - Do not add comments to the code unless necessary. The code should be self-explanatory.
 - Use meaningful variable and function names.
+- Follow good practices and code conventions.
+- Make sure that all the new code is maintanable and follows the SOLID principles.
+- Do not modify unrelated code to the task or issue you are working on.
```
frankie567, line 10: Ha, ha, that last one will be so helpful 😁 LLM are too zealous

psincraian: I hate when they start modifying random things 🤣

server/polar/order/service.py
```diff
@@ -1242,5 +1244,46 @@ async def _on_order_paid(self, session: AsyncSession, order: Order) -> None:
...
             )
 
+    async def handle_payment_failure(
+        self, session: AsyncSession, order: Order
+    ) -> Order:
+        """Handle payment failure for an order, initiating dunning if necessary."""
+        if order.subscription is None:
+            return order
+
+        if order.subscription.stripe_subscription_id is not None:
```
frankie567, line 1254: Yes, that's the solution for that. Quite simple, but effective IMO. I applied the same approach when looking for subscriptions to cycle:

https://github.com/polarsource/polar/blob/3bd15bced0a977829c68a9840493c48e132766cd/server/polar/subscription/scheduler.py#L40-L53

server/polar/order/service.py
```diff
@@ -1242,5 +1244,46 @@ async def _on_order_paid(self, session: AsyncSession, order: Order) -> None:
...
+        return await self._handle_consecutive_dunning_attempts(session, order)
+
+    async def _handle_first_dunning_attempt(
+        self, session: AsyncSession, order: Order
+    ) -> Order:
+        """Handle the first dunning attempt for an order, setting the next payment
+        attempt date and marking the subscription as past due.
+        """
+
+        first_retry_date = utc_now() + settings.DUNNING_FIRST_RETRY_DELAY
```
frankie567, line 1270: Code nitpick, but that may be nice to handle all the delays in a single list rather than having the first as a separate one.

---

I'm wondering then if we could factorize `_handle_first_dunning_attempt` and `_handle_consecutive_dunning_attempts` into a single method, where the first attempt would be a general case. Pseudo-code:

```
attempts = get_attempts(order)
if attempts >= len(settings.DUNNING_RETRY_INTERVALS):
    # Definitely failed
else:
    delay = settings.DUNNING_RETRY_INTERVALS[attempts - 1]
    next_attempt_at = (order.next_payment_attempt_at or utc_now()) + delay
```

But we might see that when we have the implementation of `_handle_consecutive_dunning_attempts` :)

server/tests/integrations/stripe/test_payment_dunning.py
```diff
@@ -0,0 +1,106 @@
+import uuid
```
frankie567: For tests, what I usually do is to mirror the name of the module. For example:

* `polar/order/service.py` -> `tests/order/test_service.py`
* `polar/integrations/stripe/payment.py` -> `tests/integrations/stripe/test_payment.py`

server/tests/integrations/stripe/test_payment_dunning.py
```diff
@@ -0,0 +1,106 @@
...
+        object="payment_intent",
+        amount=2000,
+        currency="usd",
+        status="requires_payment_method",
+        metadata={"checkout_id": str(uuid.uuid4())},
+    )
+
+
+@pytest.mark.asyncio
+class TestDunningIntegration:
```
frankie567: To be honest, I'm not sure this test brings much value. This is mainly why I didn't write tests for that `payment` module before: it's mostly glue logic. 

If we test the service methods we use under the hood, I would say it's enough. We see here that in the end, you mock lot of functions just to call `handle_payment_failure`, which is tested separately anyway.

server/tests/order/test_service.py
```diff
@@ -1257,3 +1259,144 @@ async def test_stripe_order_not_pending(
         enqueue_job_mock.assert_called_once_with(
             "order.balance", order_id=order.id, charge_id="stripe_payment_123"
         )
+
+
+@pytest.mark.asyncio
+class TestOrderServiceDunning:
```
frankie567: I usually name test classes after the method I'm testing: `TestHandlePaymentFailure`. The idea is to group all the tests of a method inside a class, just to keep things (a bit) organised.

frankie567 (changes requested):
Great work, Petru!

Just a few nitpick to mention. My main concern is regarding the test file you added for the `stripe/payment.py` module, I don't think it brings much value to the tests.

server/tests/integrations/stripe/test_payment_dunning.py
psincraian: I see tasks are at the same level as endpoints; they are exposed to the public, and they are a proxy to the service functionalities.

For me these tests do provide value. In this case, I mock Stripe responses; however, the entire service logic is activated, including payments, orders, and subscriptions. `test_service` files usually mock dependencies, so we are not testing any end-to-end flow. However, the service classes are not mocked in the `test_endpoints` files, and all of the business logic is called. One example of such test can be the `refunds/test_endpoint.py`.

server/tests/order/test_service.py
psincraian: Upss, right! I will modify that

server/tests/integrations/stripe/test_payment_dunning.py
psincraian: Right, I will change that!

server/polar/order/service.py
psincraian: I was thinking about that, at first I wanted to keep it separate because it made my life easier to read but I may merge them in the next PR

server/tests/integrations/stripe/test_payment_dunning.py
frankie567: Right, I understand. What's a bit confusing is that the function is inside that `payment.py` module and not immediately identified as the logic ran by the task — but that's entirely my doing 😬  So, I agree.

But one thing then: could we avoid then to mock `resolve_checkout`, `resolve_order` and `upsert_from_stripe_charge`? It looks to me we could make it work without mocking. I'll add more comments in this file, bear with me :)

psincraian: Understood! I will see if I can do it without mocking

server/tests/integrations/stripe/test_payment_dunning.py
```diff
@@ -0,0 +1,106 @@
...
+
+@pytest.fixture
+def stripe_charge() -> stripe_lib.Charge:
+    return stripe_lib.Charge(
+        id="ch_test_123",
+        object="charge",
+        amount=2000,
+        currency="usd",
+        status="failed",
+        metadata={"checkout_id": str(uuid.uuid4())},
```
frankie567: To resolve the order from the charge, you should set metadata like this:

```py
metadata={"order_id": str(order.id)},
```

To have the order at hand, you have two possibilities:

1. Don't create a separate fixture for `stripe_charge`, simply declare it in the test. Probably the easiest since we only have one test right now
2. Declare an order fixture first and inject it in this fixture.

---

We can discard the `checkout_id` metadata, since we're not testing the checkout resolution here (i.e. `resolve_checkout` will return `None`).

server/tests/integrations/stripe/test_payment_dunning.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+
+
+@pytest.fixture
+def stripe_payment_intent() -> stripe_lib.PaymentIntent:
```
frankie567: Looks unused ? :)

server/tests/integrations/stripe/test_payment_dunning.py
```diff
@@ -0,0 +1,106 @@
...
+
+        mocker.patch(
+            "polar.integrations.stripe.payment.resolve_checkout", return_value=None
+        )
+        mocker.patch(
+            "polar.integrations.stripe.payment.resolve_order", return_value=order
+        )
+        mocker.patch(
+            "polar.integrations.stripe.payment.payment_service.upsert_from_stripe_charge"
+        )
```
frankie567: I think we can safely remove all the three mocks here and everything should work.

server/tests/integrations/stripe/test_payment_dunning.py
psincraian: Done

server/tests/integrations/stripe/test_payment_dunning.py
psincraian: Done

server/tests/order/test_service.py
psincraian: DOne

server/tests/integrations/stripe/test_payment_dunning.py
psincraian: Done

server/tests/integrations/stripe/test_payment_dunning.py
psincraian: Done

server/tests/subscription/test_service.py
```diff
@@ -1475,3 +1476,35 @@ async def _send_confirmation_email() -> None:
             await subscription_service.send_confirmation_email(session, subscription)
 
         await watch_email(_send_confirmation_email, email_sender.path)
+
+
+@pytest.mark.asyncio
+class TestSubscriptionServiceDunning:
```
frankie567: Can we call it `TestMarkPastDue`? 😊

psincraian: Done

Issue #4785, opened 2025-01-03
Unsubscribe modal not scrolling on mobile


frankie567:
On mobile, in the customer portal, the unsubscribe modal isn't scrollable, so we can't see the submit button.

![image](https://github.com/user-attachments/assets/ced5e5e9-61d9-45ed-bb5d-817651685d16)

adiityack:
@frankie567 can I work on this bug? I would love to contribute

PR #6195, opened 2025-07-17
Make modals scrollable

pieterbeulque:
Mostly to tackle #4785 — which is why there's another commit here with some more responsive styling specifically for that page.

Only drawback with this updated styling is that the _Cancel subscription_ button coincidentally aligns nicely with the bottom of the viewport, so nothing visually hints at the thing being scrollable. However that's very viewport specific, and the current version blocks people from canceling, so it's still an improvement.

| Before | After |
|--------|------|
| image (not scrollable) | image (and scrollable!) |

Design decisions (might introduce some inconsistency that we can re-align separately in a follow-up PR):

 - Went with `p-6` consistent with components like `<Well />`
 - Render buttons full width on small viewports (happy thumbs)
 - Stack buttons vertically on small viewports

### DIFF
--- a/clients/apps/web/src/components/Subscriptions/CustomerCancellationModal.tsx
+++ b/clients/apps/web/src/components/Subscriptions/CustomerCancellationModal.tsx
@@ -105,7 +105,7 @@ const CustomerCancellationModal = ({
       {...props}
       className="md:min-w-[600px]"
       modalContent={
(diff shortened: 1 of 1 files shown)


frankie567:
Thank you @pieterbeulque, really helpful 🙏 Good to go!

PR #6196, opened 2025-07-17
(experimental) Fix pagination on mobile

pieterbeulque:
The pagination pushes past the viewport width on mobile, leading to that annoying zoom out / horizontal scroll mixture.

I did some flexbox kungfu to wrap the pagination on two lines in a way that makes sense to me.

Curious what you think, @emilwidlund, would understand if you hate it too, haha 😁 

## Desktop

image

## Mobile

image

<details>
<summary>Current version for reference</summary>
image
</details>


emilwidlund:
Cool! I don't mind that at all! Thanks!

PR #6197, opened 2025-07-17
Use range pagination

### LINKED ISSUE #4611: UX: Show range of records shown vs. pages in paginated tables
### Description

Offer more clarity on the range of records the seller is looking at in their dashboard vs. pages, i.e requiring math operations to map to the range of orders, transactions etc.

### Current Behavior

image

### Expected Behavior

Change `Page 1 of 12` to `Viewing 1-100 of 1,200`

**Pages**
- [ ] `Finance` > `Incoming`
- [ ] `Finance` > `Outgoing`
- [ ] `Sales` > `Orders`
- [ ] `Sales` > `Subscriptions`
- [ ] `Products` > `Discounts`

clemg:
* Closes #4611 

Use range pagination when useful.

We could have used the number of page to determine the total number of items, but I find it misleading.
For example:  21 items and default pagination of 20 would give "Viewing 0-20 of 40" (because ⌈21/20⌉=40), whereas we have 21 items in reality. It's would be about a 100% off.

This is optional and will fallback to page count if we have no data


frankie567:
Looks good, Clément, thanks 🙏 Just a merge conflict to solve and we'll be good to go!

pieterbeulque:
Sorry for causing the merge conflict!

I think this phrasing introduces some funny edge cases:

 - When there are no records rendering "Viewing 0–0 of 0" is a bit repetitive, maybe for empty tables we can hide this altogether?
 - "Viewing 1-12 of 12" feels verbose, maybe if `startItem === 1 && endItem === rowCount` we can show "Viewing all 12 records" instead?

frankie567:
> Sorry for causing the merge conflict!
> 
> I think this phrasing introduces some funny edge cases:
> 
> * When there are no records rendering "Viewing 0–0 of 0" is a bit repetitive, maybe for empty tables we can hide this altogether?
> * "Viewing 1-12 of 12" feels verbose, maybe if `startItem === 1 && endItem === rowCount` we can show "Viewing all 12 records" instead?

Good call! If you're willing to do that Clément, would be great :)

clemg:
Good catch @pieterbeulque! Thanks for noticing :D

frankie567:
That's great, Clément, thanks 👍 Special mention for the clean Git history, love that!

PR #6207, opened 2025-07-18
Handle discounts in new billing engine

### LINKED ISSUE #6121: Handle discounts
We should handle discounts when:

* Creating the subscription
* Cycling the subscription

The key part is to handle the discounts that have a limited validity (e.g. first three months). So we might need a specific data model to keep track of that, or maybe we could rely on the Order model.

frankie567:
Fix #6121

### DIFF
--- a/server/polar/models/discount.py
+++ b/server/polar/models/discount.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING, Literal, cast
 from uuid import UUID
 
 import stripe as stripe_lib
+from dateutil.relativedelta import relativedelta
 from sqlalchemy import (
     TIMESTAMP,
     Column,
@@ -157,6 +158,20 @@ class Discount(MetadataMixin, RecordModel):
         if len(self.products) == 0:
             return True
         return product in self.products
+
+    def is_repetition_expired(
+        self, started_at: datetime, current_period_start: datetime
+    ) -> bool:
+        if self.duration == DiscountDuration.once:
+            return True
+        if self.duration == DiscountDuration.forever:
+            return False
+        if self.duration_in_months is None:
+            return False
+
+        # -1 because the first month counts as a first repetition
+        end_at = started_at + relativedelta(months=self.duration_in_months - 1)
+        return current_period_start > end_at
 
     __mapper_args__ = {
         "polymorphic_on": "type",
--- a/server/polar/subscription/service.py
+++ b/server/polar/subscription/service.py
@@ -550,6 +550,14 @@ class SubscriptionService:
                 subscription.recurring_interval.get_next_period(current_period_end)
             )
 
+        # Check if discount is still applicable
+        if subscription.discount is not None:
+            assert subscription.started_at is not None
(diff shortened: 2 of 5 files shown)


server/polar/subscription/service.py
```diff
@@ -550,6 +550,14 @@ async def cycle(
                 subscription.recurring_interval.get_next_period(current_period_end)
             )
 
+        # Check if discount is still applicable
+        if subscription.discount is not None:
+            assert subscription.started_at is not None
+            if subscription.discount.is_repetition_expired(
```
psincraian, line 556: How does the "Maximum number of redemptions" work? I mean, if I have 10 redemptions and the discount can be applied for 10 months. Does that mean that 10 customers can use it for 10 months?

server/polar/models/discount.py
```diff
@@ -158,6 +159,20 @@ def is_applicable(self, product: "Product") -> bool:
...
+        if self.duration == DiscountDuration.once:
+            return True
+        if self.duration == DiscountDuration.forever:
+            return False
+        if self.duration_in_months is None:
+            return False
+
+        # -1 because the first month counts as a first repetition
+        end_at = started_at + relativedelta(months=self.duration_in_months - 1)
+        return current_period_start > end_at
```
psincraian, line 174: So, discounts are not valid on the end date?

server/polar/subscription/service.py
frankie567: Discount redemptions is enforced at checkout: 

https://github.com/polarsource/polar/blob/936e325328874160ac13bb235c9dd4dc5f8580d4/server/polar/checkout/service.py#L728-L737

A "redemption" represent one customer applying the discount on a purchase. How the discount is applied then (once, forever or on X months) doesn't matter. So yes, in your example, 10 customers can use it for 10 months.

server/polar/models/discount.py
frankie567: They are. But the test is inverted here: we want to know if the discount is *expired* not still *valid*. Do you think it would be clearer if we reverse the boolean logic, i.e. `is_repetition_applied`?

<details>

* `duration_in_months`: 3
* `started_at`: `2025-01-01T12:00:00`
* `end_at` = `2025-03-01T12:00:00`

**Month 2**

* `current_period_start`:  `2025-02-01T12:00:00`
* `current_period_end`: `2025-03-01T12:00:00`
* `current_period_start` > `2025-03-01T12:00:00` => `False`

**Month 3**

* `current_period_start`:  `2025-03-01T12:00:00`
* `current_period_end`: `2025-04-01T12:00:00`
* `current_period_start` > `2025-03-01T12:00:00` => `False`

**Month 4**

* `current_period_start`:  `2025-04-01T12:00:00`
* `current_period_end`: `2025-05-01T12:00:00`
* `current_period_start` > `2025-03-01T12:00:00` => `True`

</details>

psincraian: upss, my bad. You are right. I think it's clear now that I read it the second time

PR #6246, opened 2025-07-21
feat: implement retry follow ups

psincraian:
# Context
- Add a new worker to process dunning orders
- Handle the dunning process:
  - Retry the payment up to 4 times. If it fails, the subscription is marked as cancelled.
  - If any of those fail, the subscription is marked as active and the benefits are granted again.

### DIFF
--- a/server/polar/config.py
+++ b/server/polar/config.py
@@ -272,9 +272,9 @@ class Settings(BaseSettings):
     # Dunning Configuration
     DUNNING_RETRY_INTERVALS: list[timedelta] = [
         timedelta(days=2),  # First retry after 2 days
-        timedelta(days=8),  # Second retry after 10 days (2 + 8)
-        timedelta(days=6),  # Third retry after 14 days (2 + 8 + 6)
-        timedelta(days=5),  # Fourth retry after 21 days (2 + 8 + 6 + 5)
+        timedelta(days=5),  # Second retry after 7 days (2 + 5)
+        timedelta(days=7),  # Third retry after 14 days (2 + 5 + 7)
+        timedelta(days=7),  # Fourth retry after 21 days (2 + 5 + 7 + 7)
     ]
 
     model_config = SettingsConfigDict(
--- a/server/polar/config.py
+++ b/server/polar/config.py
@@ -270,7 +270,6 @@ class Settings(BaseSettings):
     ]
 
     # Dunning Configuration
-    DUNNING_FIRST_RETRY_DELAY: timedelta = timedelta(days=3)
     DUNNING_RETRY_INTERVALS: list[timedelta] = [
         timedelta(days=2),  # First retry after 2 days
         timedelta(days=8),  # Second retry after 10 days (2 + 8)
--- a/server/polar/order/service.py
+++ b/server/polar/order/service.py
@@ -63,6 +63,7 @@ from polar.models import (
 )
 from polar.models.order import OrderBillingReason, OrderStatus
 from polar.models.product import ProductBillingType
+from polar.models.subscription import SubscriptionStatus
 from polar.models.transaction import TransactionType
 from polar.models.webhook_endpoint import WebhookEventType
 from polar.notifications.notification import (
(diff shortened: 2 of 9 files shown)


server/polar/order/tasks.py
```diff
@@ -134,3 +135,27 @@ async def order_invoice(order_id: uuid.UUID) -> None:
...
+    priority=TaskPriority.MEDIUM,
+)
+async def process_dunning() -> None:
+    """Process all orders that are due for dunning (payment retry)."""
+    async with AsyncSessionMaker() as session:
+        order_repository = OrderRepository.from_session(session)
+        due_orders = await order_repository.get_due_dunning_orders()
+
+    for order in due_orders:
+        async with AsyncSessionMaker() as session:
```
psincraian: I saw that usually we are doing multiple transactions per order/subscription, but I think it's better to have 1 transaction per 1 order.

server/polar/subscription/service.py
```diff
@@ -1433,5 +1433,28 @@ async def mark_past_due(
...
+
+        previous_status = subscription.status
+        previous_ends_at = subscription.ends_at
+
+        repository = SubscriptionRepository.from_session(session)
+        subscription = await repository.update(
+            subscription, update_dict={"status": SubscriptionStatus.active}
+        )
+
+        await self._after_subscription_updated(
```
psincraian, line 1449: I think we should take a look at the communications (in another task). Once a customer payment fails:

- Stripe sends the communication that the payment failed
- We send a communication that the subscription is cancelled


When a payment is recovered:
- We send a communication that the subscription was created

server/polar/order/service.py
```diff
@@ -1282,7 +1296,84 @@ async def _handle_consecutive_dunning_attempts(
...
+        order = await repository.update(
+            order, update_dict={"next_payment_attempt_at": next_retry_date}
+        )
+
+        return order
+
+    async def _count_failed_payment_attempts(
+        self, session: AsyncSession, order: Order
+    ) -> int:
+        statement = select(func.count(Payment.id)).where(
```
psincraian: We have statements in the service, but we also have the repository. What is the logic behind it? @frankie567 🤔

server/polar/order/tasks.py
frankie567: Ah, I see 🤔 That works, but I'm more in the camp of "one task/request == one session". So if something fails during the process, the whole transaction is rollbacked and we can retry without any problem. In this context, it's not a problem since we'll retrieve a fresh list of due orders at the start, but well.

What I usually do for this kind of jobs is to have another task to process a single item. So basically this task would do:

```py
for order in due_orders:
    enqueue_job("order.process_dunning_order", order.id)
```

The nice thing about this approach is that even if one order in the middle encounters issue, it can be retried on its own; and the rest of the batch is processed properly.

server/polar/order/tasks.py
```diff
@@ -134,3 +135,27 @@ async def order_invoice(order_id: uuid.UUID) -> None:
...
+            try:
+                await order_service.process_dunning_order(session, order)
+            except Exception as e:
+                sentry_sdk.capture_exception(e, extra={"order_id": str(order.id)})
```
frankie567: This is not necessary: Sentry automatically reports error logs.

We could even drop the `try..except` block to let it raise naturally.

server/polar/subscription/service.py
frankie567: Yes, lot of emails to implement 😁

server/polar/config.py
```diff
@@ -270,12 +270,11 @@ class Settings(BaseSettings):
...
     # Dunning Configuration
-    DUNNING_FIRST_RETRY_DELAY: timedelta = timedelta(days=3)
     DUNNING_RETRY_INTERVALS: list[timedelta] = [
         timedelta(days=2),  # First retry after 2 days
-        timedelta(days=8),  # Second retry after 10 days (2 + 8)
-        timedelta(days=6),  # Third retry after 14 days (2 + 8 + 6)
-        timedelta(days=5),  # Fourth retry after 21 days (2 + 8 + 6 + 5)
+        timedelta(days=5),  # Second retry after 7 days (2 + 5)
+        timedelta(days=7),  # Third retry after 14 days (2 + 5 + 7)
+        timedelta(days=7),  # Fourth retry after 21 days (2 + 5 + 7 + 7)
```
frankie567, line 277: Noice 👏

server/polar/order/service.py
```diff
@@ -1282,7 +1296,84 @@ async def _handle_consecutive_dunning_attempts(
         self, session: AsyncSession, order: Order
     ) -> Order:
         """Handle consecutive dunning attempts for an order."""
-        # TODO : Implement logic for handling consecutive dunning attempts
+        failed_attempts = await self._count_failed_payment_attempts(session, order)
+
+        if failed_attempts >= len(settings.DUNNING_RETRY_INTERVALS):
+            # No more retries, mark subscription as unpaid and clear retry date
+            repository = OrderRepository.from_session(session)
```
frankie567: We could instantiate the repository once at the top

server/polar/order/service.py
frankie567: Mainly legacy reasons 🙈 We didn't always have the concept of repositories, so everything was inside the service, which then was a messy mix of query and business logic.

I would say this method could be a proper method in `PaymentRepository`.

frankie567 (commented):
Looks really great, awesome job!

I would suggest you consider my suggestion regarding the task and "sub-task" to process past due orders in their individual task.

server/polar/order/service.py
psincraian: Fixed

server/polar/order/tasks.py
psincraian: Good approach!

PR #6248, opened 2025-07-21
Configure meters straight from onboarding

### LINKED ISSUE #6092: Configuring meter credits during onboarding is confusing
This is probably because a new organization doesn't have meters set up yet?

So either Polar should prefill some common meters (LLM credits could be a good default) or the option should be disabled with a _Setup required_ label. Hiding the option completely could also work but it's fun to showcase everything you can do during onboarding 😁 

Image

pieterbeulque:
Fixes #6092

### Before

image

_Select a meter_ opens an empty select dropdown makes this basically a dead end in the onboarding flow.

### After

Added a brief alert that lets you create a meter straight from the metered pricing & meter credits sections.

**Metered pricing**

image

**Meter credits**

image

From there a new modal opens to create a new meter directly:

image

Upon completion, the new meter is selected and you can continue.

image

### Known issues / questions

**Usefulness in terms of UX**: This improves on the current version, which is effectively a dead end. However, I wonder if the whole setup is a bit too technical. You're basically required to set up event ingestion too for this to make sense. This can't/won't be the case in the context of onboarding, which is also I didn't include e.g. the events preview from the in-app meter configuration.  
This makes me wonder whether onboarding / initial product setup could be done differently (more of a wizard-style), I'm thinking about onboarding flows that have a similar SDK setup requirement (like Sentry or Posthog).

**Pagination**: the `GET /meters` endpoint is paginated by default, returning 10 items per page. This leads to unexpected behavior when a user has more than 10 meters. I don't know how common this is, but I've decided to not to dive into it since it's unlikely a user will have more than 10 during onboarding 😁 The laziest fix would be to return more than 10 items from the endpoint by default, which I think would be fine, but this could use a proper fix outside the scope of this issue. 

**A `requestAnimationFrame` of shame**: I'm having trouble automatically selecting the freshly created meter in the form. [It seems to be that Radix UI's select component rejects any value updates that don't exist as child items](https://github.com/radix-ui/primitives/issues/2817) of `<SelectContent>`. Combined with React Query's asynchronous cache invalidation, this is pretty difficult to fix.

Basically, in `onSelectMeter` we have to:
1. Wait until `meters` is updated from the query invalidation from the mutation callback
2. Wait for a render cycle to update the children in `<SelectContent>`
3. Only then can we call `setValue` with the new `meter.id`.
 
A `setTimeout` would work around this, but might lead to inconsistent behavior on slower networks. Explicitly awaiting a refresh and then making sure the state was reflected in the UI through a double `requestAnimationFrame` should work consistently, but it's a noticeable flash in the UX. The less hacky workaround is to not auto-select the freshly created meter, but the trade-off then is arguably worse UX.

### DIFF
--- a/clients/apps/web/src/components/Benefit/MeterCredit/BenefitForm.tsx
+++ b/clients/apps/web/src/components/Benefit/MeterCredit/BenefitForm.tsx
@@ -34,7 +34,7 @@ export const MeterCreditBenefitForm = ({
 }: {
   organization: schemas['Organization']
 }) => {
-  const { data: meters } = useMeters(organization.id, {
+  const { data: meters, refetch } = useMeters(organization.id, {
     sorting: ['name'],
   })
 
@@ -55,7 +55,18 @@ export const MeterCreditBenefitForm = ({
       //
       // This is an open issue with Radix UI since 2024
       // (https://github.com/radix-ui/primitives/issues/2817)
-      setTimeout(() => setValue('properties.meter_id', meter.id), 200)
+
+      // To work around this, we run an explicit `refetch` that we can await
+      // and then set the value in a double requestAnimationFrame callback.
+      // First rAF ensures this component is updated,
+      // second rAF ensures the <SelectContent /> was updated too.
+      await refetch()
+
+      window.requestAnimationFrame(() => {
+        window.requestAnimationFrame(() => {
+          setValue('properties.meter_id', meter.id)
+        })
+      })
     },
     [setValue],
   )
--- a/clients/apps/web/src/components/Benefit/MeterCredit/BenefitForm.tsx
+++ b/clients/apps/web/src/components/Benefit/MeterCredit/BenefitForm.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import { PlusIcon } from '@heroicons/react/24/outline'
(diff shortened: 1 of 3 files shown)


pieterbeulque:
Marking this as a draft PR until we've had a chance to discuss the open ends.

frankie567:
Thank you for tackling this @pieterbeulque 👍 

Yes, that was my concern too: configuring a meter in the onboarding context feels a bit heavy, especially since it basically requires to setup event ingestion to be functional.

I guess the ideal flow would be, as you suggest, a complete wizard which starts with a "Use-case" question (e.g. Subscription SaaS, Usage-based billed App, Digital Good, etc.) which then takes the user through configuring the related stuff in their app. But that's another beast 😅

pieterbeulque:
Looks like it [confuses some other people](https://x.com/houssam_ennoura/status/1947753620541194664) as well.

Maybe we could merge this as a quick win? My own 2 cents on the other open issues:
- Pagination is probably less of an issue (10+ meters feels like a lot)
- the `rAF` hack is awkward but fine (and self-contained/harmless + documented)

So if you don't have any other remarks, happy to mark it as ready.

frankie567:
Yes, let's go :)

PR #6259, opened 2025-07-22
hidden benefit section if benefits is not present

### LINKED ISSUE #6242: Issue with Unnecessary “Benefits” Section in Cancellation Emails
### Description
I noticed an issue with the “cancel” emails. The section “Meanwhile, you will continue to have access to the following benefits:” appears even when no benefits are listed for the product.

### Current Behavior
The “benefits” section appears in cancellation emails even if no benefits are defined for a product.

### Expected Behavior
The “benefits” section should only appear if there are actual benefits listed for the product.

### Screenshots
None

### Environment:
All

amankitsingh:
Fixes #6242

Added conditional rendering to the subscription_cancellation.tsx email template so the entire benefits section only appears when there are actual benefits to display.

### DIFF
--- a/server/emails/src/emails/subscription_cancellation.tsx
+++ b/server/emails/src/emails/subscription_cancellation.tsx
@@ -23,6 +23,23 @@ interface SubscriptionCancellationProps {
   current_year: number
 }
 
+function BenefitsSection({ benefits }: { benefits: Array<{ description: string }> }) {
+  // Only render if there are actual benefits to display
+  if (benefits.length === 0) {
+    return null
+  }
+  return (
+    <>
+      <BodyText>Meanwhile, you will continue to have access to the following benefits:</BodyText>
+      <ul className="list-disc space-y-1 pl-6">
+        {benefits.map((benefit, index) => (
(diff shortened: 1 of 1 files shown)


amankitsingh:
We can also do it in other way like below

```
// Separate component for benefits section to improve readability
function BenefitsSection({ benefits }: { benefits: Array<{ description: string }> }) {
  // Only render if there are actual benefits to display
  if (benefits.length === 0) {
    return null
  }

  return (
    <>
      <BodyText>
        Meanwhile, you will continue to have access to the following benefits:
      </BodyText>
      <ul className="list-disc space-y-1 pl-6">
        {benefits.map((benefit, index) => (
          <li key={index}>{benefit.description}</li>
        ))}
      </ul>
    </>
  )
}

```

Use it by calling

`<BenefitsSection benefits={product.benefits} />`

psincraian:
As you mentioned, I think the new component improved the redeability of the email. Feel free to make the change and I will merge the PR

amankitsingh:
@psincraian  done

PR #6260, opened 2025-07-22
fix: Hide benefits section in cancellation email if no benefits are p…

### LINKED ISSUE #6242: Issue with Unnecessary “Benefits” Section in Cancellation Emails

swarnendu19:
### Description

Fixes #6242

This PR ensures that the "Meanwhile, you will continue to have access to the following benefits:" section in cancellation emails only appears if the product has actual benefits.

### Fix

- Updated condition to check that `product.benefits` is not empty before rendering the section.

### Before

The benefits section was rendered even if no benefits existed, leading to an empty or confusing email.

### After

The section is hidden when no benefits exist. Verified with two test emails (with and without benefits).

(Here, Benefits Data are dummy) 
✅ Screenshot with benefits:
 
Screenshot 2025-07-22 200954

✅ Screenshot without benefits:

Screenshot 2025-07-22 200817
]

### DIFF
--- a/server/polar/subscription/email_templates/cancellation.html
+++ b/server/polar/subscription/email_templates/cancellation.html
@@ -5,12 +5,14 @@
 <p>We're sorry to see you go! Your subscription to <strong>{{ product.name }}</strong> will remain active until <strong>{{ subscription.ends_at.strftime("%B %d, %Y") }}</strong>, after which it will be canceled.</p>
 <p>If you change your mind, you can renew your subscription anytime before the end date.</p>
 
+{% if product.benefits %}
 <p>Meanwhile, you will continue to have access to the following benefits:</p>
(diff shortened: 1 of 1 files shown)


frankie567:
Thank you for having a stab at this @swarnendu19, but unfortunately, this version of the email has recently been removed in favor of a React Email based template.

A fix for this issue is already underway in #6259

PR #6262, opened 2025-07-22
feat: implement payment lock mechanism for order processing

psincraian:
# Context
This is to add payment locks to the order. In a next PR I will add the option for users to trigger manual payments, and this may cause to charge twice the same customer if the dunning process is running at the same time as he clicks the button.


server/polar/order/repository.py
```diff
@@ -70,6 +70,32 @@ async def get_due_dunning_orders(self, *, options: Options = ()) -> Sequence[Ord
...
         return await self.get_all(statement)
 
+    async def acquire_payment_lock(self, order_id: UUID) -> bool:
+        """
+        Attempt to acquire a payment lock for an order atomically.
+
+        Returns:
+            True if lock was acquired, False if already locked
+        """
+        from sqlalchemy import update
```
frankie567: The import should probably be at the top 😄

server/polar/order/repository.py
```diff
@@ -70,6 +70,32 @@ async def get_due_dunning_orders(self, *, options: Options = ()) -> Sequence[Ord
...
+        result = await self.session.execute(
+            update(Order)
+            .where(Order.id == order_id, Order.payment_lock_acquired_at.is_(None))
+            .values(payment_lock_acquired_at=utc_now())
+        )
+        return result.rowcount > 0
+
+    async def release_payment_lock(self, order_id: UUID) -> None:
+        """Release a payment lock for an order."""
+        from sqlalchemy import update
```
frankie567: Same comment as above :)

server/polar/order/service.py
```diff
@@ -261,6 +261,13 @@ def __init__(self, order: Order) -> None:
         super().__init__(message)
 
 
+class PaymentAlreadyInProgress(OrderError):
+    def __init__(self, order: Order) -> None:
+        self.order = order
+        message = f"Payment for order {order.id} is already in progress"
+        super().__init__(message)
```
frankie567: Tip: you can set a status code as second argument of the error class. It's handy because we have a FastAPI exception handler automatically catching our errors and build a response accordingly. By default, the status code is 500, but in this case, since it's a known error, it probably can be a 409.

server/polar/order/repository.py
```diff
@@ -70,6 +70,32 @@ async def get_due_dunning_orders(self, *, options: Options = ()) -> Sequence[Ord
...
+
+        await self.session.execute(
+            update(Order)
+            .where(Order.id == order_id)
+            .values(payment_lock_acquired_at=None)
+        )
```
frankie567, line 92: Given that we always call this method with an ORM `Order` at hand, I'm wondering if we should implement it like this:

```py
async def release_payment_lock(self, order: Order, *, flush: bool = False) -> Order:
    return await self.update(order, update_dict={"payment_lock_acquired_at": None}, flush=flush)
```

The nice thing with that:

* + We return an `Order` object with the updated `payment_lock_acquired_at` value
* + We can control the flush

server/polar/order/service.py
```diff
@@ -1276,6 +1313,15 @@ async def handle_payment_failure(
...
     ) -> Order:
         """Handle payment failure for an order, initiating dunning if necessary."""
+        # Clear payment lock on failure
+        if order.payment_lock_acquired_at is not None:
+            log.info(
+                "Clearing payment lock on order due to payment failure",
+                order_id=order.id,
+            )
+            repository = OrderRepository.from_session(session)
+            await repository.release_payment_lock(order.id)
```
frankie567: With my previous comment, we can then do:

```py
order = await repository.release_payment_lock(order)
```

server/tests/order/test_service.py
```diff
@@ -1818,3 +1819,91 @@ async def test_process_dunning_order_success(
...
+        """Test that trigger_payment acquires lock and processes payment normally."""
+        # Given
+        payment_method = await create_payment_method(save_fixture, customer=customer)
+        order = await create_order(
+            save_fixture,
+            product=product,
+            customer=customer,
+            status=OrderStatus.pending,
+        )
+        order.customer.stripe_customer_id = "cus_test123"
```
frankie567: `customer` fixture should already have a Stripe customer ID.

server/tests/order/test_service.py
```diff
@@ -1818,3 +1819,91 @@ async def test_process_dunning_order_success(
...
+        """Test that lock is released when payment processing fails."""
+        # Given
+        payment_method = await create_payment_method(save_fixture, customer=customer)
+        order = await create_order(
+            save_fixture,
+            product=product,
+            customer=customer,
+            status=OrderStatus.pending,
+        )
+        order.customer.stripe_customer_id = "cus_test123"
```
frankie567: `customer` fixture should already have a Stripe customer ID.

frankie567 (approved):
Great job 👍  A few nitpicks, but good to go otherwise!

server/polar/order/repository.py
psincraian: Fixed! I added a rule to avoid the LLM doing that

server/polar/order/repository.py
psincraian: Good catch! I added for the release_payment_lock but not the acquire lock. The support on SQLAlchemy for that one is not that good, we cannot get the number of updated rows and I think it made the code worse.

server/tests/order/test_service.py
psincraian: Foxed

server/tests/order/test_service.py
psincraian: Fixed!

server/polar/order/service.py
psincraian: Nice!

PR #6268, opened 2025-07-23
Consistent decimal usage for money input

### LINKED ISSUE #6245: Currency input validation/formatting is inconsistent
### Description

Pricing inputs somethings accept comma `,` as decimal separator, other times it expects a period `.` and rejects a comma. There's also a difference between resetting on faulty input (`NaN` error in the console) and just not accepting the keystroke.

### Current Behavior

Noticed on the onboarding flow in the pricing sections, the fixed price form accepts comma-separated decimals, while the meter form accepts period-separated decimals.

### Expected Behavior

It should be consistent, at the very least.

It'd be a good default to use US number formatting (so commas as thousands separators and periods as decimal separators). At least for inputs, I think for reporting and formatting we can use the user's locale.

### Screenshots

https://github.com/user-attachments/assets/b8112355-912d-4542-a54f-e42d66ff7faf

pieterbeulque:
Fixes #6245 

Avoid `type="number"` inputs because its behavior is pretty inconsistent across browsers/locales/…

Main changes:

 - Copied over the `keydown` handler from `UnitAmountInput` for form interactivity consistency (I like the pattern)
 - Added arrow up / down handlers to step the input per 10 cents (didn't copy to `UnitAmountInput` since that would need auto-step detection to be good UX, in my opinion)
 - Accepts both `,` and `.` as input but converts commas to periods on the fly
 - Make sure it supports pasting in differently formatted numbers (e.g. from spreadsheets)

Some of these changes are regex kungfu 101, frankly it could use some unit tests for future maintenance but that's not really set up in the `clients` workspace, can be added later. If / when multi-currency support is added this'll need another makeover (mainly because e.g. JPY doesn't have cents) and that might be a good time to solidify some of these efforts.

https://github.com/user-attachments/assets/e0e6770d-4b4e-4f3d-897a-f8c6f3d32ef8

### DIFF
--- a/clients/packages/ui/src/components/atoms/MoneyInput.tsx
+++ b/clients/packages/ui/src/components/atoms/MoneyInput.tsx
@@ -63,6 +63,19 @@ const MoneyInput = (props: Props) => {
     }
   }, [value, previousValue])
 
+  const updateValue = useCallback(
+    (newValue: string) => {
+      if (_onChange) {
+        const centsValue = getCents(newValue)
+        setPreviousValue(centsValue)
+        _onChange(centsValue)
+      }
+
+      setInternalValue(newValue)
+    },
+    [_onChange],
+  )
+
   const onChange = useCallback(
     (e: ChangeEvent<HTMLInputElement>) => {
       const input = e.target.value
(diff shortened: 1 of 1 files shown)


frankie567:
Amazing work once again, Pieter 😄 Thanks 🙏

PR #6278, opened 2025-07-24
docs: Clarify payout fees introduction

### LINKED ISSUE #6276: Clarify payout fees introduction
We start the section by:

> Polar does not add any additional fees to payouts.

Some customers misread this as "No fees for payouts", while what we want to say is "We don't make any money from it, we just apply the provider fees".

We should clarify that to avoid frustration.

rishi-raj-jain:
fix #6276


frankie567:
Thanks 👍

PR #6289, opened 2025-07-24
Avoid blocking clicks behind toast

pieterbeulque:
Noticed something annoying when toasts overlap interactive elements

### Current
https://github.com/user-attachments/assets/75fd4690-4715-457d-bfe3-f114c13d0e57

### Changes
https://github.com/user-attachments/assets/9327da55-b6dd-40f9-b547-d264173e6d1c


frankie567:
Great catch; thanks 🙏

PR #6299, opened 2025-07-27
emails: fix the link to X account

rishi-raj-jain:


psincraian:
There are some tests that need to get fixed

Issue #6302, opened 2025-07-27
License Keys: Raise descriptive error on `/activate` for a license without activations


birkjernstrom:
Better educating integrators that they should not and cannot use this endpoint for a license key that does not have the activations feature, i.e they should use `/validate` only.

amankitsingh:
@birkjernstrom , kindly review this feature

PR #6306, opened 2025-07-28
Update activate.mdx

rishi-raj-jain:
add docs for #6302

### DIFF
--- a/docs/api-reference/customer-portal/license-keys/activate.mdx
+++ b/docs/api-reference/customer-portal/license-keys/activate.mdx
@@ -1,3 +1,7 @@
 ---
 openapi: post /v1/customer-portal/license-keys/activate
 ---
+
+<Note>
+  This endpoint is only for license keys that have the **activations** on the benefit enabled. If your license key does **not** support activations, you **should** use this endpoint. Instead, you should use the [validate endpoint](/api-reference/customer-portal/license-keys/validate) to check or use the license key.
+</Note>
--- a/docs/api-reference/customer-portal/license-keys/activate.mdx
+++ b/docs/api-reference/customer-portal/license-keys/activate.mdx
@@ -2,6 +2,8 @@
 openapi: post /v1/customer-portal/license-keys/activate
 ---
 
-<Note>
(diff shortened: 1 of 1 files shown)


docs/api-reference/customer-portal/license-keys/activate.mdx
```diff
@@ -1,3 +1,7 @@
 ---
 openapi: post /v1/customer-portal/license-keys/activate
----
\ No newline at end of file
+---
+
+<Note>
+  This endpoint is only for license keys that have the **activations** on the benefit enabled. If your license key does **not** support activations, you **should** use this endpoint. Instead, you should use the [validate endpoint](/api-reference/customer-portal/license-keys/validate) to check or use the license key.
```
birkjernstrom: Reads a bit hard at the moment and a bit too disclaimer-y. I think we can shorten it and make it sound more clear and positive, e.g

```
You only need to use this endpoint if you have device **activations** enabled on the license key benefit. You then use this endpoint to reserve an allocation for a specific device. Storing the unique activation ID from the response on the device and using it as extra validation in the [`/validate`](/api-reference/customer-portal/license-keys/validate) endpoint.

Not using **activations**? Just use the [`/validate`](/api-reference/customer-portal/license-keys/validate) endpoint directly instead.
```

birkjernstrom: Perhaps even removing the `<Note>` and simply having it as the intro description for the endpoint.

birkjernstrom (changes requested):

docs/api-reference/customer-portal/license-keys/activate.mdx
rishi-raj-jain: Ah, we'd want to keep the <Note> thing since usually the intro description about the endpoint is more on the side of "using this api key to activate a particular license key"... what we mention here is more of info/note

rishi-raj-jain: we could call it a tip

birkjernstrom (approved):

PR #6348, opened 2025-08-05
Update service.py

rishi-raj-jain:
A [customer reached out](https://app.plain.com/workspace/w_01JE9TRRX9KT61D8P2CH77XDQM/thread/th_01K1QM93FK1NEXJD8TJ3EXQHD3/) that they were able to activate license keys which already had revoked status. I found that while we do check for license key status in `validate` function, we miss checking the status of the license key (or calling the validate function).
This PR aims to ensure that the license key status is checked before it goes through activation flow.


psincraian (approved):
Really nice! 👏

PR #6364, opened 2025-08-08
feat(web): improve landing page performance

Tragio:
Hi team 👋 

First off, huge thanks to the Polar team for building such a great product and saving time from people's lives. Better-Auth + Polar is truly a chef’s kiss! 🔥 

While exploring the landing page, I noticed several opportunities to reduce bandwidth costs, improve performance, and boost the chances of retaining potential customers. After some optimizations, we went from:
* Before: 101 requests, 4.9 MB transferred
* After: 54 requests, 1.2 MB transferred (📉 75% reduction)

-----

**Changes & Rationale**

1. 🧹 Removed unused preconnects  
   - There were preconnect links to `fonts.gstatic.com` and `fonts.googleapis.com`, but I couldn’t find any font usage. May have been leftovers.

2. 🖼️ Optimized background images  
   - Several large images (some ~800 KB) were loaded as CSS background images, which can’t be lazy-loaded.  
   - Converted them to use next/image (already in the project), allowing lazy-loading and modern formats (WebP/AVIF).  
   - Added sizes attributes for better responsiveness.  
   - Hero image is preloaded to improve LCP.

3. ⏳ Deferred Stripe loading  
   - `LandingCheckout` was triggering Stripe and other API calls on page load (191kb).  
   - Changed it to lazy-load when visible to avoid unnecessary network/API calls to improve initial load.  
   - Perhaps a fake image would be enough instead of recreating a dummy checkout?

4. 📦 Reduced Shiki bundle size  
   - Changed `import 'shiki/bundle/full'` to `import 'shiki/bundle/web'`, which is about half the size.  
   - Cached the highlighter instance to avoid recreating it multiple times. This reduces overhead for repeated highlight operations.

5. 👤 Optimized testimonial avatars  
   - Passed a custom image element prop to `@ui/.../Avatar.tsx` to use `next/image`, reducing testimonial avatar size from ~40 KB each to ~1 KB.

Thank you for taking the time to review this! I’m happy to make any further changes or adjustments based on your feedback 🚀

### DIFF
--- a/clients/apps/web/src/app/layout.tsx
+++ b/clients/apps/web/src/app/layout.tsx
@@ -78,12 +78,6 @@ export default async function RootLayout({
   return (
     <html lang="en" suppressHydrationWarning>
       <head>
-        <link rel="preconnect" href="https://fonts.googleapis.com" />
-        <link
-          rel="preconnect"
-          href="https://fonts.gstatic.com"
-          crossOrigin={''}
-        />
         <link
           href="/favicon.png"
           rel="icon"
--- a/clients/apps/web/src/components/Landing/Adapters.tsx
+++ b/clients/apps/web/src/components/Landing/Adapters.tsx
@@ -9,6 +9,7 @@ import {
   TabsTrigger,
 } from '@polar-sh/ui/components/atoms/Tabs'
 import { motion } from 'framer-motion'
+import Image from 'next/image'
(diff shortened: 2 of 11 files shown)


emilwidlund:
Great contribution @Tragio! Really appreciated! 🥳

PR #6397, opened 2025-08-12
Update docs_openapi.yaml

rishi-raj-jain:


frankie567:
I think we should rather fix that in the overlay, otherwise, it won't be picked up by the SDK: https://github.com/polarsource/polar/blob/main/sdk/overlays/security.yml

You can assign the ticket to me if you're unsure about how to do it :)

rishi-raj-jain:
@frankie567 sure yes! I'll learn from your work :) the TLDR is that we need to:

make sure "security": [] is in the JSON object for /validate and /activate endpoints

PR #6409, opened 2025-08-13
Add subscription prorations

malthejorgensen:
- Adds system event for subscription/plan/product changes
- Handles `credit` BillingEntries


server/polar/event/system.py
```diff
@@ -18,6 +18,7 @@ class SystemEvent(StrEnum):
     benefit_revoked = "benefit.revoked"
     subscription_cycled = "subscription.cycled"
     subscription_revoked = "subscription.revoked"
+    subscription_plan_changed = "subscription.plan_changd"
```
frankie567: Typo: `subscription.plan_changed`

---

Also, wondering if we should name in `subscription.product_changed` or `subscription.product_updated`, since "plan" is not a term we usually use in Polar

server/polar/subscription/service.py
```diff
@@ -783,26 +800,90 @@ async def update_product(
...
-            new_prices=[
-                price.stripe_price_id for price in prices if is_static_price(price)
-            ],
-            proration_behavior=proration_behavior.to_stripe(),
-            metadata={
-                "type": ProductType.product,
-                "product_id": str(product.id),
-            },
-        )
+        if organization.subscriptions_billing_engine is False:
```
frankie567: That's related to your question

> How do we resolve contention between a Subscription having stripe_subscription_id not None vs. the Organization having subscriptions_billing_engine is True? If the subscription has stripe_subscription_id one would think it should go through Stripe, but the org is "saying" something else.

My approach on this would be: just rely on `subscription.stripe_subscription_id`, and don't care about the organization setting at this point. I'm quite convinced we'll have situations with organizations that have both subscriptions on Stripe and managed by us.

server/polar/subscription/service.py
```diff
@@ -783,26 +800,90 @@ async def update_product(
...
+            entry_unused_time = BillingEntry(
+                direction=BillingEntryDirection.credit,
+                start_timestamp=cycle_start,
+                end_timestamp=now,
+                amount=round(
+                    previous_product.prices[0].price_amount * old_cycle_pct_remaining
+                ),
+                currency=subscription.currency,
+                customer_id=subscription.customer_id,
+                product_price_id=previous_product.prices[0].id,  # FIXME
```
frankie567: A bit tedious but I think the "right" way to fix that is to loop through all the prices and add a billing entry for each "static price" (see `product/guard.py`). The risk with `prices[0]` is that a metered price could be first in the list vs. a static price.

server/polar/subscription/service.py
```diff
@@ -783,26 +800,90 @@ async def update_product(
...
+            )
+            session.add(entry_unused_time)
+            session.add(entry_remaining_time)
+
+            if proration_behavior == SubscriptionProrationBehavior.invoice:
+                # Invoice immediately
+                from polar.order.service import order as order_service
+
+                await session.flush()
+                await order_service.create_subscription_order(
```
frankie567: I would recommend to trigger that in a background similar to what we do in `cycle`:

```py
enqueue_job("order.subscription_cycle", subscription.id)
```

We could introduce another task or simply rename that one and pass it the reason in argument:

```py
# In cycle
enqueue_job("order.subscription_order", subscription.id, OrderBillingReason.subscription_cycle)

# In update
enqueue_job("order.subscription_order", subscription.id, OrderBillingReason.subscription_update)
```

---

That would also avoid the circular import 😄

frankie567 (commented):
That's awesome work, Malthe 👏 Added a few comments there and there, but nothing major.

---

I'll copy/paste the questions you posted on Slack and answer here for future reference:

> How do we resolve contention between a Subscription having stripe_subscription_id not None vs. the Organization having subscriptions_billing_engine is True? If the subscription has stripe_subscription_id one would think it should go through Stripe, but the org is "saying" something else.

Answered in a comment above.

> We have OrderItem.proration but not BillingEntry.proration – currently it's hard to know if an OrderItem should have .proration set without a .proration field on BillingEntry. The functionality works as-is without any changes here I think, but wondering what we should do in [billing_entry/service.py](https://github.com/polarsource/polar/blob/main/server/polar/billing_entry/service.py#L162) (if not simply remove or ignore the proration field there).
 
Originally, I think I added that `proration` flag to match the data provided by Stripe, which we are syncing. But now that I'm revisiting it, it seems that we are not doing anything useful with it. So yes, we could very well remove it... Unless we think it may be useful in the future, but can't see how right now.

> There's a lot of overlap in my different tests right now, I'm wondering if I should consolidate the current tests even more by just expanding my current @pytest.mark.parametrize. I'm not gonna do that for discounts or more complex test cases, but for the various monthly->yearly, yearly->monthly, basic->pro, pro->basic cases it might make sense to just have those scenarios in a list of parameters for a single test function?

I'll 100% trust you on this. Your tests are very clear and structured, so if you think you can shrink them using more parametrized params, go for it 👍 

> In tests it seems a bit subtle when we're testing Stripe Subscriptions-based functionality vs. our own subscriptions billing engine. I'd probably split tests/subscription/test_service.py into multiple files. One of them being specifically for the Stripe Subscriptions-based things. I was surprised to see that subscriptions_billing_engine is True "by default" in our tests. Defaulting in either direction is fine, but it's an important "fork point" in our code so I'm wondering if we get these somewhat hidden impacts on our tests from subscriptions or organizations defaulting to one side or the other (e.g. we could make it more explicit in the fixtures/fixture creation functions to say create_subscription_using_stripe and create_subscription_not_using_stripe)

Yeah, that's where legacy becomes quite heavy. I *believe* having the flag by default on organizations should be fine (famous last words): the impact will only be on the `handle_success` method of the checkout service, where we choose whether to create the subscription on Stripe or not.

For the rest of the tests, we should be careful about the `stripe_subscription_id` on the subscription fixtures (cf question 1). The `create_subscription` fixtures functions set a value by default, but it can be overridden explicitly to `None`. I've not changed it because lot of tests still expect a Stripe subscription by default.

server/polar/subscription/service.py
malthejorgensen: All good, I can do that 👍 

How does one end up with multiple static prices though? I'm unsure what that would mean :S
Is it a leftover from the legacy model?

frankie567: Normally, you shouldn't have multiple static prices.

However, it's possible to have one static price + `1..n` metered price. In this case, it's possible that `prices[0]` is not the static price, but one of the metered price :)

server/polar/subscription/service.py
malthejorgensen: Done! 👍

server/polar/subscription/service.py
malthejorgensen: Done 👍

server/polar/event/system.py
malthejorgensen: Love that I have a spelling error in there :P `plan_changd`.
Yeah, I'll update it to `product_updated` it seems that's the wording we use most places.

malthejorgensen: Done 👍

server/polar/subscription/service.py
frankie567: Did you tweak the task in `order/tasks.py` to reflect the new name and added parameter? Don't see it there 😊

malthejorgensen: I did not! A mistake on my part. The `enqueue_job`-mock is of course shadowing the call, but even without the mock there's no error :S

I guess that's fair as it's just putting a message on the job queue

Issue #6372, opened 2025-08-08
Customer Portal: Session expiration leads to endless loading when generating invoice


rishi-raj-jain:
### Description

When using the customer portal to generate an invoice, if the user's session has expired, the user can fill out the invoice form and click “Generate Invoice.” Instead of providing an error or redirecting the user to log in again, the portal displays an endless loading state. This leaves the user without guidance or feedback on what went wrong.

### Current Behavior

- User fills out the invoice generation form in the customer portal.
- Session has expired (user is no longer authenticated).
- Upon clicking “Generate Invoice”, the UI shows an endless loading spinner.
- No error message or redirect is presented to the user.

### Expected Behavior

- If the session is expired, the user should be redirected to the login page or shown an error message indicating the session has expired.
- The loading state should not persist indefinitely.

deep-bhikadiya:
i would like to solve this issue .

PR #6415, opened 2025-08-13
fix: [console] handle loading state in DownloadInvoice

MohitTilwani15:
fix: #6372


psincraian:
Nice job @MohitTilwani15 :)

PR #6435, opened 2025-08-14
feat: add manual payment flow

psincraian:
# Context
I created this PR that handles the manual payments. In this PR we can see two important things:

image

## Frontend
- OrderPaymentRetryModal that mimics AddPaymentMethodModal.
- OrderPaymentRetry: it uses Stripe Elements to handle the credit card information. We use the confirmation token in a similar way to the checkout flow.

## Backend
- We added a new `/orders/{id}/confirm-payment` method that given a confirmation token, it will create a PaymentIntent to charge the customer.
- Added a process_retry_payment that validates the status of the order and triggers the payment

# Pending things
- [ ] Tests
- [ ] Use server side events instead of polling (new PR)
- [ ] Allow the customer to save the payment method (new PR)
- [ ] Remove payment_processor_metadata as I don't use it at the end (new PR)

### DIFF
--- a/clients/apps/web/src/components/CustomerPortal/OrderPaymentRetry.tsx
+++ b/clients/apps/web/src/components/CustomerPortal/OrderPaymentRetry.tsx
@@ -243,7 +243,7 @@ export const OrderPaymentRetry = ({
           })
 
         if (!statusError && status) {
-          if (status.status === 'paid') {
+          if (status.status === 'succeeded') {
             cleanupPolling()
             handlePaymentCompletion(true)
             return


server/polar/order/service.py
```diff
@@ -751,6 +773,164 @@ async def trigger_payment(
...
+            "order_id": str(order.id),
+        }
+        if order.tax_rate is not None:
+            metadata["tax_amount"] = str(order.tax_amount)
+            metadata["tax_country"] = order.tax_rate["country"]
+            metadata["tax_state"] = order.tax_rate["state"]
+
+        try:
+            order_repository = OrderRepository.from_session(session)
+            lock_acquired = await order_repository.acquire_payment_lock(order.id)
```
frankie567: Pedantic: we could implement that as a context manager :)

server/polar/order/service.py
```diff
@@ -751,6 +773,164 @@ async def trigger_payment(
...
+                customer=stripe_customer_id,
+                statement_descriptor_suffix=organization.statement_descriptor,
+                description=f"{organization.name} — {order.product.name}",
+                metadata=metadata,
+                return_url=settings.generate_frontend_url(
+                    f"/portal/orders/{str(order.id)}"
+                ),
+            )
+
+            if payment_intent.status == "succeeded":
```
frankie567: Question: we immediately have a "final" status after calling this? There is no kind of "pending" state where we should wait for a webhook or something?

clients/apps/web/src/components/CustomerPortal/OrderPaymentRetry.tsx
```diff
@@ -0,0 +1,406 @@
...
+      clientSecret: result.client_secret,
+    })
+
+    if (error) {
+      handlePaymentCompletion(
+        false,
+        error.message || 'Payment authentication failed',
+      )
+    } else {
+      pollForWebhookResult()
```
frankie567, line 123: Naive: since we process the payment almost synchronously now, couldn't we just wait like ~1 sec and then refetch the order, call it a day? 😄

server/polar/customer_portal/schemas/order.py
```diff
@@ -61,3 +61,33 @@ class CustomerOrderInvoice(Schema):
...
+    confirmation_token_id: str = Field(
+        ..., description="ID of the Stripe confirmation token."
+    )
+
+
+class CustomerOrderPaymentConfirmation(Schema):
+    """Response after confirming a retry payment."""
+
+    status: str = Field(..., description="Payment status after confirmation.")
+    requires_action: bool = Field(
```
frankie567: Is it really needed? It looks like we just check `status == 'requires_action'`.

If you think that's useful, I would suggest to generate it as a computed field:

```py
class CustomerOrderPaymentConfirmation(Schema):
    ...
    
    @computed_field(description="Whether the payment requires additional action (3DS).")
    def requires_action(self) -> bool:
        return status == "requires_action"
```

server/polar/customer_portal/schemas/order.py
```diff
@@ -61,3 +61,33 @@ class CustomerOrderInvoice(Schema):
...
+    status: str = Field(..., description="Current payment status.")
+    error: str | None | None = Field(
+        None, description="Error message if payment failed."
+    )
+
+
+class CustomerOrderConfirmPayment(Schema):
+    """Schema to confirm a retry payment using a Stripe confirmation token."""
+
+    confirmation_token_id: str = Field(
```
frankie567, line 79: Should we add a `payment_processor: PaymentProcessor` field? So the API can be somewhat future-proof if we add other processors.

frankie567 (approved):
Looks sensible, great job! 

A few nitpicks, but feel free to discard them :)

clients/apps/web/src/components/CustomerPortal/OrderPaymentRetry.tsx
psincraian: When status is failed or succeeded, we show that to the user immediately. If the status is `required_action` we do the pulling.

Untitled diagram _ Mermaid Chart-2025-08-19-083029

frankie567: Yes, my point was more: do we actually need to do a polling? Most likely, the order will be marked as paid shortly after a successful payment, so we could just refetch the latest order data.

psincraian: I think in the happy case you are right, just with a simple fetch after some seconds will be enough. But what will happen if the user doesn't accept the 3DS, has insufficient funds, or Stripe takes more time than expected? We still want to show a detailed error message to the user

server/polar/order/service.py
psincraian: Discussing below

clients/apps/web/src/components/CustomerPortal/OrderPaymentRetry.tsx
frankie567: Never mind, I didn't see there was the interstitial call to Stripe SDK. That's great, then 👍

server/polar/order/service.py
frankie567: Where? 😊

psincraian: Sorry, I confused it! If I look correctly at our Stripe Dashboard we only accept payments with immediate confirmation, so no bank debits. Therefore there is no "pending" status. The "required_action" it's the only thing that we should consider, and we should wait for the webhook to get it confirmed.

PR #6436, opened 2025-08-14
Update embed.ts

### LINKED ISSUE #6437: Embed Checkout: A client-side error pops up in event of closing the embedded checkout
A customer reported that when an embed checkout is closed the following error is received:

Image

which is reproducible on https://zoaibkhan.com/shop/angular-dashboard/ by the following steps:
- wait for the page to load
- click the buy now button
- close the modal that pops up
- check the console logs for the error

but the error in the dev mode occupies the whole screen (as the framework shows the client-side errors like that):

Image

**Expected Behavior**

A client-side error does not occur on closing the checkout

rishi-raj-jain:
fix #6437

### DIFF
--- a/clients/.changeset/rotten-files-sneeze.md
+++ b/clients/.changeset/rotten-files-sneeze.md
@@ -1,0 +1,5 @@
+---
+'@polar-sh/checkout': patch
+---
+
+Fix console error when the iframe is already closed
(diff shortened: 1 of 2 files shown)


frankie567:
Nice, we'll see if that makes a difference!

Since it's a package that needs to be published, you should commit a change set so the automatic CI prepares a release.

From the `clients` directory, run `pnpm run changeset` and follow the instructions to create a patch release note for the checkout package. If you're unsure, I can tackle this on Monday.

PR #6473, opened 2025-08-19
Invoice: Fix case where right-side seller is very short

malthejorgensen:
Fixes a case where the seller address would overlap with the line item table in cases where the right-side seller address was very short.

### DIFF
--- a/server/polar/config.py
+++ b/server/polar/config.py
@@ -226,9 +226,7 @@ class Settings(BaseSettings):
         state="CA",
         country=CountryAlpha2("US"),
     )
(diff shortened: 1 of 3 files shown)


frankie567:
You were faster than me 😄

I'll close mine but maybe worth to add the test case: https://github.com/polarsource/polar/pull/6474/files#diff-974409454ff59267bd894bbf35ea1c57abdc76bb0f0f4745f23069e0a818d1ba

malthejorgensen:
@frankie567 Will add your test case 😅

PR #6474, opened 2025-08-19
server/invoice: fix items table overlap with long seller info

frankie567:
- server/invoice: fix items table overlap with long seller info


frankie567:
Duplicate of #6473

PR #6489, opened 2025-08-20
server/rate-limit: introduce groups to have stricter limits by default, but more generous ones for selected orgs

frankie567:
- server/rate-limit: introduce groups to have stricter limits by default, but more generous ones for selected orgs


server/polar/rate_limit.py
```diff
@@ -7,30 +7,36 @@
...
     ],
-    # "^/v1/customer-portal/license-keys/(validate|activate|deactivate)": [
-    #     Rule(second=1, block_time=60, zone="customer-license-key")
-    # ],
-    "^/v1": [Rule(second=100, zone="api")],
+    "^/v1/customer-portal/license-keys/(validate|activate|deactivate)": [
+        Rule(second=1, block_time=60, zone="customer-license-key")
+    ],
+    "^/v1": [
+        Rule(group=RateLimitGroup.default, minute=300, zone="api"),
```
frankie567: Query from Logfire to see if current users would fall under that limit

PR #6515, opened 2025-08-22
Add impersonation for admin users

malthejorgensen:
**Todo**
- [x] Figure out what to do with backoffice -- maybe move logic of impersonate endpoint into service and just have dedicated backoffice endpoints?
- [ ] Fix `end_impersonation` endpoint
- [x] Add UI overlay

### DIFF
--- a/server/polar/auth/dependencies.py
+++ b/server/polar/auth/dependencies.py
@@ -210,7 +210,9 @@ WebUserOrAnonymous = Annotated[
     AuthSubject[Anonymous | User], Depends(_WebUserOrAnonymous)
 ]
 
-_WebUserRead = Authenticator(allowed_subjects={User}, required_scopes={Scope.web_read})
+_WebUserRead = Authenticator(
+    allowed_subjects={User}, required_scopes={Scope.web_read, Scope.web_write}
+)
 WebUserRead = Annotated[AuthSubject[User], Depends(_WebUserRead)]
 
 _WebUserWrite = Authenticator(
--- a/server/polar/benefit/auth.py
+++ b/server/polar/benefit/auth.py
@@ -9,6 +9,7 @@ from polar.auth.scope import Scope
 _BenefitsRead = Authenticator(
     required_scopes={
         Scope.web_read,
+        Scope.web_write,
         Scope.benefits_read,
         Scope.benefits_write,
     },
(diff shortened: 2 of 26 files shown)


server/polar/auth/endpoints.py
```diff
@@ -11,9 +19,176 @@
...
+        value="true",
+        expires=impersonation_session.expires_at,
+        path="/",
+        domain=settings.USER_SESSION_COOKIE_DOMAIN,
+        secure=secure,
+        httponly=False,  # JS-readable
+        samesite="lax",
+    )
+
+    await session.commit()
```
frankie567: No need to commit, we have a dependency handling this globally at request level

server/polar/auth/endpoints.py
```diff
@@ -11,9 +19,176 @@
...
+    )
+
+    # Remove impersonation indicator cookie
+    response.delete_cookie(
+        "impersonation",
+        path="/",
+        domain=settings.USER_SESSION_COOKIE_DOMAIN,
+    )
+
+    await session.commit()
```
frankie567: See comment above

frankie567 (changes requested):
Looks great, awesome job!

I think we can even go the extra mile and get rid of the `web_default` scope right now: since the scope was not stored in the session before, we can assume it's a `web:write` scope for all existing sessions.

server/polar/auth/endpoints.py
```diff
@@ -11,9 +19,176 @@
...
+            secure=secure,
+            httponly=True,
+            samesite="lax",
+        )
+
+    # Set impersonated session cookie
+    response.set_cookie(
+        settings.USER_SESSION_COOKIE_KEY,
+        value=token,
+        expires=impersonation_session.expires_at,
```
psincraian: I'm not sure how long is the session, but I would limit it to 1 hour

server/polar/auth/endpoints.py
malthejorgensen: Fixed 👍

server/polar/auth/endpoints.py
malthejorgensen: Fixed 👍

server/polar/auth/endpoints.py
malthejorgensen: Done ✅

frankie567 (approved):

PR #6649, opened 2025-09-03
server/order: don't try to revoke subscription if already cancelled on last dunning attempt

frankie567:
- server/order: don't try to revoke subscription if already cancelled on last dunning attempt


frankie567:
@psincraian Since it touches dunning, I would like to have your green light on this 😊

PR #6652, opened 2025-09-03
fix(order): ignore payment failure for already paid orders

psincraian:


frankie567:
@psincraian Curious: how can we receive payment failures if the order is already paid?

psincraian:
So, it's an edge case on what happened but it's a valid one:

1. There was a payment triggering
2. Payment fails, and Stripe sends a `charge.failed` message and we ingest it. Message is in the queue.
3. I go to the website and I do a manual payment that succeeds
4. Stripe sends `charge.succeded` and we mark the order as paid
5. The failed message is consumed

image (1)

frankie567:
~I see! Then... Could we make it so we still save that failed payment? We added the ability to record failed payments some months ago because merchants were confused sometimes as to why an order was not paid yet.~

~In this case, it feels like we would lose some interesting information if we just discard it.~

EDIT: Ignore me, it's already the case since it's handled before in the task; nothing to do with `handle_payment_failure` 😊

PR #6706, opened 2025-09-06
fix: webhooks retry on events that are no longer selected

veryayskiy:
## Problem
When users update their webhook endpoint configuration to remove certain event types, failed webhook events that were created before the configuration change continue to retry (up to 10 attempts). This causes users to receive unwanted webhook deliveries for events they've explicitly unchecked during the retry period.

Screenshot 2025-09-05 at 21 39 11

## Solution
Added validation in the webhook task to check if the endpoint still supports the event type before sending. If the event type is no longer supported, the event is skipped and marked as succeeded to prevent further retries.

## Changes:

- Modified polar/webhook/tasks.py to validate event types against current endpoint configuration
- Added graceful error handling for malformed payloads
- Added comprehensive test coverage for the validation logic

Result: Users will no longer receive webhook events for event types they've removed from their webhook configuration, while maintaining normal functionality for supported events.

### DIFF
--- a/server/polar/webhook/tasks.py
+++ b/server/polar/webhook/tasks.py
@@ -1,4 +1,5 @@
 import base64
+import json
 from collections.abc import Mapping
 from ssl import SSLError
 from uuid import UUID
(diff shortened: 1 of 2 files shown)


zantaaaa (approved):

frankie567:
Hi @veryayskiy 👋 Thank you for this proposal, that's interesting!

Ideally, I would go the extra mile and actually store the `type` of the event directly in the `WebhookEvent` model, so we don't have to parse the payload again. I feel it'll be more future-proof and efficient. So basically:

* Add the `type` column to  `WebhookEvent` model
* Set it when creating the `WebhookEvent` (`webhook/service.py::WebhookService::send`)
* Create an Alembic database migration that backfills that column by extracting the type from the payload JSONB column

If you're up to this task, that would be awesome! But no hard feelings if you prefer to pass :)

frankie567:
I'm doing this in a combined effort to solve #6753, thank you for your time 🙏

Issue #6665, opened 2025-09-03
Webhooks: product.updated webhook is fired multiple times on single update


rishi-raj-jain:
### Description

A `product.updated` webhook is being fired multiple times even though the product is only updated once. This leads to duplicate notifications and potentially unintended side effects for webhook consumers.

### Current Behavior

- When a product is updated (single update action), the `product.updated` webhook triggers multiple times.
- Consumers of the webhook receive duplicate events for a single change.

### Expected Behavior

- The `product.updated` webhook should only fire once per product update event.
- Consumers should receive a single webhook notification for each product change.

### Screenshots

_No screenshots provided._

### Environment:

-   Operating System: 
-   Browser (if applicable):

frankie567:
So, this behaviour is kinda expected: under the hood, we first update the product information and _then_ update the associated benefits (two distinct API calls). That's why we trigger two webhooks; which is technically and semantically correct.

In #6832, I tweak the UI so it only trigger the benefits update if they are changes there (newly enabled/disabled benefits). This will then naturally reduce the number of webhooks when updating the product via the dashboard.

PR #6708, opened 2025-09-07
Fix double webhook call

DomiR:
This PR would fix: https://github.com/polarsource/polar/issues/6665

Main issue is that: https://github.com/DomiR/polar/blob/6ec6d23aed2dc88ccc2cda2540cb5c2cc971b780/clients/apps/web/src/components/Products/ProductPage/ProductPageContextView.tsx#L83 calls product update and update benefits.

On the backend, updating benefits will also trigger the `product.update` webhook, which is why the webhook is called when creating a product as well (which is probably also not desired, as there is a product.create webhook).

In the frontend `useUpdateProductBenefits` is only called in 3 places:
- [ProductStep.tsx](https://github.com/DomiR/polar/blob/6ec6d23aed2dc88ccc2cda2540cb5c2cc971b780/clients/apps/web/src/components/Onboarding/ProductStep.tsx#L77): Product creation → product.create webhook, then product.update because of the benefits update
- [CreateProductPage.tsx](https://github.com/DomiR/polar/blob/6ec6d23aed2dc88ccc2cda2540cb5c2cc971b780/clients/apps/web/src/components/Products/CreateProductPage.tsx#L64): Product creation → product.create webhook, then product.update because of the benefits
- [ProductPageContextView](https://github.com/DomiR/polar/blob/6ec6d23aed2dc88ccc2cda2540cb5c2cc971b780/clients/apps/web/src/components/Products/ProductPage/ProductPageContextView.tsx#L59) Product Update and **Culprit** → will call product.update hook for updating product, then again for benefit update

I don't have a complete picture of your architecture, so I'm proposing to remove the product.update webhook for benefits update. Another solution could be to create a benefits.update webhook, if necessary.

### DIFF
--- a/server/.vscode/launch.json
+++ b/server/.vscode/launch.json
@@ -15,8 +15,10 @@
       "name": "API",
       "type": "python",
       "env": {
-        "PYDEVD_DISABLE_FILE_VALIDATION": "1"
+        "PYDEVD_DISABLE_FILE_VALIDATION": "1",
+        "AUTHLIB_INSECURE_TRANSPORT": "true"
       },
+      "envFile": "${workspaceFolder}/.env",
       "request": "launch",
       "module": "uvicorn",
       "args": [
@@ -29,7 +31,7 @@
         "--port",
         "8000"
       ],
-      "justMyCode": false
+      "justMyCode": false,
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


frankie567:
Thank you for your proposal; but unfortunately, it's important to trigger the webhook when we update the benefits of a product: the integration needs to be informed the product has added/removed benefits.

PR #6714, opened 2025-09-07
fix(checkout): Use minimum amount for PWYW placeholder

### LINKED ISSUE #6450: Checkout Links: Minimum value placeholder should reflect configured minimum, not default $10
### Description

When creating a checkout link with "Pay What You Want" and only a minimum value set (no suggested amount), the placeholder in the checkout reflects the default value of $10 instead of the configured minimum value.

Example: https://buy.polar.sh/polar_cl_8j8MmIQFV9zSq7Z4VnNoRRPBXdWdNG3Yb8ZEj4GkKpq

### Current Behavior

- If a minimum value is set but the suggested amount is left empty, the checkout placeholder shows "$10".
- If both minimum and suggested amount are set, the placeholder works as expected.

### Expected Behavior

- If only a minimum value is set (suggested amount is empty), the checkout placeholder should show the minimum value, not the default $10.

Dhanus3133:
### Changes

Updates the checkout creation logic to correctly prioritize the `minimum_amount` from the product's price settings before falling back to the default.

Fixes #6450

### DIFF
--- a/server/tests/checkout/test_endpoints.py
+++ b/server/tests/checkout/test_endpoints.py
@@ -13,7 +13,6 @@ from polar.auth.scope import Scope
 from polar.checkout.repository import CheckoutRepository
 from polar.checkout.schemas import CheckoutProductCreate
 from polar.checkout.service import checkout as checkout_service
-from polar.config import settings
 from polar.enums import SubscriptionRecurringInterval
 from polar.integrations.stripe.service import StripeService
 from polar.kit.tax import calculate_tax
@@ -28,7 +27,6 @@ from polar.models import (
     UserOrganization,
 )
 from polar.models.checkout import CheckoutStatus
-from polar.models.organization import Organization
 from polar.postgres import AsyncSession
 from tests.fixtures.auth import AuthSubjectFixture
 from tests.fixtures.database import SaveFixture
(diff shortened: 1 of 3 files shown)


server/tests/checkout/test_endpoints.py
```diff
@@ -358,6 +399,60 @@ async def test_external_customer_id(
         assert json["external_customer_id"] == "external_customer_id_value"
         assert json["customer_external_id"] == "external_customer_id_value"
 
+    @pytest.mark.auth(AuthSubjectFixture(scopes={Scope.checkouts_write}))
```
frankie567: Great that you added tests. If possible, I would like you to add them in `test_service.py`; as I prefer to have them on the business logic layer. You'll find `TestCreate`, `TestClientCreate` and `TestCheckoutLinkCreate` classes where those cases can fit nicely.

If you can also use a parametrized test (`@pytest.mark.parametrize`) to have those three cases into on test function, that would be awesome!

frankie567 (changes requested):
Great work! If you can address my comment regarding the tests, would be awesome!

server/tests/checkout/test_endpoints.py
Dhanus3133: @frankie567 I’ve made the changes as you suggested.

frankie567:
Thank you, that's really helpful :pray:

Issue #6300, opened 2025-07-27
`{CHECKOUT_ID}` in the success url auto-escapes to `%7BCHECKOUT_ID%7D` upon `Save Link`


rishi-raj-jain:
### Description

Sporadically, using `{CHECKOUT_ID}` in the success URL auto-escapes it to `%7BCHECKOUT_ID%7D` as `Save Link` button is clicked. A customer reported this via email as well.

### Expected Behavior

{CHECKOUT_ID} is not auto-escaped.

### Screenshots

https://github.com/user-attachments/assets/94f96c3a-07e4-4ab0-aea0-613e1de939f2

PR #6715, opened 2025-09-07
fix(checkout-links): Prevent auto-escaping {CHECKOUT_ID} in success URL

Dhanus3133:
### Summary
Fixes an issue where `{CHECKOUT_ID}` in the success URL was auto-escaped to `%7BCHECKOUT_ID%7D` after saving a checkout link.

### Why This Change Was Made
- Libraries like **pypandic** follow RFC 3986 and escape `{}` as `%7B`/`%7D`.
- This ensures the placeholder is kept as-is and replaced correctly when generating the final URL.

Fixes: #6300

### DIFF
--- a/clients/apps/web/src/components/CheckoutLinks/CheckoutLinkForm.tsx
+++ b/clients/apps/web/src/components/CheckoutLinks/CheckoutLinkForm.tsx
@@ -265,7 +265,7 @@ export const CheckoutLinkForm = ({
                   <Input
                     placeholder="https://example.com/success?checkout_id={CHECKOUT_ID}"
                     {...field}
-                    value={field.value || ''}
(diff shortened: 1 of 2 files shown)


frankie567:
Thank you for having a look into this, but I don't think this is the right approach: calling `unquote` on the whole URL opens a security risk, as _everything_ can be decoded, even malicious payloads and stuff like that.

IMO, we should do the following:

* Find a way to "unescape" `{CHECKOUT_ID}` (and only that) when creating/updating that URL. We can probably rely on a custom Pydantic validator for this. Don't forget to use it both in `checkout/schemas.py` _and_ `checkout_link/schemas.py`.
* Create a database migration to unescape `{CHECKOUT_ID}` (and only that) from existing checkout links.

PR #6730, opened 2025-09-08
Add date and date-range filters for webhooks 

### LINKED ISSUE #6560: Add ability to filter webhooks by date (range and specific date)
## Feature Overview

Requesting the ability to filter webhooks by date within the Polar platform. This feature should allow users to:

- Filter webhooks that were triggered between a custom date range
- Filter webhooks on a specific date

## Use Cases

- Auditing webhook activity for specific periods
- Debugging issues that occurred on particular days

## Feature Details

- UI should support filtering webhooks by start and end date
- Optionally, allow filtering by a single calendar date
- Should work with all existing webhook event types

Manas-Kenge:
Fixes #6560

image

Adds support for filtering webhooks by a custom date range or a specific date,

### DIFF
--- a/clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/webhooks/ClientPage.tsx
+++ b/clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/webhooks/ClientPage.tsx
@@ -2,15 +2,32 @@
 
 import { DashboardBody } from '@/components/Layout/DashboardLayout'
 import WebhookSettings from '@/components/Settings/Webhook/WebhookSettings'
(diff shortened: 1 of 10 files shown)


Manas-Kenge:
It should work now

frankie567 (changes requested):
After having a closer look, I now understand what the problem is: you added the filter on the webhooks list page. But what we _really_ want here is to filter the _deliveries_, on the dedicated webhook page.

That's why you have difficulties applying the filter on `useListWebhooksDeliveries`: you're not on the right page.

Manas-Kenge:
Oh, got it! Should I still keep the filter on the webhook list page as well, so we have filters on both pages?

frankie567:
@Manas-Kenge I don't think so, since we likely won't really need to filter them.

PR #6758, opened 2025-09-10
Add support for customizing note in purchase confirmation emails

### LINKED ISSUE #6612: Emails: Ability to customize thank you note in the successful purchase email
> From [the ticket](https://app.plain.com/workspace/w_01JE9TRRX9KT61D8P2CH77XDQM/thread/th_01K34JY0V39P1K442WGQAG0HKH/)

A merchant requested the ability to suggest the thank you note that's visible in the email of successful purchase.

This is different from creating a thank you custom note as it's visible on the payment screen and not in the email sent to the end customer.

This was requested as the user is migration from LS to Polar and they have the following option:
Image
Which shows up as following in emails to the user:
Image

Manas-Kenge:
Fixes #6612 

Provide merchants with a simple way to personalize purchase confirmation emails by adding a custom thank-you message.


frankie567:
We'll adopt a different approach: https://github.com/polarsource/polar/issues/6612#issuecomment-3410322030

Thank you for taking time to look at this 🙏

PR #6790, opened 2025-09-12
subscription/service: Enqueue benefit grants in `product_update()`

malthejorgensen:


frankie567:
Good catch, thank you 🙏

PR #6799, opened 2025-09-12
Remove :w file 👀

patrick91:
Just noticed this while taking a look at the repo 😊


frankie567:
Strange 😄 Good catch, thank you 🙏

PR #6807, opened 2025-09-14
List benefits on customers page

### LINKED ISSUE #6167: List benefits (granted and revoked) on the customer details
It might be interesting when reviewing a customer in the dashboard to view the list of the benefits they have been granted/revoked; to make sure everything works as expected (e.g. after a subscription cancellation or a refund).

Manas-Kenge:
Fixes #6167

### DIFF
--- a/clients/apps/web/src/components/Customer/CustomerBenefit.tsx
+++ b/clients/apps/web/src/components/Customer/CustomerBenefit.tsx
@@ -1,11 +1,7 @@
-// https://github.com/polarsource/polar/issues/6167
 'use client'
 
 import { useCustomerBenefitGrantsList } from '@/hooks/queries/benefits'
-import {
-  DataTablePaginationState,
-  DataTableSortingState,
-} from '@/utils/datatable'
+import { DataTableSortingState } from '@/utils/datatable'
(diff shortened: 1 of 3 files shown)


frankie567 (changes requested):
Thank you @Manas-Kenge, that's interesting 🙏 

Of course, we're limited here with the API we have — that's why we need to loop over the benefits to get that info.

I think it's time we introduce a proper `/v1/benefit-grants/` set of endpoint in the Admin API, where can we basically query all the grants, and filter them by customer, organization, benefit, etc. Similar to what we do in `/v1/customer-portal/benefit-grants`.

I know it's quite of a task so feel free to pass if you don't have time to do it :)

Manas-Kenge:
No worries, I will do it when I get some time.

Manas-Kenge:
@frankie567 can you elaborate more on what you mean when you say “admin API”? Also, where should this API be placed in terms of folder structure?

frankie567:
Hey, sorry @Manas-Kenge, but it looks like @pieterbeulque solved that in a separate effort (#6907). Thank you for your time :pray:

Manas-Kenge:
No worries at all, appreciate you taking the time

PR #6808, opened 2025-09-14
Guide to perform subscription upgrades by merchants and customers

### LINKED ISSUE #6569: docs: Create a guide on how to perform subscription upgrades
$75

as merchants and as customers

(only after #6566 is resolved)

shubhi2000:
fix #6569 
 
This guide walks you through how to upgrade a subscription, both from the merchant’s and the customer’s perspective.

### DIFF
--- a/docs/integrate/sdk/adapters/better-auth.mdx
+++ b/docs/integrate/sdk/adapters/better-auth.mdx
@@ -10,6 +10,7 @@ A [Better Auth](https://github.com/better-auth/better-auth) plugin for integrati
 ## Features
 
 - [Automatic Customer creation on signup](#automatic-customer-creation-on-signup)
+- [Sync customer deletion](#sync-customer-deletion)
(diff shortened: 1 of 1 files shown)


shubhi2000:
Sure, thanks

docs/docs.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://mintlify.com/docs.json",
-  "theme": "palm",
+  "theme": "maple",
```
rishi-raj-jain: can you take pull from the polar's main latest in your branch? go to the GitHub UI > select your branch > sync your branch

basically this change should not appear since docs.json already has maple as noted under https://github.com/polarsource/polar/blob/main/docs/docs.json#L3

but it's not just about this change, I believe other changes show here are also already present

shubhi2000: Thanks for pointing this out. I have created a new PR for this guide where this is resolved. Could you review that?

PR #6825, opened 2025-09-15
Export customers to CSV

pieterbeulque:
image

Current CSV headers:

```csv
ID,External ID,Created At,Email,Name,Tax ID,Billing Address Line 1,Billing Address Line 2,Billing Address City,Billing Address State,Billing Address Zip,Billing Address Country,Metadata
```

### DIFF
--- a/server/polar/customer/repository.py
+++ b/server/polar/customer/repository.py
@@ -104,11 +104,14 @@ class CustomerRepository(
     async def stream_by_organization(
         self,
         auth_subject: AuthSubject[User | Organization],
-        organization_id: Sequence[UUID],
+        organization_id: Sequence[UUID] | None,
     ) -> AsyncGenerator[Customer]:
-        statement = self.get_readable_statement(auth_subject).where(
-            Customer.organization_id.in_(organization_id),
-        )
+        statement = self.get_readable_statement(auth_subject)
+
+        if organization_id is not None:
+            statement = statement.where(
+                Customer.organization_id.in_(organization_id),
+            )
 
         async for customer in self.stream(statement):
             yield customer


rishi-raj-jain:
Can we export all columns (vs selected ones) as done in https://github.com/polarsource/examples/blob/main/export-all-customers-to-csv/index.tsx#L26-L31? or give option in the front-end to deselect/select specific columns?

emilwidlund:
> Can we export all columns (vs selected ones) as done in https://github.com/polarsource/examples/blob/main/export-all-customers-to-csv/index.tsx#L26-L31? or give option in the front-end to deselect/select specific columns?

I'd rather like to avoid too many choices in the frontend for a feature like this.

emilwidlund:
Great job!

pieterbeulque:
@rishi-raj-jain If I understand that code correctly, doesn't that just export a list of emails (one per row)?

The columns that are part of the CSV export in this PR are all fields on the customer model. They're the same as the ones you can see on the customer detail 👇 
 
image

pieterbeulque:
@emilwidlund - Thank you! I moved the export button to align with the _Search Customers_ input because I didn't want to tinker with the section header too much—I like its consistency across other sections of the app. 

Thoughts? I think it's not good not bad… Happy to change it later.

image

pieterbeulque:
@frankie567 - Can you skim the Python part real quick? It's mostly copied over from `subscriptions/export`. My main concern is running this at runtime for big accounts (request timeouts or memory issues), but we can fix that when it's an issue?

emilwidlund:
> @emilwidlund - Thank you! I moved the export button to align with the _Search Customers_ input because I didn't want to tinker with the section header too much—I like its consistency across other sections of the app.
> 
> Thoughts? I think it's not good not bad… Happy to change it later.
> 
> image

Hmm, I would probably prefer if we could add a context-menu-button in the header, in between the plus & sorting button. 

Clicking on that would show the option to download the CSV. It's fine if it's the only option in there for now. It would scale better once we add more actions on Customers as a whole.

Something like -

ray-so-export - 2025-09-15T233916 614

pieterbeulque:
image

Better indeed. I think we can still do better but let's keep it at this for now. Thanks again.

server/polar/customer/endpoints.py
```diff
@@ -77,6 +82,75 @@ async def list(
...
+                "Billing Address Line 2",
+                "Billing Address City",
+                "Billing Address State",
+                "Billing Address Zip",
+                "Billing Address Country",
+                "Metadata",
+            )
+        )
+
+        (customers, _) = await customer_service.list(
```
frankie567: I think we should rely on the repository here rather than a service method (we see it's a bit weird with the pagination parameters and count output)[^1]. Besides, we could leverage streaming capabilities[^2]. This can look like this:

```py
# server/polar/customer/repository.py
class CustomerRepository(...):
    # ...
    async def stream_by_organization(
        self, organization_id: UUID
    ) -> AsyncGenerator[Customer]:
        statement = self.get_base_statement().where(
            Customer.organization_id == organization_id
        )
        async for customer in self.stream(statement):
            yield customer
```

[^1]: I know this is how it's done now for subscriptions 😅 I should probably revisit that too!
[^2]: Actually, I discovered by accident recently our current usage of `stream` doesn't really "stream" records in batches from the DB like I thought; but I'm currently fixing that 😄 (#6839)

frankie567 (changes requested):

pieterbeulque:
@frankie567 - Thanks for the review! I updated the code to reflect your example (taking into account DB session sharing & authentication from the other requests) and make the organization ID match the `MultipleQueryFilter` input for the endpoint.

frankie567 (approved):
Perfect 🙌

server/polar/customer/repository.py
```diff
@@ -101,6 +101,18 @@ async def get_by_stripe_customer_id_and_organization(
         )
         return await self.get_one_or_none(statement)
 
+    async def stream_by_organization(
+        self,
+        auth_subject: AuthSubject[User | Organization],
+        organization_id: Sequence[UUID],
```
frankie567: Actually 🤓 I saw mypy complaining because `organization_id` can actually be `None` from the request. So the solution is to handle that here as well:

```py
organization_id: Sequence[UUID] | None,
) -> AsyncGenerator[Customer]:
    # ...
    if organization_id is not None:
        statement = statement.where(Customer.organization_id.in_(organization_id))
``

frankie567 (changes requested):

PR #6845, opened 2025-09-16
Pieterbeulque/metered pricing with credits

pieterbeulque:
Ignore the first few commits, that fixes some styling inconsistencies that rubbed me the wrong way while trying to debug this issue, haha.

The actual fix for #5972 is the single line in bce56a5:

```diff
  const { data: benefitGrants } = useCustomerBenefitGrants(api, {
 -  order_id: order.id,
 +  checkout_id: order.checkout_id,
    limit: 100,
    sorting: ['type'],
  })
```

As mentioned in #5972, when benefits are granted through a subscription, the `order_id` is set to `null` on the `benefits_grant` table, so we can't retrieve the benefits by `order_id`.

By searching for checkout ID, this will work, because `customer_benefit_grant_service.list` implements filtering on `checkout_id` by an `OR` on `subscription_id` or `order_id`.

### DIFF
--- a/clients/apps/web/src/components/Checkout/CheckoutCard.tsx
+++ b/clients/apps/web/src/components/Checkout/CheckoutCard.tsx
@@ -36,10 +36,8 @@ export const CheckoutCard = ({
       />
 
       {product.benefits.length > 0 ? (
-        <div className="flex flex-col gap-4">
-          <div className="flex flex-col gap-2">
-            <h1 className="font-medium dark:text-white">Included</h1>
-          </div>
+        <div className="flex flex-col gap-2">
+          <h1 className="font-medium dark:text-white">Included</h1>
           <div className="flex flex-col gap-y-2">
             {/* @ts-ignore */}
             <BenefitList benefits={product.benefits} toggle={true} />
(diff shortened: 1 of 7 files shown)


psincraian:
Nice! Have you tested this locally?

pieterbeulque:
@psincraian 

>Nice! Have you tested this locally?

Yep, it appears to be working, showing the granted benefit on each order generated from a subscription cycle.

pieterbeulque:
The only thing I worry about is that `checkout_id` is a nullable column on `orders` so this might not work all the time.

psincraian:
@pieterbeulque on subscription renewals I expect order.checkout_id to be null as there is no checkout involved. Only a payment

frankie567:
Sorry to chime in, but what if we do this: 😄

```ts
  const { data: benefitGrants } = useCustomerBenefitGrants(api, {
    limit: 100,
    sorting: ['type'],
    ....order.subscription_id ? { subscription_id: order.subscription_id } : { order_id: order.id }
  })
  ```

pieterbeulque:
Yeah, checkout isn't going to work then. I think Francois is right, I'll push a revision in a bit

Issue #6357, opened 2025-08-07
Customer portal: Estimated Total does not include tax


rishi-raj-jain:
### Description

The customer portal does not show an accurate Estimated Total because it does not include tax. Even when the order that created the subscription includes tax, and subscription renewals are also expected to be charged tax, the Estimated Total displayed omits this amount.

### Current Behavior

- The Estimated Total shown in the customer portal does not account for tax, leading to inaccurate totals for users.
- Orders that created subscriptions include tax, but this is not reflected in the Estimated Total.
- Future subscription renewals are also likely to charge tax, but the portal does not indicate this in the estimate.

### Expected Behavior

- The Estimated Total in the customer portal should accurately reflect the total amount, including any applicable tax, both for initial orders and future renewals.

### Screenshots

<img width="1638" height="634" alt="Image" src="https://github.com/user-attachments/assets/449a3a26-fb5c-4932-9fc2-5f5a1a3963d5" />

<img width="1050" height="1567" alt="Image" src="https://github.com/user-attachments/assets/452107ce-fd61-4408-b3aa-8ab4675bb544" />

PR #6867, opened 2025-09-17
Add endpoint to preview upcoming subscription charge including taxes

pieterbeulque:
Related to #6357

https://github.com/user-attachments/assets/a9fb9830-a8f1-4ff6-815a-0e7d0f736d46


server/polar/customer_portal/endpoints/subscription.py
```diff
@@ -104,6 +105,72 @@ async def get(
...
+    subtotal_amount: int
+    tax_amount: int
+    total_amount: int
+
+
+@router.get(
+    "/{id}/charge-preview",
+    summary="Preview Next Charge For Active Subscription",
+    response_model=SubscriptionChargePreviewResponse,
+    responses={404: SubscriptionNotFound},
```
frankie567, line 121: Perhaps add the tag `APITag.private`, so it's not (yet) considered as a public API.

server/polar/customer_portal/endpoints/subscription.py
```diff
@@ -104,6 +105,72 @@ async def get(
...
+    subscription = await customer_subscription_service.get_by_id(
+        session, auth_subject, id
+    )
+
+    if subscription is None:
+        raise ResourceNotFound()
+
+    if subscription.status != "active":
+        ## FIXME Is a 404 the correct behavior?
+        raise ResourceNotFound()
```
frankie567, line 139: We could indeed define a specific 400 error, but not a big deal

server/polar/customer_portal/endpoints/subscription.py
```diff
@@ -104,6 +105,72 @@ async def get(
...
+
+    base_price = next(p for p in subscription.prices if p.amount_type == "fixed")
```
frankie567: This won't work with Custom (PWYW) and Free prices. I suggest you go through `subscription_product_prices` and use amount instead:

```py
base_price = sum(p.amount for p in subscription.subscription_product_prices)
```

`subscription.prices` is actually a proxy shortcut that "hop" through `subscription_product_prices`. But we do store the "real" amount directly in `subscription_product_prices`.

server/polar/customer_portal/endpoints/subscription.py
```diff
@@ -104,6 +105,72 @@ async def get(
...
+            subscription.customer.billing_address,
+            [subscription.customer.tax_id]
+            if subscription.customer.tax_id is not None
+            else [],
+            subscription.tax_exempted,
+        )
+
+        tax_amount = tax["amount"]
+
+    total = subtotal + tax_amount
```
frankie567: We should also probably consider the discount. The tricky part is that we'll need to check if the discount is still applicable, like we do in `cycle`.

https://github.com/polarsource/polar/blob/e0e7b3339b6703969c4a24c8049f325c97d81dc5/server/polar/subscription/service.py#L590-L596

https://github.com/polarsource/polar/blob/e0e7b3339b6703969c4a24c8049f325c97d81dc5/server/polar/subscription/service.py#L613-L617

pieterbeulque:
image

server/polar/customer_portal/endpoints/subscription.py
```diff
@@ -104,6 +105,93 @@ async def get(
...
+
+    subtotal_amount = base_price + metered_amount
+
+    discount_amount = 0
+
+    # Ensure the discount has not expired yet
+    if subscription.discount is not None:
+        assert subscription.started_at is not None
+        if subscription.discount.is_repetition_expired(
+            subscription.started_at, subscription.current_period_start
```
frankie567: Since we're looking at the _next_ invoice we should check if the discount expires on the next period. So it should rather be:

```py
subscription.discount.is_repetition_expired(subscription.started_at, subscription.current_period_end)
```

server/polar/customer_portal/endpoints/subscription.py
```diff
@@ -104,6 +105,93 @@ async def get(
...
+        ):
+            subscription.discount = None
```
frankie567: Could we assign `discount` as a local variable? It's a bit scary like this 😄

frankie567 (changes requested):

server/polar/customer_portal/endpoints/subscription.py
pieterbeulque: Haha, yeah, good call, I copied this over without thinking too much, sorry.

PR #6900, opened 2025-09-18
Cancellation metrics

pieterbeulque:
image


pieterbeulque:
@frankie567 mind taking a look at the Python part? It's basically using the same CTE as the active subscriptions metrics but with some other WHERE clauses.

PR #6907, opened 2025-09-19
Show benefit grants on customer page

pieterbeulque:
Adds a new table on the customers page.

image

I decided to add a specific `benefit-grants` endpoint in line with the other endpoints, the existing endpoint to check for grants was `/benefits/{id}/grants` which would require looping over all existing benefits to see if a grant existed for that customer. I think this is better then. Not sure if we should expose it as a public API but I don't think there's much harm in doing so.

### DIFF
--- a/clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/sales/subscriptions/ClientPage.tsx
+++ b/clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/sales/subscriptions/ClientPage.tsx
@@ -4,7 +4,7 @@ import { DashboardBody } from '@/components/Layout/DashboardLayout'
 import { SubscriptionStatus as SubscriptionStatusComponent } from '@/components/Subscriptions/SubscriptionStatus'
 import SubscriptionStatusSelect from '@/components/Subscriptions/SubscriptionStatusSelect'
 import SubscriptionTiersSelect from '@/components/Subscriptions/SubscriptionTiersSelect'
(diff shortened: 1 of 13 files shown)


clients/apps/web/src/components/Benefit/BenefitGrantStatus.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const statusDescription = {
+    Revoked: 'The customer does not have access to this benefit',
+    Granted: 'The customer has access to this benefit',
+    Pending: 'The benefit grant is currently being processed',
+    Error: grant.error?.message ?? 'An unknown error occurred',
+  }
+
+  const statusClassNames = {
+    Revoked: 'bg-red-100 text-red-500 dark:bg-red-950',
+    Granted: 'bg-emerald-200 text-emerald-500 dark:bg-emerald-950',
```
psincraian, line 36: For what I see we don't use a lot of green colors in polar.sh, I think Emil and the team tries to keep it more monochrome

psincraian (approved):
Nice PR! 👏 Have you tested on how it looks on a dark mode?

clients/apps/web/src/components/Benefit/BenefitGrantStatus.tsx
pieterbeulque: Yeah, it's a bit of an odd look but it was pre-existing on the other benefits table, so I kept it in for consistency.

Issue #6892, opened 2025-09-18
Show benefits in the customers view


rishi-raj-jain:
## Is your feature request related to a problem?

> From [the ticket](https://app.plain.com/workspace/w_01JE9TRRX9KT61D8P2CH77XDQM/thread/th_01K5DM9RRXF672R30H11Q38WR8/)

Currently, there's no way to find the benefits granted to a specific customer (from in the customers view). The only way is to go to a specific benefit and look for the customer but this approach fails to work for high number of customers + multiple benefits per customer in case of multiple products (one-time and a subscription for example that both grant some benefits).

## Describe the solution you'd like

Would like that the customers view (/customers) in the Polar dashboard shows the benefits available to that customer.

## Describe alternatives you've considered

NA

## Additional context

NA

naaa760:
@rishi-raj-jain 
i hope you can now see all customer benefits directly in the customers view without having to jump between pages
please let me know!

pieterbeulque:
This was closed in #6907 the day after your issue, but somehow never got closed 😁

PR #6920, opened 2025-09-20
feat: show customer benefits in dashboard view

naaa760:
fixes: #6892

### **Description:**
Added a benefits section to the customer overview page so users can quickly see what benefits each customer has without having to navigate away. The benefits are displayed in a clean table showing the benefit type, when it was granted, and any associated properties. This makes it much easier to manage customers who have multiple benefits from different products or subscriptions.

**What changed:**
- Added a new benefits table between subscriptions and orders
- Shows benefit type, granted date, and properties for each customer
- Uses existing backend API that already had this data available
- Matches the existing design patterns of the customer page


pieterbeulque:
Thanks for looking into this!

This is a duplicate of #6907. Also, the customer state endpoint isn't sufficient because from an admin standpoint it also makes sense to see revoked grants and other information.

nehaaprasad:
@pieterbeulque 
thankyou for letting me know

PR #6921, opened 2025-09-20
Add confirmation modal for deleting discounts

### LINKED ISSUE #6918: Add confirmation modals before destructive actions
For example: https://x.com/thevinodpatidar/status/1969110054281298313

Manas-Kenge:
Fixes #6918

image


pieterbeulque:
@psincraian Do you know what happens when you delete a discount? What happens if it's applied to subscriptions? I assume it'll be gone for the next charge? It'd be good if the description in this modal would be a bit more explanatory about the consequences of deleting a discount.

psincraian:
I think @frankie567 can answer all this questions better

frankie567:
@pieterbeulque The behavior is unfortunately a bit _undefined_. What happens is that we toggle the soft-deletion flag on the discount. But most likely, when handling subscriptions, we retrieve it through a relationship, so it'll still work.

PR #6946, opened 2025-09-22
Update Footer.tsx

rishi-raj-jain:

PR #6949, opened 2025-09-22
Fix `/docs` link

pieterbeulque:
Next.js' prefetching trips up over the fact that `docs` isn't an existing route. `/docs` gets pointed to Mintlify on Cloudflare level.


pieterbeulque:
Replaces #6946

PR #6961, opened 2025-09-23
Always show taxes on Checkout

emilwidlund:
- **checkout: always show taxes, and disable button when update is inflight**

### DIFF
--- a/clients/packages/checkout/src/components/CheckoutForm.tsx
+++ b/clients/packages/checkout/src/components/CheckoutForm.tsx
@@ -87,7 +87,7 @@ interface BaseCheckoutFormProps {
   loading: boolean
   loadingLabel: string | undefined
   disabled?: boolean
-  updateIsPending?: boolean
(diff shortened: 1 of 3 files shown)


clients/packages/checkout/src/components/CheckoutForm.tsx
```diff
@@ -886,6 +894,7 @@ interface CheckoutFormProps {
   loading: boolean
   loadingLabel: string | undefined
   disabled?: boolean
+  updateIsPending?: boolean
```
pieterbeulque: I like `isUpdatePending` better as a pattern to signal it's a boolean

emilwidlund: @pieterbeulque Good idea 👍

emilwidlund: Done ✅

PR #6978, opened 2025-09-24
Forever discounts don't need credit card anymore

### LINKED ISSUE #6934: Subscriptions: Forever discounted still require credit card
### Description

> From [the ticket](https://app.plain.com/workspace/w_01JE9TRRX9KT61D8P2CH77XDQM/thread/th_01K5GXTPE9X1NY5EZHBNG38MX2/)

Subscriptions with a 'forever' discount are still requiring users to update their credit card details even though since the discount code is set to forever, it turns the subscription into a free purchase.

### Current Behavior

- Users with subscriptions that have a 100% discount set to forever require their credit card to be entered during the checkout.
- The system treats these subscriptions as if a valid/updated payment method is required, even though the discount covers the full amount indefinitely.

### Expected Behavior

- Subscriptions with a 'forever' discount should not require users to enter a credit card on checkout

Manas-Kenge:
Fixes #6934

When a forever discount is applied, the credit card section disappears.

### DIFF
--- a/clients/packages/checkout/src/components/CheckoutForm.tsx
+++ b/clients/packages/checkout/src/components/CheckoutForm.tsx
@@ -88,6 +88,7 @@ interface BaseCheckoutFormProps {
   loadingLabel: string | undefined
   disabled?: boolean
   isUpdatePending?: boolean
+  shouldShowPaymentForm?: boolean
   themePreset: ThemingPresetProps
 }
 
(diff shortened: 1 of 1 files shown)


frankie567:
I don't think this will work. Yes, the payment form will be hidden; but I'm fairly certain the backend won't be happy about it when submitting the checkout.

I think this would be better handled by tweaking `is_payment_setup_required` directly in the backend.

Manas-Kenge:
@frankie567 should I open a separate PR with the changes, or can you reopen this PR so I can push the changes here?

PR #7043, opened 2025-09-27
fix: issue 6972

MarioCadenas:
## 📋 Summary

**Related Issue**: Fixes #

## 🧪 Testing

### Test Instructions

1.
2.
3.

## ✅ Pre-submission Checklist

- [ ] All tests pass locally


clients/packages/ui/src/components/ui/select.tsx
```diff
@@ -67,36 +67,84 @@ const SelectScrollDownButton = React.forwardRef<
...
+        onLoadMore?.();
+      }
+    };
+
+    if (onLoadMore) {
+      viewportEl.addEventListener("scroll", loadMoreHandler);
+    }
+
+    return () => {
+      if (onLoadMore) {
```
pieterbeulque: If this `useEffect` triggers because `onLoadMore` is removed, then we'll have a dangling event because `onLoadMore` is falsey in the clean-up code?

```ts
useEffect(() => {
  // ... existing code

  let loadMoreHandler;
  
  if (onLoadMore) {
    loadMoreHandler = () => {
      // Handle the scroll handler
    }
  }
  
  if (loadMoreHandler) {
    viewportEl.addEventListener("scroll", loadMoreHandler)
  }
  
  return () => {
    if (loadMoreHandler) {
      viewportEl.removeEventListener("scroll", loadMoreHandler)
    }
  }
}, [onLoadMore])
```

clients/packages/ui/src/components/ui/select.tsx
```diff
@@ -67,36 +67,84 @@ const SelectScrollDownButton = React.forwardRef<
...
+    const loadMoreHandler = () => {
+      const { scrollTop, scrollHeight, clientHeight } = viewportEl;
+
+      if (scrollTop + clientHeight >= scrollHeight - MAX_SCROLL_THRESHOLD) {
+        onLoadMore?.();
+      }
+    };
+
+    if (onLoadMore) {
+      viewportEl.addEventListener("scroll", loadMoreHandler);
```
pieterbeulque: I prefer doing these with an IntersectionObserver but I wonder if that's still actually relevant for rendering performance these days.

clients/packages/ui/src/components/ui/select.tsx
pieterbeulque: Regardless of that, since `onLoadMore` changes on every render, I wonder if this `useEffect` may cause some rendering hiccups.

clients/packages/ui/src/components/ui/select.tsx
```diff
@@ -67,36 +67,84 @@ const SelectScrollDownButton = React.forwardRef<
...
+            "p-1",
+            position === "popper" &&
+              "h-[var(--radix-select-trigger-height)] w-full min-w-[var(--radix-select-trigger-width)]"
+          )}
+        >
+          {children}
+          {loading && (
+            <div className="flex items-center justify-center py-2 text-sm text-muted-foreground">
+              Loading...
+            </div>
```
pieterbeulque: I appreciate adding a loader state but I don't like how it flickers in your demo, not sure what we could do about it.

clients/packages/ui/src/components/ui/select.tsx
MarioCadenas: Yeah, I think that was a bit hacky on my side. The main problem is that the useEffect gets triggered even when the select has not even opened yet. I changed the implementation, tracking the open/close state, so it only sets up the intersection observer once it's opened (so it can get the element for it, because Radix doesn't mount it until it's opened) and then tracking the onLoadMore function with a ref, so the latest ref gets used in the intersection observer

I think is much cleaner now, thanks for the suggestions 😄

clients/packages/ui/src/components/ui/select.tsx
MarioCadenas: I fully removed it, looks better now

PR #7093, opened 2025-09-30
Handle negative customer balance

frankie567:
We had a case with a customer who had a negative balance, because Stripe can't process payments less than 50 cents. In this case, we should mark the order as paid and carry over the balance to the next invoice.

With those changes, we basically generalize the concept `from_customer_balance` we had to handle positive balances, so it can also handle this case of negative balances. We rename the property and make sure to use the right sign when applying the balance to the invoice.

We also defer the 0.50$ minimum check at payment time.


server/polar/order/schemas.py
```diff
@@ -39,9 +39,16 @@ class OrderBase(TimestampedSchema, IDSchema):
 
     tax_amount: int = Field(description="Sales tax amount in cents.")
     total_amount: int = Field(description="Amount in cents, after discounts and taxes.")
-    from_balance_amount: int = Field(
```
psincraian, line 42: we stop returning this attribute in the API. If someone is using an older SDK version, will it fail?

psincraian (approved):
This starts to become complex! Let's document negative orders and the balance.

I will create something in the handbook or in here.

server/polar/order/schemas.py
frankie567: Indeed, that's why I re-added it as a `computed_field` below:

https://github.com/polarsource/polar/blob/6f198e8366c3e518ceb7bffe2c96132ca792ef22/server/polar/order/schemas.py#L93-L95

The field will still be output in the API (taking its value from `applied_balance_amount`), but `SkipJsonSchema` will hide it from the OpenAPI schema, making it disappear from the docs and newer versions of SDK.

PR #7116, opened 2025-10-01
feat: allow to assign seats to customers

psincraian:


janvaniperen:
This may be just what I need... how do I create a product that allows these seats? Has the documentation been updated?

PR #7143, opened 2025-10-02
Fix for issue #6780 : benefits: file upload of 2 GB fails with correc…

### LINKED ISSUE #6780: benefits: file upload of 2 GB fails
## Description

> From [the ticket](https://app.plain.com/workspace/w_01JE9TRRX9KT61D8P2CH77XDQM/thread/th_01K4T1JW702B43YV7VK1ZAKVW8/)

The file downloads benefit upload for a 5 GB file just does not work as can be seen in the following video:

https://github.com/user-attachments/assets/4245d5c0-cdb3-44e5-af2b-28e1ad2606b9

## Expected Behavior

As in the docs, 10 GB file uploads work fine.

Sample files to test upload with available at https://testfile.org/

vin-dKR:
### Description
This PR fixes issue #6780 where file uploads larger than 2 GB were failing.

Closes #6780

### Changes
- Updated upload handler to slice directly from the `File` object per chunk (no full-file reads).
- Removed whole-file checksum at create; compute per-chunk SHA‑256 instead.
- Optimized memory usage during multipart uploads (O(chunk) instead of O(file)).
- Added loader and surfaced error state in the UI while processing.
- Increased downloadable files limit to 10 GB (frontend maxSize, backend validation).
- Removed the `buffer` and implemented through `file` instead in `Upload.ts`

### Testing
- Uploaded files of 50 MB, 1 GB, 2 GB and 6GB successfully.
- Verified smaller files still work as expected.

https://github.com/user-attachments/assets/5d5f5ff6-60cc-42de-8ccd-13fd8848f510

### DIFF
--- a/clients/apps/web/src/components/FileUpload/Upload.ts
+++ b/clients/apps/web/src/components/FileUpload/Upload.ts
@@ -12,37 +12,37 @@ interface UploadProperties {
   organization: schemas['Organization']
   service: schemas['FileServiceTypes']
   file: File
-  buffer: ArrayBuffer
-  onFileCreate: (response: schemas['FileUpload'], buffer: ArrayBuffer) => void
+  onFileCreate: (response: schemas['FileUpload']) => void
   onFileUploadProgress: (file: schemas['FileUpload'], uploaded: number) => void
   onFileUploaded: (response: FileRead) => void
+  onError: (message: string, error?: Error) => boolean
 }
 
 export class Upload {
   organization: schemas['Organization']
   service: schemas['FileServiceTypes']
   file: File
-  buffer: ArrayBuffer
-  onFileCreate: (response: schemas['FileUpload'], buffer: ArrayBuffer) => void
+  onFileCreate: (response: schemas['FileUpload']) => void
   onFileUploadProgress: (file: schemas['FileUpload'], uploaded: number) => void
   onFileUploaded: (response: FileRead) => void
+  onError: (message: string, error?: Error) => boolean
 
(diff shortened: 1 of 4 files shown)


vin-dKR:
hi @psincraian this is my latest PR i have closed the previous PR  due to some conflicts. Please review this new PR.

server/polar/file/schemas.py
```diff
@@ -26,6 +26,12 @@ class DownloadableFileCreate(FileCreateBase):
     """Schema to create a file to be associated with the downloadables benefit."""
 
     service: Literal[FileServiceTypes.downloadable]
+    size: int = Field(
```
psincraian, line 29: Why is this attribute needed? If you add it here, you should provide a value in the model if not you cannot download it.

psincraian (changes requested):
Sorry for the delay, I missed this.

Left a comment.

clients/apps/web/src/components/FileUpload/Upload.ts
```diff
@@ -64,7 +63,6 @@ export class Upload {
       name: this.file.name,
       size: this.file.size,
       mime_type: mimeType,
-      checksum_sha256_base64: sha256base64,
```
psincraian, line 67: The checksum has been removed; this creates a data integrity risk for corrupted files. 

My recommendation is to implement an incremental checksum per chunk and verify it server side.

psincraian:
No new activity, closing this PR

PR #7151, opened 2025-10-03
feat: Ability to edit organization slug

### LINKED ISSUE #6992: dashboard: ability to edit organization slug
Many merchants have reached out via X DMs and Email for the ability to edit their slugs. While this is a simple request, it ends up consuming quite the bandwidth and requires me to be present in various platforms. Would be great if we could have this in-built feature!

Dhanus3133:
Closes #6992


pieterbeulque:
Thanks for this PR, but changing organization slugs is a phishing attack vector that we want to handle more carefully (i.e. keep old organization slugs in a quarantine, check the new slugs against a list of forbidden slugs, …), so this needs more speccing out.

Dhanus3133:
> Thanks for this PR, but changing organization slugs is a phishing attack vector that we want to handle more carefully (i.e. keep old organization slugs in a quarantine, check the new slugs against a list of forbidden slugs, …), so this needs more speccing out.

@pieterbeulque Thanks for the feedback. I’m still interested in working on this and would love to update the PR with the safeguards you mentioned. Could you clarify a few things, so I can start working on this slowly.

*   How long should we quarantine an old slug?
*   Should the old slug redirect to the new one only during the quarantine and it is not permanent right?
*   Is there an existing package or list you'd prefer for blocking forbidden slugs?

PR #7154, opened 2025-10-03
feat: add a separate high_priority queue

psincraian:
We should be fine to deploy this PR:

* Postgres has a limit of 400 connections [ref](https://render.com/docs/postgresql-creating-connecting?_gl=1*7lmyd0*_gcl_au*ODMwMDc5NDAuMTc1NzkyOTM4NA..*_ga*MjAwNzQwNTA0NC4xNzU3OTI5Mzg0*_ga_QK9L9QJC5N*czE3NTk0OTkxNzUkbzMkZzEkdDE3NTk0OTkxNzkkajU2JGwwJGgw#connection-limits) and we are using ~80
* Redis has a limit of 1k connections. We are only using ~220.

### DIFF
--- a/render.yaml
+++ b/render.yaml
@@ -184,7 +184,7 @@ services:
         fromRegistryCreds:
           name: Registry Credentials (by fvoron)
     plan: pro
-    dockerCommand: uv run dramatiq -p 2 -t 4 --queues high_priority -f polar.worker.scheduler:start polar.worker.run
+    dockerCommand: uv run dramatiq -p 2 -t 4 --queues high_priority polar.worker.run
     healthCheckPath: /
     region: ohio
     numInstances: 1
--- a/render.yaml
+++ b/render.yaml
@@ -96,9 +96,9 @@ services:
       - fromGroup: logfire-server
       - fromGroup: openai-production
 
-  # Worker - High Priority
+  # Worker
   - type: web # Use a web service to allow health checks
-    name: worker-high-priority
+    name: worker
     runtime: image
     image:
       url: ghcr.io/polarsource/polar:latest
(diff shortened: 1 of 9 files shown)


render.yaml
```diff
@@ -106,7 +106,7 @@ services:
         fromRegistryCreds:
           name: Registry Credentials (by fvoron)
     plan: pro
-    dockerCommand: uv run dramatiq -p 4 -t 8 -f polar.worker.scheduler:start polar.worker.run
+    dockerCommand: uv run dramatiq -p 2 -t 4 --queues default -f polar.worker.scheduler:start polar.worker.run
```
frankie567, line 109: We shouldn't start the scheduler on both workers, otherwise we'll get duplicated jobs 🙂

Next stop would be to have a dedicated server _just_ for the scheduler, so we can independently scale the worker horizontally

server/polar/worker/__init__.py
```diff
@@ -178,17 +178,29 @@ class TaskPriority(IntEnum):
...
     priority: TaskPriority = TaskPriority.LOW,
     broker: dramatiq.Broker | None = None,
     **options: Any,
 ) -> Callable[[Callable[P, Awaitable[R]]], Callable[P, Awaitable[R]]]:
+    if queue_name is None:
+        queue_name = (
+            TaskQueue.HIGH_PRIORITY
+            if priority == TaskPriority.HIGH
+            else TaskQueue.DEFAULT
+        )
```
frankie567, line 202: Clever! I like the idea that it's made transparently :)

psincraian: I hate duplicated code :D

render.yaml
psincraian: Aha! Good catch

psincraian: Scheduler is considered a high or normal/medium priority? 🤔

psincraian: pushed to default queue for now

PR #7180, opened 2025-10-06
RFC: Signed JWT as short-lived OAT for roundtrip API calls

pieterbeulque:
As discussed on Slack, signed JWT's as an alternative for organization access tokens. Wanted to take a look at what it'd take and it's actually pretty straight-forward.


pieterbeulque:
Replaced with #7185

PR #7198, opened 2025-10-08
server/payment: handle case where payment_intent error doesn't have a code (thank you Stripe...)

frankie567:
- server/payment: handle case where payment_intent error doesn't have a code (thank you Stripe...)


Kludex:
I've been there. 👀

PR #7210, opened 2025-10-09
fix: non-summable metrics should take into account the full billing period events

psincraian:


server/polar/billing_entry/repository.py
```diff
@@ -91,6 +110,22 @@ async def get_pending_ids_by_subscription_and_price(
...
+            self.get_pending_by_subscription_statement(subscription_id)
+            .join(
+                ProductPriceMeteredUnit,
+                BillingEntry.product_price_id == ProductPriceMeteredUnit.id,
+            )
+            .with_only_columns(BillingEntry.id)
+            .where(ProductPriceMeteredUnit.meter_id == meter_id)
+        )
+        results = await self.session.execute(statement)
+        return results.scalars().unique().all()
```
frankie567, line 126: You could use as a shortcut:

```py
return await self.get_all(statement)
```

server/polar/billing_entry/repository.py
```diff
@@ -58,25 +58,44 @@ async def get_static_pending_by_subscription(
...
+                ProductPriceMeteredUnit.meter_id,
                 func.min(BillingEntry.start_timestamp),
                 func.max(BillingEntry.end_timestamp),
+                latest_price_created_at,
+            )
+            .group_by(BillingEntry.product_price_id, ProductPriceMeteredUnit.meter_id)
+            .order_by(None)  # Clear existing ORDER BY from base statement
+            .order_by(
+                ProductPriceMeteredUnit.meter_id.asc(),
+                latest_price_created_at.desc(),
```
frankie567: I think looking at the price creation date is not the right call: a customer may have switched to another Product that may have been created earlier than its current price.

The source of truth for the current price is `SubscriptionProductPrice`. Maybe the solution is not to sort the price and take the first one, but rather than finding the active price for this meter when iterating through the tuples in `compute_pending_subscription_line_items`.

psincraian: Ahhh, you are right. I confused the entity meanings here 🥲

Issue #6817, opened 2025-09-15
Introduce chat widget


rishi-raj-jain:
@psincraian 

Can we use the following code snippet in the Polar dashboard to load a live chat widget with Polar when someone clicks "Chat with us" under "Support" like as in Plan as follows:

<img width="237" height="342" alt="Image" src="https://github.com/user-attachments/assets/d86ab4eb-5f6d-424d-be8e-21372795ac8b" />

```html
<script>
(function(d, script) {
  script = d.createElement('script');
  script.async = false;
  script.onload = function(){
    Plain.init({
      appId: 'liveChatApp_01K55GCP6V4FJ0DKDP604MZYDV',
    });
  };
  script.src = 'https://chat.cdn-plain.com/index.js';
  d.getElementsByTagName('head')[0].appendChild(script);
}(document));
</script>
```

and can we also link the Documentation + API Reference as other items?

emilwidlund:
Not 100% onboard with this. Let me think this through a bit.

rishi-raj-jain:
What do you mean? Can you share more?

rishi-raj-jain:
@emilwidlund agreed to do this as per the conversation in https://polar-olw4768.slack.com/archives/C0942KS0BD2/p1758536589096499. @psincraian can we ship this with a little explicit mention that the chats via widget is not a real-time one.

PR #7213, opened 2025-10-09
feat: Introduce chat widget

rishi-raj-jain:
fix https://github.com/polarsource/polar/issues/6817

CSP as per https://plain.support.site/article/live-chat-overview#content-security-policy-csp

https://github.com/user-attachments/assets/b8daae5e-dcba-4df4-bf5a-a81f60dd8506


pieterbeulque:
Nice, are we launching in-app Spotify? 😁

rishi-raj-jain:
> What is the network error that is displaying in the dashboard? 🤔

That's because I shut down my terminal that ran the APIs while recording. I can re-record if required :)

PR #7224, opened 2025-10-10
dash: slighly adjust lower dashboard nav footer

emilwidlund:


rishi-raj-jain:
@emilwidlund can you also handle https://github.com/polarsource/polar/pull/7222 in this? I've closed that one since it's at the same section and we can save time.

PR #7294, opened 2025-10-16
Allow searching customers by external ID

pieterbeulque:
QOL improvement for people looking to find a customer by their external ID

### DIFF
--- a/server/polar/customer/endpoints.py
+++ b/server/polar/customer/endpoints.py
@@ -63,7 +63,9 @@ async def list(
         None, title="OrganizationID Filter", description="Filter by organization ID."
     ),
     email: str | None = Query(None, description="Filter by exact email."),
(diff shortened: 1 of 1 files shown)


frankie567:
@pieterbeulque Wouldn't we want an exact filter rather than a fuzzy search?

pieterbeulque:
@frankie567 I think incomplete matches should work so that you can search `e7c24ef3` to match on `e7c24ef3-2151-4641-b565-4fd276891cee`. But I changed the query to prefix `ilike "{query}%"` so that we can use the existing index efficiently.

frankie567:
Ok, so it's more of a QoL for the dashboard, that was the thing I was missing :)

frankie567 (approved):

pieterbeulque:
Ah, yes, sorry!

PR #7297, opened 2025-10-16
Create meters for new customers if there are existing events

pieterbeulque:
Customers with pre-existing events for their external_id will not have meters linked until they ingest another event. This fixes that.

This code errs on the safe side of things to only run when it will be relevant because I didn't want to stray too much from the current implementation. Maybe that's unnecessary and we can skip the `has_events` logic and just run `touch_meters` always (or when there's an `external_id` – if `external_id` is not null the meters will definitely all be empty)

### DIFF
--- a/server/polar/customer/repository.py
+++ b/server/polar/customer/repository.py
@@ -3,7 +3,7 @@ from collections.abc import AsyncGenerator, Iterable, Sequence
 from typing import Any
 from uuid import UUID
 
-from sqlalchemy import Select, exists, func, select, update
+from sqlalchemy import Select, func, select, update
 
 from polar.auth.models import AuthSubject, Organization, User, is_organization, is_user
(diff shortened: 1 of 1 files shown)


server/polar/customer/repository.py
```diff
@@ -41,6 +41,22 @@ async def create_context(
         customer = await self.create(object, flush=flush)
         yield customer
         assert customer.id is not None, "Customer.id is None"
+
+        # If the customer has an external_id, check if there are existing events
+        # for that external_id. If so, touch meters to trigger meter calculations.
+        if customer.external_id is not None:
```
frankie567, line 47: I would do something even simpler:

```py
if customer.external_id is not None:
    enqueue_job("customer_meter.update_customer", customer.id)
```

This way, we don't need to worry here about the existence of events or touching meters.

PR #7312, opened 2025-10-16
fix: checkout query param prefill

### LINKED ISSUE #7156: Checkout Links: amount query param is not working in the pay what you want product
> From [the ticket](https://app.plain.com/workspace/w_01JE9TRRX9KT61D8P2CH77XDQM/thread/th_01K6JRGRVX75GJXJGZBK2HRA1B/)

As shared on https://polar.sh/docs/features/checkout/links#query-parameters, adding the query param ?amount=40 for example in the checkout link should change the amount for the checkout session generated but the query param isn't currently being respected in the Polar system.

Sample checkout link - https://sandbox-api.polar.sh/v1/checkout-links/polar_cl_LWsLiCeLWDm2UBbJqCXAR8JNVdGkvg9tpe9HH4HAHmX/redirect?amount=40

Yopi:
## 📋 Summary

**Related Issue**: Fixes #7156

Fix the prefill/override of Checkout form with query params by actually storing the overridden values in the CheckoutFormProvider.

It looks like the problem is that we are setting the prefilled values to the CheckoutForm `form`, but not the `checkout`, and in `CheckoutPWYWForm` we are reading the value from the `checkout` object. 

This updates the `CheckoutPWYWForm` to use the CheckoutForm `form`, and ensures that we set the prefilled values on mount. Additionally it reuses the getCents helper from the ui library (now exposed) so that we have uniform handling of cents.

## 🧪 Testing

- [X] I have tested these changes locally
- [X] All existing tests pass (`uv run task test` for backend, `pnpm test` for frontend)
- [X] I have added new tests for new functionality
- [X] I have run linting and type checking (`uv run task lint && uv run task lint_types` for backend)

### Test Instructions

1. Add a query param to the checkout link.
2.
3.

## ✅ Pre-submission Checklist

- [X] All tests pass locally

### DIFF
--- a/clients/packages/checkout/src/components/CheckoutPWYWForm.tsx
+++ b/clients/packages/checkout/src/components/CheckoutPWYWForm.tsx
@@ -10,8 +10,9 @@ import {
   FormMessage,
 } from '@polar-sh/ui/components/ui/form'
 import { ThemingPresetProps } from '@polar-sh/ui/hooks/theming'
-import { useEffect } from 'react'
-import { useForm } from 'react-hook-form'
+import { getCents } from '@polar-sh/ui/lib/money'
+import { useEffect, useRef } from 'react'
+import { UseFormReturn } from 'react-hook-form'
 import useDebouncedCallback from '../hooks/debounce'
 import { formatCurrencyNumber } from '../utils/money'
 
@@ -36,48 +37,66 @@ const DollarSignIcon = ({ className }: { className?: string }) => {
 }
 
 export interface CheckoutPWYWFormProps {
-  update: (data: CheckoutUpdatePublic) => void
+  update: (data: CheckoutUpdatePublic) => Promise<CheckoutPublic>
   checkout: CheckoutPublic
+  form: UseFormReturn<CheckoutUpdatePublic>
   productPrice: ProductPriceCustom
   themePreset: ThemingPresetProps
+  prefilledParameters?: Record<string, string>
 }
 
 export const CheckoutPWYWForm = ({
   update,
   checkout,
+  form,
   productPrice,
   themePreset,
+  prefilledParameters,
 }: CheckoutPWYWFormProps) => {
-  const { amount } = checkout
+  const { control, trigger, reset, watch, getValues } = form
+  const appliedPrefilledAmountRef = useRef<string | null>(null)
 
-  const form = useForm<{ amount: number }>({
-    defaultValues: { amount: amount || 0 },
-  })
-  const { control, trigger, reset, watch } = form
-
-  const debouncedAmountUpdate = useDebouncedCallback(
-    async (amount: number) => {
-      const isValid = await trigger('amount')
-      if (isValid) {
-        update?.({ amount })
-      }
-    },
-    600,
-    [update, trigger],
-  )
+  const updateAmount = async (amount: number) => {
+    const isValid = await trigger('amount')
+    if (isValid) {
+      update?.({ amount })
+    }
+  }
+  const debouncedUpdateAmount = useDebouncedCallback(updateAmount, 600, [
+    update,
+    trigger,
+  ])
 
   useEffect(() => {
     const subscription = watch(async (value, { name }) => {
-      if (name === 'amount' && value.amount !== undefined) {
-        debouncedAmountUpdate(value.amount)
+      if (
+        name === 'amount' &&
+        value.amount !== undefined &&
+        value.amount !== null
+      ) {
+        debouncedUpdateAmount(value.amount)
       }
     })
     return () => subscription.unsubscribe()
-  }, [watch, debouncedAmountUpdate])
+  }, [watch, debouncedUpdateAmount])
 
   useEffect(() => {
-    reset({ amount: amount || 0 })
-  }, [amount, reset])
+    const prefilledAmountString = prefilledParameters?.amount
+    if (!prefilledAmountString) {
+      return
+    }
+
+    if (appliedPrefilledAmountRef.current !== prefilledAmountString) {
+      appliedPrefilledAmountRef.current = prefilledAmountString
+      const amountInCents = getCents(prefilledAmountString)
+
+      // Use reset to update the form with the correct cent value
+      const currentValues = getValues()
+      reset({ ...currentValues, amount: amountInCents })
+
+      updateAmount(amountInCents)
+    }
+  }, [prefilledParameters?.amount, reset, getValues, updateAmount])
 
   let customAmountMinLabel = null
   let customAmountMaxLabel = null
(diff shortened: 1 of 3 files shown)


clients/packages/checkout/src/providers/CheckoutFormProvider.tsx
```diff
@@ -74,23 +80,46 @@ export const CheckoutFormProvider = ({
...
-      ...checkout,
+      ...baseCheckout,
       customerBillingAddress:
-        checkout.customerBillingAddress as AddressInput | null,
-      discountCode: checkout.discount ? checkout.discount.code : undefined,
+        baseCheckout.customerBillingAddress as AddressInput | null,
+      discountCode: baseCheckout.discount
+        ? baseCheckout.discount.code
+        : undefined,
       ...(prefilledParameters ? unflatten(prefilledParameters) : {}),
```
pieterbeulque, line 89: We can use `prefilledData` here as well then to avoid running `unflatten` twice.

pieterbeulque:
> If that's not the case, I think we should be reading the value from the form in CheckoutPWYWForm instead of from the checkout object.

I think this is the correct interpretation. `checkout` is the representation of the checkout session in the database, `form` is the interactive form representation for the UI. Changing the `form` can go to the API which will then update the `checkout`. I believe that's the hierarchy for the data.

So instead of tinkering with `checkout` I'd pass `form` to the `CheckoutPWYWForm` similar to how the `CheckoutForm` is also passed `form` and `checkout` as two separate props.

But honestly the code in `CheckoutPWYWForm` weirds me out a bit:

```ts
  useEffect(() => {
    reset({ amount: amount || 0 })
  }, [amount, reset])
```

which I think is meant to reset the input to the amount set from the API validation after `debouncedAmountUpdate` pushed it to the API.

So the hairy part about this issue is that if the `amount` is set from the querystring to be completely consistent that'd immediately have to trigger a `debouncedAmountUpdate` to run the server-side validation of the amount.

Yopi:
> > If that's not the case, I think we should be reading the value from the form in CheckoutPWYWForm instead of from the checkout object.
> 
> I think this is the correct interpretation. `checkout` is the representation of the checkout session in the database, `form` is the interactive form representation for the UI. Changing the `form` can go to the API which will then update the `checkout`. I believe that's the hierarchy for the data.
> 
> So instead of tinkering with `checkout` I'd pass `form` to the `CheckoutPWYWForm` similar to how the `CheckoutForm` is also passed `form` and `checkout` as two separate props.
> 
> But honestly the code in `CheckoutPWYWForm` weirds me out a bit:
> 
> ```ts
>   useEffect(() => {
>     reset({ amount: amount || 0 })
>   }, [amount, reset])
> ```
> 
> which I think is meant to reset the input to the amount set from the API validation after `debouncedAmountUpdate` pushed it to the API.
> 
> So the hairy part about this issue is that if the `amount` is set from the querystring to be completely consistent that'd immediately have to trigger a `debouncedAmountUpdate` to run the server-side validation of the amount.

All right, I think this is working now. I tried to push as much of the implementation into the PWYW component, and make it lean on the existing functionality to not hack it in.

clients/packages/checkout/src/providers/CheckoutFormProvider.tsx
```diff
@@ -79,13 +87,24 @@ export const CheckoutFormProvider = ({
   const [loadingLabel, setLoadingLabel] = useState<string | undefined>()
   const [isUpdatePending, setIsUpdatePending] = useState(false)
 
+  const prefilledData = useMemo(() => {
+    if (!prefilledParameters) return {}
+
+    const data = unflatten(prefilledParameters)
+    if (data.amount !== undefined) {
+      data.amount = getCents(String(data.amount))
+    }
```
pieterbeulque: I believe this isn't used and only the code in PWYWForm is relevant?

clients/packages/checkout/src/components/CheckoutPWYWForm.tsx
```diff
@@ -36,48 +37,72 @@ const DollarSignIcon = ({ className }: { className?: string }) => {
...
-  }, [amount, reset])
+  }, [amount, reset, prefilledParameters?.amount, getValues])
+
+  useEffect(() => {
+    const prefilledAmountString = prefilledParameters?.amount
+    if (!prefilledAmountString) {
+      return
+    }
+
+    if (appliedPrefilledAmountRef.current !== prefilledAmountString) {
```
pieterbeulque, line 89: This is pretty painful 😁 is the reason we have to do this that the `useEffect` with `amount, reset` dependencies triggers on render and it resets the amount to `0` before the client has had the chance to run `amountUpdate`

clients/packages/checkout/src/components/CheckoutPWYWForm.tsx
```diff
@@ -36,48 +37,72 @@ const DollarSignIcon = ({ className }: { className?: string }) => {
...
   const { amount } = checkout
 
-  const form = useForm<{ amount: number }>({
-    defaultValues: { amount: amount || 0 },
-  })
-  const { control, trigger, reset, watch } = form
+  const { control, trigger, reset, watch, getValues } = form
+  const appliedPrefilledAmountRef = useRef<string | null>(null)
 
+  const nonDebouncedAmountUpdate = async (amount: number) => {
```
pieterbeulque: Nitpick and I understand how you landed on this but probably `updateAmount` and `debouncedUpdateAmount` are more obvious method names than `nonDebouncedXYZ` 😬

clients/packages/checkout/src/components/CheckoutPWYWForm.tsx
pieterbeulque: It's fine but I think this will eventually bite back

clients/packages/checkout/src/components/CheckoutPWYWForm.tsx
Yopi: Hahah, yes, that does make more sense

clients/packages/checkout/src/providers/CheckoutFormProvider.tsx
Yopi: That was my assumption as well, but it gets used on page refresh. That might be indicative on something being wrong though, because I agree that it shouldn't be needed anymore.

clients/packages/checkout/src/components/CheckoutPWYWForm.tsx
Yopi: Yes, exactly the case 🤔

Yopi:
I moved in all of the logic into PWYWForm, updated the naming to be a bit more sensical and kept the useEffect with the ref to avoid a loop on updating it. Feels like it would be worth to revisit this at some point to make the full flow a bit nicer, but I think this works now.

Yopi:
#7366 will be a better way to solve it

PR #7315, opened 2025-10-16
Re-implement PostHog according to docs

pieterbeulque:
Simplify the implementation using the full JS SDK & its included React provider.

### DIFF
--- a/clients/apps/web/package.json
+++ b/clients/apps/web/package.json
@@ -71,7 +71,7 @@
     "next": "15.5.4",
     "next-themes": "^0.4.6",
     "nuqs": "^2.7.1",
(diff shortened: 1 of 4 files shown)


pieterbeulque:
cc @frankie567 - the full PostHog JS SDK is tree-shakeable enough that it doesn't make an impact on bundle size, apparently.

frankie567:
Ok, great! Guarantee you this wasn't the case a few months before 😅

frankie567:
~10% of the page size: https://github.com/polarsource/polar/pull/4897

pieterbeulque:
That's wild! Well, I hope the bundle analysis script above is correct then 😁

PR #7333, opened 2025-10-17
Create placeholder customers when ingesting events for unknown external ID's

pieterbeulque:


server/polar/customer/service.py
```diff
@@ -314,3 +314,7 @@ async def webhook(
 
 
 customer = CustomerService()
+customer = CustomerService()
```
pieterbeulque: Ignore—this is a weird bug with my auto-formatting in VS Code.

server/migrations/versions/2025-10-17-1145_make_customer_email_nullable.py
```diff
@@ -0,0 +1,32 @@
...
+down_revision = "59a5d45ae3fd"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
+    # Make email nullable - NULL email indicates a placeholder customer
+    op.alter_column(
+        "customers", "email", existing_type=sa.String(length=320), nullable=True
+    )
```
Yopi, line 25: In this case I think it might make sense. I'm not a huge fan of null values, but it might be riskier to assign them to a "unknown" customer as well as would be the alternative.

Yopi: Or if we should have an additional field being "placeholder". Thinking about the balance of being explicit and having a more compressed database schema.

server/polar/models/customer.py
```diff
@@ -258,3 +269,39 @@ def billing_name(self, value: str | None) -> None:
...
+
+    if TYPE_CHECKING:
+        # Type narrowing: email is always None for PlaceholderCustomer
+        email: Mapped[None]
+
+    @property
+    def legacy_user_public_name(self) -> str:
+        if self.name:
+            return self.name[0]
+        return "?"
```
Yopi, line 339: Where do the `legacy_user_public_name` show up? Is it basically never?

pieterbeulque: I have no clue 😁

frankie567: It's used in Order & Subscription schemas where we may still refer to a user. Legacy story: some months ago, we didn't have a Customer entity: everybody was a Polar user!

pieterbeulque:
Now that we have the option to subscribe your customers to a free plan, maybe it's better to make that the recommended way to track usage on free users and not go through with this PR.

That means we don't properly support anonymous usage yet (email remains the only required field) but I think that's a fair trade-off to make compared to this.

frankie567:
Agreed, let's avoid too much complex changes for now if we have a decent workaround. We can always revisit that if the needs arise.

PR #7336, opened 2025-10-17
feat: charge the billing owner for seat + usage based

psincraian:
In this PR we allow charging the billing manager for the seat subscriptions. Mainly we do the following:

Get the Customer Seats events, and for each one of those events, create a BillingEntry for the billing manager. This way, SubscriptionMeters get updated correctly, and the billing manager is charged accordingly.

CustomerMeters are still per customer seat holder, as in a next step we may show individualized usage. 

> **Important**
> Seats based pricing act as a pooling pricing. So credit grants and usages are combined across all customers.

### DIFF
--- a/server/polar/customer_seat/repository.py
+++ b/server/polar/customer_seat/repository.py
@@ -1,5 +1,4 @@
 from collections.abc import Sequence
-from typing import TYPE_CHECKING
 from uuid import UUID
 
 from sqlalchemy import Select, select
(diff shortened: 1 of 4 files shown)


server/polar/customer_seat/repository.py
```diff
@@ -199,6 +198,29 @@ async def get_by_subscription_and_auth_subject(
         )
         return await self.get_one_or_none(statement)
 
+    async def get_active_seat_for_customer(
+        self,
+        customer_id: UUID,
+        *,
+        options: tuple["_AbstractLoad", ...] = (),
```
frankie567: We have a type alias for that in `server/polar/kit/repository/base.py`: 

https://github.com/polarsource/polar/blob/5efee7ad48432bd0e7da496fc23763c2a3dd3b79/server/polar/kit/repository/base.py#L31

frankie567 (commented):
I think it's working! 

I'm wondering though, in the sake of readability, if we could abstract all that inside `subscription_product_price` repository/service. Because, ultimately, what we need in `create_billing_entries` is two things: the paying customer and the price.

So I think we could have something like:

```py
try:
    holder_customer, subscription_product_price = customer_price_map[customer.id]
except KeyError:
    holder_customer, subscription_product_price = await subscription_product_price_service.get_by_customer_and_meter(
        customer.id, meter.id  # Do all the work with seats inside there.
    )
    customer_price_map[customer.id] = holder_customer, subscription_product_price
```

And then keep the rest of the logic untouched. So most of the tedious work would be around `get_by_customer_and_meter`. Maybe we could even get that data in one call with a clever SQL query 😄

PR #7344, opened 2025-10-19
"Get started with metering" code snippet doesn't work on the Meter Page (Sandbox)

adasq:
## 📋 Summary

Hey Team,

I've noticed, that for `sandbox` web interface, the meter code snippet from Meter Page (`https://sandbox.polar.sh/dashboard/{orgId}/usage-billing/meters?selectedMeter={meterId}`) doesn't work.

image

It throws with the following issue, even though the TOKEN is valid.
```
API error occurred: Status 401
Body: {"error": "invalid_token", "error_description": "The access token provided is expired, revoked, malformed, or invalid for other reasons."}
```

It's likely because of default `server` points to `production`, rather than `sandbox`. Defining it in config does the trick:

```
const polar = new Polar({
    accessToken: "{TOKEN}",
    server: "sandbox"
});
```

The following PR changes extend the page code snippet with the missing field (`server: 'sandbox'`) in SDK configuration.

### DIFF
--- a/clients/apps/web/src/components/Meter/MeterGetStarted.tsx
+++ b/clients/apps/web/src/components/Meter/MeterGetStarted.tsx
@@ -1,5 +1,5 @@
-import { schemas } from '@polar-sh/client'
 import { CONFIG } from '@/utils/config'
+import { schemas } from '@polar-sh/client'
 
 import {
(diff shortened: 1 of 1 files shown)


frankie567:
Thanks, that's helpful 👍

PR #7356, opened 2025-10-20
feat(order): Add trial over label in statement description

### LINKED ISSUE #6962: Add a "TRIAL OVER" label to the statement descriptor after trial ends
According to [Stripe](https://docs.stripe.com/billing/subscriptions/trials#compliance), some payment networks require to ad a specific label to the statement descriptor when performing the first payment after a free trial period.

We should probably do that as well.

Yopi:
## 📋 Summary

**Related Issue**: Fixes #6962

This PR adds a "TRIAL OVER" suffix to the statement descriptor for the first payment after the trial is over.

## 🎯 What

Introduces an internal BillingReason enum that is mapped toward the existing/external BillingReason for backward compatability. We could decide to introduce it as a breaking change going forward, but this allows us to decouple the two for now.

## 🧪 Testing

- [X] I have tested these changes locally
- [X] All existing tests pass (`uv run task test` for backend, `pnpm test` for frontend)
- [X] I have added new tests for new functionality
- [X] I have run linting and type checking (`uv run task lint && uv run task lint_types` for backend)

### Test Instructions

1.
2.
3.

## ✅ Pre-submission Checklist

- [X] All tests pass locally

### DIFF
--- a/server/polar/order/service.py
+++ b/server/polar/order/service.py
@@ -318,16 +318,6 @@ def _is_empty_customer_address(customer_address: dict[str, Any] | None) -> bool:
 
 
 class OrderService:
-    def get_statement_descriptor_suffix(
-        self, order: Order, organization: Organization
-    ) -> str:
-        if (
-            order.billing_reason
-            == OrderBillingReasonInternal.subscription_cycle_after_trial
-        ):
-            return f"{organization.statement_descriptor} TRIAL OVER"
-        return organization.statement_descriptor
-
     @asynccontextmanager
     async def acquire_payment_lock(
         self, session: AsyncSession, order: Order, *, release_on_success: bool = True
@@ -903,9 +893,7 @@ class OrderService:
                         customer=stripe_customer_id,
                         confirm=True,
                         off_session=True,
-                        statement_descriptor_suffix=self.get_statement_descriptor_suffix(
-                            order, order.organization
-                        ),
+                        statement_descriptor_suffix=order.statement_descriptor_suffix,
                         description=f"{order.organization.name} — {order.product.name}",
                         metadata=metadata,
                     )
@@ -1014,9 +1002,7 @@ class OrderService:
                         payment_method=saved_payment_method.processor_id,
                         customer=customer.stripe_customer_id,
                         confirm=True,
-                        statement_descriptor_suffix=self.get_statement_descriptor_suffix(
-                            order, organization
-                        ),
+                        statement_descriptor_suffix=order.statement_descriptor_suffix,
                         description=f"{organization.name} — {order.product.name}",
                         metadata=metadata,
                         return_url=settings.generate_frontend_url(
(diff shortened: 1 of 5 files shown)


server/polar/order/service.py
```diff
@@ -314,6 +318,16 @@ def _is_empty_customer_address(customer_address: dict[str, Any] | None) -> bool:
...
 class OrderService:
+    def get_statement_descriptor_suffix(
+        self, order: Order, organization: Organization
+    ) -> str:
+        if (
+            order.billing_reason
+            == OrderBillingReasonInternal.subscription_cycle_after_trial
+        ):
+            return f"{organization.statement_descriptor} TRIAL OVER"
+        return organization.statement_descriptor
```
Yopi: I am guessing that we have a max limit here, is it a hard 22 characters?

frankie567: Yes, that's the limit imposed by Stripe (and the payment request fails if we overflow it). So we should take that into account. Maybe `organization.statement_descriptor` can become a method taking  a suffix in parameter.

server/polar/order/service.py
```diff
@@ -314,6 +318,16 @@ def _is_empty_customer_address(customer_address: dict[str, Any] | None) -> bool:
 
 
 class OrderService:
+    def get_statement_descriptor_suffix(
```
frankie567: Could it be a method on the Order model?

server/polar/order/schemas.py
```diff
@@ -68,6 +79,14 @@ class OrderBase(TimestampedSchema, IDSchema):
     )
     billing_address: Address | None
 
+    @field_serializer("billing_reason")
+    def serialize_billing_reason(
```
frankie567, line 83: Nice! Could we add a simple test for that? A simple schema test in `tests/order/test_schemas.py` can do.

Yopi: Absolutely!

server/polar/order/service.py
Yopi: Went a bit back and forth on making it a method or adding a separate one, but in the end I also think it's nicer that there is just the one that also takes a suffix.

server/tests/order/test_schemas.py
```diff
@@ -0,0 +1,79 @@
...
+        list(BILLING_REASON_SERIALIZATION_MAP.items()),
+    )
+    def test_billing_reason_serialization(
+        self,
+        internal_reason: OrderBillingReasonInternal,
+        expected_output: OrderBillingReason,
+    ) -> None:
+        """Test that billing reasons are serialized correctly."""
+        # Use model_construct to bypass validation
+        test_uuid = UUID("12345678-1234-5678-1234-567812345678")
```
frankie567: Another approach could be to create an Order fixture, validate it as the Pydantic schema and then test the output:

```py
order = await create_order(
        save_fixture,
        product=product,
        customer=customer,
        billing_reason=internal_reason,
    )

order_schema = Order.model_validate(order)
serialized = order.model_dump(mode="json")
assert serialized["billing_reason"] == expected_output
```

Yopi: Makes sense! I wasn't sure if that was idiomatic when doing the schema tests, but I like it a lot more

frankie567:
Will add you to the Vercel team so the build can trigger automatically 😄

EDIT: you're already there :D Maybe you need to link your GitHub account then :)

Yopi:
> Will add you to the Vercel team so the build can trigger automatically 😄
> 
> EDIT: you're already there :D Maybe you need to link your GitHub account then :)

Ah! Is that what I need to do. Done now 😄

PR #7366, opened 2025-10-21
Prefill checkout link query parameters server-side

pieterbeulque:


server/polar/checkout/service.py
```diff
@@ -768,13 +772,43 @@ async def checkout_link_create(
...
+                except PolarRequestValidationError:
+                    pass
+
+            custom_field_data_value = query_prefill.get("custom_field_data")
+            if custom_field_data_value is not None and isinstance(
+                custom_field_data_value, dict
+            ):
+                checkout.custom_field_data = {
+                    **(checkout.custom_field_data or {}),
+                    **custom_field_data_value,
```
pieterbeulque: Should this be filtered for existing custom fields for the configured keys?

server/polar/checkout/service.py
```diff
@@ -723,8 +724,11 @@ async def checkout_link_create(
             currency = price.price_currency
         elif is_custom_price(price):
             currency = price.price_currency
+            query_amount = query_prefill.get("amount") if query_prefill else None
             amount = (
-                price.preset_amount
+                int(query_amount) * 100
```
pieterbeulque: `float`? I guess you should be able to pass `amount=19.99` or would we require cents? It's not documented right now, but it also doesn't work right now 😁 so not sure if switching formats here would be considered a breaking change.

Yopi: Good question, it feels like the most similar way to how it's handled in the input field would be `int(float(query_amount) * 100)`, and allowing the user to specify 19.99 like you say

frankie567: The API expects cents (int) everywhere, so I would recommend to keep the same behaviour for the query params.

pieterbeulque: Yes, I agree (will also more gracefully work with `JPY` in the future).

Also, we should probably validate the price against `price.minimum_amount` or would that be handled elsewhere?

server/polar/checkout_link/endpoints.py
```diff
@@ -180,24 +186,49 @@ async def redirect(
         raise ResourceNotFound()
 
     ip_address = request.client.host if request.client else None
+
+    # Build query_prefill dictionary from explicit parameters
+    query_prefill: dict[str, str | dict[str, str] | None] = {
```
pieterbeulque: Let's put `product_id` in these as well then for consistency.

server/polar/checkout/service.py
frankie567: Probably, but what should we do if it's invalid? Just ignore it? It would be a bit brutal to fail the redirection 😅

Yopi: I'm thinking that we just fall back to the other prices, and we just don't allow it to be prefilled in that case. Since we already have a bunch of fall back prices

pieterbeulque: Invalid query parameters should be ignored, yes (i.e. act the same as if they were not set)

Yopi: Handled price validation and fall through now, and made it be cents

server/polar/checkout/service.py
Yopi: I'm honestly not sure what the repercussions would be. Right now I think they would get stored, but not really used, so it shouldn't matter. But if we validate them, we could also refrain from removing them to make it clear that it's not being used? If that would help with users debugging their setup.

pieterbeulque: Should I read that as "if it wasn't stripped, it wasn't parsed correctly so you're passing something wrong"?

Yopi: Haha, yes. Sorry for the unclear wording.

If we validate it, we can do ☝️

pieterbeulque: I think that'd be neat!

pieterbeulque:
I cleaned up the front-end but I'm slightly bummed this isn't a net-negative LOC diff 😁

Yopi:
> I cleaned up the front-end but I'm slightly bummed this isn't a net-negative LOC diff 😁

The backend tests do their part in pulling up the LOC count 😅

PR #7368, opened 2025-10-21
Polish seat management in customer portal

emilwidlund:


pieterbeulque:
I dropped Heroicons from the repo so you may have some merge conflicts here, sorry.

PR #7069, opened 2025-09-30
7010/Support more flexible subscription intervals

Yopi:
## 📋 Summary

**Related Issue**: Fixes #7010

This PR adds a recurring_interval_count in addition to the recurring_interval on the product and subscription models. Additionally it makes the cycle calculations for billing take the recurring_interval_count in consideration.

## 🎯 What

Backend:
* [X]  Add the necessary properties to the Product and Subscription models
* [X]  Create the database migration to add those fields and migrate existing Products and Subscriptions
* [X]  Expose and handle those properties correctly in the Product, Subscription and Checkout API
* [X] Handle the configuration correctly when handling a cycle.

Frontend
* [X]  Adapt the Product creation/update form to configure the recurring interval
    * [?]  Nice to have: show quick and common configurations by default like `1 month` or `1 year`[ ]  
* [x] Display the interval nicely across the dashboard (Product details, Subscription details…)
* [x] Display the interval nicely in the Customer Portal
* [ ]  Display the interval nicely in the Checkout form

## 🤔 Why

To allow for more flexibility in billing intervals.

## 🧪 Testing

- [X] I have tested these changes locally
- [X] All existing tests pass (`uv run task test` for backend, `pnpm test` for frontend)
- [X] I have added new tests for new functionality
- [X] I have run linting and type checking (`uv run task lint && uv run task lint_types` for backend)

### Test Instructions

1.
2.
3.

## 🖼️ Screenshots/Recordings
Screen Recording 2025-10-01 at 16 10 09

Screenshot 2025-10-01 at 16 47 06

Screenshot 2025-10-01 at 16 46 57

## ✅ Pre-submission Checklist

- [WIP] My code follows the project's style guidelines
- [X] I have performed a self-review of my code
- [X] I have commented my code where necessary
- [X I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have updated the relevant tests
- [X] All tests pass locally
- [X] **AI/LLM Policy**: If I used AI assistance, I have tested and executed the code locally (not just "vibe-coded")

PR #7410, opened 2025-10-22
Flexible subscription intervals (pt1)

### LINKED ISSUE #7010: Support more flexible subscription intervals
Now that our billing engine is enabled on every organizations, we can consider implementing more flexible features around billing cycles. In particular, we want to add more flexibility to **recurring intervals**. Currently, we only support **monthly** and **yearly**.

## Proposal

Similar to what we did for trials, we propose to configure the recurring interval using two properties:

* An interval: day, week, month, year
* An interval count: the number determining the length of the cycle. For example `2 month` would mean the subscription would be billed every two months, `14 days`, every 14 days, etc.

> \[!NOTE\]
>
> During our billing engine rollout, for testing purposes, we added rough support for daily and weekly intervals. Don't be surprised if you see them and if they don't show up properly in the UI. They are currently limited to Polar admins, but this limitation should of course be removed with the current effort.

## Backend tasks

- [X] Add the necessary properties to the Product and Subscription models
- [X] Create the database migration to add those fields and migrate existing Products and Subscriptions
- [X] Expose and handle those properties correctly in the Product, Subscription and Checkout API
- [X] Handle the configuration correctly when handling a cycle. Tip: `SubscriptionRecurringInterval.get_next_period`already has most of the logic necessary.

## Frontend tasks

- [ ] Adapt the Product creation/update form to configure the recurring interval
  - [ ] Nice to have: show quick and common configurations by default like `1 month` or `1 year`
- [ ] Display the interval nicely across the dashboard (Product details, Subscription details…)
- [ ] Display the interval nicely in the Customer Portal
- [ ] Display the interval nicely in the Checkout form
  * A bit more touchy since this requires to generate the public JS SDK ([https://github.com/polarsource/polar-js](https://github.com/polarsource/polar-js)) with the latest API changes. Feel free to reach out for guidance.

Yopi:
## 📋 Summary

**Related Issue**: Fixes #7010

This PR supersedes https://github.com/polarsource/polar/pull/7069:
Backend:
* [X]  Add the necessary properties to the Product and Subscription models
* [X]  Create the database migration to add those fields and migrate existing Products and Subscriptions
* [X]  Expose and handle those properties correctly in the Product, Subscription and Checkout API
* [X] Handle the configuration correctly when handling a cycle.

Frontend
* [ ]  Adapt the Product creation/update form to configure the recurring interval
    * [ ]  Nice to have: show quick and common configurations by default like `1 month` or `1 year`  
* [x] Display the interval nicely across the dashboard (Product details, Subscription details…)
* [x] Display the interval nicely in the Customer Portal

It adds all of configuration, formatting and API changes so that we can roll this out and enable it at a later point. This PR will enable daily and weekly intervals, but not yet allowing the user to specify an interval count.

## 🧪 Testing

### Test Instructions

1.
2.
3.

## ✅ Pre-submission Checklist

- [X] All tests pass locally

### DIFF
--- a/clients/apps/web/src/components/Subscriptions/SubscriptionDetails.tsx
+++ b/clients/apps/web/src/components/Subscriptions/SubscriptionDetails.tsx
@@ -7,6 +7,39 @@ import { formatCurrencyAndAmount } from '@polar-sh/ui/lib/money'
 import { DetailRow } from '../Shared/DetailRow'
 import { SubscriptionStatus } from './SubscriptionStatus'
 
+const formatRecurringSchedule = (
+  interval: schemas['SubscriptionRecurringInterval'],
+  intervalCount: number | null,
+): string => {
+  const count = intervalCount && intervalCount > 1 ? intervalCount : null
+
+  if (count) {
+    const intervalLabel =
+      interval === 'day'
+        ? 'Day'
+        : interval === 'week'
+          ? 'Week'
+          : interval === 'month'
+            ? 'Month'
+            : 'Year'
+    const pluralLabel = `${intervalLabel}${count > 1 ? 's' : ''}`
+    return `Every ${count} ${pluralLabel}`
+  }
+
+  switch (interval) {
+    case 'day':
+      return 'Daily'
+    case 'week':
+      return 'Weekly'
+    case 'month':
+      return 'Monthly'
+    case 'year':
+      return 'Yearly'
+    default:
+      return interval
+  }
+}
+
 const CANCELLATION_REASONS: {
   [key: string]: string
 } = {
@@ -85,8 +118,11 @@ const SubscriptionDetails = ({ subscription }: SubscriptionDetailsProps) => {
         )}
 
         <DetailRow
-          label="Recurring Interval"
-          value={subscription.recurring_interval === 'month' ? 'Month' : 'Year'}
+          label="Billing Schedule"
+          value={formatRecurringSchedule(
+            subscription.recurring_interval,
+            subscription.recurring_interval_count,
+          )}
         />
 
         <DetailRow


server/migrations/versions/2025-09-29-1146_add_recurring_interval_count_to_product_and_subscription.py
```diff
@@ -0,0 +1,51 @@
...
+        "products",
+        sa.Column("recurring_interval_count", sa.Integer(), nullable=True),
+    )
+    op.add_column(
+        "subscriptions",
+        sa.Column(
+            "recurring_interval_count",
+            sa.Integer(),
+            nullable=False,
+            server_default="1",
```
frankie567: I think we should remove the `server_default` and let us handle that on the Python side.

To make it pass during the migration, first create the column as nullable, execute the `UPDATE` below and then alter the column to make it non-nullable.

server/polar/models/subscription.py
```diff
@@ -108,6 +108,9 @@ class Subscription(CustomFieldDataMixin, MetadataMixin, RecordModel):
     recurring_interval: Mapped[SubscriptionRecurringInterval] = mapped_column(
         StringEnum(SubscriptionRecurringInterval), nullable=False, index=True
     )
+    recurring_interval_count: Mapped[int] = mapped_column(
+        Integer, nullable=False, default=0
```
frankie567: I don't think we need a default here

server/polar/subscription/schemas.py
```diff
@@ -62,6 +62,13 @@ class SubscriptionBase(IDSchema, TimestampedSchema):
         description="The interval at which the subscription recurs.",
         examples=["month"],
     )
+    recurring_interval_count: int | None = Field(
```
frankie567: Type annotation should be `int` since it's not nullable.

server/polar/subscription/service.py
```diff
@@ -560,9 +565,15 @@ async def create_or_update_from_checkout_stripe(
 
         if product.is_legacy_recurring_price:
             subscription.recurring_interval = prices[0].recurring_interval
+            subscription.recurring_interval_count = 0
```
frankie567: I think the value should be `1` in this case.

server/polar/subscription/service.py
```diff
@@ -560,9 +565,15 @@ async def create_or_update_from_checkout_stripe(
...
         else:
             assert product.recurring_interval is not None
             subscription.recurring_interval = product.recurring_interval
+            subscription.recurring_interval_count = (
```
frankie567: I think we should do like above:

```py
assert product.recurring_interval_count is not None
subscription.recurring_interval_count = product.recurring_interval_count
```

server/polar/subscription/service.py
```diff
@@ -946,7 +959,13 @@ async def update_product(
             SubscriptionProductPrice.from_price(price, seats=subscription.seats)
             for price in prices
         ]
+        assert product.recurring_interval is not None
         subscription.recurring_interval = product.recurring_interval
+        subscription.recurring_interval_count = (
```
frankie567: I think we should do like above:

```py
assert product.recurring_interval_count is not None
subscription.recurring_interval_count = product.recurring_interval_count
```

server/tests/fixtures/random_objects.py
```diff
@@ -940,16 +945,22 @@ async def create_subscription(
     prices = prices or product.prices
 
     recurring_interval = product.recurring_interval
+    recurring_interval_count = product.recurring_interval_count
     if product.is_legacy_recurring_price:
         recurring_interval = product.prices[0].recurring_interval
+        recurring_interval_count = 0
```
frankie567: I think the value should be `1`

server/polar/subscription/service.py
Yopi: My idea was that this would allow us to roll out the frontend/checkout changes independent on these changes. If we assert it but don't pass it in from the checkout I think it will fail, and then we are time pressured for the follow up change.

server/polar/subscription/service.py
Yopi: I agree. I made it like this in the case we wanted to differentiate/notice the legacy recurring price in the interval count.

server/migrations/versions/2025-09-29-1146_add_recurring_interval_count_to_product_and_subscription.py
Yopi: I agree with you if we want to make it non null. You don't think it's a problem that recurring_interval is nullable but recurring_interval_count is not? E.g. what does it mean semantically if you have recurring_interval none and recurring_interval 1

server/polar/subscription/service.py
Yopi: Now that I think about it, you're right if we make it non nullable this won't be a problem

Yopi:
@frankie567 mind taking another look?

PR #7418, opened 2025-10-22
Bar charts

pieterbeulque:
@emilwidlund - This is a hill I'm willing to die on, though I can agree bar charts are a bit less aesthetically pleasing.

Line graphs are used for continuous data (e.g. your MRR, ACV, … fluctuating over time).
Bar graphs are used for data that resets every interval (e.g. revenue: yesterday you make $40, today you make $20, …)

This switches to bar charts for the relevant metrics on analytics and for all meters.

Thoughts?

image


frankie567:
It's indeed way less aesthetically pleasing 😅

I get the argument about continuous vs discrete data; even though I had to think about it a minute before understanding. Do you think a user arriving on the analytics page will have this mindset out-of-the-box? I suspect they might just think "WTF, Polar, why 1/2 charts are ugly?"

pieterbeulque:
It's a battle I fought and lost at Baremetrics too 😁

emilwidlund:
Alright, I'm good with this. However, can you set the `maxWidth` on the `<Bar />` to `10`? Not a fan of these fat ones, hehe.

pieterbeulque:
Merging it in but keeping the dashboard to line charts only. I'm using the bar chart for count aggregated meters only, for now. We'll see where this goes.

PR #7460, opened 2025-10-24
Organization and customer based invoice numbers

### LINKED ISSUE #6994: Randomize the invoice order number
> From [the ticket](https://app.plain.com/workspace/w_01JE9TRRX9KT61D8P2CH77XDQM/thread/th_01K5YRH07DKN37GHPDWJPS14VG/)

Currently, invoices display ORG-X where X is the number of the sales made till that purchase. This basically tells the end-customer the number of sales made by the org which the merchant is not comfortable sharing.

Yopi:
## 📋 Summary

**Related Issue**: Fixes #6994

Introduces an organization setting to switch between customer based invoice numbers (each customer increments from 1 independently from each other) or organization wide invoice numbers (each order increments the same counter).

Organization wide is what we have today. Customer based will no longer make it possible to understand volume information if that is sensitive.

## 🎯 What

* Introduce invoice_next_number on `customer`
* Introduce `invoice_numbering` on `organization` subscription settings
* Use the configured invoice next number from either organization or customer when making an order.

## 🧪 Testing

### Test Instructions

1.
2.
3.

## ✅ Pre-submission Checklist

- [ ] All tests pass locally


server/polar/organization/service.py
```diff
@@ -361,17 +365,34 @@ async def get_next_invoice_number(
...
-            },
-        )
-        return invoice_number
+        from polar.customer.repository import CustomerRepository
+
+        match organization.invoice_numbering:
+            case InvoiceNumbering.customer:
+                customer_suffix = str(customer.id).split("-")[0].upper()
+                invoice_number = f"{organization.customer_invoice_prefix}-{customer_suffix}-{customer.invoice_next_number:04d}"
+                customer_repository = CustomerRepository.from_session(session)
```
Yopi: This will get fairly long. Should we generate a shorter ID to store on customers as well?

server/migrations/functions.py
```diff
@@ -0,0 +1,29 @@
+# PostgreSQL functions that are applied, that needs to be both migrated and recreated
+# in our tests.
+
+SEQUENCE_CREATE_CUSTOMER_SHORT_ID = "CREATE SEQUENCE customer_short_id_seq START 1;"
+
+FUNCTION_GENERATE_CUSTOMER_SHORT_ID = """
+CREATE OR REPLACE FUNCTION generate_customer_short_id()
+RETURNS bigint AS $$
+DECLARE
```
Yopi, line 10: This is based on the Instagram ID generation algorithm. We get fairly small, ordered but not leaking user volume, IDs that we can base26-encode to get an alphabetical identifier we can use for invoice numbers.

server/migrations/functions.py
```diff
@@ -0,0 +1,29 @@
+# PostgreSQL functions that are applied, that needs to be both migrated and recreated
```
Yopi, line 1: I wasn't entirely sure how to play this. I wanted to do the ID generation in postgres to avoid having to retry saving customers if we got unlucky in the random ID generation. This is not really following any existing conventions. I put them here so that we can load them into the test runner as well, and it becomes fairly clear for the future how that works.

server/migrations/functions.py
Yopi: The Instagram ID generation algorithm also has a portion of the ID as a machine shard. We could put that in from the start, as it will become a bit annoying to do a migration in the future, if we believe we will need it.

server/polar/models/organization.py
```diff
@@ -63,12 +63,14 @@ class OrganizationSubscriptionSettings(TypedDict):
...
     allow_customer_updates: bool
     proration_behavior: SubscriptionProrationBehavior
+    invoice_numbering: InvoiceNumbering
 
 
 _default_subscription_settings: OrganizationSubscriptionSettings = {
     "allow_multiple_subscriptions": False,
     "allow_customer_updates": True,
     "proration_behavior": SubscriptionProrationBehavior.prorate,
+    "invoice_numbering": InvoiceNumbering.organization,
```
frankie567, line 82: Semantically speaking, it's a bit unexpected to see that under `subscription_settings`, since it affects all kind of orders, not only subscriptions.

server/polar/organization/service.py
```diff
@@ -361,17 +365,33 @@ async def get_next_invoice_number(
...
-        repository = OrganizationRepository.from_session(session)
-        organization = await repository.update(
-            organization,
-            update_dict={
-                "customer_invoice_next_number": organization.customer_invoice_next_number
-                + 1
-            },
-        )
-        return invoice_number
+        from polar.customer.repository import CustomerRepository
```
frankie567: We should import that at the top

server/polar/organization/service.py
```diff
@@ -361,17 +365,33 @@ async def get_next_invoice_number(
...
+
+        match organization.invoice_numbering:
+            case InvoiceNumbering.customer:
+                invoice_number = f"{organization.customer_invoice_prefix}-{customer.short_id_str}-{customer.invoice_next_number:04d}"
+                customer_repository = CustomerRepository.from_session(session)
+                customer.invoice_next_number = customer.invoice_next_number + 1
+                await customer_repository.update(
+                    customer,
+                    update_dict={"invoice_next_number": customer.invoice_next_number},
+                    flush=True,
```
frankie567: We don't necessarily need to flush here

server/tests/organization/test_service.py
```diff
@@ -163,19 +163,126 @@ async def test_valid_with_none_subscription_settings(
...
+
+    next_invoice_number = await organization_service.get_next_invoice_number(
+        session, organization, customer
+    )
+
+    assert (
+        next_invoice_number
+        == f"{organization.customer_invoice_prefix}-{customer.short_id_str}-0001"
+    )
+    await session.commit()
```
frankie567, line 212: Not sure it's necessary to commit/refresh everywhere.

Besides, I would _highly_ recommend to prefer `session.flush()` over `session.commit()` in most cases :D

clients/apps/web/src/components/Settings/OrganizationSubscriptionSettings.tsx
```diff
@@ -127,6 +128,28 @@ const OrganizationSubscriptionSettings: React.FC<
               )}
             />
           </SettingsGroupItem>
+
+          <SettingsGroupItem
```
frankie567, line 132: I don't think we should let users choose this. Can we consider that existing merchants keep the current behaviour and new merchants get the customer-sequence?

server/migrations/functions.py
frankie567: I was a bit reluctant to have a Postgres function at first but... Can't deny it, that's pretty neat and efficient 😄 I like it!

server/migrations/functions.py
frankie567: What would that look like in our current setup? Just an arbitrary fixed ID?

Yopi:
> Main concern for me is the public option setting; I think this should be kept internal.

Do you think we should keep it optional (e.g. keep it as it is for current customers, default to the new way) but just not make it an exposed option? Or do you think we should migrate everyone over to the new approach and commit to it?

server/migrations/functions.py
Yopi: Exactly, we'd set it to 0 or 1 for now, and then a component of the id would just be fixed. It would allow us to introduce it at some point without migrating the IDs or having two realities.

server/polar/models/organization.py
Yopi: Haha, you got me. I was being a bit lazy, and this is where I thought it made the _most_ sense. Fully agree with you that it doesn't entirely fit here.

server/polar/organization/service.py
Yopi: I was struggling a little bit with ensuring that the customer invoice next number was updated in the subsequent calls in the test. I can remove the flush here, but will need to keep the session commit in the test (or vice-versa), but that probably makes the most sense

server/migrations/functions.py
frankie567: Ok, let's add it then 👍

frankie567:
> > Main concern for me is the public option setting; I think this should be kept internal.
> 
> Do you think we should keep it optional (e.g. keep it as it is for current customers, default to the new way) but just not make it an exposed option? Or do you think we should migrate everyone over to the new approach and commit to it?

I think existing users should stay on the current approach; and only switch them manually if they specifically ask for it. New users will get that new approach.

server/migrations/functions.py
Yopi: I was thinking this over (implications on the size of the ID and size of the base26-encoded), and thought maybe it would make sense to go in the opposite direction. Switching over to seconds instead of milliesconds, and not adding the shard gives us a capacity of 1024 customers/second, and alphabetical IDs that are ~8 characters.

Keeping milliseconds and adding shard IDs we would be on ~11 characters. Not sure where we should be aiming right now, but it does feel a bit overkill.

clients/apps/web/src/components/Settings/OrganizationSubscriptionSettings.tsx
Yopi: Removed the user settings!

server/migrations/functions.py
frankie567: Your call :)

frankie567 (approved):

server/polar/models/organization.py
```diff
@@ -74,6 +74,15 @@ class OrganizationSubscriptionSettings(TypedDict):
 }
 
 
+class OrganizationOrderSettings(TypedDict):
+    invoice_numbering: InvoiceNumbering
+
+
+_default_order_settings: OrganizationOrderSettings = {
+    "invoice_numbering": InvoiceNumbering.organization,
```
Yopi, line 82: Leaving this on organization for now, so we can roll this out and change the defaults after this PR.

server/migrations/functions.py
Yopi: I increased it to millisecond precision again, in the event of a spike I don't want to be guilty of it not working. I'll merge this now!

PR #7486, opened 2025-10-25
Events: Implement Customer System Events

emilwidlund:


server/polar/customer/service.py
```diff
@@ -152,6 +155,21 @@ async def create(
                 ),
             )
         ) as customer:
+            await event_service.create_event(
```
frankie567: Since we may implicitly create customers in various places (e.g. Checkout), it's more practical to put that in the `create_context` method of `CustomerRepository`. We already have logic like enqueuing webhooks there.

server/polar/customer/service.py
```diff
@@ -218,17 +236,86 @@ async def update(
...
             update_dict=customer_update.model_dump(exclude_unset=True, by_alias=True),
         )
 
+        organization_repository = OrganizationRepository.from_session(session)
+        organization = await organization_repository.get_by_id(
+            updated_customer.organization_id
+        )
+        assert organization is not None
+
+        await event_service.create_event(
```
frankie567: Same thing, should be done in `CustomerRepository.update`.

server/polar/customer/service.py
```diff
@@ -218,17 +236,86 @@ async def update(
...
+            session,
+            build_system_event(
+                SystemEvent.customer_updated,
+                customer=updated_customer,
+                organization=organization,
+                metadata={
+                    "customer_id": str(updated_customer.id),
+                    "updated_fields": {
```
frankie567: Do we really need to store the updated fields? Feels like a darling.

server/polar/customer/service.py
```diff
@@ -218,17 +236,86 @@ async def update(
...
-        return await repository.soft_delete(customer)
+        soft_deleted_customer = await repository.soft_delete(customer)
+
+        organization_repository = OrganizationRepository.from_session(session)
+        organization = await organization_repository.get_by_id(
+            soft_deleted_customer.organization_id
+        )
+        assert organization is not None
+
+        await event_service.create_event(
```
frankie567: Same thing, should be done in `CustomerRepository.soft_delete`.

server/polar/event/system.py
```diff
@@ -80,6 +83,54 @@ class BenefitRevokedEvent(Event):
         user_metadata: Mapped[BenefitGrantMetadata]  # type: ignore[assignment]
 
 
+class CustomerCreatedMetadata(TypedDict):
+    customer_id: str
+    customer_email: str
+    customer_name: NotRequired[str | None]
+    customer_external_id: NotRequired[str | None]
```
frankie567: Type for `customer_name`/`customer_external_id` should be `str | None`: the key will always be there, possibly with a `None` value, but it'll be there.

server/polar/event/system.py
```diff
@@ -80,6 +83,54 @@ class BenefitRevokedEvent(Event):
...
+class CustomerUpdatedFields(TypedDict):
+    email: NotRequired[str | None]
+    name: str | None
+    billing_address: NotRequired[dict[str, str] | None]
+    tax_id: NotRequired[str | None]
+    metadata: NotRequired[dict[str, str | int | bool] | None]
+
+
+class CustomerUpdatedMetadata(TypedDict):
+    customer_id: str
```
frankie567, line 112: For consistency with created and deleted, we should probably have name, email and external ID as well.

frankie567 (changes requested):
A few tweaks to make to make sure it's working properly + concern about the updated_fields data.

server/polar/customer/service.py
emilwidlund: It makes sense for an Update Event. It wouldn't be of much help if you just could see that it happened, but not what data was added to it imo.

server/polar/customer/service.py
emilwidlund: Fixed! 👍

server/polar/customer/service.py
emilwidlund: Fixed! 👍

server/polar/customer/service.py
emilwidlund: Fixed! 👍

server/polar/event/system.py
emilwidlund: yup, agree!

emilwidlund:
@frankie567 Mind giving this another look? Had to do inline imports to handle some circular import issues.

frankie567:
I'll revisit!

frankie567:
@emilwidlund Did some changes to the backend to make it more robust. All good to me, you can merge it whenever you want 👍

emilwidlund (commented):
Great work! Thanks!

PR #7490, opened 2025-10-26
Guide for Theme switch in Checkout

### LINKED ISSUE #7100: docs: document how to open a dark themed checkout
$40

We need a new guide on "Switching themes in the checkout"

section 1. In this we need to show how a checkout session can be created using the API and then appending the theme query param as dark or light would allow to change the theme of the genearted checkout session so that the end user sees it in that mode

section 2. Similarly, we need to show that in checkout links the theme can be adjusted by using the parameter ?theme=dark or light (no need to show an API call here)

having images would be helpful

@shubhi2000 do you want to attempt this ahead of other issues?

shubhi2000:
fix #7100 

Learn how to switch theme for a checkout session created using the API as well as for checkout link.

### DIFF
--- a/docs/guides/theme-switch-in-checkout.mdx
+++ b/docs/guides/theme-switch-in-checkout.mdx
@@ -1,6 +1,6 @@
 ---
-title: "How to Switch Theme in Checkout"
-sidebarTitle: "Switch theme in checkout"
+title: "How to Switch Checkout Theme"
+sidebarTitle: "Switch Checkout Theme"
 description: "Learn how to switch the theme in a checkout session and a checkout link."
 ---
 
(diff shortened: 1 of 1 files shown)


shubhi2000:
Hi @rishi-raj-jain 
Could you review my PR?

I have created three sections, section 1 describes how to create a checkout session using API. 
Section 2 describes how to add theme query param to checkout session to switch the theme. 
Section 3 describes how to add theme query param to checkout link to switch the theme.

Is it inconsistent to show how to create a checkout session step-by-step but not show the steps of creating a checkout link and just referring to a link to follow for that?

Thanks

PR #7491, opened 2025-10-27
Events: Implement Order System Events

emilwidlund:
Implements System events for
- Order Paid
- Order Refunded

### DIFF
--- a/server/polar/event/system.py
+++ b/server/polar/event/system.py
@@ -146,7 +146,7 @@ class OrderPaidEvent(Event):
 
 class OrderRefundedMetadata(TypedDict):
     order_id: str
-    amount: int
+    refunded_amount: int
     currency: str
 
 
--- a/clients/apps/web/src/components/Events/utils.tsx
+++ b/clients/apps/web/src/components/Events/utils.tsx
@@ -59,6 +59,15 @@ export const useEventCard = (event: schemas['Event']) => {
   }, [event])
 }
 
+export const isOrderEvent = (
+  event: schemas['Event'],
+): event is schemas['OrderPaidEvent'] | schemas['OrderRefundedEvent'] => {
+  return (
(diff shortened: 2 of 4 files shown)


server/polar/event/system.py
```diff
@@ -129,6 +131,32 @@ class SubscriptionSeatsUpdatedEvent(Event):
...
+class OrderPaidEvent(Event):
+    if TYPE_CHECKING:
+        source: Mapped[Literal[EventSource.system]]
+        name: Mapped[Literal[SystemEvent.order_paid]]
+        user_metadata: Mapped[OrderPaidMetadata]  # type: ignore[assignment]
+
+
+class OrderRefundedMetadata(TypedDict):
+    order_id: str
+    amount: int
```
frankie567: Maybe call it `refunded_amount` to disambiguate with the actual order amount?

emilwidlund: Makes sense! Done :)

server/polar/event/schemas.py
```diff
@@ -286,6 +288,24 @@ class SubscriptionProductUpdatedEvent(SystemEventBase):
...
+class OrderPaidEvent(SystemEventBase):
+    """An event created by Polar when an order is paid."""
+
+    name: Literal[SystemEventEnum.order_paid] = Field(description=_NAME_DESCRIPTION)
+    metadata: OrderPaidMetadata = Field(
+        validation_alias=AliasChoices("user_metadata", "metadata")
+    )
+
+
+class OrderRefundedEvent(SystemEventBase):
```
Yopi, line 300: @emilwidlund would it make sense to track whether this refund event was partial or fully refunded? As it is now we mark all events as `refunded` even if a user makes multiple refunds. Or is that uninteresting?

PR #7516, opened 2025-10-27
feat(order): Add task to backfill order events

Yopi:
Backfills the events added in https://github.com/polarsource/polar/pull/7491

### DIFF
--- a/server/polar/event/system.py
+++ b/server/polar/event/system.py
@@ -1,5 +1,5 @@
 from enum import StrEnum
-from typing import TYPE_CHECKING, Any, Literal, overload
+from typing import TYPE_CHECKING, Any, Literal, NotRequired, overload
 
 from sqlalchemy.orm import Mapped
 from sqlalchemy.util.typing import TypedDict
@@ -135,6 +135,7 @@ class OrderPaidMetadata(TypedDict):
     order_id: str
     amount: int
     currency: str
+    backfilled: NotRequired[bool]
 
 
 class OrderPaidEvent(Event):
@@ -148,6 +149,7 @@ class OrderRefundedMetadata(TypedDict):
     order_id: str
     refunded_amount: int
     currency: str
+    backfilled: NotRequired[bool]
 
 
 class OrderRefundedEvent(Event):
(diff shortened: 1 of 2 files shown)


emilwidlund:
Looking good to me. Maybe @frankie567 wants to take a quick look.

server/polar/event/system.py
```diff
@@ -135,6 +135,7 @@ class OrderPaidMetadata(TypedDict):
     order_id: str
     amount: int
     currency: str
+    backfilled: NotRequired[bool]
```
frankie567, line 138: Do we really need that info? + It'll be a bit weird for the user to see that

server/polar/order/tasks.py
```diff
@@ -223,3 +232,147 @@ async def process_dunning_order(order_id: uuid.UUID) -> None:
...
+
+async def _build_events_for_orders(
+    session: AsyncSession, orders: list[Order]
+) -> list[dict[str, Any]]:
+    """Build events for a batch of orders using transaction timestamps."""
+    order_ids = [order.id for order in orders]
+    orders_by_id = {order.id: order for order in orders}
+
+    transactions_result = await session.execute(
+        select(Transaction)
```
frankie567, line 318: A bit surprising to see Transaction involved here (Order.created_at / Refund.created_at could have worked). But that's fine :)

Yopi: Haha, I was scratching my head on this. But maybe a classic case of going across the river for the water

server/polar/event/system.py
Yopi: I couldn't see that all of this was actually exposed to the user other than on the API, but maybe that's what you mean.

The main thing that I got a bit worried about was our possibilities of re-running this if something went wrong, which is why I added it. But I can remove it again if we think it's too much and unnecessary.

emilwidlund (approved):

server/polar/event/system.py
frankie567: In the API, but also in the dashboard, where you can query the Events and see their data.

I see what you mean. Maybe we can consider a migration afterwards to remove the key once everything is done.

Yopi: Yeah I'd like that. It's probably not _too wild_ for a user to see this if they notice it during the time we have it, given that the events will have popped up after the fact

PR #7521, opened 2025-10-27
Respect soft-deletes on customer meters

pieterbeulque:
- Exclude archived meters from Customer Detail → Usage tab
- Exclude archived meters on the customer portal

### DIFF
--- a/server/polar/customer_meter/service.py
+++ b/server/polar/customer_meter/service.py
@@ -88,10 +88,7 @@ class CustomerMeterService:
         repository = MeterRepository.from_session(session)
         statement = (
             repository.get_base_statement()
-            .where(
-                Meter.organization_id == customer.organization_id,
-                Meter.archived_at.is_(None),
-            )
+            .where(Meter.organization_id == customer.organization_id)
             .order_by(Meter.created_at.asc())
         )
 


server/polar/customer_meter/service.py
pieterbeulque, line 59: Probably we shouldn't set `Meter.archived_at.is_(None)` if `meter_id is not None` so you can access a customer meter for an archived meter, but I'm not 100% if that's needed.

server/polar/customer_meter/service.py
```diff
@@ -85,7 +88,10 @@ async def update_customer(
         repository = MeterRepository.from_session(session)
         statement = (
             repository.get_base_statement()
-            .where(Meter.organization_id == customer.organization_id)
+            .where(
+                Meter.organization_id == customer.organization_id,
+                Meter.archived_at.is_(None),
```
pieterbeulque: @frankie567 - Do you think it'd be a breaking change to stop updating archived meters? You can still view them & their usage so in that sense you'd expect the numbers to be accurate…

server/polar/customer_meter/service.py
Yopi: I think it makes sense, because going in via Usage Billing -> Meters you have the filter for archived / non-archived, where it would list the customers that has billing for the different meters.

emilwidlund:
Does it make sense to add an "archived" boolean option to the list-API instead? That's something we do for products API for instance. We don't return them unless explicitly asked for via params.

pieterbeulque:
@emilwidlund Yeah, that already exists on the meters itself

PR #7530, opened 2025-10-28
Add grace period for benefit revocation

Yopi:
Adds a configuration to subscription settings for an organization whether to revoke benefits immediately (like the current behaviour) or to give a grace period similar to our payment retry functionality.

This only applies to failed payments, not for cancellation or natural ends of subscriptions.

This applies to customer seat benefits as well as regular benefits.

### DIFF
--- a/server/polar/subscription/service.py
+++ b/server/polar/subscription/service.py
@@ -8,7 +8,7 @@ from urllib.parse import urlencode
 
 import stripe as stripe_lib
 import structlog
-from sqlalchemy import desc, select
+from sqlalchemy import select
 from sqlalchemy.orm import contains_eager, selectinload
 
 from polar.auth.models import AuthSubject
@@ -54,7 +54,6 @@ from polar.models import (
     Customer,
     Discount,
     Event,
-    Order,
     Organization,
     Payment,
     PaymentMethod,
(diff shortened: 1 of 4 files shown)


server/polar/models/organization.py
```diff
@@ -63,12 +63,14 @@ class OrganizationSubscriptionSettings(TypedDict):
     allow_multiple_subscriptions: bool
     allow_customer_updates: bool
     proration_behavior: SubscriptionProrationBehavior
+    benefit_revocation_grace_period: BenefitRevocationGracePeriod
```
frankie567: Could we have a simple `int` instead of an enum? Feels a bit more simple; even if we keep predefined values in the UI.

server/polar/subscription/service.py
```diff
@@ -2202,6 +2203,65 @@ async def _send_webhook(
...
+            return False
+
+        grace_period_days = int(organization.benefit_revocation_grace_period)
+
+        # If grace period is 0 (immediate), no grace period applies
+        if grace_period_days == 0:
+            return False
+
+        # Get the latest pending order for this subscription
+        statement = (
```
frankie567: What would you think of having a `past_due_at` timestamp on Subscription which would tell us when the subscription has gone past due?

frankie567 (commented):
Great! Have two suggestions on the approach, but feel free to discard them if you think it's not relevant :)

server/polar/subscription/service.py
Yopi: Haha, I was thinking the same, but didn't feel comfortable bloating the models more. I think it makes sense. I'll add it!

server/polar/models/organization.py
Yopi: Sure!

PR #7543, opened 2025-10-28
Update OrganizationProfileSettings.tsx

rishi-raj-jain:


rishi-raj-jain:
Screenshot 2025-10-29 at 12 04 21 PM

pieterbeulque:
You can make the linter pass by running `pnpm format` on your machine.

I'd use a `<a>` instead of a `<Link>` since it's an external link (we don't want Next.js prefetching)

The subtext isn't super clear to me, it's not phrased from an end user perspective. If the links aren't visible, why would I add them? Why would I use the Organizations API to read my own social media links from here?

What do you want to communicate? Is it maybe something like:

> Social media links help with your account verification. They will not be shown publicly.

rishi-raj-jain:
> The subtext isn't super clear to me, it's not phrased from an end user perspective. If the links aren't visible, why would I add them? Why would I use the Organizations API to read my own social media links from here?

These are required with account reviews, since we've started to enforce the requirement of sharing your social media with us so that we can identify and associate with you as a merchant on Polar.

rishi-raj-jain:
> Social media links help with your account verification. They will not be shown publicly.

This is so well written! Creating a PR.

rishi-raj-jain:
Created https://github.com/polarsource/polar/pull/7550.

PR #7555, opened 2025-10-29
Show upcoming charge on subscription detail

pieterbeulque:
**Trial**
image

**Active subscription**
image


server/polar/subscription/service.py
```diff
@@ -2030,6 +2042,77 @@ async def update_meters(
...
+                if subscription.customer.tax_id is not None
+                else [],
+                subscription.tax_exempted,
+            )
+
+            tax_amount = tax["amount"]
+
+        total = taxable_amount + tax_amount
+
+        return {
```
frankie567: You could directly instantiate  and return `SubscriptionChargePreview` schema instead of passing through a TypedDict :D

pieterbeulque: That's way cleaner! Thanks.

PR #7562, opened 2025-10-29
fix MoneyInput bug when deleting last integer part

### LINKED ISSUE #7561: MoneyInput component jumps caret when trying to edit
### Description

MoneyInput component misbehaves when deleting the only digit before the decimal separator. It makes the caret jump to the end which makes it hard to continue inputting the amount you wanted.

 ### Current Behavior

Typing 8.99, clicking after the 8, and pressing backspace leaves 0.99 in the field and drops the caret at the end, so continuing to type overwrites the fractional part and leaves the whole number without a leading zero.

 ### Expected Behavior

Removing the leading digit should leave .99 in place while the user types and the caret should stay in place so user can continue typing.

### Screenshots

<details>
<summary>Video of current buggy behaviour</summary>

https://github.com/user-attachments/assets/a5d2e141-4469-4625-a1e0-5e96cabf14d4

</details>

### Environment:

-   Operating System: macOS
-   Browser (if applicable): Chrome

simonxabris:
Related Issue: Fixes #7561

  ## 🎯 What

 Ensure fractional-only entries (.99) no longer force a leading zero while typing, but reintroduce it on blur so the displayed amount remains a valid currency string.

  ## 🤔 Why

Removing the first digit before the decimal was inserting 0, which pushed the caret to the end and interrupted editing. Users should be able to keep typing in place, yet still see a normalized currency value after leaving the field.

  ## 🔧 How

Modified the `onChange` of `MoneyInput` so that after it calls `.toFixed` we check if it would insert a leading `0` and strip it if it does. This allows user to continue typing without interruptions.

  ## 🧪 Testing

  - [x] All existing tests pass (uv run task test for backend, pnpm test for frontend)
  - [x] I have run linting and type checking (uv run task lint && uv run task lint_types for backend)

  ### Test Instructions

  1. Open a screen that uses `MoneyInput`.
  2. Enter 8.99, click after 8, press backspace, and continue typing to confirm the caret stays in place.
  3. Leave the field to verify .99 normalizes to 0.99.

## 🖼️ Screenshots/Recordings

https://github.com/user-attachments/assets/0c4f39d2-ffd0-45cf-a989-72e5e83e619e

## ✅ Pre-submission Checklist

- [x] All tests pass locally

### DIFF
--- a/clients/packages/ui/src/components/atoms/MoneyInput.tsx
+++ b/clients/packages/ui/src/components/atoms/MoneyInput.tsx
@@ -152,10 +152,12 @@ const MoneyInput = (props: Props) => {
       if (internalValue) {
         let nextValue = internalValue
 
+        // Add 0 as integer part if value starts with `.`
         if (nextValue.startsWith('.')) {
           nextValue = `0${nextValue}`
         }
 
+        // Strip trailing decimal point
         if (nextValue.endsWith('.')) {
           nextValue = nextValue.replace(/\.$/, '')
         }


pieterbeulque:
Good catch, appreciate it!

PR #7582, opened 2025-10-30
Orders backfill use orders and refunds tables

Yopi:


Yopi:
If you want to have a look at this attempt @frankie567 🙏

server/polar/order/tasks.py
```diff
@@ -235,6 +235,58 @@ async def process_dunning_order(order_id: uuid.UUID) -> None:
         await order_service.process_dunning_order(session, order)
 
 
+@actor(
+    actor_name="order.remove_backfilled_events",
+    priority=TaskPriority.LOW,
+    time_limit=3600_000,
+)
+async def remove_backfilled_events(
+    batch_size: int = settings.DATABASE_STREAM_YIELD_PER, rate_limit_delay: float = 0.1
```
frankie567: I think we can afford a bigger batch size: 1000-2000 sounds reasonable

server/polar/order/tasks.py
```diff
@@ -235,6 +235,58 @@ async def process_dunning_order(order_id: uuid.UUID) -> None:
...
+                    Event.user_metadata["backfilled"].as_boolean().is_(True),
+                )
+                .limit(batch_size)
+            )
+            event_ids = [row[0] for row in result.all()]
+
+            if not event_ids:
+                break
+
+            await session.execute(delete(Event).where(Event.id.in_(event_ids)))
```
frankie567: We could probably perform that in one query by putting the "select" as a subquery:

```py
delete(Event).where(
    Event.id.in_(
        select(Event.id) # ...
    )
)
```

frankie567 (commented):
Suggested a few tweaks, but otherwise looks good to me 👍

server/polar/order/tasks.py
Yopi: Agree! When you run it in the interface you need to set a value either way, but can update it for the person checking the code to see what a sane default is.

server/polar/order/tasks.py
Yopi: Thanks!

PR #7597, opened 2025-10-30
server: bump dependencies

frankie567:
- server: bump dependencies

### DIFF
--- a/server/uv.lock
+++ b/server/uv.lock
@@ -8,16 +8,16 @@ resolution-markers = [
 
 [[package]]
 name = "alembic"
-version = "1.17.0"
+version = "1.17.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "mako" },
     { name = "sqlalchemy" },
     { name = "typing-extensions" },
 ]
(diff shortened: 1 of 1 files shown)


server/uv.lock
```diff
@@ -2410,14 +2436,14 @@ sdist = { url = "https://files.pythonhosted.org/packages/48/c8/4c9705c0499c8b5ac
 
 [[package]]
 name = "starlette"
-version = "0.48.0"
+version = "0.49.1"
```
Kludex, line 2439: Good job 😎

frankie567: Thank you for handling the CVE ❤️ Are you keeping an eye on all our commits? 🤣

Kludex: I don't really know how GitHub handles the feed, but I see your commits...

Screenshot 2025-10-31 at 10 46 38

I think it's because I have starred this repository.

---

But I'm everywhere... 😈

frankie567: Wow, given my activity here, I guess I'm flooding your feed; sorry about that 😅

PR #7600, opened 2025-10-31
Handle free subscriptions in analytics

pieterbeulque:
Free → paid subscription changes now count as 'new revenue', not 'renewed revenue'.

There's a similar error in _Renewed subscriptions_ (the count metric), either with trials cycling or with free → paid switches. 

I'm not sure if we should include free subscriptions in these metrics at all. It's more likely that you're interested in your paying customers. Or is this another case of "it depends", or, even more fun, "let's do both".

### DIFF
--- a/server/polar/metrics/metrics.py
+++ b/server/polar/metrics/metrics.py
@@ -23,6 +23,7 @@ from polar.kit.time_queries import TimeInterval
 from polar.models import Checkout, Order, Subscription
 from polar.models.checkout import CheckoutStatus
 from polar.models.event import Event
+from polar.models.order import OrderBillingReasonInternal
 from polar.models.subscription import CustomerCancellationReason
 
 from .queries import MetricQuery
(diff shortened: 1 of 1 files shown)


server/polar/metrics/metrics.py
```diff
@@ -295,10 +298,12 @@ def get_sql_expression(
         cls, t: ColumnElement[datetime], i: TimeInterval, now: datetime
     ) -> ColumnElement[int]:
         return func.sum(Order.payout_amount).filter(
-            i.sql_date_trunc(
-                cast(SQLColumnExpression[datetime], Subscription.started_at)
+            Order.billing_reason.in_(
+                [
+                    OrderBillingReasonInternal.subscription_create,
+                    OrderBillingReasonInternal.subscription_cycle_after_trial,
```
Yopi, line 304: Just my ignorance, but is `subscription_create` only for non-trial subscriptions?

pieterbeulque: Starting a trial doesn't create an order, if I recall correctly, so it's only on trial conversion that the order is created with one of these two billing reasons and that's when we do the charge.

frankie567: Starting a trial does create an Order, with the `subscription_create` billing reason :)

pieterbeulque: Another one bites the dust 🫣

pieterbeulque:
Logged in polarsource/feedback#96 instead

PR #7605, opened 2025-10-31
Bump API paginatation max limit to align with events limit

Yopi:
With this change https://github.com/polarsource/polar/commit/f03b0d6c404300520d4ec14db824261585165e36 if we have between 100 and 300 events the user is not seeing any pages and also not the events.

### DIFF
--- a/server/polar/config.py
+++ b/server/polar/config.py
@@ -250,7 +250,7 @@ class Settings(BaseSettings):
     PAYOUT_INVOICES_PREFIX: str = "POLAR-"
 
     # Application behaviours
-    API_PAGINATION_MAX_LIMIT: int = 100
+    API_PAGINATION_MAX_LIMIT: int = 300
 
     ACCOUNT_PAYOUT_DELAY: timedelta = timedelta(seconds=1)
     ACCOUNT_PAYOUT_MINIMUM_BALANCE: int = 1000


frankie567:
Do we really need to show 300 events by default on the events page? I'm worried this could be a big stress on the API if some users are abusing it.

Yopi:
> Do we really need to show 300 events by default on the events page? I'm worried this could be a big stress on the API if some users are abusing it.

Yeah, I'm not sure either. I just saw that we are not showing the pagination right now with the FE-change. I would also be hesitant to just merge this, but one of the two needs to be changed

pieterbeulque:
Revert it on the events page and throw it my way to render pagination 👋

frankie567:
Selfishly, I would prefer the UI to be tweaked rather than changing a default that impacts the whole API.

Implementing an infinite scroll shouldn't be too much work, we've already done it before: 

```ts
export const useCheckoutLinks = (
  organizationId: string,
  parameters?: Omit<
    NonNullable<operations['checkout-links:list']['parameters']['query']>,
    'organization_id'
  >,
) =>
  useInfiniteQuery({
    queryKey: ['checkout_links', { organizationId, ...(parameters || {}) }],
    queryFn: ({ pageParam }) =>
      unwrap(
        api.GET('/v1/checkout-links/', {
          params: {
            query: {
              organization_id: organizationId,
              ...(parameters || {}),
              page: pageParam,
            },
          },
        }),
      ),
    initialPageParam: 1,
    getNextPageParam: (lastPage, _allPages, lastPageParam) => {
      if (
        lastPageParam === lastPage.pagination.max_page ||
        lastPage.items.length === 0
      ) {
        return null
      }

      return lastPageParam + 1
    },
    retry: defaultRetry,
  })
  ```

Yopi:
Sounds good to me!

PR #7638, opened 2025-11-03
Skip tests if no changes

pieterbeulque:
Save the planet 🌍

### DIFF
--- a/.github/workflows/test_client.yaml
+++ b/.github/workflows/test_client.yaml
@@ -3,8 +3,14 @@ name: Client
 on:
   push:
     branches: ["main"]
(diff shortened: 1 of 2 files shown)


frankie567:
The thing is we have rules that _require_ tests to pass before being able to merge a PR. I'll see if there is a setting for when a workflow is skipped, but otherwise, I'm afraid we can't do that 😅

frankie567:
😖 https://github.com/orgs/community/discussions/44490

pieterbeulque:
Ha, okay, I did make an assumption that that was just going to work (skip = green). That's lame.

PR #7670, opened 2025-11-04
ops: Stagger deployment of migrations

Yopi:
If we have migrations, we can run the API deployment first since it will run the migration. If that succeeds we can deploy the workers, and they will be able to be deployed without any issues with mismatching schemas.

### DIFF
--- a/.github/workflows/deploy-environment.yml
+++ b/.github/workflows/deploy-environment.yml
@@ -13,10 +13,19 @@ on:
         description: "Docker image digest to deploy"
         required: true
         type: string
-      render-service-ids:
-        description: "Comma-separated list of Render service IDs to deploy to"
+      render-api-service-id:
+        description: "Render API service ID"
         required: true
         type: string
+      render-worker-service-ids:
+        description: "Comma-separated list of Render worker service IDs"
+        required: true
+        type: string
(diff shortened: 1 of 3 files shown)


Yopi:
Opened this up for the sake of discussion @frankie567, @psincraian. As I understand it, there's not really a problem in having the workers fail for a few seconds given that they will retry, but it would feel better if we could reduce the noise during a deployment of a migration. 

I think this coupled with splitting migrations and code changes as much as we can will make deployments a lot less volatile.

psincraian:
@Yopi if we want to reduce the errors and noise during DB migrations, I think the solution is to enforce an expand-contract pattern to be applied in every migration 🤔

If I'm not mistaken, this will make the errors and noise worse, as we are going to deploy changes sequentially, so the worker will try to use a removed column for longer.

Yopi:
Yes, that is the other half of the solution. The code using new columns needs to come after the migration, and vice-versa with removed columns.

If I'm not fully mistaken, the reason why we got issues yesterday is that we need to add new columns to the models for the DB migration to run, but the code with the models containing the new columns will be deployed before the migration has run. That causes the workers to fail to query data using the models.

PR #7689, opened 2025-11-05
settings: Implement autoSave mechanism instead of multiple save-buttons

emilwidlund:


pieterbeulque:
Isn't a full second delay rather annoying? Especially if we don't have a loader indicator anymore, you don't really know the save is pending from the UI. Or am I missing something?

emilwidlund:
> Isn't a full second delay rather annoying? Especially if we don't have a loader indicator anymore, you don't really know the save is pending from the UI. Or am I missing something?

If you change settings in Linear for example, any change you make automatically saves in the background without any visual cues, like a toaster.

Successful save toaster will still fire here though, but after 1 second. I wonder if we should remove them, and only show toaster if something unexpected happens during the save. 

Designing for deterministic outcomes is probably best.

emilwidlund:
Removed successful toasters now for a more seamless experience.

pieterbeulque:
Yeah, that's a good change!

I meant lowering the debounce delay to ±150–200ms to catch double clicks but I guess it doesn't matter that much.

PR #7711, opened 2025-11-07
feat(events): Add transitive closure

Yopi:
The closure allows us to easily query a sub-hierarchy of events (either from the root, or from any point in the tree), to be able to count events, or aggregate different attributes on the events. By using and keeping a closure we can do queries such as

```
SELECT * FROM events
LEFT JOIN events_closure ON events_closure.ancestor_id=events.id
```

Instead of having to do a recursive CTE.

Additionally we add a root_event_id to the events, as an equivalent to a trace ID in opentelemetry. This will allow us to query all events that belong together without even having to join in the closure.


server/polar/models/event.py
```diff
@@ -109,6 +111,14 @@ def is_not(self, other: Any) -> BinaryExpression[bool]:
 
 class Event(Model, MetadataMixin):
     __tablename__ = "events"
+    __table_args__ = (
+        Index(
+            "ix_events_org_timestamp_id",
+            "organization_id",
+            literal_column("timestamp DESC"),
```
frankie567, line 118: Curiosity: why it's important to index by descending order? :)

Yopi: The question is I think if we should have an index in both directions. I was optimizing for the events listing which will be descending by default, and that's how we will mostly query it.

Yopi:
> Feels sensible at first look, but curious to see how we'll use that when inserting and querying events :)

Tagged you in the follow up PR 😄

PR #7714, opened 2025-11-07
Add aggregation to Events list and a statistics endpoint for Events

Yopi:


server/polar/event/endpoints.py
```diff
@@ -88,6 +91,10 @@ async def list(
         False,
         description="When true, filters by parent_id (root events if not specified). When false, returns all events regardless of hierarchy.",
     ),
+    aggregate_costs: bool = Query(
```
Yopi: Not 100% sure about this, or we should make it more generic (aggregate_attribute: MultipleQueryFilter?) to be a bit more forward looking

server/polar/event/endpoints.py
```diff
@@ -138,6 +146,100 @@ async def list(
     )
 
 
+@router.get(
+    "/hierarchy-stats",
```
Yopi, line 151: Naming tbd

server/polar/event/endpoints.py
```diff
@@ -138,6 +146,100 @@ async def list(
...
+    summary="Get Hierarchy Statistics",
+    openapi_extra={"parameters": [get_metadata_query_openapi_schema()]},
+)
+async def get_hierarchy_stats_endpoint(
```
Yopi, line 156: @emilwidlund This endpoint might have been what we were discussing the other day

server/polar/event/endpoints.py
```diff
@@ -138,6 +146,100 @@ async def list(
     )
 
 
+@router.get(
+    "/hierarchy-stats",
+    summary="Get Hierarchy Statistics",
+    openapi_extra={"parameters": [get_metadata_query_openapi_schema()]},
+)
```
frankie567, line 155: Until we're 100% sure of the API format _and_ that it's useful to officially expose, I would advocate to make it a private endpoint (i.e. excluded from documentation and SDK). To do so add the argument: `    tags=[APITag.private]`

server/polar/event/sorting.py
```diff
@@ -32,3 +32,23 @@ class EventNamesSortProperty(StrEnum):
         )
     ),
 ]
+
+
+class EventHierarchyStatsSortProperty(StrEnum):
+    name = "name"
```
frankie567: I think Pyright is usually not happy with `name` as key. I usually workaround like this:

https://github.com/polarsource/polar/blob/e1adf61e6eaf01b6979285a5bddccf65f67bef3f/server/polar/product/sorting.py#L11-L12

server/polar/event/repository.py
```diff
@@ -191,3 +273,211 @@ def get_by_pending_entries_for_meter_statement(
...
+        statement: Select[tuple[Event]],
+        limit: int,
+        page: int,
+        aggregate_costs: bool = False,
+    ) -> tuple[Sequence[Event], int]:
+        """
+        List events using closure table to get a correct children_count.
+        Optionally aggregates costs from descendants into user_metadata._cost.amount.
+        """
+        descendant_event = Event.__table__.alias("descendant_event")
```
frankie567: I think you can use `aliased` as a shortcut: https://docs.sqlalchemy.org/en/20/orm/queryguide/api.html#sqlalchemy.orm.aliased

server/polar/event/repository.py
```diff
@@ -44,16 +58,84 @@ async def insert_batch(
...
 
+        events_needing_parent_lookup = []
+
+        # Set root_id for root events before insertion
+        for event in events:
+            if event.get("root_id") is not None:
+                continue
+            elif event.get("parent_id") is None:
+                # Root event - generate ID if not present and set root_id to itself
+                if "id" not in event:
```
frankie567: `id` is not present in the create schema, so this is always `True`.

server/polar/event/repository.py
frankie567, line 57: I guess we would benefit to have a `TypedDict` to define this "ingestable event" more precisely

server/polar/event/repository.py
```diff
@@ -44,16 +58,84 @@ async def insert_batch(
...
+                {
+                    "ancestor_id": event_id,
+                    "descendant_id": event_id,
+                    "depth": 0,
+                }
+            )
+
+            if parent_id is not None:
+                # Get parent's closure entries to copy
+                parent_closures_result = await self.session.execute(
```
frankie567: What if we didn't inserted the closures of the parent yet? How can we be sure we handled the parent before?

server/polar/event/repository.py
```diff
@@ -44,16 +58,84 @@ async def insert_batch(
...
+                            "depth": depth + 1,
+                        }
+                    )
+
+        if closure_entries:
+            await self.session.execute(
+                insert(EventClosure)
+                .values(closure_entries)
+                .on_conflict_do_nothing(index_elements=["ancestor_id", "descendant_id"])
+            )
```
frankie567: It looks like this closure linking operation is a bit expensive, isn't it? Maybe we should move it to a background task after the actual ingestion?

server/polar/event/endpoints.py
emilwidlund: Agree!

server/polar/event/endpoints.py
Yopi: Yes, that's a great idea! I didn't realise we had that 😄. Exactly what I would argue for during the development phase of this

server/polar/event/repository.py
Yopi: Ah, yeah that makes sense. Maybe it's too defensive

server/polar/event/repository.py
Yopi: Yeah, that makes sense. It's fine for it to lag slightly. I'll move it. If we move it we can probably combine the two flows that are basically the same (this one and the one in the model after insert)

server/polar/event/repository.py
Yopi: Essentially we have a requirement already in the ingestion layer that we need to have ingested a parent before its children. We could start to think about what it would mean if we don't have that requirement, but given https://github.com/polarsource/polar/pull/7714/files/6ddbcdd73de6d6aaf40c1e0fed2799bf10287136#diff-ab9916f2746e2f54c336e1e62cf87d62793ec90a226c59cb5eb68e6f95d9a170R561-R570 I am fine with not handling it here

server/polar/event/schemas.py
```diff
@@ -424,4 +424,50 @@ class EventName(Schema):
...
+class EventWithAggregations(Schema):
+    """An event with aggregated values from its descendants."""
+
+    event: Event = Field(description="The event.")
+    aggregations: EventAggregations = Field(
+        description="Aggregated values from all descendant events."
+    )
+
+
+class EventHierarchyStats(Schema):
```
pieterbeulque: This is a bit of a funny name. `RootEventStatistics`?

PR #7753, opened 2025-11-11
Increase minimum payout amount for certain currencies

Yopi:
Basically following the footsteps introduced in https://github.com/polarsource/polar/pull/4664. I don't think this is the best solution, but I also think this is something that will be continued to be worked on given the support for currencies that is coming soon (https://github.com/polarsource/handbook/pull/18). Hence, I opted to do something simple to now, to just hinder the problem from appearing. Given that it's not been raised in the past 11 months, it's not occurring very often.

I added a table for completeness to the docs, but it's kind of huge. Might not be the way we want to document it.
Screenshot 2025-11-11 at 11 38 32

### DIFF
--- a/docs/features/finance/payouts.mdx
+++ b/docs/features/finance/payouts.mdx
@@ -4,7 +4,46 @@ sidebarTitle: "Payouts"
 description: "Easily withdraw money from your Polar account at any time"
 ---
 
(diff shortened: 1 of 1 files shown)


frankie567 (approved):
Awesome, Jesper 👏

Issue #5798, opened 2025-05-30
Investigate memory leak in worker


frankie567:
It looks like our memory usage is slowly growing over time.

<img src="https://uploads.linear.app/5cb99480-3e0b-4cc4-a2f7-c61944b50350/6bd360c0-e66d-409b-9ddb-8b88cdc1465e/0a285040-d3e6-431b-a0e6-ee1bc40b2929?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzVjYjk5NDgwLTNlMGItNGNjNC1hMmY3LWM2MTk0NGI1MDM1MC82YmQzNjBjMC1lNjZkLTQwOWItOWRkYi04Yjg4Y2RjMTQ2NWUvMGEyODUwNDAtZDNlNi00MzFiLWEwZTYtZWUxYmM0MGIyOTI5IiwiaWF0IjoxNzQ4NTg4MTAxLCJleHAiOjMzMzE5MTQ4MTAxfQ.ogKzNTdBlK7Q6928hHeIBLz1q3EgDgOSEOa4hSvl2lg " alt="Capture d’écran 2025-05-30 à 08.53.26.png" width="866" data-linear-height="307" />

A few ideas:

* HTTPX clients
* Redis clients

frankie567:
Seems solved, even though I'm not sure if it's because of things I implemented or it was nothing after all.<br><br>We still see that slow increase over time, but seems to stabilise at some point:

[...]

PR #7754, opened 2025-11-11
server/sentry: Reenable Dramatiq integration

Yopi:
Given Francois comment here https://github.com/polarsource/polar/issues/5798\#issuecomment-2987077486, we should reenable the Dramatiq integration. Right now we are a bit blind for things happening in the workers.

### DIFF
--- a/server/polar/sentry.py
+++ b/server/polar/sentry.py
@@ -37,7 +37,7 @@ def configure_sentry() -> None:
         environment=settings.ENV,
         integrations=[
             FastApiIntegration(transaction_style="endpoint"),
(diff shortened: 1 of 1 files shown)


frankie567:
Didn't remember that! Worth to note that errors ([example](https://polar-sh.sentry.io/issues/6683367160/?project=4505046565519360&query=is%3Aunresolved&referrer=issue-stream)) were still reported correctly from the workers, but I guess some metadata/breadcrumbs were missing.

frankie567 (approved):

Yopi:
Interesting, then that's not what I was chasing either 😄

PR #7856, opened 2025-11-17
Persist business name across checkouts

### LINKED ISSUE #7854: Business name is not prefilled when creating a checkout session with a customer ID

pieterbeulque:
Fixes #7854 

Feels a bit hacky since we don't store `is_business_customer` directly from the checkout on the customer, so I use the required fields from the checkouts when `is_business_customer` is set to know if `is_business_customer` was checked previously 😵

### DIFF
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -496,11 +496,7 @@ class CheckoutService:
 
             # Auto-select business customer if they have both a billing name & billing address
             # since that means they've previously checked the is_business_customer checkbox
-            if (
-                checkout.customer.billing_name is not None
-                and checkout.customer.billing_address is not None
-                and checkout.customer.billing_address.has_address()
-            ):
+            if checkout.customer.billing_name is not None:
                 checkout.is_business_customer = True
 
         if checkout.payment_processor == PaymentProcessor.stripe:


server/polar/checkout/service.py
```diff
@@ -493,6 +494,15 @@ async def create(
...
 
+            # Auto-select business customer if they have both a billing name & billing address
+            # since that means they've previously checked the is_business_customer checkbox
+            if (
+                checkout.customer.billing_name is not None
+                and checkout.customer.billing_address is not None
+                and checkout.customer.billing_address.has_address()
+            ):
+                checkout.is_business_customer = True
+
```
frankie567, line 501: It's good that we prefill the business name value if we have it.

However, I would say we should avoid overloading the `is_business_customer` flag with implicit magic: since this property can be explicitly set from the Create Checkout API (https://polar.sh/docs/api-reference/checkouts/create-session#body-is-business-customer), I think we should not get in the way.

pieterbeulque: The weird thing is that if I don't add this line then the business name & address is prefilled but the checkbox isn't checked, so as a customer you're submitting information that you don't see if you never check the box.

frankie567: Could we then prefill if/when `is_business_customer` is set to true? And clear if set to false?

pieterbeulque: Aha, in a way that makes sense, but then you have to have a way to know that when calling Polar, while you don't know what they used during their previous checkout (that's all Polar state).

We can know from their previous checkout but that feels too silly.

frankie567: But the customer can still check it. If the customer is already on Polar's record, that doesn't make a big difference.

pieterbeulque: I'm afraid I don't understand what you're saying then, sorry. This prefill happens during a checkout session create for an existing customer.

If it's unchecked by default then we'd always clear it / not prefill? Or are you saying capture when the user checks the `is_business_customer` checkbox, and then prefill everything in the checkout session update? But that's inconsistent with how e.g. `billing_address` is currently implemented? 🤔

frankie567: Yes, that was my idea: if `is_business_customer` becomes checked, then in the update handler, we can prefill with the data we have.

> But that's inconsistent with how e.g. billing_address is currently implemented? 🤔

How so?

PR #7858, opened 2025-11-17
Remove churn rate from metrics for now

pieterbeulque:
It's visible on the homepage & customers page but we haven't properly tested & built this out yet.


Yopi (approved):

pieterbeulque:
@copilot remove churn rate from the test too

pieterbeulque:
@Yopi Why would these specific tests (hierarchical events) fail if I skip an unrelated test? 🤔

Yopi:
It really shouldn't, I wonder if it is flaky? 🤔

Not getting the same locally

PR #7860, opened 2025-11-17
metrics: cumulative revenue number

Yopi:
We want to include all of history in the cumulative revenue numbers.

### DIFF
--- a/server/polar/metrics/queries.py
+++ b/server/polar/metrics/queries.py
@@ -12,6 +12,7 @@ from sqlalchemy import (
     and_,
     cte,
     func,
+    literal,
     or_,
     select,
 )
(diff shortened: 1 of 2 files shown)


Yopi:
@frankie567 before I merge this, I just want to check in with you that this is actually correct -- I.e. the cumulative revenue should include all of the revenue (from history) up until the point of the period. Not only the revenue within the period?

I updated the test case to show what I mean

server/polar/metrics/queries.py
```diff
@@ -142,6 +143,42 @@ def get_orders_metrics_cte(
...
+                ),
+                func.coalesce(func.sum(Order.payout_amount), 0).label(
+                    "hist_net_cumulative_revenue"
+                ),
+            )
+            .select_from(Order)
+            .join(
+                Subscription,
+                isouter=True,
+                onclause=Order.subscription_id == Subscription.id,
```
frankie567: Do we need this join to compute that metric?

Yopi: No we don't, good catch! Thank you

PR #7896, opened 2025-11-18
Refresh organization cache after profile picture update

pieterbeulque:


pieterbeulque:
@sebastianekstrom how much of a code smell do you think all these `router.refresh()`'es are? 😇

PR #7928, opened 2025-11-19
metrics: Handle non-1-aligned periods for joining cumulative revenue

Yopi:
Otherwise we get no records to join on and thus no results.


pieterbeulque (approved):
🐛🔨

PR #7966, opened 2025-11-20
Validate custom field number constraints

pieterbeulque:
Users could enter wrong numbers in the custom field validation constraints and that threw a 500 upon rendering the checkout.

### DIFF
--- a/server/polar/custom_field/schemas.py
+++ b/server/polar/custom_field/schemas.py
@@ -1,13 +1,6 @@
-from typing import Annotated, Any, Literal
+from typing import Annotated, Literal
 
-from pydantic import (
-    UUID4,
-    Discriminator,
-    Field,
-    StringConstraints,
-    TypeAdapter,
-    field_validator,
-)
+from pydantic import UUID4, Discriminator, Field, StringConstraints, TypeAdapter
 
 from polar.kit.metadata import (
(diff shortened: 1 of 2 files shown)


server/polar/custom_field/schemas.py
```diff
@@ -40,6 +47,28 @@
...
+
+    form_label: str | None = None
+    form_help_text: str | None = None
+    form_placeholder: str | None = None
+    ge: int | None = None
+    le: int | None = None
+
+    @field_validator("le")
+    @classmethod
+    def validate_ge_le_relationship(cls, v: int | None, info: Any) -> int | None:
```
frankie567: I think we can avoid to have `CustomFieldNumberPropertiesInput` by building a custom validator and set it in annotation directly on `ComparableProperties`:

```py
def validate_ge_le(v: int, info: ValidationInfo[Any]) -> int:
    ge = info.data.get("ge")
    if ge is not None and v is not None and ge > v:
        raise ValueError(
            "Greater than or equal (ge) must be less than or equal to "
            "Less than or equal (le)"
        )
    return v


class ComparableProperties(TypedDict):
    ge: NotRequired[int]
    le: NotRequired[Annotated[int, AfterValidator(validate_ge_le)]]

```

PR #7968, opened 2025-11-20
event: Add labels to events

Yopi:
This moves the system event labels to the backend, with the introduction of user event labels.

All events now have a label that can either be updated by the user or set by the system.

The labels for system events are not typed as a hardcoded string for two reasons:
1. We want to be able to change the copy without breaking the API
2. We dont want to have multiple definitions of the labels

### DIFF
--- a/clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/usage-billing/events/ClientPage.tsx
+++ b/clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/usage-billing/events/ClientPage.tsx
@@ -4,7 +4,6 @@ import { CustomerSelector } from '@/components/Customer/CustomerSelector'
 import { EventCreationGuideModal } from '@/components/Events/EventCreationGuideModal'
 import { EventMetadataFilter } from '@/components/Events/EventMetadataFilter'
 import { Events } from '@/components/Events/Events'
-import { useEventDisplayName } from '@/components/Events/utils'
 import { DashboardBody } from '@/components/Layout/DashboardLayout'
 import DateRangePicker from '@/components/Metrics/DateRangePicker'
 import { Modal } from '@/components/Modal'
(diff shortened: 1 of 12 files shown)


server/polar/event/schemas.py
```diff
@@ -215,6 +215,7 @@ class BaseEvent(IDSchema):
         default=None,
         description="The ID of the parent event.",
     )
+    label: str = Field(description="Human readable label of the event type.")
```
pieterbeulque, line 218: Wouldn't this be nullable? 🤔

clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/usage-billing/events/ClientPage.tsx
```diff
@@ -48,6 +47,46 @@ import z from 'zod'
 
 const PAGE_SIZE = 100
 
+// Replace this after migrating /v1/events/names to use labels.
+const useEventDisplayName = (eventName: schemas['Event']['name']) => {
```
Comment (pieterbeulque) [inline clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/usage-billing/events/ClientPage.tsx:51]: Is this blocked by something I'm not seeing or are you just planning to do this in a separate PR?

Comment (Yopi) [reply clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/usage-billing/events/ClientPage.tsx:51]: I wanted to limit the blast radius of this change. We'll want to switch the events listing here to not use the events/names endpoint but rather one that gets us the event types (Potentially? I'm not 100% sure how we should handle the system events in that endpoint). 

It's not blocked, but it's also not really impacting anything if we merge this for now (other than the events will be using their names in the sidebar in the events page).

server/polar/event/schemas.py
Yopi: I'm thinking that we enforce it. All user events will get an event_types_id and a corresponding entry in the events_types table with the label set to the name initially. Then we handle the system events specifically in the model now

clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/usage-billing/events/ClientPage.tsx
Comment (pieterbeulque) [reply clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/usage-billing/events/ClientPage.tsx:51]: Ah, okay, I get it now, I thought `event.label` wasn't set for system events.

Would it be cleaner to add an optional `label` field to the output of `event/names` as well and return these strings from the back-end? Single source of truth and all that.

PR #7981, opened 2025-11-21
Validate minimum_limit

pieterbeulque:
I think this can break the analytics in the native app, as monthly is no longer a valid interval for date ranges < 60 days.

 cc @sebastianekstrom

### DIFF
--- a/server/polar/event/endpoints.py
+++ b/server/polar/event/endpoints.py
@@ -234,8 +234,8 @@ async def list_statistics_timeseries(
                 {
                     "loc": ("query",),
                     "msg": (
-                        "The interval is too big. "
-                        "Try to change the interval or reduce the date range."
+                        "The interval does not match the date range. "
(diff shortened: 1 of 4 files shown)


frankie567:
@pieterbeulque Should we merge this?

pieterbeulque:
@sebastianekstrom - Can I pass this to you? It'll need some context grabbing but basically on our analytics we now have a minimum date range too. So you can't show your metrics monthly if total date range is fewer than 60 days, for example.

How I'd go about this is first verifying how the analytics are shown & what the supported date ranges are. Do we have date picker + hourly/daily/weekly/monthly selector in the native app? That functionality should match the limitations we now have in `/dashboard/[organization]/analytics`. Once the native app can't send in any unsupported combinations, we can merge this.

Feel free to pull me in on Slack!

pieterbeulque:
F— it.

PR #8006, opened 2025-11-23
Update theming.ts

rishi-raj-jain:
from the [internal conversation](https://polar-olw4768.slack.com/archives/C0942KS0BD2/p1763906061244909)


emilwidlund:
We'd want to strip away the custom theme support entirely though.

rishi-raj-jain:
@emilwidlund 

I've updated the code to reflect that.

PR #8013, opened 2025-11-24
Trigger `subscription.updated` webhook on trial updates

### LINKED ISSUE #7962: Fix subscription.updated webhook not firing when trial date is extended
> from the [ticket](https://app.plain.com/workspace/w_01JE9TRRX9KT61D8P2CH77XDQM/thread/th_01KAEY1BQB2TSNBP427FXVEX7P/)

The `subscription.updated` webhook is not being fired when the trial date is extended for a subscription. This affects customers who rely on webhooks to track subscription changes.

pieterbeulque:
The update_trial method was missing the webhook trigger that other subscription update methods use. This commit adds the _after_subscription_updated call to ensure the subscription.updated webhook fires when a trial date is extended.

Also adds a test to verify the webhook is triggered.

Fixes #7962

### DIFF
--- a/server/tests/subscription/test_service.py
+++ b/server/tests/subscription/test_service.py
@@ -2340,32 +2340,6 @@ class TestUpdateTrial:
         save_fixture: SaveFixture,
         product: Product,
         customer: Customer,
-    ) -> None:
-        subscription = await create_trialing_subscription(
-            save_fixture, product=product, customer=customer
-        )
-
-        assert subscription.trial_end is not None
-        original_trial_end = subscription.trial_end
-
-        new_trial_end = original_trial_end + timedelta(days=30)
-
-        updated_subscription = await subscription_service.update_trial(
-            session, subscription, trial_end=new_trial_end
-        )
-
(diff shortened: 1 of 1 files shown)


server/tests/subscription/test_service.py
```diff
@@ -2360,6 +2360,34 @@ async def test_trialing_subscription_extending(
...
+        reset_hooks(subscription_hooks)
+        updated_subscription = await subscription_service.update_trial(
+            session, subscription, trial_end=new_trial_end
+        )
+
+        assert updated_subscription.status == SubscriptionStatus.trialing
+        assert updated_subscription.trial_end == new_trial_end
+
+        # Verify that the webhook was triggered
+        assert_hooks_called_once(subscription_hooks, {"updated"})
```
frankie567, line 2366: We could add those assertions in the other tests rather than adding a dedicated test :)

frankie567 (commented):
All good! Small personal nitpick on the test, but do as you see fit :)

PR #8025, opened 2025-11-24
event-types: Add label property selector

Yopi:
As an addition to event labels, we allow for an label property selector for data in the metadata, to allow for a bit more specific event labels

Screenshot 2025-11-24 at 11 05 13
Screenshot 2025-11-24 at 11 05 25


pieterbeulque:
Beautiful

PR #8038, opened 2025-11-24
refactor: remove unnecessary twMerge calls with static strings

pieterbeulque:
Replace all className={twMerge('static strings')} with className="static strings" for better rendering performance.

cc @emilwidlund


emilwidlund:
Ah, yes lol. Thanks for cleaning up my shit 🥇

PR #8091, opened 2025-11-26
Add customers (cost) analytics

Yopi:
- Add private (for now) customers analytics endpoint
- Add table on costs page


Yopi:
@pieterbeulque taking this over into his own PR

PR #8132, opened 2025-11-27
refactor(events): Faster query to get customers latest event

birkjernstrom:


birkjernstrom:
Haven't been able to test this locally (dev environment is completely broken). But this should *theoretically* work and at least in draft mode, and serve for collaboration/discussion.

Another challenge: This query gets expanded upon in `customer_meter.service.update_customer_meter:137+` and `customer_meter.service.update_customer_meter:172+`

birkjernstrom:
Before I forget: In regards to the extra conditionals that are set and modifications to the query... Perhaps they need to be set on the inner queries of the UNION. So solution might have to be a final "wrapper" of the ultimate query that creates a union of it.

PR #8142, opened 2025-11-28
logfire: Use Render service name for workers

Yopi:


Yopi:
I think this would be useful for debugging, unless there is a reason to not do it @frankie567 ?

frankie567:
Yep, good catch. I'm wondering then if we shouldn't just get rid of the `service_name` argument and always read `RENDER_SERVICE_NAME`?

Yopi:
I think that would be nice. I wasn't sure if we had anything depending on the name being `server` as opposed to `api` which is the render service name

Yopi:
I'll follow up

PR #8162, opened 2025-11-28
Add `dirty` boolean on customers

pieterbeulque:
This would exclude customers from being queued again during customer meter updates when tasks take longer than the debounce threshold.

### DIFF
--- a/server/migrations/versions/2025-11-28-1506_add_dirty_flag_to_customer.py
+++ b/server/migrations/versions/2025-11-28-1506_add_dirty_flag_to_customer.py
@@ -1,0 +1,51 @@
+"""add dirty flag to customer
+
+Revision ID: 856fff9b3dfe
+Revises: f5223877acbb
+Create Date: 2025-11-28 15:06:40.120064
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# Polar Custom Imports
+
+# revision identifiers, used by Alembic.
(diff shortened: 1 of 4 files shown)


pieterbeulque:
I expect the migration to need a bit of a manual push on a customer or two to avoid someone lingering in the wrong state (`meters_dirtied_at` not matching `dirty`)

server/migrations/versions/2025-11-28-1506_add_dirty_flag_to_customer.py
```diff
@@ -0,0 +1,51 @@
...
+down_revision = "f5223877acbb"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
+    # Add dirty column with default False
+    op.add_column(
+        "customers",
+        sa.Column("dirty", sa.Boolean(), nullable=False, server_default=sa.false()),
```
Yopi, line 25: Does the name make sense on the customer? I.e. what does it mean that a customer is dirty outside of processing customer_meters.

Yopi, line 25: I think you said it, but will we allow for more functionality if we were to make this a datetime from the get go? Even if we use it the same way?

pieterbeulque:
I'm an idiot

PR #8169, opened 2025-12-01
Expand system events

Yopi:
This PR expands system events with:

More information to `order.paid`, `subscription.revoked` events. Basically information on amounts and subscription periods.

New events:
- `subscription.created`
- `subscription.canceled`
- `checkout.created`

The idea behind these events is that we should be able do generate metrics based on the system events if we want to, instead of having to query all individual tables.

As well as a backfill script. I replaced the previous backfill script, we could add the previous back fill into the new script if we want to be able to keep that part as well.

### DIFF
--- a/server/polar/event/repository.py
+++ b/server/polar/event/repository.py
@@ -26,6 +26,7 @@ from sqlalchemy.orm import aliased, joinedload
 from polar.auth.models import AuthSubject, Organization, User, is_organization, is_user
 from polar.kit.repository import RepositoryBase, RepositoryIDMixin
 from polar.kit.repository.base import Options
+from polar.kit.sorting import Sorting
 from polar.kit.utils import generate_uuid
 from polar.models import (
     BillingEntry,
@@ -38,6 +39,7 @@ from polar.models import (
 from polar.models.event import EventClosure, EventSource
 from polar.models.product_price import ProductPriceMeteredUnit
 
+from .sorting import EventSortProperty
 from .system import SystemEvent
 
 
@@ -247,6 +249,9 @@ class EventRepository(RepositoryBase[Event], RepositoryIDMixin[Event, UUID]):
         depth: int | None = None,
         parent_id: UUID | None = None,
         cursor_pagination: bool = False,
+        sorting: Sequence[Sorting[EventSortProperty]] = (
+            (EventSortProperty.timestamp, True),
+        ),
     ) -> tuple[Sequence[Event], int]:
         """
         List events using closure table to get a correct children_count.
@@ -376,6 +381,14 @@ class EventRepository(RepositoryBase[Event], RepositoryIDMixin[Event, UUID]):
                         text("true"),
                     )
 
+        order_by_clauses: list[UnaryExpression[Any]] = []
+        for criterion, is_desc in sorting:
+            clause_function = desc if is_desc else asc
+            if criterion == EventSortProperty.timestamp:
+                order_by_clauses.append(
+                    clause_function(paginated_events_subquery.c.timestamp)
+                )
+
         final_query = (
             select(Event)
             .select_from(paginated_events_subquery)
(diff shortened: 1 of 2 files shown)


server/polar/checkout/service.py
```diff
@@ -2208,6 +2214,20 @@ async def _create_ad_hoc_custom_price(
     async def _after_checkout_created(
         self, session: AsyncSession, checkout: Checkout
     ) -> None:
+        metadata = CheckoutCreatedMetadata(
+            checkout_id=str(checkout.id),
+            checkout_status=checkout.status,
```
frankie567, line 2219: Is it for consistency with potential `checkout.updated`, `checkout.confirmed` events? Because at this stage, `status` will always be `open`.

server/polar/event/system.py
```diff
@@ -157,6 +163,38 @@ class CustomerDeletedEvent(Event):
...
+
+class SubscriptionCanceledEvent(Event):
+    if TYPE_CHECKING:
+        source: Mapped[Literal[EventSource.system]]
+        name: Mapped[Literal[SystemEvent.subscription_canceled]]
+        user_metadata: Mapped[SubscriptionCanceledMetadata]  # type: ignore[assignment]
+
+
 class SubscriptionCycledMetadata(TypedDict):
     subscription_id: str
```
frankie567, line 203: Could it be worth to add more metadata (amount, currency...) to this one as well?

server/polar/event/system.py
```diff
@@ -157,6 +163,38 @@ class CustomerDeletedEvent(Event):
...
+        name: Mapped[Literal[SystemEvent.subscription_created]]
+        user_metadata: Mapped[SubscriptionCreatedMetadata]  # type: ignore[assignment]
+
+
+class SubscriptionCanceledMetadata(TypedDict):
+    subscription_id: str
+    customer_cancellation_reason: NotRequired[str]
+    customer_cancellation_comment: NotRequired[str]
+    canceled_at: str
+    ends_at: NotRequired[str]
```
frankie567, line 192: `created` and `revoked` have access to amount, currency and recurring interval. Wouldn't it be interesting for this one as well?

server/polar/event/system.py
```diff
@@ -223,7 +265,15 @@ class OrderPaidMetadata(TypedDict):
...
-    backfilled: NotRequired[bool]
+    net_amount: int
+    tax_amount: int
+    applied_balance_amount: int
+    discount_amount: int
+    discount_id: NotRequired[str]
+    platform_fee: int
+    subscription_id: NotRequired[str]
+    recurring_interval: NotRequired[str]
+    recurring_interval_count: NotRequired[int]
```
frankie567, line 284: For clarity, shouldn't we name those `subscription_recurring_interval` or `subscription_recurring_interval_count`? Maybe too verbose.

server/polar/checkout/service.py
Yopi: Yeah that's what I was thinking, but I wasn't sure if it was worth creating those events as well. I don't think we have a need right now

server/polar/event/system.py
Yopi: I was thinking that it's implied by the event name, and that this is what they are called on the db table as well. Could of course rename them if we think it is clearer!

frankie567: Yes, you're right 👍

server/polar/event/system.py
Yopi: Yes!

server/polar/event/system.py
Yopi: Added it!

server/polar/checkout/service.py
Yopi: I don't know, maybe it's entirely unnecessary like you say. It will always be the same 🤔

PR #8171, opened 2025-12-01
Terraform split workspaces and add linting

Yopi:
* Split TF into a workspace per environment
* Add formatting check + linting

### DIFF
--- a/.github/workflows/terraform.yaml
+++ b/.github/workflows/terraform.yaml
@@ -1,0 +1,45 @@
+name: Terraform
+
+on:
+  push:
+    branches: ["main"]
+    paths:
+      - "terraform/**"
+      - ".github/workflows/terraform.yaml"
+  pull_request:
+    types: [opened, synchronize]
+    paths:
+      - "terraform/**"
+      - ".github/workflows/terraform.yaml"
+
+permissions:
+  contents: read
+
+jobs:
+  fmt:
+    name: "Terraform: Lint 📝"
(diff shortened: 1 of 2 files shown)


terraform/production/aws.tf
```diff
@@ -115,7 +115,7 @@ resource "aws_s3_bucket" "backups" {
   bucket = "polar-sh-backups"
 }
 
-resource "aws_s3_bucket_lifecycle_configuration" "backups-lifecycle" {
+resource "aws_s3_bucket_lifecycle_configuration" "backups_lifecycle" {
```
frankie567, line 118: Won't that destroy the existing bucket? 😊

terraform/production/aws.tf
frankie567, line 154: Probably we'll need to have sandbox buckets under sandbox at some point?

terraform/production/aws.tf
Yopi: It will destroy the lifecycle config at least, but I'll move the resource so we won't get a problem

terraform/production/aws.tf
Yopi: Yeah, I think that makes sense. I was thinking that can happen in the next step, depending on what we want to do on the AWS side (one or multiple accounts there as well)

terraform/production/aws.tf
Yopi: But good point, I'll do that separately first!

PR #8173, opened 2025-12-01
Add more details to order email

### LINKED ISSUE #7984: More details in new sale email
## Is your feature request related to a problem?

Current email:

Image

I offer PPP discounts depending on the country that the user is visiting from, and it would be nice to quickly confirm if the billing country matches the used discount when an email comes in.

Also I'm always just curious to learn about a new customer when a sale is made :)

## Describe the solution you'd like

Either more details in the email or a link to the Polar app where I can validate such details would be nice. The link of the email address is a mailto: link.

## Describe alternatives you've considered

Looking up the customer manually.

sebastianekstrom:
## 📋 Summary

**Related Issue**: Fixes #7984

## 🎯 What

This PR will add some more data to the order confirmation email to give our users a quick glance of the purchase that has been made:

- Product image
- CTA that will take you to the order
- Customer details, including name and address

I also took the opportunity to redesign the email a bit.

## 🖼️ Screenshots/Recordings

Frame 3822


server/emails/src/emails/notification_new_sale.tsx
```diff
@@ -1,36 +1,141 @@
...
+  })
+
+  const addressParts = [
+    billing_address_line1,
+    billing_address_city,
+  ].filter(Boolean)
+  const formattedAddress =
+    addressParts.length > 0 ? addressParts.join(', ') : null
+
+  const orderUrl = `https://polar.sh/dashboard/${organization_slug}/sales/${order_id}`
```
Yopi: I think we send emails in both environments, in which case this needs to handle sandbox / prod

server/polar/notifications/notification.py
```diff
@@ -89,15 +90,41 @@ class MaintainerNewPaidSubscriptionNotification(NotificationBase):
 
 
 class MaintainerNewProductSaleNotificationPayload(NotificationPayloadBase):
-    customer_name: str
+    customer_email: str
+    customer_name: str | None = None
```
Yopi, line 96: Won't we still always have a `customer_name`? So this should just be `: str`

Yopi: Ah, or was the previous guarantee based on that we were passing in the `customer_email`?

server/polar/order/service.py
```diff
@@ -1542,16 +1543,44 @@ async def send_admin_notification(
...
+                        else None,
                         product_name=product.name,
                         product_price_amount=order.net_amount,
-                        organization_name=organization.slug,
+                        product_image_url=product_image_url,
+                        order_id=str(order.id),
+                        order_date=order.created_at.isoformat(),
+                        organization_name=organization.name,
+                        organization_slug=organization.slug,
+                        billing_reason=order.billing_reason,
```
Yopi, line 1583: I think this is typed as a non-str (`OrderBillingReasonInternal`), so you should be able to pass through the enum instead of a string

server/polar/notifications/notification.py
```diff
@@ -89,15 +90,41 @@ class MaintainerNewPaidSubscriptionNotification(NotificationBase):
...
+    organization_slug: str
+    billing_reason: str
 
     @computed_field
     def formatted_price_amount(self) -> str:
         return format_currency(self.product_price_amount / 100, "USD", locale="en_US")
 
+    @computed_field
+    def formatted_billing_reason(self) -> str:
+        reasons = {
```
Yopi: Passing through the enum (as per the other comment) will make this a bit more complete. I think you can even get an exhaustive switch in that case (if you opt to not use a dict)

server/polar/notifications/notification.py
sebastianekstrom: > Ah, or was the previous guarantee based on that we were passing in the customer_email?

Exactly!

server/polar/order/service.py
sebastianekstrom: Good call!

server/emails/src/emails/notification_new_sale.tsx
sebastianekstrom: Nice catch!

PR #8228, opened 2025-12-02
Replace `hierarchical:true` with `depth` parameter

pieterbeulque:


server/polar/sentry.py
```diff
@@ -28,6 +28,9 @@ def setup_once() -> None:
 
 
 def configure_sentry() -> None:
+    if not settings.SENTRY_DSN:
```
Yopi: Intentional to be part of this commit?

pieterbeulque: Nope, sorry

PR #8253, opened 2025-12-03
Accept `aggregate_fields` on `GET /v1/events/:id`

pieterbeulque:
Cherry-picked from #8182

### DIFF
--- a/server/polar/event/repository.py
+++ b/server/polar/event/repository.py
@@ -421,6 +421,21 @@ class EventRepository(RepositoryBase[Event], RepositoryIDMixin[Event, UUID]):
 
         return events, total_count
 
+    async def get_with_aggregation(
+        self,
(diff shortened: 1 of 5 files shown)


server/polar/event/repository.py
```diff
@@ -421,6 +421,21 @@ async def list_with_closure_table(
...
+        id: UUID,
+        aggregate_fields: Sequence[str],
+    ) -> Event | None:
+        """Get a single event with aggregated metadata from descendants."""
+        statement = self.get_readable_statement(auth_subject).where(Event.id == id)
+
+        events, _ = await self.list_with_closure_table(
+            statement, limit=1, page=1, aggregate_fields=aggregate_fields
+        )
+
```
Yopi, line 436: Don't you need to set the depth here to get the aggregation from the children? Otherwise you'll just get a single event

PR #8292, opened 2025-12-04
Add `logo.dev` for business emails

pieterbeulque:
We'd need to purchase a license to do it in prod, but it looks way nicer for B2B.

### DIFF
--- a/server/polar/customer/schemas/customer.py
+++ b/server/polar/customer/schemas/customer.py
@@ -130,11 +130,14 @@ class CustomerBase(MetadataOutputMixin, TimestampedSchema, IDSchema):
     def avatar_url(self) -> str:
         domain = self.email.split("@")[-1].lower()
 
-        if not settings.LOGO_DEV_TOKEN or domain in settings.PERSONAL_EMAIL_DOMAINS:
+        if (
+            not settings.LOGO_DEV_PUBLISHABLE_KEY
+            or domain in settings.PERSONAL_EMAIL_DOMAINS
+        ):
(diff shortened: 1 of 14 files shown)


pieterbeulque:
@frankie567 Care to take a look at the Python part? 🙇

server/polar/customer/schemas/customer.py
```diff
@@ -40,6 +41,23 @@
 _name_description = "The name of the customer."
 _name_example = "John Doe"
 
+_personal_email_domains: set[str] = {
```
frankie567: Could we put that under `config.py` in `Settings`? Would allow us to easily override it dynamically if needed.

server/polar/organization/service.py
```diff
@@ -260,6 +260,12 @@ async def update(
             },
         )
 
+        # Ignore logo.dev URLs for avatar_url - these are computed values
+        # that the frontend may inadvertently send back
+        if avatar_url := update_dict.get("avatar_url"):
```
frankie567: That logic should be made as a validator and applied at schema level.

```
def _discard_logo_dev_url(url: HttpUrl) -> HttpUrl | None:
    if url.host and url.host.endswith("logo.dev")
        return None
    return url

AvatarUrl = Annotated[HttpUrlToStr, AfterValidator(_discard_logo_dev_url)]

class OrganizationCreate(Schema): # and OrganizationUpdate
    avatar_url: AvatarUrl | None = None
```

frankie567 (changes requested):

pieterbeulque:
Good call on the validator, @frankie567, that's a lot cleaner. Can you take another look?

PR #8343, opened 2025-12-08
Add threshold and spread time for orders

pieterbeulque:
@frankie567 - This is more conceptual than anything else, but I thought it was an interesting problem so I wanted to poke around with it a bit.

It's a balancing act of a few factors. I tried to juggle this by first introduce a threshold that flips into "bulk mode" which will introduce this delay spreading, allowing small spikes for immediate updates below that threshold.

Beyond that threshold, try to ensure a target "spread time" (i.e. time of completion; currently five minutes) so that changes still feel relatively "instant". Allow some elasticity in this with a target & minimum delay (200ms & 50ms respectively). This will aim to keep the spread time (leaving more time in between on smaller sets, compressing up to the minimum delay on bigger payloads) and eventually grow beyond the spread time on really large sets.

With the current configuration, up to 50 jobs are queued immediately, beyond that:

Count | Delay/item | Last item delay | Total spread
-- | -- | -- | --
100 | 200ms | 19.8s | 0.3 min
500 | 200ms | 99.8s | 1.7 min
1500 | 200ms | 299.8s | 5.0 min
3000 | 100ms | 299.9s | 5.0 min
6000 | 50ms | 299.9s | 5.0 min
10000 | 50ms | 499.9s | 8.3 min
100000 | 50ms | 4999.9s | 83.3 min

My biggest concern is that storing a job in Redis for that long risks data loss / inconsistencies if something were to happen.

### DIFF
--- a/server/polar/worker/_enqueue.py
+++ b/server/polar/worker/_enqueue.py
@@ -221,6 +221,6 @@ def calculate_bulk_job_delay(index: int, total_count: int) -> int | None:
         delay_per_item = target_delay_ms
     else:
         # Compress to fit, but enforce minimum floor
-        delay_per_item = max(min_delay_ms, max_spread_ms / total_count)
+        delay_per_item = max(min_delay_ms, int(max_spread_ms / total_count))
 
     return int(index * delay_per_item)
--- a/server/tests/subscription/test_service.py
+++ b/server/tests/subscription/test_service.py
@@ -991,6 +991,7 @@ class TestCycle:
             customer_id=customer.id,
             product_id=product.id,
             subscription_id=subscription.id,
+            delay=None,
         )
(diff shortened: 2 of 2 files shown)


server/polar/order/service.py
```diff
@@ -1771,27 +1771,34 @@ async def send_confirmation_email(
     async def update_product_benefits_grants(
         self, session: AsyncSession, product: Product
     ) -> None:
-        statement = select(Order).where(
+        # Skip seat-based orders - benefits are granted when seats are claimed
```
frankie567, line 1774: Unrelated but we should probably open an issue for that: if a merchant adds a new benefit to a product, we expect all the claimed seats get that new benefit.

PR #8359, opened 2025-12-08
Track AI onboarding on organizations

pieterbeulque:
I'm not sure how we feel about adding an extra field like this on the organization.

Is there a way to exclude this from the public API?

The main reason to add this is to track % of successful setups / time to first sale / … on companies that manage to set up their agent through the AI onboarding assistant.

### DIFF
--- a/server/migrations/versions/2025-12-08-1609_add_ai_onboarding_completed_at_to_.py
+++ b/server/migrations/versions/2025-12-08-1609_add_ai_onboarding_completed_at_to_.py
@@ -1,7 +1,7 @@
 """add ai_onboarding_completed_at to organization
 
 Revision ID: 1b73becdc178
-Revises: c4b8b4d63805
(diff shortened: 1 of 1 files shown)


Yopi (approved):
The APITag.private will exclude it from the external SDK at least.

I think it's fine to add it, since we have so many other `timestamps_at` on the organization, it sort of aligns with those.

pieterbeulque:
Yes, but that's for the endpoint, the field is still on the public schema for `Organization` if I'm not mistaken. I know you have used `include_in_schema=False` previously, maybe I should add that as well.

We should look into these timestamps at some point (e.g. there is an existing `onboarded_at` that doesn't get set anymore with the new onboarding, I believe).

Yopi:
Ahh, yes. My wires got crossed, but yes, you can add the `include_in_schema=False`

Issue #8335, opened 2025-12-05
Add a delay when enqueuing benefit grant updates for a Product


frankie567:
When a merchant adds a new Benefit to a Product, all existing subscribers and customers get access to that new Benefit. For this, we enqueue a benefit grant update for all the customers. 

However, it might hit us hard if the merchant has lot of customers: it'll fill up the queue immediately, and nothing else will be processed before we're able to exhaust it.

## Proposal

The functions performing this operation are the following:

https://github.com/polarsource/polar/blob/7836113c892eb6a04d9160e5a3e66760e9d2fe4e/server/polar/order/service.py#L1771-L1794

https://github.com/polarsource/polar/blob/7836113c892eb6a04d9160e5a3e66760e9d2fe4e/server/polar/subscription/service.py#L2475-L2487

In this implementation, every orders/subscriptions are enqueued at once.

What we can do is to add an increasing delay between each iteration, so that other tasks get a chance to be processed in between. This is supported by Dramatiq: https://dramatiq.io/guide.html#scheduling-messages

rishi-raj-jain:
@frankie567 but wouldn't delaying this basically delay the time by which all the customers would receive the benefits?

frankie567:
No, because it only concerns the case of adding a benefit to existing customers

rishi-raj-jain:
Sounds good, thank you!

PR #8360, opened 2025-12-08
Fix delay by manually setting `eta`

pieterbeulque:
The fix in #8343 didn't actually work, since `delay` isn't an option on `message_with_options` unlike on `send_with_options` in Dramatiq. Thank you @Yopi for pointing that out.

This mimics the delay implementation from `send_with_options` to properly implement the delay. This is agnostic of Redis /  RabbitMQ / …

I have tested it locally with a 30-second delay and you can see them execute one-by-one now.

### DIFF
--- a/server/polar/worker/_enqueue.py
+++ b/server/polar/worker/_enqueue.py
@@ -87,6 +87,7 @@ class JobQueueManager:
             )
 
             # Handle delay: convert to eta and use delayed queue
+            # See https://github.com/Bogdanp/dramatiq/blob/aa91cdfcfa6d8ad957ca0afe900266617f2661f8/dramatiq/brokers/stub.py#L107-L116
             if delay is not None and delay > 0:
                 current_millis = int(time.time() * 1000)
                 eta = current_millis + delay


server/polar/worker/_enqueue.py
```diff
@@ -76,12 +78,23 @@ async def flush(self, broker: dramatiq.Broker, redis: Redis) -> None:
...
+
+            # Build base message without delay
             message = fn.message_with_options(
                 args=args,
                 kwargs=kwargs,
                 redis_message_id=redis_message_id,
-                delay=delay,
             )
+
+            # Handle delay: convert to eta and use delayed queue
```
Yopi, line 89: We could add a link to https://github.com/Bogdanp/dramatiq/blob/aa91cdfcfa6d8ad957ca0afe900266617f2661f8/dramatiq/brokers/stub.py#L107-L116 here just for completeness

pieterbeulque, line 89:
```suggestion
            # Handle delay: convert to eta and use delayed queue
            # See https://github.com/Bogdanp/dramatiq/blob/aa91cdfcfa6d8ad957ca0afe900266617f2661f8/dramatiq/brokers/stub.py#L107-L116
```

Yopi:
Closes [#8335](https://github.com/polarsource/polar/issues/8335)

PR #8379, opened 2025-12-09
customer meter: be smarter about our querying

Yopi:
- Instead of returning a list of events in each branch of the UNION query, which we then aggregate and sum,
we can let the branches return the values to be summed, so that postgres can avoid looping over the events.

- Since the credit events are all system events, we don't need to do the union all branching. Instead we can
filter out only system events

### DIFF
--- a/server/tests/customer_meter/test_service.py
+++ b/server/tests/customer_meter/test_service.py
@@ -653,6 +653,135 @@ class TestUpdateCustomerMeter:
         assert customer_meter.balance == Decimal(10)
         assert updated is True
 
+    async def test_unique_aggregation_with_external_id(
+        self,
+        save_fixture: SaveFixture,
+        session: AsyncSession,
+        locker: Locker,
+        customer_with_external_id: Customer,
+        organization: Organization,
+    ) -> None:
+        from polar.meter.aggregation import UniqueAggregation
+
+        meter = await create_meter(
+            save_fixture,
+            name="Unique Users",
+            filter=Filter(
+                conjunction=FilterConjunction.and_,
+                clauses=[
+                    FilterClause(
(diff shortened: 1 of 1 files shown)


server/polar/customer_meter/service.py
```diff
@@ -331,5 +314,115 @@ def _build_latest_event_statement(
             .limit(1)
         )
 
+    async def _get_usage_quantity(
```
psincraian, line 317: This is a smart approach! 👏

server/polar/customer_meter/service.py
```diff
@@ -331,5 +314,115 @@ def _build_latest_event_statement(
...
+        self, session: AsyncSession, customer: Customer, meter: Meter
+    ) -> float:
+        """
+        Get the aggregated usage quantity for a customer's meter.
+
+        This method avoids the expensive nested loop that occurs when fetching
+        IDs first and then re-querying to aggregate. For summable aggregations
+        (COUNT, SUM), we aggregate each branch and sum the results. For
+        non-summable aggregations (MAX, MIN, AVG, UNIQUE), we aggregate over
```
pieterbeulque, line 326: For min & max you can still get the min / max for each branch and min/max these two return values again, right?

pieterbeulque: Not sure if that'd actually make a difference though.

Yopi: Yeah, I think a question is if that would get optimized by the query planner no matter if we do it manually or not

psincraian:
## Auto code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #8387, opened 2025-12-09
Events: Migrate to cursor pagination

Yopi:
Adds a new concept of CursorPagination which has a next page instead of a count. This allows you to do a previous / next type pagination. We no longer have a max page count, or a total number of events.

This should speed up events heavy pages with 100-1000x.

* Screenshot 2025-12-09 at 14 50 39  
* Screenshot 2025-12-09 at 14 50 42  
* Screenshot 2025-12-09 at 14 50 50  
* Screenshot 2025-12-09 at 14 50 53  
 
  

**Updated**
Screenshot 2025-12-11 at 18 12 08
Screenshot 2025-12-11 at 18 12 19
Screenshot 2025-12-11 at 18 14 28

### DIFF
--- a/clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/analytics/events/EventsPage.tsx
+++ b/clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/analytics/events/EventsPage.tsx
@@ -8,13 +8,12 @@ import { DashboardBody } from '@/components/Layout/DashboardLayout'
 import DateRangePicker from '@/components/Metrics/DateRangePicker'
 import { Modal } from '@/components/Modal'
 import { useModal } from '@/components/Modal/useModal'
-import Pagination from '@/components/Pagination/Pagination'
-import { useEventNames, useEvents } from '@/hooks/queries/events'
+import { useEventNames, useInfiniteEvents } from '@/hooks/queries/events'
 import useDebounce from '@/utils/useDebounce'
 import AddOutlined from '@mui/icons-material/AddOutlined'
 import RefreshOutlined from '@mui/icons-material/RefreshOutlined'
 import Search from '@mui/icons-material/Search'
-import { operations, schemas } from '@polar-sh/client'
+import { schemas } from '@polar-sh/client'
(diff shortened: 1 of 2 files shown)


pieterbeulque:
I understand where you're coming from but I do think it's a slight UX regression.

For more tabular views, I definitely think it works, and we can add a _Showing first XYZ items - click to load more_. Coincidentally, this is the current pagination for the span detail page too. That way you can just append more and more rows to the table until all data is shown (i.e. `has_next_page` is null).

With page-based pagination `[←] Page N [→]` it's less good, in my opinion.

We should definitely consider this if the performance improvements are this significant, but I'd then also move away from `[←] Page N [→]` navigation all together (and thus also removing pages from the querystring). This then implies that a "page" is meaningless (and shouldn't be directly linkable), which for an event log is something I can live with.

Yopi:
> I understand where you're coming from but I do think it's a slight UX regression.

Haha, yes it is. I added you two because it's not great, but we can't count the events anymore unfortunately. That's why I added you two because I wanted some feedback on where we should take this.

I entertained the idea of having a break point to avoid regressing this for everyone (e.g. < 10 000 events, you get pages, and above you get this), but that also didn't sound great.

> I'd then also move away from [←] Page N [→] navigation all together (and thus also removing pages from the querystring). This then implies that a "page" is meaningless (and shouldn't be directly linkable), which for an event log is something I can live with.

Is it because the page-concept feels flawed without the ability of going to the last page? Personally I like it when I can actually jump to a page to try to get somewhere faster. A bit similar to GitHub's commit listing API, but I can understandthat it feels like we are exposing a flaw in the structure to give some but not all of the functionality.

Yopi:
I split out the count query here: https://github.com/polarsource/polar/pull/8389, and it looks like it made it at least somewhat bearable. Still too slow though!

pieterbeulque:
> actually jump to a page to try to get somewhere faster

But you're just blindly guessing to land somewhere hopefully closer to what you're looking for, while in fact you should use filters to narrow down the view to find what you're after. e.g. you shouldn't list "All time" events and then jump to page 7 to land at December 3, you should just… filter for December 3. No?

PR #8414, opened 2025-12-10
Skip archived meter calculation

pieterbeulque:
This skips calculation of archived meters and then enqueues all meters to be recalculated when a meter is unarchived.

This will introduce a big burst of customer meter updates though, so I'm not very confident if we should ship that part or make it more refined.

### DIFF
--- a/server/polar/meter/service.py
+++ b/server/polar/meter/service.py
@@ -21,6 +21,7 @@ from sqlalchemy.orm import joinedload
 from polar.auth.models import AuthSubject, Organization, User
 from polar.billing_entry.repository import BillingEntryRepository
 from polar.config import settings
+from polar.customer.repository import CustomerRepository
 from polar.event.repository import EventRepository
 from polar.exceptions import PolarError, PolarRequestValidationError, ValidationError
 from polar.kit.metadata import MetadataQuery, apply_metadata_clause, get_metadata_clause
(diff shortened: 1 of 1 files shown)


server/polar/meter/service.py
```diff
@@ -249,7 +249,20 @@ async def archive(self, session: AsyncSession, meter: Meter) -> Meter:
...
+        # Queue customer meter updates for all customers in the organization
+        from polar.customer.repository import CustomerRepository
+
+        customer_repository = CustomerRepository.from_session(session)
+        statement = customer_repository.get_base_statement().where(
+            Customer.organization_id == meter.organization_id
+        )
+        customers = await customer_repository.get_all(statement)
+        if customers:
+            await customer_repository.touch_meters(customers)
```
pieterbeulque: This is a bit of a shotgun blast…

server/polar/meter/service.py
```diff
@@ -249,7 +249,20 @@ async def archive(self, session: AsyncSession, meter: Meter) -> Meter:
 
     async def unarchive(self, session: AsyncSession, meter: Meter) -> Meter:
         repository = MeterRepository.from_session(session)
-        return await repository.update(meter, update_dict={"archived_at": None})
+        meter = await repository.update(meter, update_dict={"archived_at": None})
+
+        # Queue customer meter updates for all customers in the organization
+        from polar.customer.repository import CustomerRepository
```
Yopi: Should this be top of file?

server/polar/meter/service.py
Yopi: I guess it's fine / better than what we have today. Possibly that we could be a bit smarter here, but probably not necessary

PR #8432, opened 2025-12-10
server: use alembic-utils to manage Postgres functions and triggers

frankie567:
- server: use alembic-utils to manage Postgres functions and triggers


frankie567:
@Yopi Found `alembic-utils` to help manage PostgreSQL functions/triggers migrations. I think it'll be helpful :) Took the liberty to move the Customer Short ID generator function inside the model with this syntax.

PR #8433, opened 2025-12-10
Simplify and incrementally count meters

Yopi:
Sum, count, min and max are all possible to do incrementally,
this would save us from needing to recompute the entire meter for
those types of meters.

However, if we have two events that get the same ingested_at, we will start miss one of them.

### DIFF
--- a/server/polar/customer_meter/service.py
+++ b/server/polar/customer_meter/service.py
@@ -17,6 +17,7 @@ from polar.kit.utils import utc_now
 from polar.locker import Locker
 from polar.meter.aggregation import (
     AggregationFunction,
+    CountAggregation,
     PropertyAggregation,
     UniqueAggregation,
 )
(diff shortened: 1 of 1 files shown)


server/polar/customer_meter/service.py
```diff
@@ -404,6 +453,17 @@ def _get_outer_aggregation(self, meter: Meter, column: Any) -> Any:
...
+    def _combine_incremental(
+        self, meter: Meter, current: float, delta: float
+    ) -> Decimal:
+        if isinstance(meter.aggregation, PropertyAggregation):
+            match meter.aggregation.func:
+                case AggregationFunction.max:
+                    return Decimal(max(current, delta))
+                case AggregationFunction.min:
+                    return Decimal(min(current, delta))
+        return Decimal(current + delta)
```
pieterbeulque: It may be more clear to use explicit `case` for `.sum` and `.count` and raise an error if we try to do this for avg / unique to surface broken code paths.

Yopi: Good idea, I made it into an exhaustive match instead

PR #8446, opened 2025-12-11
rate-limit: Call the middleware twice (before and after auth)

Yopi:
This will rate limit all requests by IP address, and if they authenticate we will also rate limit them by the auth subject.

If they fail authentication we abort the middleware execution order.

### DIFF
--- a/server/polar/app.py
+++ b/server/polar/app.py
@@ -172,6 +172,7 @@ def create_app() -> FastAPI:
     if not settings.is_testing():
         app.add_middleware(rate_limit.get_middleware)
         app.add_middleware(AuthSubjectMiddleware)
+        app.add_middleware(rate_limit.get_middleware)
         app.add_middleware(FlushEnqueuedWorkerJobsMiddleware)
         app.add_middleware(AsyncSessionMiddleware)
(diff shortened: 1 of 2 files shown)


frankie567:
I'm not sure I get the problem 🤔 `AuthSubjectMiddleware` always runs before the rate limit middleware (reverse order), so if the user is not authenticated, we'll have a `AuthSubject[Anonymous]` in the scope.

Yopi:
> I'm not sure I get the problem 🤔 `AuthSubjectMiddleware` always runs before the rate limit middleware (reverse order), so if the user is not authenticated, we'll have a `AuthSubject[Anonymous]` in the scope.

That's exactly the problem. Since `AuthSubjectMiddleware` runs first, if you supply an invalid credential you will hit

```
        try:
            auth_subject = await get_auth_subject(request, session)
        except OAuth2Error as e:
            response = await oauth2_error_exception_handler(request, e)
            return await response(scope, receive, send)
```

Which will never trigger the `RateLimitMiddleware`. Additionally, we will always run the 4 DB queries no matter how many requests you make. The aim of this change is to rate limit anonymous users more quickly to avoid them draining resources.

PR #8453, opened 2025-12-11
order-items: Add order id index on order items

Yopi:
To speed up the order -> order_item relationship
given the fact that we are eagerly loading them.

### DIFF
--- a/server/migrations/versions/2025-12-11-1051_add_index_to_order_items_order_id.py
+++ b/server/migrations/versions/2025-12-11-1051_add_index_to_order_items_order_id.py
@@ -1,0 +1,32 @@
+"""add index to order_items.order_id
+
+Revision ID: aa7611c80a41
+Revises: cb7c757fe36e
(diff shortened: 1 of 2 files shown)


server/migrations/versions/2025-12-11-1051_add_index_to_order_items_order_id.py
```diff
@@ -0,0 +1,27 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "aa7611c80a41"
+down_revision = "cb7c757fe36e"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
+    op.create_index(
```
frankie567: Might be worth to create it concurrently to avoid locks 😊

Yopi: Yes!

Issue #8484, opened 2025-12-11
Product description images broken with blob storage URLs


rishi-raj-jain:
> from [the ticket](https://app.plain.com/workspace/w_01JE9TRRX9KT61D8P2CH77XDQM/thread/th_01KC6BH8Y9YTK3GYA3JHEY5N1H/)

User reported that markdown image syntax fails to display images in product long descriptions when using Vercel blob storage URLs.

<img width="3024" height="1714" alt="Image" src="https://github.com/user-attachments/assets/f09d8e73-b0ad-4a44-8cfc-231a6d9184eb" />

The markdown syntax used:
```
![](https://y4pdgnepgswqffpt.public.blob.vercel-storage.com/components/description-image-k5u8v4wm-mAL0h04Iy8fKZasbZIJksALzgq241r)
```

Image appears as broken instead of displaying correctly. Based on the screenshot, there are Content Security Policy violations that may be blocking the image from loading.

Error shown in console indicates CSP directive `img-src 'self' blob: data:` violations, suggesting the external blob storage domain is not allowlisted.

<sub>[Untitled Canvas](https://app.stilla.ai/m/memo_01kc7hwcmwe6g8bmg3ar0etyx2). Ask @stilla for more context.</sub>

PR #8485, opened 2025-12-11
Update next.config.mjs

rishi-raj-jain:
fix https://github.com/polarsource/polar/issues/8484


rishi-raj-jain:
@Yopi can you take a look at this and confirm if this is the right way to allow external images being displayed on the checkouts?

emilwidlund:
Not super comfortable with this. 

Anyone could upload anything to a public Vercel bucket, and this would explicitly whitelist it. 

We should only allow & whitelist our own Polar buckets.

PR #8499, opened 2025-12-12
auth: Validate token checksums

Yopi:
Log an error if the checksum is not matching even though the token was valid.


frankie567:
@Yopi I think we can discard that for now :)

PR #8506, opened 2025-12-12
Add iOS widget

sebastianekstrom:
### 📋 Summary

This PR will add an iOS widget based on the `@bacons/apple-targets` package.

| Dark mode | Light mode | Lock screen |
|--------|--------|--------|
| Simulator Screenshot - iPhone 17 #15 - 2025-12-12 at 13 46 57 | Simulator Screenshot - iPhone 17 #15 - 2025-12-12 at 13 47 02 | Simulator Screenshot - iPhone 17 #15 - 2025-12-12 at 13 47 11 |


clients/apps/app/app/_layout.tsx
```diff
@@ -61,6 +63,14 @@ export default Sentry.wrap(function RootLayout() {
     InstrumentSerif_400Regular_Italic,
   })
 
+  useEffect(() => {
+    AppState.addEventListener('change', (state) => {
+      if (state === 'background') {
+        ExtensionStorage.reloadWidget()
```
sebastianekstrom, line 71: According to the [docs](https://developer.apple.com/documentation/widgetkit/keeping-a-widget-up-to-date):

> A widget’s budget applies to a 24-hour period. WidgetKit tunes the 24-hour window to the user’s daily usage pattern, which means the daily budget doesn’t necessarily reset at exactly midnight. For a widget the user frequently views, a daily budget typically includes from 40 to 70 refreshes. This rate roughly translates to widget reloads every 15 to 60 minutes, but it’s common for these intervals to vary due to the many factors involved.

This `reloadWidget()` call forces the widget to reload as soon as the app enters the background to ensure it's up-to-date.

clients/apps/app/app.config.js
```diff
@@ -0,0 +1,102 @@
+const IS_WIDGET_BUILD = process.env.EXPO_WIDGET_BUILD === "1";
```
sebastianekstrom: Our Sentry plugin caused a few issues so I had to convert our `app.json` to JavaScript and conditionally add the Sentry plugin.

PR #8531, opened 2025-12-13
worker: MemoryTraceMiddleware using tracemalloc

birkjernstrom:


server/polar/worker/_memory.py
```diff
@@ -70,3 +74,83 @@ def after_process_message(
...
+            threshold=self.threshold,
+        )
+
+    def before_process_message(
+        self, broker: dramatiq.Broker, message: dramatiq.Message[Any]
+    ) -> None:
+        if not self.enabled:
+            return
+
+        self._before_message = tracemalloc.take_snapshot()
```
Yopi, line 110: I think we need to store these in a thread safe way, either in a thread local variable or on the message. E.g.

```
message.options['_memory_snapshot'] = tracemalloc.take_snapshot()
```

Alterantively in `__init__` we could do
```
self._state = threading.local()
```
and store it there

Yopi (approved):
I think this will be useful to get out

PR #8546, opened 2025-12-15
events: Add cursor_pagination query parameter

Yopi:
To get page-less pagination and instead just the next cursor


frankie567:
@Yopi Do you still need this?

Yopi:
Lets hold off on this, we might not need it with the other changes we are making.

PR #8583, opened 2025-12-16
try/catch checkForUpdateAsync()

sebastianekstrom:
## 📋 Summary

Simulators/emulators doesn't support `checkForUpdateAsync()`, so every time we do a pull-to-refresh on home it will throw an error which is annoying.


clients/apps/app/components/Orders/OrderRow.tsx
```diff
@@ -55,7 +55,7 @@ export const OrderRow = ({
             <Image
               source={{ uri: product?.medias?.[0]?.public_url }}
               style={{ width: '100%', height: '100%' }}
-              resizeMode="cover"
+              contentFit="cover"
```
sebastianekstrom, line 58: Drive-by fix, `resizeMode` is deprecated

PR #8596, opened 2025-12-16
system-events: Add balance.{order,refund,refund_reversal,dispute} events

Yopi:
These can be used to build balance metrics in the correct currency

### DIFF
--- a/server/polar/event/system.py
+++ b/server/polar/event/system.py
@@ -35,6 +35,7 @@ class SystemEvent(StrEnum):
     balance_refund = "balance.refund"
     balance_refund_reversal = "balance.refund_reversal"
     balance_dispute = "balance.dispute"
+    balance_dispute_reversal = "balance.dispute_reversal"
 
 
 SYSTEM_EVENT_LABELS: dict[str, str] = {
@@ -60,6 +61,7 @@ SYSTEM_EVENT_LABELS: dict[str, str] = {
     "balance.order": "Balance Order",
     "balance.refund": "Balance Refund",
     "balance.dispute": "Balance Dispute",
+    "balance.dispute_reversal": "Balance Dispute Reversal",
 }
 
 
@@ -424,6 +426,13 @@ class BalanceDisputeEvent(Event):
         user_metadata: Mapped[BalanceDisputeMetadata]  # type: ignore[assignment]
 
 
+class BalanceDisputeReversalEvent(Event):
+    if TYPE_CHECKING:
+        source: Mapped[Literal[EventSource.system]]
+        name: Mapped[Literal[SystemEvent.balance_dispute_reversal]]
+        user_metadata: Mapped[BalanceDisputeMetadata]  # type: ignore[assignment]
+
+
 @overload
 def build_system_event(
     name: Literal[SystemEvent.meter_credited],
@@ -616,6 +625,15 @@ def build_system_event(
 @overload
 def build_system_event(
     name: Literal[SystemEvent.balance_dispute],
+    customer: Customer,
+    organization: Organization,
+    metadata: BalanceDisputeMetadata,
+) -> Event: ...
+
+
+@overload
+def build_system_event(
+    name: Literal[SystemEvent.balance_dispute_reversal],
     customer: Customer,
     organization: Organization,
     metadata: BalanceDisputeMetadata,
--- a/server/polar/transaction/service/dispute.py
+++ b/server/polar/transaction/service/dispute.py
@@ -186,12 +186,10 @@ class DisputeTransactionService(BaseTransactionService):
         if dispute_reversal_transaction is not None:
             all_fees += dispute_reversal_fees
 
-        polar_fees: int = 0
         try:
-            fee_transactions = await self._create_dispute_fees_balances(
+            await self._create_dispute_fees_balances(
                 session, payment_transaction=payment_transaction, dispute_fees=all_fees
             )
-            polar_fees = sum(f[1].amount for f in fee_transactions)
         except NotBalancedPaymentTransaction:
             log.warning(
                 "Dispute fees balances could not be created for payment transaction",
@@ -220,7 +218,7 @@ class DisputeTransactionService(BaseTransactionService):
                     "tax_amount": settlement_tax_amount,
                     "tax_state": payment_transaction.tax_state,
                     "tax_country": payment_transaction.tax_country,
-                    "fee": polar_fees,
+                    "fee": sum(-fee.amount for fee in dispute_fees),
                 }
                 if order is not None:
                     metadata["order_id"] = str(order.id)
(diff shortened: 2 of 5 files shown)


server/polar/transaction/service/refund.py
```diff
@@ -116,6 +123,42 @@ async def create(self, session: AsyncSession, refund: Refund) -> Transaction:
...
+            if order is not None:
+                metadata["order_id"] = str(order.id)
+                metadata["product_id"] = str(order.product_id)
+                if order.subscription_id is not None:
+                    metadata["subscription_id"] = str(order.subscription_id)
+                refund_total = refund.amount + refund.tax_amount
+                order_total = order.total_amount
+                total_refunded = order.refunded_amount + order.refunded_tax_amount
+                metadata["is_partial"] = refund_total < order_total
+                metadata["completes_refund"] = total_refunded >= order_total
```
Yopi: Should this only be looking at `refunded_amount` and not include `refunded_tax_amount`?

server/polar/event/system.py
```diff
@@ -341,6 +347,64 @@ class CheckoutCreatedEvent(Event):
...
+    refund_id: str
+    order_id: NotRequired[str]
+    product_id: NotRequired[str]
+    subscription_id: NotRequired[str]
+    amount: int
+    currency: str
+    presentment_amount: NotRequired[int]
+    presentment_currency: NotRequired[str]
+    is_partial: NotRequired[bool]
+    completes_refund: NotRequired[bool]
```
frankie567: What does this one represent?

server/polar/order/service.py
```diff
@@ -1494,6 +1499,28 @@ async def create_order_balance(
...
+            "product_id": str(order.product_id),
+            "amount": payment_transaction.amount,
+            "currency": payment_transaction.currency,
+        }
+        if order.subscription_id is not None:
+            metadata["subscription_id"] = str(order.subscription_id)
+        if payment_transaction.presentment_amount is not None:
+            metadata["presentment_amount"] = payment_transaction.presentment_amount
+        if payment_transaction.presentment_currency is not None:
+            metadata["presentment_currency"] = payment_transaction.presentment_currency
```
frankie567: On a Transaction of type `payment`, `refund` and `dispute`, presentment amount and currency will be filled. So I think we can make this metadata mandatory. You can add an `assert` here to make the type checker happy.

server/polar/transaction/service/dispute.py
```diff
@@ -191,6 +199,36 @@ async def create_dispute(
 
         await session.flush()
 
+        customer = payment_transaction.payment_customer
+        organization = payment_transaction.payment_organization
```
frankie567: `payment_organization` is an old thing from the issue funding era that doesn't mean what you want here :) It'll probably be `None` in all cases.

I would recommend to load `customer.organization` instead.

server/polar/transaction/service/dispute.py
```diff
@@ -191,6 +199,36 @@ async def create_dispute(
...
+                metadata["order_id"] = str(order.id)
+                metadata["product_id"] = str(order.product_id)
+                if order.subscription_id is not None:
+                    metadata["subscription_id"] = str(order.subscription_id)
+            if dispute_transaction.presentment_amount is not None:
+                metadata["presentment_amount"] = dispute_transaction.presentment_amount
+            if dispute_transaction.presentment_currency is not None:
+                metadata["presentment_currency"] = (
+                    dispute_transaction.presentment_currency
+                )
```
frankie567, line 234: Same comment as above re: presentment being mandatory

server/polar/transaction/service/refund.py
```diff
@@ -116,6 +123,42 @@ async def create(self, session: AsyncSession, refund: Refund) -> Transaction:
             payment_transaction=payment_transaction,
             refund_amount=settlement_amount - settlement_tax_amount,
         )
+
+        customer = payment_transaction.payment_customer
+        organization = payment_transaction.payment_organization
```
frankie567: Same comment as above re: payment_organization

server/polar/transaction/service/refund.py
frankie567: Since you have an order at hand here, you can use `remaining_balance` property to quickly know if there is something left to refund.

server/polar/event/system.py
```diff
@@ -341,6 +347,64 @@ class CheckoutCreatedEvent(Event):
         user_metadata: Mapped[CheckoutCreatedMetadata]  # type: ignore[assignment]
 
 
+class BalanceOrderMetadata(TypedDict):
+    transaction_id: str
+    order_id: str
+    product_id: str
```
frankie567: `product_id` is nullable on `Order` (in preparation for Wallet top-ups), so it should probably be `NotRequired`.

frankie567 (changes requested):
A few things there and there, but that's the right approach :)

On another level, I'm wondering if we should add `tax_amount` in the events metadata as well: merchants indeed only care about the raw amount, but I'm wondering if we might need the tax amount in the future. That said, it can be a rabbit hole as it'll also require to add `tax_country`/`tax_state` to fully make sense 🤔

server/polar/event/system.py
Yopi: I was thinking that we would want is_partial, and if you refund multiple times it might complete the full refund. But as I was looking at this I was also thinking that maybe we should use the naming / concept from order status instead?

frankie567: I see 🤔 Yeah, it's not crystal clear like this. What if we simply output `refundable_amount`, i.e. the amount that can still be refunded (if 0, then the refund is complete).

Yopi: I like that!

Yopi:
> A few things there and there, but that's the right approach :)

Thank you for the good feedback!
 
> On another level, I'm wondering if we should add `tax_amount` in the events metadata as well: merchants indeed only care about the raw amount, but I'm wondering if we might need the tax amount in the future. That said, it can be a rabbit hole as it'll also require to add `tax_country`/`tax_state` to fully make sense 🤔

That's a good question. If we think that it would be useful we might as well add it now I guess.

I also realised that we do revert_refund transactions which we might want to have an event for. Either as a negative amount with the same refund event, or a new event.

server/polar/order/service.py
Yopi: I added assertions, but also got a bit worried so I wrapped these in a try for now. I am thinking that we can remove it later (as it shouldn't happen) but to mitigate the roll out here a bit.

frankie567:
> That's a good question. If we think that it would be useful we might as well add it now I guess.

I gotta feeling (_woohooo_) that it'll be useful in the future. So I would say let's add it (+tax country/state).

> I also realised that we do revert_refund transactions which we might want to have an event for. Either as a negative amount with the same refund event, or a new event.

Yeah let's do `balance.refund_reversal`. It's quite rarely used (6 transactions as of today!), but let's be complete.

Yopi:
> > That's a good question. If we think that it would be useful we might as well add it now I guess.
> 
> I gotta feeling (_woohooo_) that it'll be useful in the future. So I would say let's add it (+tax country/state).

Perfect, I'll add it.
> 
> > I also realised that we do revert_refund transactions which we might want to have an event for. Either as a negative amount with the same refund event, or a new event.
> 
> Yeah let's do `balance.refund_reversal`. It's quite rarely used (6 transactions as of today!), but let's be complete.

Should we do the same with the `balance.dispute_reversal` as well? 😅

frankie567:
> Should we do the same with the `balance.dispute_reversal` as well? 😅

So, the trick is that we only write the transaction when the dispute is closed (won or lost). If it's won, we write a Dispute + Dispute Reversal. If it's lost, just a Dispute. ~So I would be tempted to say that we only create an event `balance.dispute` when it's `lost` (because the merchant lost money) and be silent about won disputes.~

EDIT: Actually, if we want to be clear about fees (see below), we probably need to have `balance.dispute_reversal` as well (because fees are applied even in case of won dispute).

---

Now that I think of it, there is also the fees dimension 😅 In this implementation, we store the amounts before Polar fees. If we want to show graphs with "net" amounts, we probably also need to store `balance.fees` events ☺️☺️

Yopi:
I think this is good to go now!

PR #8640, opened 2025-12-18
Redesign benefits

sebastianekstrom:
## 📋 Summary

Redesign the benefits section to scale better for users with a lot of benefits.

## 🖼️ Screenshots/Recordings

Frame 44


server/polar/benefit/endpoints.py
```diff
@@ -51,6 +51,12 @@ async def list(
     type: MultipleQueryFilter[BenefitType] | None = Query(
         None, title="BenefitType Filter", description="Filter by benefit type."
     ),
+    id_in: MultipleQueryFilter[BenefitID] | None = Query(
```
psincraian: To filter by values we usually use the param name, so in this case `id` and you can pass a list.

Similar, for excluding we use `exclude_{param_name}`.

server/polar/benefit/service.py
```diff
@@ -33,6 +33,8 @@ async def list(
         *,
         type: Sequence[BenefitType] | None = None,
         organization_id: Sequence[uuid.UUID] | None = None,
+        id_in: Sequence[uuid.UUID] | None = None,
+        id_not_in: Sequence[uuid.UUID] | None = None,
```
psincraian, line 37: On the service layer, you can keep the same ones for now :)

PR #8650, opened 2025-12-18
portal: Settings to enable/disable certain features

birkjernstrom:
image

Introduces `customer_portal_settings` on organizations:
```json
{
    “usage”: { “show”: true },
    "subscription”: {
      "update_seats": true,
      "update_plan": false
    }
}
```
- [x] Toggle usage insights
- [x] Toggle ability to change seats
- [x] Toggle ability to update subscription plans  

Subscription updates was previously configured in `subscription_settings.allow_customer_updates`, but moving here. But it's being kept in the data model. Migration happens in code with backward compatibility. Once this PR is out, we can separately deploy clean-up/removal of the current attribute.


server/polar/models/organization.py
```diff
@@ -111,6 +112,29 @@ class OrganizationCustomerEmailSettings(TypedDict):
 }
 
 
+class CustomerPortalUsageSettings(TypedDict, total=False):
```
frankie567: I think we can drop `total=False` on each schema: with the migration, all existing orgs will have the settings filled.

server/polar/models/organization.py
```diff
@@ -314,7 +342,7 @@ def allow_multiple_subscriptions(self) -> bool:
 
     @property
     def allow_customer_updates(self) -> bool:
```
frankie567: Since it's now just an alias to `customer_portal_subscription_update_plan`, can we replace that property wherever it's used and drop it?

clients/apps/web/src/components/Settings/OrganizationCustomerPortalSettings.tsx
```diff
@@ -0,0 +1,153 @@
...
+  organization: schemas['Organization']
+}
+
+const OrganizationCustomerPortalSettings: React.FC<
+  OrganizationCustomerPortalSettingsProps
+> = ({ organization }) => {
+  const form = useForm<schemas['OrganizationCustomerPortalSettings']>({
+    defaultValues: {
+      usage: {
+        show: organization.customer_portal_settings?.usage?.show ?? true,
```
frankie567: Once you'll remove `total=False` on the Python schema (see comment above), we'll be able to remove lot of optional chaining operators.

frankie567 (changes requested):
Looks great, just a few typing nitpicks.

Still concerned as I mentioned in the original issue by the number of settings we introduce vs. having a global switch to disable the portal and let merchants implement their own portal.

PR #8678, opened 2025-12-22
Optimize benefit grant queuing

pieterbeulque:
When processing benefit grants (e.g., when a product is updated with new benefits), skip enqueueing jobs for:

1. Benefits that are already granted - no need to re-grant
2. Benefits with errors (BenefitActionRequiredError) - these require customer action (like connecting OAuth) and should only retry when that action is taken, not on every product update

This significantly reduces queue pressure for merchants with large customer bases and many benefits per product. For example, a product with 92 benefits and 516 customers previously queued 47,472 jobs on every product update; now it only queues jobs for new/changed benefits.

Also optimizes revoke task to only process benefits that are actually granted, avoiding unnecessary revoke attempts.

### DIFF
--- a/server/polar/benefit/grant/service.py
+++ b/server/polar/benefit/grant/service.py
@@ -374,7 +374,7 @@ class BenefitGrantService(ResourceServiceReader[BenefitGrant]):
         errored_benefit_ids = {
             g.benefit_id
             for g in existing_grants
-            if g.error and g.error.get("type") == "BenefitActionRequiredError"
+            if g.error and g.error.get("type") == BenefitActionRequiredError.__name__
         }
 
         if task == "grant":


server/polar/benefit/grant/service.py
```diff
@@ -364,12 +364,32 @@ async def enqueue_benefits_grants(
...
-        for benefit in product.benefits:
+        # Get existing grants for this customer and scope to avoid redundant jobs
+        existing_grants = await repository.list_by_customer_and_scope(customer, **scope)
+        granted_benefit_ids = {g.benefit_id for g in existing_grants if g.is_granted}
+        # Don't retry grants that failed due to required customer action -
+        # they should only be retried when the customer takes that action
+        errored_benefit_ids = {
+            g.benefit_id
+            for g in existing_grants
+            if g.error and g.error.get("type") == "BenefitActionRequiredError"
```
psincraian: Can we do something fancier with typings ? Maybe using the class name?

server/polar/benefit/grant/service.py
```diff
@@ -364,12 +364,32 @@ async def enqueue_benefits_grants(
...
+        }
+
+        if task == "grant":
+            benefits_to_process = [
+                b
+                for b in product.benefits
+                if b.id not in granted_benefit_ids and b.id not in errored_benefit_ids
+            ]
+        else:
+            # Only revoke benefits that are actually granted
```
psincraian, line 387: what about the ones that are errored? Can the customer claim those ones later?

pieterbeulque: Yes, they can never resolve through this path. The customer has to connect their GitHub account in the customer portal to resolve the error and that will allow them to claim the repositories.

PR #8681, opened 2025-12-22
Avoid unnecessary customer meter update triggers

pieterbeulque:
Prevents unnecessary `customer_meter.update_customer` tasks for orgs without meters and non-meter-related system events (like `benefit_granted`, which is how I discovered this)

Changes:

- Only call `touch_meters()` for user events or meter-related system events (`meter_credited`, `meter_reset`)
- Check if organizations have active meters before touching

---

It actually caught me by surprise that we call `ingested` on system events at all. Is this by design or did this slip in unnoticed? If I read the code for `ingested()`, there's actually nothing in there that has to do with system events.

I was wondering if instead of this fix, we could also just do a simple check on `event.source == "user"` before calling the `ingested` method.

Edit: I found that this change was intentional in https://github.com/polarsource/polar/commit/c2525565a1fb5d0a61686c5b243106d85ad503b9. From that lens, I think this PR is worth pursuing, as we've added some things in this `ingested` callback that is really user-event specific and we should opt-out of it when the created system events are irrelevant to that logic.

### DIFF
--- a/server/tests/event/test_service.py
+++ b/server/tests/event/test_service.py
@@ -23,7 +23,11 @@ from polar.exceptions import PolarRequestValidationError
 from polar.kit.pagination import PaginationParams
 from polar.kit.time_queries import TimeInterval
 from polar.kit.utils import utc_now
-from polar.meter.aggregation import AggregationFunction, PropertyAggregation
+from polar.meter.aggregation import (
+    AggregationFunction,
+    CountAggregation,
+    PropertyAggregation,
+)
 from polar.meter.filter import Filter, FilterClause, FilterConjunction, FilterOperator
 from polar.models import (
     Customer,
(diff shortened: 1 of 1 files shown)


server/polar/event/service.py
```diff
@@ -723,20 +725,54 @@ async def ingested(
...
+            orgs_with_meters_statement = (
+                select(Meter.organization_id)
+                .where(
+                    Meter.organization_id.in_(organization_ids),
+                    Meter.archived_at.is_(None),
+                )
+                .distinct()
+            )
+            result = await session.execute(orgs_with_meters_statement)
+            orgs_with_meters = set(result.scalars().all())
```
pieterbeulque, line 763: Maybe this is all getting a bit too granular here and I'm installing booby traps left & right 😬

PR #8684, opened 2025-12-22
Only send confirmation email once order is paid

pieterbeulque:
Move confirmation email sending from `_on_order_created` to `_on_order_paid`.

Previously, when a subscription cycled, the "subscription renewed" email was sent immediately when the order was created (in pending state), before payment was attempted. This caused customers with failed payments to receive both a "renewed" email and a payment failure email.

Before I fix the tests, I'm well aware that this is a bit of a shift in how subscriptions work (i.e. optimistic renewing versus taking the money first) so I'm not sure if this is a change we actually _want_ to make.

[Slack thread](https://polar-olw4768.slack.com/archives/C096F9P7E0P/p1766411700590599) for additional context.

### DIFF
--- a/server/tests/order/test_service.py
+++ b/server/tests/order/test_service.py
@@ -2074,10 +2074,11 @@ class TestHandlePayment:
         assert updated_order.status == OrderStatus.paid
         assert updated_order.tax_transaction_processor_id == "tax_txn_456"
 
-        # Verify enqueue_job was called to balance the order
-        enqueue_job_mock.assert_called_once_with(
+        # Verify enqueue_job was called to balance the order and send confirmation email
+        enqueue_job_mock.assert_any_call(
             "order.balance", order_id=order.id, charge_id="stripe_payment_123"
         )
+        enqueue_job_mock.assert_any_call("order.confirmation_email", order.id)
 
         # Verify stripe tax transaction was created
         stripe_service_mock.create_tax_transaction.assert_called_once_with(


server/polar/order/service.py
```diff
@@ -1566,7 +1566,8 @@ async def send_webhook(
         await webhook_service.send(session, organization, event_type, order)
 
     async def _on_order_created(self, session: AsyncSession, order: Order) -> None:
-        enqueue_job("order.confirmation_email", order.id)
+        # Confirmation email is sent in _on_order_paid when payment completes
+        # (for both immediately-paid orders and orders that become paid later)
```
frankie567: Verbose LLM comment 😄

server/polar/order/service.py
```diff
@@ -1596,6 +1597,10 @@ async def _on_order_updated(
     async def _on_order_paid(self, session: AsyncSession, order: Order) -> None:
         assert order.paid
 
+        # Send confirmation email when order becomes paid
+        # (for orders created as pending, e.g. subscription cycles)
```
frankie567: Verbose LLM comment 😄

pieterbeulque:
I'm still on the fence if we should merge it or not because it doesn't really match the revocation behavior (which is we renew when we create the order, not when the order is paid–and then we revoke once the payment never succeeds).

pieterbeulque:
Picking it back up, I don't think we should do it because user notifications should match the actual behavior… So this should consider the grace period and all that.

PR #8690, opened 2025-12-23
Disable failing webhooks faster

pieterbeulque:
In main, we don't exclude pending webhook events (`succeeded is None`) from the check on consecutive failures, which makes it so that in burst of webhook triggers, some webhooks collect up to 400-500 events and a multiple of that in delivery attempts before there's a long enough pause that the 10 latest events are all effectively failed without any new ones in a pending state.

This fixes it by looking at the 10 latest completed events (succeeded or failed) which I think is correct behavior and was an initial oversight in implementing the auto-disable. We also mark pending webhooks as skipped to make sure they won't be (re-)tried.

### DIFF
--- a/server/polar/webhook/repository.py
+++ b/server/polar/webhook/repository.py
@@ -72,6 +72,21 @@ class WebhookEventRepository(
         )
         return await self.get_all(statement)
 
+    async def get_pending_by_endpoint(
+        self, endpoint_id: UUID
+    ) -> Sequence[WebhookEvent]:
+        """
+        Get all pending events for an endpoint.
+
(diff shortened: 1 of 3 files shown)


pieterbeulque:
One thing that I don't like about this change is that the 11th pending webhook can succeed and then also be marked as skipped, which is a weird state no webhook delivery should end up in.

frankie567:
@pieterbeulque Good catch on the check, I think it's indeed the right way.

However, marking the pending events as skipped is a bit redundant here since it's done by the task itself anyway:

https://github.com/polarsource/polar/blob/main/server/polar/webhook/tasks.py#L60-64

PR #8702, opened 2025-12-26
Prevent billing address changes on VAT invoices

pieterbeulque:
VAT is calculated based on the billing address at order creation time. Changing the country/state after an order has been paid would result in an inconsistent invoice where the address doesn't match the VAT charged.

### DIFF
--- a/server/polar/customer_portal/endpoints/order.py
+++ b/server/polar/customer_portal/endpoints/order.py
@@ -12,7 +12,6 @@ from polar.models.product import ProductBillingType
 from polar.openapi import APITag
 from polar.order.schemas import OrderID
 from polar.order.service import (
-    InvoiceBillingAddressUpdateError,
     MissingInvoiceBillingDetails,
     NotPaidOrder,
     PaymentAlreadyInProgress,
@@ -114,13 +113,7 @@ async def get(
     "/{id}",
     summary="Update Order",
     response_model=CustomerOrder,
-    responses={
-        404: OrderNotFound,
-        422: {
-            "description": "Cannot update billing address country/state after order is paid.",
-            "model": InvoiceBillingAddressUpdateError.schema(),
-        },
-    },
+    responses={404: OrderNotFound},
 )
 async def update(
     id: OrderID,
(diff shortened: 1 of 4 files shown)


server/polar/customer_portal/endpoints/order.py
```diff
@@ -113,7 +114,13 @@ async def get(
     "/{id}",
     summary="Update Order",
     response_model=CustomerOrder,
-    responses={404: OrderNotFound},
+    responses={
+        404: OrderNotFound,
+        422: {
```
frankie567: No need to document 422 error codes, it's out of the box already

server/polar/order/endpoints.py
```diff
@@ -173,7 +177,13 @@ async def get(
     "/{id}",
     summary="Update Order",
     response_model=OrderSchema,
-    responses={404: OrderNotFound},
+    responses={
+        404: OrderNotFound,
+        422: {
```
frankie567: No need to document 422 error codes, it's out of the box already

server/polar/order/service.py
```diff
@@ -339,6 +349,21 @@ async def update(
...
+            existing_address = order.billing_address
+
+            # Check if country or state is being changed
+            new_country = new_address.country if new_address else None
+            new_state = new_address.state if new_address else None
+            existing_country = existing_address.get("country") if existing_address else None
+            existing_state = existing_address.get("state") if existing_address else None
+
+            if new_country != existing_country or new_state != existing_state:
+                raise InvoiceBillingAddressUpdateError(order)
```
frankie567: Rather than raising a custom error, we should directly raise a `PolarRequestValidationError`, like here for example:

https://github.com/polarsource/polar/blob/93404fadeaa712ac2ceecc64e8e4f43bfbf1a937/server/polar/product/service.py#L430-L439

frankie567 (changes requested):
Good catch, just a few changes needed on the im plementation itself

pieterbeulque:
Superseded by #8892

frankie567:
Wow sorry, @pieterbeulque, completely forgot about this one 🙈

pieterbeulque:
hehe no all good 🫶  this had merge conflicts anyway so better to do it from scratch

PR #8737, opened 2026-01-01
fix(refund): use quoted forward references for TYPE_CHECKING imports

### LINKED ISSUE #8736: Bug: NameError in refund.py prevents database migrations and server startup
### Description

After pulling the latest changes from the `main` branch, running database migrations (`uv run task db_migrate`) or starting the API server fails with a `NameError` in [server/polar/models/refund.py](cci:7://file:///home/priyanshu/repos/polar/server/polar/models/refund.py:0:0-0:0). The issue was introduced in commit `8fed8aae6d` ("server/refund: add Refund.payment and refactor service").

### Current Behavior

Running any command that imports the models (e.g., `uv run task db_migrate` or `uv run task api`) fails with:
NameError: name 'Order' is not defined

### Expected Behavior

Database migrations and server startup should work without errors. The fix is to wrap the type annotations in quotes to make them forward references:

### Environment:

Operating System: Linux (Ubuntu)
Browser: brave
Python Version: 3.13

priyanshuharshbodhi1:
The Order and Payment types are imported under TYPE_CHECKING but were used without quotes in the from_stripe method signature, causing a NameError at runtime during module import.

This prevented database migrations and server startup.

fixes #8736

### DIFF
--- a/server/polar/models/refund.py
+++ b/server/polar/models/refund.py
@@ -240,7 +240,7 @@ class Refund(MetadataMixin, RecordModel):
 
     @classmethod
     def from_stripe(
-        cls, stripe_refund: stripe_lib.Refund, order: Order, payment: Payment
+        cls, stripe_refund: stripe_lib.Refund, order: "Order", payment: "Payment"
     ) -> Self:
         amount, tax_amount = order.calculate_refunded_tax_from_total(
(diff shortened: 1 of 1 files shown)


frankie567:
Great catch, thanks 🙏

PR #8764, opened 2026-01-05
fe: Add logout button to onboarding

### LINKED ISSUE #8741: Feature request: Allow users to log out or delete their account
> from [the ticket](https://app.plain.com/workspace/w_01JE9TRRX9KT61D8P2CH77XDQM/thread/th_01KDZ8XG0ECH3Y1NDSP5CFXPG1/)

Users should be able to log out of their account or delete their account entirely when on the onboarding:

Image

Yopi:
Fix #8741

If the user does not have an existing org.

### DIFF
--- a/clients/apps/web/src/components/Onboarding/OrganizationStep.tsx
+++ b/clients/apps/web/src/components/Onboarding/OrganizationStep.tsx
@@ -5,6 +5,7 @@ import { useAuth } from '@/hooks'
 import { usePostHog } from '@/hooks/posthog'
 import { useCreateOrganization } from '@/hooks/queries'
 import { setValidationErrors } from '@/utils/api/errors'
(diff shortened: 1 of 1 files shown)


clients/apps/web/src/components/Onboarding/OrganizationStep.tsx
```diff
@@ -351,6 +352,17 @@ export const OrganizationStep = ({
                     </Button>
                   </Link>
                 )}
+                {!hasExistingOrg && (
```
pieterbeulque: Can this be a `<Link>` like above? Maybe with an explicit `prefetch={false}` even though it should not do that for external URL's.

`onClick` feels dirty.

Yopi: Agree!

PR #8770, opened 2026-01-05
Fix avg & unique by reverting to old cartesian join

pieterbeulque:
Just to test. Follow-up to #8769

To be correct, I think we should do something like this, but could run some numbers to assess the impact on performance etc?

### DIFF
--- a/server/polar/meter/service.py
+++ b/server/polar/meter/service.py
@@ -346,16 +346,24 @@ class MeterService:
         day_column = interval.sql_date_trunc(Event.timestamp)
         truncated_timestamp = interval.sql_date_trunc(timestamp_column)
 
-        # Determine the appropriate SQL function for the running total calculation.
-        # For summable aggregations (count, sum), we can sum the daily values.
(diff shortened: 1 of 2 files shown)


frankie567:
@pieterbeulque Should we discard this?

pieterbeulque:
… well. It's not calculating correctly in production right now, but merging this would be a performance regression. @Yopi - since you "broke" it / fixed the performance issue (half empty / half full kind of way to look at it, haha), what do you think?

Not pointing fingers 😊 

Actually, not sure how it'd go together with the other optimizations made in main since this PR.

Yopi:
I'm also not sure. Maybe we should merge this just to get it working correctly. Given the other changes we've made it might be fine out of a performance improvement perspective.

pieterbeulque:
@Yopi - I rebased this against main so maybe we can merge it now? I think the issue is still in main.

Yopi:
> @Yopi - I rebased this against main so maybe we can merge it now? I think the issue is still in main.

Lets do it

PR #8771, opened 2026-01-06
Restore Fee Credits feature

emilwidlund:
This commit restores the work for Fee Credits, previously reverted because a migration issue.


server/migrations/versions/2026-01-07-1634_add_account_credits.py
```diff
@@ -0,0 +1,91 @@
+"""Add new account_credits table and ability to set fee_credit on campaigns
```
Yopi, line 1: Rename these three to be more accurate in date of merge (and sorting) and change the down revision on this one to the latest (3d212567b9a6)

Yopi (commented):
I think it could be good to de-risk the PR to split the three migrations out into its own PR that you can merge beforehand. So that you decouple changing the DB and using the DB change

server/migrations/versions/2026-01-07-1634_add_account_credits.py
emilwidlund: Ah! Fixed!

PR #8772, opened 2026-01-06
Speed up scope approval screen in OAuth flow

pieterbeulque:
When testing MCP, I noticed this page renders really slow, and it's a server-side render after selecting the organization, which causes you to think it's broken and clicking multiple times, basically poor UX all over.

Not reading the full OpenAPI spec should speed that up significantly (it has an average 3.5 second load time if I access it directly, generating 833kB of JSON).

If you'd rather not include the scope display names in the response, I can whip up a quick endpoint that outputs them on `oauth/scope-names` or something instead.


frankie567:
Fine by me :)

Issue #8714, opened 2025-12-29
Apply discounts to existing subscriptions


pieterbeulque:
When a "once" discount is applied to an existing active subscription mid-billing-cycle, the discount is immediately considered expired and will never be applied to any invoice.

```python
# server/polar/models/discount.py:144-147

def is_repetition_expired(self, started_at, current_period_start, trial_ended=False):
    if self.duration == DiscountDuration.once:
        return not trial_ended
```

For repeating discounts, it's influenced by the subscription's start date too, so it also won't work as expected. Applying a 3 month 50% off discount in the fifth month of the subscription will just consider the discount expired.

You'd expect it to apply to the next invoice, which is useful for cases like customer support granting next month for free as compensation for an issue, or trying to prevent churn with next X charges 80% off.

pieterbeulque:
Fixed in #8826 and #8827

PR #8776, opened 2026-01-06
Apply discounts to existing subscriptions

pieterbeulque:
Issue #8714 

The original issue was that discounts applied to existing subscriptions mid-billing-cycle were immediately considered expired and never actually applied. The root cause was that `is_repetition_expired()` was using the subscription start date to calculate discount expiration, which doesn't work when a discount is added after the subscription started.

The fix involved tracking when a discount is first actually applied to a billing entry. For this I added a new `discount_applied_at` field on the `Subscription` model. This field is null when a discount is added but hasn't been used yet, and gets set during `cycle()` when the discount is first applied to a billing entry. The `is_repetition_expired()` method now uses this field to determine if "once" discounts have been used up and when "repeating" discount durations should start counting. This PR also fixes `subscription.amount` to only reflect the discounted price after the discount is actually applied, not immediately when added mid-cycle.

When a discount with `max_redemptions` is claimed but never actually used (subscription cancelled or discount removed before billing), the redemption slot was being wasted. In this PR I address this by soft-deleting the `DiscountRedemption` when a discount is removed before use or when a subscription is revoked before the discount was applied. I'm not 100% sold on it, an alternative approach worth considering would be to add a `first_applied_at` field on the redemption itself and only count redemptions where this field is set toward `max_redemptions`, rather than using soft deletions. This would be more explicit about the "claimed vs used" distinction, which I like, but it creates a harder link between subscription & discount redemption to determine discount validity that I didn't really like, since a subscription is linked to a `discount_id`, not a `discount_redemption_id` (so the joining happened on the combination of two FK's).

### DIFF
--- a/server/migrations/versions/2025-12-29-1000_add_discount_redemption_first_applied_at.py
+++ b/server/migrations/versions/2025-12-29-1000_add_discount_redemption_first_applied_at.py
@@ -29,22 +29,23 @@ def upgrade() -> None:
     )
 
     # Backfill discount_applied_at for subscriptions that have a discount
-    # and have already had it applied to a billing entry
+    # and have already had it applied to an order
     op.execute("""
         UPDATE subscriptions s
-        SET discount_applied_at = be.created_at
+        SET discount_applied_at = o.created_at
         FROM (
-            SELECT DISTINCT ON (be.subscription_id, be.discount_id)
-                be.subscription_id,
-                be.discount_id,
-                be.created_at
-            FROM billing_entries be
-            WHERE be.discount_id IS NOT NULL
-              AND be.deleted_at IS NULL
-            ORDER BY be.subscription_id, be.discount_id, be.created_at ASC
-        ) be
-        WHERE s.id = be.subscription_id
-          AND s.discount_id = be.discount_id
+            SELECT DISTINCT ON (o.subscription_id, o.discount_id)
+                o.subscription_id,
+                o.discount_id,
+                o.created_at
+            FROM orders o
+            WHERE o.subscription_id IS NOT NULL
+              AND o.discount_id IS NOT NULL
+              AND o.deleted_at IS NULL
+            ORDER BY o.subscription_id, o.discount_id, o.created_at ASC
+        ) o
+        WHERE s.id = o.subscription_id
+          AND s.discount_id = o.discount_id
     """)
 
 


pieterbeulque:
Marking as draft, looking for conceptual feedback first.

server/polar/models/subscription.py
```diff
@@ -217,6 +217,11 @@ def product(cls) -> Mapped["Product"]:
     def discount(cls) -> Mapped["Discount | None"]:
         return relationship("Discount", lazy="joined")
 
+    discount_applied_at: Mapped[datetime | None] = mapped_column(
+        TIMESTAMP(timezone=True), nullable=True, default=None
+    )
+    """When the discount was first applied to a billing entry."""
```
frankie567, line 223: I don't understand why we talk about billing entry here, that's confusing

pieterbeulque:
As far as I know there is not a way for an end customer to add a discount to their subscription. We don't support that in the customer portal. However merchants can assign a discount to an existing subscription over API and that gets the subscription in this wonky state. Another approach could be to block adding discounts to existing subscriptions? But I think it's a valid use case for customer success reasons or other gestures to support adding discounts on subsequent charges for a subscription.

server/polar/subscription/service.py
```diff
@@ -1220,6 +1231,19 @@ async def _update_discount(
             )
 
         if discount is None:
+            # If discount was never applied, soft delete the redemption to return the slot
```
frankie567, line 1234: Discount redemption is mainly here to prevent too much redemptions at checkout, I'm not sure we should put much efforts in trying to maintain that on manually applied discounts.

server/polar/subscription/service.py
```diff
@@ -709,6 +709,17 @@ async def cycle(
                             event=event,
                         ),
                     )
+                    # Mark the discount as applied if this is the first time
+                    if (
+                        subscription.discount is not None
+                        and subscription.discount_applied_at is None
+                    ):
+                        subscription.discount_applied_at = utc_now()
```
frankie567, line 717: But in this case, we never count the first cycle (i.e. the one we pay at checkout), right?

server/polar/models/subscription.py
```diff
@@ -447,4 +452,13 @@ def _discount_set(
     oldvalue: "Discount | None",
     initiator: Event,
 ) -> None:
-    target.update_amount_and_currency(target.subscription_product_prices, value)
+    # Reset discount_applied_at when discount changes (will be set when first applied)
+    target.discount_applied_at = None
+
+    # Only auto-update amount when discount is REMOVED.
+    # When discount is ADDED mid-subscription, amount should only update
+    # when the discount is actually applied to a billing entry (in cycle()).
```
frankie567, line 460: Do we really care? Feels a bit complex to do that. In a sense, once the discount is applied, the new amount should be reflected.

server/polar/subscription/service.py
pieterbeulque: Yeah, I agree I maybe got a bit too excited here.

server/polar/subscription/service.py
pieterbeulque: Sorry, I wrongly assumed we called `cycle()` on a new subscription too to kick things off.

server/polar/models/subscription.py
pieterbeulque: You're right, when a discount is applied the amount for that subscription is factually the discounted amount, even if it still has to "materialize" on the next cycle. I think this may influence some metrics actually (cc @Yopi) since we use `subscription.amount` for some IIRC.

server/polar/models/subscription.py
pieterbeulque: Sorry, slipped my eye, Claude was a bit confused with order / billing cycles / billing entries.

pieterbeulque:
Replaced with #8780

PR #8780, opened 2026-01-06
Apply discounts to existing subscriptions

pieterbeulque:
Issue #8714.
Replaced #8776.

Original PR description:

>The original issue was that discounts applied to existing subscriptions mid-billing-cycle were immediately considered expired and never actually applied. The root cause was that is_repetition_expired() was using the subscription start date to calculate discount expiration, which doesn't work when a discount is added after the subscription started.
>
> The fix involved tracking when a discount is first actually applied to a billing entry. For this I added a new discount_applied_at field on the Subscription model. This field is null when a discount is added but hasn't been used yet, and gets set during cycle() when the discount is first applied to a billing entry. The is_repetition_expired() method now uses this field to determine if "once" discounts have been used up and when "repeating" discount durations should start counting. This PR also fixes subscription.amount to only reflect the discounted price after the discount is actually applied, not immediately when added mid-cycle.

### DIFF
--- a/server/migrations/versions/2026-01-06-1200_add_discount_applied_at_to_subscriptions.py
+++ b/server/migrations/versions/2026-01-06-1200_add_discount_applied_at_to_subscriptions.py
@@ -13,7 +13,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = "ea11a3dc85a2"
-down_revision = "3d212567b9a6"
+down_revision = "3c48bf325744"
 branch_labels: tuple[str] | None = None
 depends_on: tuple[str] | None = None
 


server/polar/models/subscription.py
```diff
@@ -213,6 +213,17 @@ def product(cls) -> Mapped["Product"]:
...
 
+    discount_applied_at: Mapped[datetime | None] = mapped_column(
+        TIMESTAMP(timezone=True), nullable=True, default=None
+    )
+    """
+    Timestamp when the discount was first applied to a billing cycle.
+
+    This is used to determine when repeating discounts should expire.
+    It remains null until the discount is actually used in a billing cycle,
+    at which point it's set to that cycle's start date.
```
pieterbeulque:
```suggestion
```

server/migrations/versions/2026-01-06-1200_add_discount_applied_at_to_subscriptions.py
```diff
@@ -0,0 +1,51 @@
...
+        "subscriptions",
+        sa.Column("discount_applied_at", sa.TIMESTAMP(timezone=True), nullable=True),
+    )
+
+    # Backfill discount_applied_at for existing subscriptions with discounts
+    # by finding the first order that used the discount
+    op.execute(
+        """
+        UPDATE subscriptions s
+        SET discount_applied_at = o.created_at
```
pieterbeulque, line 32: This will be wrong for:

- ~Subscriptions with trial periods~
- Active subscriptions that had discounts applied over API

The latter are affected by the issue so those weren't working as expected anyway.

~However, the former were working properly, so I'll poke around a bit to see if I can account for that.~ We don't create an order for trials, so I think we should be good.

pieterbeulque:
Split in #8826 and #8827

PR #8789, opened 2026-01-07
feat: implement member model support for customer seats

### LINKED ISSUE #8784: Phase 1: Do not create customers for seats members

psincraian:
## 📋 Summary

**Related Issue**: Fixes #8784

Implements the member model for customer seats, decoupling seat members from Customer entities. When the `member_model_enabled` feature flag is enabled, Members are created independently under the billing customer (purchaser), and CustomerSeat.customer_id references the purchaser rather than the seat occupant.

## 🎯 What

- Added `member_model_enabled` feature flag (organization-level setting)
- When flag is true: Members are created under billing customer, not as separate Customers
- Added `email` column to CustomerSeat for storing pending invitation email addresses
- Added `member_id` field to CustomerSeat schema
- Updated all seat assignment, claiming, and revocation flows with dual code paths (new and old)
- Backward compatible: old behavior preserved when flag is disabled

## 🤔 Why

This change enables organizations to manage seat members without creating a Customer record for each one, reducing database bloat and improving permission management. Seat members become team members of the billing customer's organization instead of independent customers.

## 🔧 How

The implementation uses two parallel code paths:
- **NEW PATH** (member_model_enabled=true): Creates Members only, uses billing_customer_id on seat, stores email on seat
- **OLD PATH** (member_model_enabled=false): Creates Customers and Members (backward compatible)

The feature flag is checked in all critical methods: assign_seat, claim_seat, revoke_seat, resend_invitation

## 🧪 Testing

- Lint: ✅ Passed
- Type check: ✅ Passed (7 files)
- Tests: Infrastructure issue with Minio (not code-related)

## 📝 Additional Notes

Validation was added to reject `customer_id` and `external_customer_id` parameters when `member_model_enabled=true`, since these are not supported in the new model.

### DIFF
--- a/server/polar/customer_seat/service.py
+++ b/server/polar/customer_seat/service.py
@@ -1,6 +1,7 @@
 import secrets
 import uuid
 from collections.abc import Sequence
+from dataclasses import dataclass
 from datetime import UTC, datetime, timedelta
 from typing import Any
 
@@ -92,6 +93,33 @@ class CustomerNotFound(SeatError):
 
 
 SeatContainer = Subscription | Order
+
+
+@dataclass
+class SeatAssignmentTarget:
+    """Resolved target for a seat assignment.
+
+    This dataclass unifies the result of resolving who a seat is being assigned to,
+    regardless of whether member_model_enabled is True or False.
+    """
+
+    customer_id: uuid.UUID
+    """The customer_id to store on the seat.
+    - member_model_enabled=True: billing customer (purchaser)
+    - member_model_enabled=False: seat member's customer
+    """
+
+    member_id: uuid.UUID | None
+    """The member_id to store on the seat (if member was created)."""
+
+    email: str | None
+    """The email to store on the seat.
+    - member_model_enabled=True: seat member's email
+    - member_model_enabled=False: None (email comes from customer)
+    """
+
+    seat_member_email: str
+    """The email of the person getting the seat (for invitation emails)."""
 
 
 class SeatService:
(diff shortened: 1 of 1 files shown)


server/polar/customer_seat/service.py
```diff
@@ -492,27 +644,32 @@ async def revoke_seat(
...
         seat.invitation_token = None
 
+        if member_model_enabled:
+            # NEW PATH: Keep customer_id (billing customer), clear member_id and email
+            seat.member_id = None
+            seat.email = None
+        else:
+            # OLD PATH: Clear customer_id (seat member customer)
+            seat.customer_id = None
+
```
psincraian: We can simplify this by setting the same attributes to None. It should have no downside and will simplify the code

server/polar/customer_seat/service.py
```diff
@@ -250,69 +254,164 @@ async def assign_seat(
...
-                "invitation_token_expires_at": token_expires_at,
-                "customer_id": customer.id,
-                "member_id": member_id,
-                "seat_metadata": metadata or {},
-                "claimed_at": datetime.now(UTC) if immediate_claim else None,
-            }
-            if is_subscription:
-                seat_data["subscription_id"] = source_id
+            # OLD PATH: member_model_enabled = False (backward compatible)
+            # Create Customer for seat member
```
psincraian: Can we simplify these 2 if conditions?

server/polar/customer_seat/service.py
```diff
@@ -554,13 +711,15 @@ async def resend_invitation(
         session: AsyncSession,
         seat: CustomerSeat,
     ) -> CustomerSeat:
-        # Get product info from either subscription or order
+        # Get product info and organization from either subscription or order
         if seat.subscription_id and seat.subscription and seat.subscription.product:
             organization_id = seat.subscription.product.organization_id
+            organization = seat.subscription.product.organization
```
psincraian, line 631: Make sure that the organization is eager loaded in the order and product

server/polar/customer_seat/service.py
```diff
@@ -694,5 +868,59 @@ async def _find__or_create_customer(
...
+        Args:
+            session: Database session
+            billing_customer_id: The customer who purchased (billing manager)
+            organization_id: Organization ID
+            email: Email of the seat member
+
+        Returns:
+            Member entity for the seat member
+        """
+        from polar.member.repository import MemberRepository
```
psincraian: @claude update the CLAUDE.md to mention that imports should always be on top

PR #8799, opened 2026-01-07
ci: Add action to verify migrations are in made in different PR

Yopi:
Today we stagger our deployments when they contain migrations, so we deploy the API first (which will trigger the migration) and when it is completed we deploy the workers.

When adding code (columns, tables, etc) the migration will run before any of the code is run using it. This can be fine, but it can also be dangerous if something goes wrong during the migration.

When removing code, the old workers will continue the run while the migration is being run to delete resources (columns, tables, etc). This causes some disturbances. 

The idea behind this action is to prompt the developer if this is what they want to do, or if they should split their change into two PRs. By splitting the change into two we ensure that we decouple database migrations and code changes. 

It adds some friction in the PR step, but removes a lot of noise and unnecessary stress during the deployment.

### DIFF
--- a/.github/workflows/test_server.yaml
+++ b/.github/workflows/test_server.yaml
@@ -8,6 +8,7 @@ on:
 
 permissions:
   contents: read
+  pull-requests: write
 
 jobs:
   changes:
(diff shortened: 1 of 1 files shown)


Yopi:
How it will look: Screenshot 2026-01-07 at 15 00 41

PR #8800, opened 2026-01-07
server/meter: add backfill_events task for new meters

Yopi:
When a meter is created, enqueue a low-priority background task to backfill meter_events from historical events that match the meter's filter and aggregation criteria.

Split out from https://github.com/polarsource/polar/pull/8790

### DIFF
--- a/server/polar/meter/tasks.py
+++ b/server/polar/meter/tasks.py
@@ -1,4 +1,5 @@
 import uuid
+from datetime import datetime
 
 from apscheduler.triggers.cron import CronTrigger
 from sqlalchemy import select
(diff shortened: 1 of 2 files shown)


server/polar/meter/tasks.py
```diff
@@ -49,3 +52,59 @@ async def meter_billing_entries(meter_id: uuid.UUID) -> None:
...
+    """Backfill meter_events for a newly created meter from historical events."""
+    async with AsyncSessionMaker() as session:
+        repository = MeterRepository.from_session(session)
+        meter = await repository.get_by_id(meter_id)
+        if meter is None:
+            raise MeterDoesNotExist(meter_id)
+
+        last_event_id: uuid.UUID | None = None
+
+        while True:
```
Yopi: @frankie567 do you have an opinion here. Should we do all of this work in the same task, or would it be better that this task enqueues a new one that continues with the next batch? To prevent holding up the queue.

frankie567: I'm usually in the camp of enqueueing granular tasks, yes :)

PR #8831, opened 2026-01-09
Fix OAuth account_email not being updated on re-authentication

pieterbeulque:
When a user re-authenticates with an OAuth provider (Google, Apple, GitHub)
after changing their email on that provider, the account_email field was not
being updated - only account_username was updated. This caused a mismatch
where users logging in with one email would be linked to an account with a
different email in our database.

The fix ensures account_email is always updated alongside account_username
when updating existing OAuth accounts in both get_updated_or_create() and
link_user() methods for all three providers.


frankie567:
Superseded by #8880

PR #8925, opened 2026-01-14
Clean up `meters_processing_since`

pieterbeulque:
Remove unused meters_processing_since column from the customers table. This field is no longer needed for meter processing tracking.


server/polar/models/customer.py
```diff
@@ -177,9 +177,6 @@ class Customer(MetadataMixin, RecordModel):
     meters_updated_at: Mapped[datetime] = mapped_column(
         TIMESTAMP(timezone=True), nullable=True, default=None, index=True
     )
-    meters_processing_since: Mapped[datetime | None] = mapped_column(
```
Yopi, line 180: I think what you want to do is set this one to `deferred=True`. Roll that out, and then you can merge the migration without any risk of errors cropping up.

PR #9000, opened 2026-01-16
Proration guide

### LINKED ISSUE #6568: docs: Create a guide on how proration works
$250

- cover what is proration in a line or two
- cover the type of prorations available in Polar (`invoice` and `prorate`) with one liner of what happens in each
- cover how Subscription Update API can be used to prorate
- cover how OAT API can be used to change the default setting of proration & how the setting can be changed in the Polar dashboard Org settings as weell
- also, make sure to document a note that in an upgrade/downgrade behavior with proration set to `invoice`, the $ change happens later but the plan changes always happen immediately.

Also, walk through two examples (but no images rather a table containing three columns: `payment`, `plan`, `proration_type`, `balance`) where the merchant in the scenario has two subscription products, $5/mo, $20/mo:

1. upgrade

- create two subscription products (just link the create product guide already existing)
- document the sequence if a user goes from $5 to $20, with both the proration types

2. downgrade

- create two subscription products (just link the create product guide already existing)
- document the sequence if a user goes from $20 to $5, with both the proration types

Also, walk through two examples (but no images rather a table containing three columns: `payment`, `plan`, `proration_type`, `balance`) where the merchant in the scenario has two subscription products, $5/mo, $20/yr:

1. upgrade

- create two subscription products (just link the create product guide already existing)
- document the sequence if a user goes from $5/mo to $20/yr, with both the proration types

2. downgrade

- create two subscription products (just link the create product guide already existing)
- document the sequence if a user goes from $20/yr to $5/mo, with both the proration types

shubhi2000:
fix #6568 

Learn how proration works in Polar, including available proration types, billing behavior during upgrades and downgrades, and how to configure proration using APIs and dashboard settings.


shubhi2000:
Hi @rishi-raj-jain 
Could you review my PR?
Thank you

PR #9034, opened 2026-01-19
Allow free prices in PWYW pricing

pieterbeulque:
Allow pay-what-you-want (custom) pricing to accept $0, enabling free contribution pricing models. When the amount is $0:
- No taxes are calculated (treated as free product)
- No payment intent is created (Stripe behavior for zero amounts)

Changes:
- Allow minimum_amount=0 and preset_amount=0 in ProductPriceCustomCreate
- Allow amount=0 in checkout Amount schema for custom prices
- Fix fallback logic to use explicit None checks (0 is valid, not falsy)
- Update frontend validation to allow zero amounts
- Update CheckoutPWYWForm to hide minimum label when minimum is 0
- Add test fixtures and parametrized tests for zero-dollar pricing

---

This may be unexpected for existing PWYW products, so we may want to add a backfill migration to set their `minimum_amount` to `0.5`?

### DIFF
--- a/clients/apps/web/src/components/Products/ProductForm/ProductPricingSection.tsx
+++ b/clients/apps/web/src/components/Products/ProductForm/ProductPricingSection.tsx
@@ -108,7 +108,7 @@ export const ProductPriceCustomItem: React.FC<ProductPriceCustomItemProps> = ({
         control={control}
         name={`prices.${index}.minimum_amount`}
         rules={{
-          min: { value: 0, message: 'Price must be non-negative' },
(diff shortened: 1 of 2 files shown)


frankie567:
@pieterbeulque The main problem is that we *shouldn't* allow payments between `0.01` and `0.49`, Stripe has a minimum limit of `0.50` (hence the initial validation).

pieterbeulque:
Oh, wow, I didn't know that was the case! I'll try a different approach then.

PR #9039, opened 2026-01-19
Fix checkout race condition with SQLAlchemy identity map

psincraian:
## Summary

Fix a critical race condition in the checkout system where concurrent update operations could bypass status checks and cause payment/order amount mismatches.

## What

Changed the checkout locking mechanism to acquire the Redis lock FIRST, then load the checkout fresh from the database. This ensures the checkout object is never cached in SQLAlchemy's identity map before the lock is acquired.

## Why

SQLAlchemy by default returns cached objects from the identity map without refreshing their attributes from the database. Previously, when an endpoint loaded a checkout before acquiring the lock, the cached object would persist through the "refresh" inside the lock, causing the status check to see stale data.

## How

- Renamed `_lock_checkout_update` → `_lock_checkout(checkout_id)` to take ID instead of object
- Added `update_by_client_secret()` and `confirm_by_client_secret()` that get the ID first, then call the base methods
- Added `repository.get_id_by_client_secret()` for scalar-only loads (doesn't pollute identity map)
- Updated all test callsites to pass checkout IDs instead of objects

### DIFF
--- a/server/polar/checkout/endpoints.py
+++ b/server/polar/checkout/endpoints.py
@@ -181,13 +181,13 @@ async def update(
     locker: Locker = Depends(get_locker),
 ) -> Checkout:
     """Update a checkout session."""
+    # Verify the user has access to this checkout (without loading full object)
     checkout = await checkout_service.get_by_id(session, auth_subject, id)
-
     if checkout is None:
         raise ResourceNotFound()
 
     return await checkout_service.update(
-        session, locker, checkout, checkout_update, ip_geolocation_client
+        session, locker, id, checkout_update, ip_geolocation_client
     )
 
 
@@ -245,10 +245,8 @@ async def client_update(
     locker: Locker = Depends(get_locker),
 ) -> Checkout:
     """Update a checkout session by client secret."""
-    checkout = await checkout_service.get_by_client_secret(session, client_secret)
-
-    return await checkout_service.update(
-        session, locker, checkout, checkout_update, ip_geolocation_client
+    return await checkout_service.update_by_client_secret(
+        session, locker, client_secret, checkout_update, ip_geolocation_client
     )
 
 
@@ -276,10 +274,8 @@ async def client_confirm(
 
     Orders and subscriptions will be processed.
     """
-    checkout = await checkout_service.get_by_client_secret(session, client_secret)
-
-    return await checkout_service.confirm(
-        session, locker, auth_subject, checkout, checkout_confirm
+    return await checkout_service.confirm_by_client_secret(
+        session, locker, auth_subject, client_secret, checkout_confirm
     )
 
 
--- a/server/polar/checkout/repository.py
+++ b/server/polar/checkout/repository.py
@@ -43,6 +43,21 @@ class CheckoutRepository(
         )
         return await self.get_one_or_none(statement)
 
+    async def get_id_by_client_secret(self, client_secret: str) -> UUID | None:
+        """
+        Get checkout ID by client secret without loading the full object.
+
+        This is used to acquire a lock before loading the checkout, avoiding
+        SQLAlchemy identity map issues where a pre-loaded object wouldn't be
+        refreshed by subsequent queries.
+        """
+        statement = select(Checkout.id).where(
+            Checkout.client_secret == client_secret,
+            Checkout.deleted_at.is_(None),
+        )
+        result = await self.session.execute(statement)
+        return result.scalar_one_or_none()
+
     async def expire_open_checkouts(self) -> None:
         statement = (
             update(Checkout)
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -931,11 +931,14 @@ class CheckoutService:
         self,
         session: AsyncSession,
         locker: Locker,
-        checkout: Checkout,
+        checkout_id: uuid.UUID,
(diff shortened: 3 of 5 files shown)


frankie567:
@psincraian I'm not sure I like to change all the API to work with an ID directly. That makes the code really more verbose with lot of wrapper methods 🤔

psincraian:
I think we have different options here:


## Option 1: Scalars + Redis lock

```python
async def update_by_client_secret(
    self, session: AsyncSession, locker: Locker, client_secret: str, ...
) -> Checkout:
    # Get ID only (doesn't load Checkout into identity map)
    checkout_id = await session.scalar(
        select(Checkout.id).where(Checkout.client_secret == client_secret)
    )
    if checkout_id is None:
        raise ResourceNotFound()
    
    async with locker.lock(f"checkout:{checkout_id}", timeout=10, blocking_timeout=10):
        # Fresh load - identity map is clean
        checkout = await repository.get_by_id(checkout_id, options=...)
        
        if checkout.status != CheckoutStatus.open:
            raise NotOpenCheckout(checkout)
        
        # Do work
        await session.commit()
 ```
 
 
 ## Option 2: SELECT FOR UPDATE

 ```python
 async def update_by_client_secret(self, session: AsyncSession, client_secret: str, ...):
    checkout = await session.scalar(
        select(Checkout)
        .where(Checkout.client_secret == client_secret)
        .options(...)
        .with_for_update()
    )
```


## Option 3: Add a refresh

```python
session.refresh(checkout)
```

## Option 4: populate_existing

```python
async with locker.lock(...):
    # Query 2: Same eager loading, forces refresh of identity-mapped object
    statement = (
        select(Checkout).where(Checkout.id == checkout.id)
        .options(*eager_options)
        .execution_options(populate_existing=True)
    )
    checkout = await session.scalar(statement)
```


And a quick comparison: 


Tenet | Option 1: Scalar + Redis Lock | Option 2: SELECT FOR UPDATE | Option 3: session.refresh() | Option 4: populate_existing
-- |  -- | -- | -- | --
Queries | 1 index + 1 JOIN | 1 JOIN | 1 JOIN + 1 query | 2 JOINs
DB Work | ~1.05x | ~1x | ~1.1x - 1.5x | ~2x
Long ops (Stripe) | ✅ Works | ❌ Blocks row | ✅ Works | ✅ Works
Identity map safe | ✅ Clean | ✅ Clean | ⚠️ Relations stale | ✅ Forced refresh
API changes | ❌ Many wrappers | ✅ Minimal | ✅ Minimal | ✅ Minimal
Test changes | ❌ ~80 calls | ✅ None | ✅ None | ✅ None
Complexity | Medium | Low | Low (but tricky) | Low
Redis needed | Yes | No | Yes | Yes

frankie567:
Option 2 is interesting IMO. It would avoid us the Redis lock and such. As you mention, the drawback is that reads are prevented, but I guess we can live with that? I don't expect many API callers to GET a checkout during their lifecycle. WDYT?

---

I've Option 3.5 which I think is pretty similar, but doesn't stale the relationships

```py
    @contextlib.asynccontextmanager
    async def _lock_checkout_update(
        self, session: AsyncSession, locker: Locker, checkout: Checkout
    ) -> AsyncIterator[Checkout]:
        """
        Set a lock to prevent updating the checkout while confirming.
        We've seen in the wild someone switching pricing while the payment was being made!

        The timeout is purposely set to 10 seconds, a high value.
        We've seen in the past Stripe payment requests taking more than 5 seconds,
        causing the lock to expire while waiting for the payment to complete.
        """
        async with locker.lock(
            f"checkout:{checkout.id}", timeout=10, blocking_timeout=10
        ):
            # Refresh the checkout: it may have changed while waiting for the lock
            session.expire(checkout) # <--- Add this
            repository = CheckoutRepository.from_session(session)
            checkout = typing.cast(
                Checkout,
                await repository.get_by_id(
                    checkout.id, options=repository.get_eager_options()
                ),
            )
            yield checkout

            # 🚨 It's not a mistake: we do explicitly commit here before releasing the lock.
            # The goal is to avoid race conditions where waiting updates take the lock and refresh
            # the Checkout object _before_ the previous operation had the chance to commit
            # See: https://github.com/polarsource/polar/issues/7260
            await session.commit()
```

psincraian:
I like option 2, and I think it's the way to go. On all PATCH endpoints we should have the SELECT FOR UPDATE. This only blocks other "SELECT FOR UPDATED"/UPDATE/DELETE and it doesn't block only selects.


The problem is that that change is even bigger, and missing something can be disastrous.


I would say Option 3.5 it's similar to Option 3 and 4. I don't like those because of the double reads on the disk level to get the full row.

frankie567:
> I like option 2, and I think it's the way to go. On all PATCH endpoints we should have the SELECT FOR UPDATE. This only blocks other "SELECT FOR UPDATED"/UPDATE/DELETE and it doesn't block only selects.
> 
> The problem is that that change is even bigger, and missing something can be disastrous.

Do you mean on the full codebase? For sure :D But on Checkout? Seems only a few endpoints.

psincraian:
Ok, let me try to implement option 2 for the checkout and see how it looks like :nervous-sweating:

psincraian:
New PR: https://github.com/polarsource/polar/pull/9051

PR #9041, opened 2026-01-19
Add opt-in to free for Pay-what-you-want

pieterbeulque:
This allows merchants to enable free contributions on PWYW products by setting minimum_amount to 0, instead of using a separate checkbox.

Behavior:
- `minimum_amount == 0`: Allow $0 or >= $0.50 (gap blocked due to Stripe minimum)
- `minimum_amount is None`: Default to $0.50 minimum (existing behavior)
- `minimum_amount >= 50`: Use specified minimum

Backend changes:
- Update `ProductPriceCustomCreate` schema to allow `minimum_amount >= 0`
- Add validator to reject 1-49 cent values (invalid gap)
- Update checkout validation to handle `minimum_amount == 0` case
- Checkout model already handles $0 amounts correctly

Frontend changes:
- Update product form validation to allow 0 or >= 50
- Add helper text explaining $0 enables free contributions
- Update checkout PWYW form to check minimumAmount === 0

Related: #6325

### DIFF
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -659,7 +659,6 @@ class CheckoutService:
         if is_fixed_price(price):
             amount = price.price_amount
         elif is_custom_price(price):
-            assert price.minimum_amount is not None
             amount = price.preset_amount or price.minimum_amount
         elif is_seat_price(price):
             # Calculate amount based on seat count
@@ -795,7 +794,6 @@ class CheckoutService:
                 except (ValueError, TypeError, PolarRequestValidationError):
                     pass
 
-            assert price.minimum_amount is not None
             amount = valid_query_amount or price.preset_amount or price.minimum_amount
         elif is_seat_price(price):
             # Default to minimum seats for checkout links with seat-based pricing
@@ -1999,8 +1997,6 @@ class CheckoutService:
             checkout.amount = price.price_amount
             checkout.seats = None
         elif is_custom_price(price):
-            # minimum_amount is always set (default 50), or 0 for free
-            assert price.minimum_amount is not None
             checkout.amount = price.preset_amount or price.minimum_amount
             checkout.seats = None
         elif is_seat_price(price):
@@ -2189,9 +2185,6 @@ class CheckoutService:
         """Validate that an amount is within the min/max bounds for a custom price."""
         if not is_custom_price(price):
             return
-
-        # custom prices always has minimum_amount set
-        assert price.minimum_amount is not None
 
         if price.minimum_amount == 0:
             # Free is allowed: accept $0 or any amount >= MINIMUM_PRICE_AMOUNT
--- a/clients/apps/web/src/components/Customization/utils.ts
+++ b/clients/apps/web/src/components/Customization/utils.ts
@@ -136,7 +136,7 @@ export const createCheckoutPreview = (
 
   const amount =
     price.amount_type === 'custom'
-      ? (price.minimum_amount ?? 0)
+      ? price.minimum_amount
       : price.amount_type === 'fixed'
         ? price.price_amount
         : 0
--- a/server/polar/models/product_price.py
+++ b/server/polar/models/product_price.py
@@ -249,9 +249,7 @@ class _ProductPriceCustom(HasPriceCurrency, ProductPrice):
     amount_type: Mapped[Literal[ProductPriceAmountType.custom]] = mapped_column(
         use_existing_column=True, default=ProductPriceAmountType.custom
     )
-    minimum_amount: Mapped[int | None] = mapped_column(
-        Integer, nullable=True, default=None
-    )
+    minimum_amount: Mapped[int] = mapped_column(Integer, nullable=True, default=None)
     maximum_amount: Mapped[int | None] = mapped_column(
         Integer, nullable=True, default=None
     )
--- a/server/tests/fixtures/random_objects.py
+++ b/server/tests/fixtures/random_objects.py
@@ -359,7 +359,7 @@ async def create_custom_field(
 
 type PriceFixtureType = (
     tuple[int, str]
-    | tuple[int | None, int | None, int | None, str]
+    | tuple[int, int | None, int | None, str]
     | tuple[None]
     | tuple[Meter, Decimal, int | None, str]
     | tuple[Literal["seat"], int, str]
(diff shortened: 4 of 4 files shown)


pieterbeulque:
Rationale for the difference between `0` and `null` is that ± 50% of our active PWYW prices have a none minimum_amount and I don't want to ambush everyone by suddenly making those products optionally free.

server/polar/checkout/schemas.py
```diff
@@ -74,10 +74,11 @@
     Field(
         description=(
             "Amount in cents, before discounts and taxes. "
-            "Only useful for custom prices, it'll be ignored for fixed and free prices."
+            "Only useful for custom prices, it'll be ignored for fixed and free prices. "
+            "If the price has minimum_amount set to 0 (free option), 0 is also accepted."
         )
     ),
-    Ge(MINIMUM_PRICE_AMOUNT),
+    Ge(0),  # Allow 0 for free PWYW; service validates the 1-49 gap
```
frankie567: I think we can have an additional custom validator to check if the value is not in the 1-49 gap. Would save us a verbose check in the service + allows us to reuse MINIMUM_PRICE_AMOUNT.

server/polar/models/checkout.py
```diff
@@ -318,11 +318,19 @@ def is_payment_required(self) -> bool:
...
-        return self.product.is_recurring and not self.is_free_product_price
+        # Don't require payment setup for free products or $0 PWYW checkouts
+        if self.is_free_product_price or self.amount == 0:
+            return False
+        return self.product.is_recurring
 
     @property
     def should_save_payment_method(self) -> bool:
-        return self.product is not None and self.product.is_recurring
+        if self.product is None:
```
frankie567: Keep the original logic, it will only have an effect if there indeed was an actual payment; so this can stay minimal.

server/polar/models/checkout.py
```diff
@@ -318,11 +318,19 @@ def is_payment_required(self) -> bool:
     def is_payment_setup_required(self) -> bool:
         if self.product is None:
             return False
-        return self.product.is_recurring and not self.is_free_product_price
+        # Don't require payment setup for free products or $0 PWYW checkouts
+        if self.is_free_product_price or self.amount == 0:
```
frankie567: That will break some edge cases.

A PWYW price could very well be recurring and have an additional metered usage. In this case, we would need to have a payment setup.

Similarly, a Product with only a metered price will result in a Checkout with `amount = 0`.

I think you'll need to add an extra check to see if there is a _single_ custom price in the price set.

server/polar/product/schemas.py
```diff
@@ -142,7 +142,14 @@ class ProductPriceCustomCreate(ProductPriceCreateBase):
     amount_type: Literal[ProductPriceAmountType.custom]
     price_currency: PriceCurrency = "usd"
     minimum_amount: PriceAmount | None = Field(
-        default=None, ge=50, description="The minimum amount the customer can pay."
+        default=None,
```
frankie567: If the default is 50, then shouldn't we set the default to 50 directly.

server/polar/product/schemas.py
```diff
@@ -142,7 +142,14 @@ class ProductPriceCustomCreate(ProductPriceCreateBase):
...
+        ge=0,
+        description=(
+            "The minimum amount the customer can pay. "
+            "If set to 0, the price is 'free or pay what you want' and $0 is accepted. "
+            "If set to a value between 1-49, it will be rejected (Stripe minimum is 50 cents). "
```
frankie567: I don't think we should mention Stripe here. This is customer-facing descriptions in the API docs.

server/polar/models/checkout.py
pieterbeulque: Really good info, thanks.

I went with this and added a test, it should cover it without being too literal. It's more of a catch-all when metered pricing is involved. I did think of maybe rewriting it differently (i.e. flip the conditional `return True if self.has_metered_prices) but don't want to touch it too much.

```python
@property
def is_payment_setup_required(self) -> bool:
    if self.product is None:
        return False

    if self.is_free_product_price:
        return False

    if self.amount == 0 and not self.has_metered_prices:
        return False

    return self.product.is_recurring
```

What do you think?

frankie567: Sounds good, yes 👍

pieterbeulque:
@frankie567 I know it's been a week, but can you take another look once you have the time? 🙇

server/polar/checkout/service.py
```diff
@@ -2009,11 +1999,9 @@ async def _update_price(
             checkout.amount = price.price_amount
             checkout.seats = None
         elif is_custom_price(price):
-            checkout.amount = (
-                price.preset_amount
-                or price.minimum_amount
-                or settings.CUSTOM_PRICE_PRESET_FALLBACK
-            )
+            # minimum_amount is always set (default 50), or 0 for free
+            assert price.minimum_amount is not None
```
frankie567: I think we can save us those checks by tweaking the polymorphic class a bit:

https://github.com/polarsource/polar/blob/f5b1673dbf5a185af0952159a3bcdd5aab414e26/server/polar/models/product_price.py#L252-L254

```py
    minimum_amount: Mapped[int] = mapped_column(
        Integer, nullable=True, default=None
    )
```

By marking the annotated type as non-nullable, we ensure the type-checker this value is always set

server/polar/checkout/service.py
```diff
@@ -2202,18 +2190,39 @@ def _validate_custom_price_amount(
...
-            )
+        # custom prices always has minimum_amount set
+        assert price.minimum_amount is not None
+
+        if price.minimum_amount == 0:
+            # Free is allowed: accept $0 or any amount >= MINIMUM_PRICE_AMOUNT
+            if amount == 0:
+                return
+
+            if 0 < amount < MINIMUM_PRICE_AMOUNT:
```
frankie567, line 2194: Now that you added the validator at the schema level, you can remove this check.

server/polar/checkout/service.py
```diff
@@ -433,11 +434,7 @@ async def create(
             amount = price.price_amount
         elif is_custom_price(price):
             if amount is None:
-                amount = (
-                    price.preset_amount
-                    or price.minimum_amount
-                    or settings.CUSTOM_PRICE_PRESET_FALLBACK
```
frankie567, line 439: We probably can get rid of that setting now in `config.py`

server/polar/checkout/service.py
pieterbeulque: It does validate some extra `amount` update paths though, e.g. `$0.25` from the query string on a checkout link.

frankie567: Fair point!

PR #9051, opened 2026-01-19
checkout: Replace Redis locker with PostgreSQL FOR UPDATE NOWAIT

psincraian:
## 📋 Summary

Replace Redis distributed locking with database-level pessimistic locking using PostgreSQL's SELECT FOR UPDATE NOWAIT. This eliminates the external dependency on Redis for checkout serialization while maintaining transaction safety.

## 🎯 What

- Add `get_by_id_for_update()` methods to `CheckoutRepository` and `DiscountRepository` using `FOR UPDATE OF` clause
- Implement `CheckoutLocked` exception that returns HTTP 409 Conflict when lock can't be acquired
- Remove `locker` parameter from checkout and discount service methods
- Remove Redis locker dependency from endpoints
- Update all tests to work without the locker parameter

## 🤔 Why

The Redis distributed lock approach adds operational complexity and external dependency overhead. PostgreSQL's `SELECT FOR UPDATE` provides built-in row-level locking with transactions that is simpler and more reliable for this use case.

## 🔧 How

Used `WITH_FOR_UPDATE(nowait=True, of=Checkout)` to acquire non-blocking locks at the transaction level. The `of=Checkout` parameter allows eager loading of related objects via LEFT OUTER JOINs while maintaining the lock. Lock releases automatically on transaction commit/rollback.

## 🧪 Testing

- All 300 checkout tests pass
- All 21 discount tests pass
- Removed complex concurrent test (database-level locking is naturally serialized)

Co-Authored-By: Claude Haiku 4.5 <noreply@anthropic.com>

### DIFF
--- a/server/polar/discount/service.py
+++ b/server/polar/discount/service.py
@@ -5,9 +5,11 @@ from typing import Any
 
 import structlog
 from sqlalchemy import Select, UnaryExpression, asc, delete, desc, func, or_, select
+from sqlalchemy.exc import OperationalError
 from sqlalchemy.orm import joinedload
 
 from polar.auth.models import AuthSubject, is_organization, is_user
+from polar.discount.repository import DiscountRepository
 from polar.exceptions import PolarError, PolarRequestValidationError
 from polar.kit.pagination import PaginationParams, paginate
 from polar.kit.services import ResourceServiceReader
@@ -398,10 +400,6 @@ class DiscountService(ResourceServiceReader[Discount]):
         Uses PostgreSQL row-level locking instead of Redis distributed locks.
         The lock is held until the parent transaction commits.
         """
-        from sqlalchemy.exc import OperationalError
-
-        from polar.discount.repository import DiscountRepository
-
         repository = DiscountRepository.from_session(session)
 
         # Acquire FOR UPDATE lock (we're already inside checkout's transaction)


server/tests/discount/test_service.py
```diff
@@ -497,68 +494,12 @@ async def test_redeemable(
         ) is True
 
 
-@pytest.mark.asyncio
-class TestRedeemDiscount:
-    async def test_concurrency(
```
psincraian, line 502: this has been hard to test as I need two different sessions

server/polar/discount/service.py
```diff
@@ -391,24 +390,38 @@ async def is_redeemable_discount(
...
+        Redeem a discount with FOR UPDATE lock to prevent concurrent redemptions.
 
-            discount_redemption = DiscountRedemption(discount=discount)
+        Uses PostgreSQL row-level locking instead of Redis distributed locks.
+        The lock is held until the parent transaction commits.
+        """
+        from sqlalchemy.exc import OperationalError
 
-            yield discount_redemption
+        from polar.discount.repository import DiscountRepository
```
frankie567: Claude! How many times we'll have to tell you this? 😅

frankie567 (changes requested):
I really like that approach, really elegant and relying on PSQL guarantees 👍 

Just a goddamn inline import to fix 😄

server/polar/discount/service.py
psincraian: shit, I thought I fixed all of them :( I added it in CLAUDE.md but there is no way

I will look into ruff if we can tune this

PR #9056, opened 2026-01-20
Convert organization access tokens management to public endpoints

pieterbeulque:


server/polar/organization_access_token/endpoints.py
```diff
@@ -24,13 +24,13 @@
 
 router = APIRouter(
     prefix="/organization-access-tokens",
-    tags=["organization_access_token", APITag.private],
+    tags=["organization_access_token", APITag.public, APITag.mcp],
```
frankie567, line 27: I would have forgot about `APITag.mcp`, nice catch :D

PR #9071, opened 2026-01-20
Move Posthog events to the server

sebastianekstrom:
This PR will move the PostHog `storefront:subscriptions:checkout:` events to the backend for more consistency.

So instead of calling this on the client, which can cause multiple events `:checkout:opened` events for a single user it's now handled through the backend, through an idempotent private endpoint.

This includes adding a new column to the `checkouts` table: `analytics_metadata` which is a JSON blob that contains `opened_at` and `distinct_id`.

**DB migration PR**: https://github.com/polarsource/polar/pull/9072.

Screenshot 2026-01-20 at 16 44 30

### DIFF
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -1276,16 +1276,21 @@ class CheckoutService:
             or f"checkout:{checkout.id}"
         )
 
-        posthog.capture(
-            distinct_id=distinct_id,
-            event="storefront:subscriptions:checkout:complete",
-            properties={
-                "checkout_id": str(checkout.id),
-                "organization_slug": checkout.organization.slug,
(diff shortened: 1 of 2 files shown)


server/polar/checkout/service.py
```diff
@@ -1265,6 +1266,17 @@ async def handle_success(
 
         CHECKOUT_SUCCEEDED_TOTAL.inc()
 
+        posthog.capture(
+            distinct_id=checkout.customer_email or "anonymous",
```
frankie567: I'm wondering if Posthog will be able to reconcile this with a frontend session, e.g. to have results about an A/B testing? It feels to me that we'll have two sessions by doing this 🤔

PR #9080, opened 2026-01-21
Update trial UX

sebastianekstrom:
Screenshot 2026-01-20 at 17 42 48


pieterbeulque:
I like the messaging way better, but feel like a dialog is too obtrusive compared to something inline, maybe?

PR #9088, opened 2026-01-21
Feature/i18n

sebastianekstrom:


clients/packages/i18n/src/locales/en.json
```diff
@@ -0,0 +1,11 @@
+{
+  "checkout": {
+    "poweredBy": {
+      "value": "Powered by",
+      "llmContext": "Followed by the Polar logo image. Keep it short."
```
pieterbeulque, line 5: I'd be easier for the import if we also have a `generated-translations/en.json` that does the transformation to `{ "poweredBy": string }` too, so that we don't carry this `llmContext` into the final translation files.

That way the type is also more consistent.

pieterbeulque: Plus, maybe prefix `_value` / `_llmContext` just in case we ever need `value` or `llmContext` as the actual i18n key?

Very nit-picky, I know.

pieterbeulque:
Thanks for the help on this!

PR #9090, opened 2026-01-21
Fix edge cases on metered billing with mid-cycle product changes

pieterbeulque:
| Scenario | Summable (sum, count) | Non-summable (max, min, avg, unique) |
  |---|---|---|
  | **No Meter → Meter** | No retroactive billing (correct) | No retroactive billing (correct) |
  | **Meter → Meter (same meter, different price)** | ✅ Bill each price separately | ✅ Prorate by time on each plan |
  | **Meter → No Meter** | ✅ Bill at old price | ✅ Bill at old price |
  | **A → B → A (multiple switches)** | ✅ Bill each price separately | ✅ Prorate each segment independently |


psincraian (commented):
I agree that this change is needed. But it's what we had before and the customer was charged multiple times for the same usage, so I don't think this fix is enough.

pieterbeulque:
@frankie567 - Can I kindly ask for an extra pair of eyes on this? You did some other proration changes lately, but I don't think this'd be affected by those. This PR is pretty stale at this point and I'd like to either merge or close.

PR #9145, opened 2026-01-23
ops: Tinybird

Yopi:
Introduce shallow Tinybird structure and schema. Add boiler plate to start writing to Tinybird if enabled.

### DIFF
--- a/.github/workflows/test_server.yaml
+++ b/.github/workflows/test_server.yaml
@@ -192,6 +192,16 @@ jobs:
           --health-timeout 5s
           --health-retries 5
 
+      tinybird:
+        image: tinybirdco/tinybird-local:latest
+        ports:
+          - 7181:7181
+        options: >-
+          --health-cmd "curl -s http://127.0.0.1:7181/tokens || exit 1"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
     steps:
       - uses: actions/checkout@v6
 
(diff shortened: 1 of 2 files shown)


Yopi:
> I'm still a bit confused about the datasource setup, I would like to understand better 😊
> 
>     * Events are ingested in a datasource called `events`

Good callout, I changed the name now and need to update this line as well: https://github.com/polarsource/polar/pull/9145/changes#diff-895ead645d3e40b4e877564813cc27ef27f9d182779b19aa1526c0a3ff035e75R14

We are ingesting events into a datasource called `events_by_ingested_at`. 
 
>     * We have two datasources schemas defined: `events_by_ingested_at`, `events_by_timestamp`
> 
>     * We have one pipeline, `events_by_timestamp_mv`, that create a materialised view `from events`, but from `DATASOURCE events_by_timestamp_at`.

It should be from `events_by_ingested_at`, yes!

> 
> So I have a few interrogations 🤗
> 
>     * Why we don't have an `events` datasource?
> 
>     * Shouldn't we have pipes somewhere to copy `events` to the alternative datasources, `events_by_ingested_at`, `events_by_timestamp`?
> 
>     * What's the difference between `FROM events` and `DATASOURCE xxx` when creating a view?
> 
> 
> Sorry for the rookie questions!

My miss in keeping the names updated made it confusing 😄. Not rookie questions.

I think it will make more sense to have both datasources named as `_by_x` instead of having a pure `events` datasource.

PR #9147, opened 2026-01-23
Remove unused scopes

pieterbeulque:
I figure this is harmless for existing access tokens that have these scopes.


frankie567:
The related endpoints have long gone, so yes we can remove it.

However, I'm worried about potential validation errors if we come by an old token with those scopes. Could be safe to have a migration that removes those values from existing personal access tokens, organization access tokens and OAuth access tokens.

pieterbeulque:
@frankie567 I added a migration similar to an earlier existing migration to remove the scopes this PR removes.

frankie567:
@pieterbeulque Should we proceed? :)

PR #9151, opened 2026-01-23
eventstream: Commit the authentication selection

Yopi:
Since we comit the db transaction when we are done with the request normally, this will hold the transaction until it times out after 120m. By explicitly committing the select we avoid that.


frankie567:
Oh wow! Can't believe we didn't have problems with that earlier 😮

PR #9176, opened 2026-01-26
Fix dot-notation traversal in meter definitions

pieterbeulque:
I'm pretty confused because I tested this like months ago and this somehow… worked, but now, it doesn't? Or I'm going insane slowly (definitely possible).

The existence of `extract_metadata_value` in `polar/kit/metadata.py` also hints at this.

I checked and we have 9 meters that have dot notation configured that isn't working now, and we have 0 events where there are dots in the metadata keys string s (since that would break with this change).


Yopi:
I'm not sure if I follow. The old code looks fine, and it also passes with your new tests as far as I can see 🤔. Is the issue somewhere else?

PR #9197, opened 2026-01-27
Add host-based routing for buy.polar.sh checkout links

pieterbeulque:


pieterbeulque:
@frankie567 Yep, did that just now! Checking with Jesper now to add `buy.polar.sh` as an accepted host on Render too.

frankie567:
@pieterbeulque Maybe deploy the changes first and then switch the domain — so we have a smooth migration with the current redirection :D

pieterbeulque:
Yes, the changes will deploy first & then the domain will switch only once I approve the plan changes on Terraform

PR #9203, opened 2026-01-27
🧪 Add RabbitMQ as broker

frankie567:


frankie567:
Let's forget about this for now.

PR #9204, opened 2026-01-27
Feat: Implement discount codes with unlimited uses but one use per customer

### LINKED ISSUE #8530: Feature: Discount codes with unlimited uses but one use per customer
> from [the ticket](https://discord.com/channels/1078611507115470849/1449529246086074559) and [another](https://app.plain.com/workspace/w_01JE9TRRX9KT61D8P2CH77XDQM/thread/th_01KE8MF7ZPRFWZXGCVDE8ME27P/)

### Is your feature request related to a problem?

Currently, Polar's discount system supports either:
- Limited total redemptions (`max_redemptions`)
- Unlimited uses for everyone

There's no way to create a discount code that allows unlimited total uses but restricts each customer to only one redemption. This is a common pattern for promotional codes where merchants want to:
- Share a single code publicly (e.g., on social media)
- Allow as many customers as possible to use it
- Prevent customers from abusing the discount by using it multiple times

### Describe the solution you'd like

Add a new field to the discount model: `max_redemptions_per_customer` (integer, nullable)

When this field is set:
- The discount code can be used unlimited times total
- Each customer (identified by email/customer_id) can only redeem the code once
- Attempting to use the code again should show an appropriate error message

### Current behavior

The discount model currently supports:
- `max_redemptions`: Global limit on total uses
- No per-customer restrictions

### Expected behavior

A discount with `max_redemptions_per_customer: 1` would:
- Allow unlimited customers to use the code
- Prevent any single customer from using it more than once
- Track redemptions per customer in the `discount_redemptions` table

### Additional context

This feature would complement the existing `max_redemptions` field and could be used together (e.g., max 1000 total uses AND max 1 per customer).

DavideSegullo:
## 📋 Summary

**Related Issue**: Fixes polarsource/feedback#86

This PR introduces the possibility to add a usage limit per user for a discount code.

## 🎯 What

To achieve this goal, changes were made to the database.

The `customer_id` property was added to the `discount_redemptions` table in order to know which user made each redemption.

The `max_redemptions_per_customer` property was added to the `discounts` table, an integer that indicates how many times a single user can use that discount code.

To do this, checks have been added to the checkout services and utilities that verify whether a code has been redeemed.

**IMPORTANT**: in order to associate the `customer_id` in the `discount_redemptions` table, this association must only be made when the customer is created and associated with the checkout. This only happens when the `POST /client/{client_secret}/confirm` service is invoked, but only after the discount code has been validated, so it was necessary to add the check within the `_confirm_inner` function inside the checkout services. 
Is this practice considered correct? 

## 🤔 Why

We are developing a SaaS where we would like to give some customers a discount code, but we want to ensure that the same user cannot abuse this code by allowing a maximum number of uses.

## 🔧 How

After analyzing the code with Claude Code and understanding how it is structured, I applied the changes suggested by Claude and then refactored the code, moving it to where I thought it made the most sense based on what I understood from analyzing and reading Polar's code.

## 🧪 Testing

- [x] All existing tests pass (`uv run task test` for backend, `pnpm test` for frontend)
- [x] I have added new tests for new functionality
- [x] I have run linting and type checking (`uv run task lint && uv run task lint_types` for backend)

### Test Instructions

1. Create a product
2. Create a discount code and specify a max_redemptions_per_customer value, 1 is fine
3. Create a checkout linked to the product create in the step 1, insert your email and apply the coupon and complete the checkout, it should successfully complete checkout.
4. Create another checkout insert the same email you added previously and apply the coupon, it should fail and tell you that you can no longer redeem that coupon
5. Create another checkout insert another email and apply the coupon,  it should successfully complete checkout.

## 🖼️ Screenshots/Recordings

https://youtu.be/AAtPi55AsVM

## 📝 Additional Notes

I did all the tests using a 100% discount code because I didn't configure the entire Stripe purchase flow. I don't know if this is a problem.

## ✅ Pre-submission Checklist

- [x] All tests pass locally


pieterbeulque:
I like the feature request! It definitely makes sense and I believe this implementation looks fine.

Let me share it internally to move it forward.

pieterbeulque:
Hey Davide, one problem with this approach is that this will only work if the customer ID is provided to the checkout over API when creating a new checkout session. If the customer checks out using the same / similar email (e.g. a `pieter+new@polar.sh`), they'd effectively work around this limitation. You can find similar heuristics in our trial abuse code `server/polar/trial_redemption/service.py`. For this to work in all cases you'd expect it to work, I figure we'd have to do something similar.

DavideSegullo:
> Hey Davide, one problem with this approach is that this will only work if the customer ID is provided to the checkout over API when creating a new checkout session. If the customer checks out using the same / similar email (e.g. a `pieter+new@polar.sh`), they'd effectively work around this limitation. You can find similar heuristics in our trial abuse code `server/polar/trial_redemption/service.py`. For this to work in all cases you'd expect it to work, I figure we'd have to do something similar.

Done, do I need to generate clients so test will work?

pieterbeulque:
Hi Davide, sorry, I missed this. You can regenerate the client by running `pnpm generate` in `clients`, if you're still interested in getting this over the finish line.

Sorry again for letting this go stale. Appreciate you spending your time on this.

PR #9232, opened 2026-01-29
Add `lock_customer_email` flag to checkouts

pieterbeulque:
This won't actually work until a new version of the SDK is released, but wanted to check how we feel about doing something like this.

Related:

- https://github.com/orgs/polarsource/discussions/6172
- https://x.com/levelsio/status/1828836290789294219

I think `lock_customer_email` is a bit of a weird naming but I couldn't come up with anything else.

### DIFF
--- a/server/polar/checkout/schemas.py
+++ b/server/polar/checkout/schemas.py
@@ -1,5 +1,5 @@
 from datetime import datetime
-from typing import Annotated, Any, Literal
+from typing import Annotated, Any, Literal, Self
 
 from annotated_types import Ge, Le, MaxLen, MinLen
 from pydantic import (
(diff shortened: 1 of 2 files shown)


server/migrations/versions/2026-01-28_add_lock_customer_email.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def upgrade() -> None:
+    op.add_column(
+        "checkouts",
+        sa.Column(
+            "lock_customer_email",
+            sa.Boolean(),
+            nullable=False,
+            server_default=sa.sql.expression.false(),
```
pieterbeulque, line 28: Avoid locking the DB when adding this column

psincraian: don't make my mistake, you need to do this in multiple migrations :(

First, you should make it nullable. If you add it non nullable, a new DB check will be created (locking the db) until all the rows have been updated.

pieterbeulque:
For now, the recommended way to do this is by creating a customer for that email first, and then using the customer_id parameter when creating the checkout session. This will have the same effect of pre-filling & disabling the email input field.

PR #9259, opened 2026-01-29
Use UPDATE RETURNING to avoid race conditions

### LINKED ISSUE #7830: Fix IntegrityError race condition in invoice number generation
Fix race condition causing IntegrityError when acquiring next invoice number during high throughput order processing. 
<sub>[Fix invoice number issue](https://app.stilla.ai/m/memo_01ka249m7afxvs39bvfsaesffn). Ask @stilla for more context.</sub>

pieterbeulque:
Resolves #7830 

Increment when retrieving, then use `returning - 1` as the invoice number. It's a bit of a funny gymnastic but nothing too odd, I think.

### DIFF
--- a/server/polar/customer/repository.py
+++ b/server/polar/customer/repository.py
@@ -3,7 +3,7 @@ from collections.abc import AsyncGenerator, Sequence
 from typing import Any
 from uuid import UUID
 
-from sqlalchemy import Select, func, select
+from sqlalchemy import Select, func, select, update
 from sqlalchemy import inspect as orm_inspect
 from sqlalchemy.orm import InstanceState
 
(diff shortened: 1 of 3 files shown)


Yopi (approved):
Nice solution. Have you managed to stress test this to ensure that the db transactions are working as expected?

pieterbeulque:
I didn't, no, because I don't really know how 😬

psincraian:
sorry, missed that! But looks good :D

psincraian:
I think that operation is atomic, so there is no need for a different transaction level

pieterbeulque:
@psincraian Jesper mentioned you may have some tooling laying around to stress test this. Worth doing or not?

PR #9261, opened 2026-01-29
feat: Implement webhook for checkout.expired

### LINKED ISSUE #8693: Add checkout.expired webhook event for checkout expiration
> from [the ticket](https://x.com/nota_codeur/status/2003784302904148165)

Add a `checkout.expired` webhook event that fires when a checkout expires.

Currently, when a checkout expires, there's no webhook event to notify about this state change. The `checkout.updated` event doesn't fire when a checkout expires, which prevents developers from implementing email notifications or other automation when checkouts expire.

This would allow developers to:

* Send reminder emails when checkouts expire
* Track checkout conversion rates more accurately
* Implement custom expiration handling logic

muddlebee:
## 📋 Summary

**Related Issue**: Fixes #8693 

## 🎯 What
Added a checkout.expired webhook event that fires when a checkout expires.

## 🧪 Testing

### Test Instructions

1.
2.
3.

## ✅ Pre-submission Checklist

- [x] All tests pass locally

### DIFF
--- a/server/tests/webhooks/test_webhooks.py
+++ b/server/tests/webhooks/test_webhooks.py
@@ -1,4 +1,3 @@
-from datetime import timedelta
 from typing import cast
 from unittest.mock import MagicMock
 
@@ -10,14 +9,9 @@ from dramatiq import Retry
 from pytest_mock import MockerFixture
 from standardwebhooks.webhooks import Webhook as StandardWebhook
 
-from polar.checkout.repository import CheckoutRepository
-from polar.checkout.tasks import checkout_expired
 from polar.config import settings
 from polar.kit.db.postgres import AsyncSession
-from polar.kit.utils import utc_now
-from polar.models.checkout import CheckoutStatus
 from polar.models.organization import Organization
-from polar.models.product import Product
 from polar.models.subscription import Subscription
 from polar.models.webhook_endpoint import (
     WebhookEndpoint,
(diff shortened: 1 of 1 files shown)


muddlebee:
does this look good? @pieterbeulque  :eyes: 
added tests in further commits

server/polar/checkout/repository.py
```diff
@@ -62,17 +62,22 @@ async def get_by_id_for_update(
...
                 Checkout.deleted_at.is_(None),
                 Checkout.expires_at <= utc_now(),
                 Checkout.status == CheckoutStatus.open,
             )
-            .values(status=CheckoutStatus.expired)
+            .options(*self.get_eager_options())
         )
-        await self.session.execute(statement)
+        result = await self.session.execute(statement)
+        return list(result.scalars().all())
```
muddlebee: The current checkout expiration mechanism uses a bulk update query in expire_open_checkouts().
To trigger webhooks, I have modified this to fetch expired checkouts individually and send webhooks for each one.

frankie567:
@muddlebee Thank you for tackling this @muddlebee, that's great :) 

I would like us to take a slightly different approach to make sure we don't have a too heavy task to run. Here is what I suggest:

* Keep the `UPDATE` query to batch update the checkouts, _but_ add a `RETURNING` clause that gives the ID of the affected checkout
* For each of those returned checkout ID, enqueue a new task (e.g. `checkout.expired`) with the checkout ID.
* Implement that `checkout.expired` task that:
    * Fetch the checkout by the given ID
    * Double-check that the status is expired
    * Trigger the webhook
 
What do you think? Could you tackle that? :)

muddlebee:
@frankie567 thank you for valuable feedback. I was sceptical about the approach too, and your advice makes sense..

muddlebee:
@frankie567 I made the changes as per your suggestions. does this work? :eyes:

server/tests/webhooks/test_webhooks.py
```diff
@@ -255,3 +262,48 @@ async def test_webhook_standard_webhooks_compatible(
     request = route_mock.calls.last.request
     w = StandardWebhook(secret.encode("utf-8"))
     assert w.verify(request.content, cast(dict[str, str], request.headers)) is not None
+
+
+@pytest.mark.asyncio
+async def test_checkout_expired_webhook(
```
frankie567: We can remove that test, doesn't add much given the other tests we already have

muddlebee:
@frankie567 fixed! thank you for your time and guidance :pray: appreciate it!

PR #9291, opened 2026-01-30
Skip renewal emails for completely free subscriptions

### LINKED ISSUE #9287: Option to disable "Subscription cycled" email only for free tier
## Is your feature request related to a problem?

I want to automatically subscribe my users to a free plan so I can use meters to track their usage without forcing them to go through a subscription screen. Currently, sending renewal emails for $0 invoices does not make sense and may confuse users. 

However, I still want to ensure that paid customers continue to receive their renewal emails.

## Describe the solution you'd like

An option to specifically disable "Subscription Cycled" emails for free products/plans only.

## Describe alternatives you've considered

I have looked into disabling all subscription emails, but that would prevent my paid customers from receiving necessary billing information.

pieterbeulque:
Fixes #9287

Sending subscription renewals for free subscriptions is confusing for customers, especially if they've been 'invisibly' put on a free plan over API for easier entitlement management. Hiding it for everyone feels better than adding yet another toggle in the settings, in my opinion.

I decided to keep the `subscription_create` email for free subscriptions, customers may rely on that email to get access to their benefits. I think that's still a bit of a weird use case for the 'invisible free plan over API' use case, but then for now these merchants would have to disable the `subscription_create` email altogether, and implement their own communication. A fair trade-off, for now, I think.

### DIFF
--- a/server/polar/order/service.py
+++ b/server/polar/order/service.py
@@ -79,7 +79,6 @@ from polar.payment_method.repository import PaymentMethodRepository
 from polar.payment_method.service import payment_method as payment_method_service
 from polar.product.guard import (
     is_custom_price,
-    is_free_price,
     is_seat_price,
     is_static_price,
 )
@@ -1484,7 +1483,7 @@ class OrderService:
             subscription = order.subscription
             if subscription is not None:
                 is_free_product_price = all(
-                    is_free_price(spp.product_price)
+                    spp.product_price.is_free
                     for spp in subscription.subscription_product_prices
                 )
                 if is_free_product_price:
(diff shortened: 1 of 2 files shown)


pieterbeulque:
@copilot This PR should use the new `price.is_free` added in #10017.

It should also exclude free products from the emails added in #10103.

server/polar/order/service.py
```diff
@@ -1385,6 +1390,20 @@ async def send_confirmation_email(
...
 
+        # Skip subscription cycle emails for completely free subscriptions
+        if order.billing_reason in (
+            OrderBillingReasonInternal.subscription_cycle,
+            OrderBillingReasonInternal.subscription_cycle_after_trial,
+        ):
+            subscription = order.subscription
+            if subscription is not None:
+                is_free_product_price = all(
+                    is_free_price(spp.product_price)
```
frankie567: Should we update this with the new `is_free` method on price? Otherwise, good to go.

frankie567: Too late :p

PR #9322, opened 2026-02-02
Mark products as private

pieterbeulque:
Private products can only be purchased directly and won't show up in plan switchers.

Loom: https://www.loom.com/share/9d9c42c8ce2d4847832a71fdba44d2d0

### DIFF
--- a/server/polar/product/service.py
+++ b/server/polar/product/service.py
@@ -144,21 +144,6 @@ class ProductService:
         )
         return await repository.get_one_or_none(statement)
 
-    async def get_embed(
-        self, session: AsyncReadSession, id: uuid.UUID
-    ) -> Product | None:
-        repository = ProductRepository.from_session(session)
-        statement = (
-            repository.get_base_statement()
-            .where(
-                Product.id == id,
-                Product.is_archived.is_(False),
-                Product.visibility != ProductVisibility.draft,
-            )
-            .options(selectinload(Product.product_medias))
-        )
-        return await repository.get_one_or_none(statement)
-
     async def create(
         self,
         session: AsyncSession,


server/polar/checkout/service.py
```diff
@@ -620,6 +621,18 @@ async def client_create(
                 ]
             )
 
+        if product.visibility == ProductVisibility.draft:
```
pieterbeulque, line 623: I didn't add draft support in the front-end yet, but this is already a safety measure to not allow checkout sessions for draft products

server/polar/product/endpoints.py
```diff
@@ -74,6 +76,10 @@ async def list(
         title="BenefitID Filter",
         description="Filter products granting specific benefit.",
     ),
+    visibility: builtins.list[ProductVisibility] | None = Query(
+        default=[ProductVisibility.public, ProductVisibility.private],
```
frankie567: I usually avoid to have defaults on those query parameters. My rationale behind this is it's difficult to avoid the default if we need to (e.g. in this case, it would require to explicitly pass all the visibility values) while it's easy for us to add a filtering on the frontend side.

server/polar/customer_portal/service/organization.py
```diff
@@ -18,7 +18,13 @@ async def get_by_slug(
                 Organization.slug == slug,
             )
             .options(
-                selectinload(Organization.products).options(
+                selectinload(
+                    Organization.products.and_(
+                        Product.deleted_at.is_(None),
+                        Product.is_archived.is_(False),
+                        Product.visibility == ProductVisibility.public,
```
frankie567, line 25: I think we should also enforce that on the update endpoint in `server/polar/customer_portal/endpoints/subscription.py`

server/polar/product/service.py
frankie567: Just noticed this method is completely unused nowadays. If we can take this chance to remove it :)

PR #9338, opened 2026-02-03
improve notification conversion

emilwidlund:


pieterbeulque:
I wouldn't do this, except for when formatted pricing is `$0.00`, then it looks silly (I believe Sebastian removed it on the email too in that specific case only).

I don't think it'll hit the same dopamine receptors if the amount is no longer in there 😁 + it helps you know from a glance which plan (if you have multiple). And I think people like sharing a screenshot of a bunch of these notifications to flex their money muscles.

emilwidlund:
> I wouldn't do this, except for when formatted pricing is `$0.00`, then it looks silly (I believe Sebastian removed it on the email too in that specific case only).
> 
> I don't think it'll hit the same dopamine receptors if the amount is no longer in there 😁 + it helps you know from a glance which plan (if you have multiple). And I think people like sharing a screenshot of a bunch of these notifications to flex their money muscles.

Fair! I've been a bit back & forth on this too. I think I'll just go back and address the parenthesis thing which looks really ugly.

PR #9419, opened 2026-02-06
benefits: Revoke benefits immediately on product change

### LINKED ISSUE #9410: Fix event ordering in subscription upgrade benefit grant flow
> from [the example](https://polar.sh/dashboard/kentauros-ai/customers/933faf52-9a83-4f44-a47b-e7faebd133d7) and the [slack conversation](https://polar-olw4768.slack.com/archives/C096F9P7E0P/p1770380560991229)

## Problem

When a subscription is upgraded, the benefit grant events are firing in the wrong order, causing incorrect meter credit calculations.

**Expected flow:**

1. Old benefits revoked
2. Meter reset
3. New benefits granted

**Current flow:**

1. Meter reset
2. New benefits granted
3. Old benefits revoked ❌

## Impact

This results in negative credit balances. Example:

* Initial purchase: 100 credits granted
* Upgrade: Meter reset, 7500 credits granted, then 100 credits revoked
* Result: -100 total credits (should be 7500)

Yopi:
Fix #9410

Today we have an issue where a meter reset will happen, and benefit revocation and benefit grants will happen after the fact. This means that someone might lose credits if they are granted 100 credits today, the meter resets, they receive 1000 credits and get the 100 credits removed. This means that they end up with 900 credits.

If we instead revoke the benefits, reset the meters and then grant the benefits they will end up with the 1000 credits as expected.

The change here is that the benefit revocation happens synchronously instead of in a task. The ideal would be to chain tasks together, or order them, but it's not possible today.

### DIFF
--- a/server/polar/benefit/grant/service.py
+++ b/server/polar/benefit/grant/service.py
@@ -7,6 +7,7 @@ import structlog
 from sqlalchemy import select
 from sqlalchemy.orm import joinedload
 
+from polar.benefit.repository import BenefitRepository
 from polar.customer.repository import CustomerRepository
 from polar.event.service import event as event_service
 from polar.event.system import BenefitGrantMetadata, SystemEvent, build_system_event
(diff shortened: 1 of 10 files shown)


frankie567:
I understand the problem, but this has quite a big impact on the code and potential performance issues in case of products with loads of benefits (think Lexington Themes). It'll also probably fail the whole process if the benefit revocation logic fails (e.g. GitHub or Discord error). IMO, it's crucial to keep benefit granting logic independent. 

I think we can solve that by setting up a result backend in Dramatiq, and then use a group or pipeline: https://dramatiq.io/reference.html#dramatiq.pipeline We could set up a dedicated task for that case that triggers that pipeline to make sure everything is done in the correct order.

Yopi:
> I understand the problem, but this has quite a big impact on the code and potential performance issues in case of products with loads of benefits (think Lexington Themes). It'll also probably fail the whole process if the benefit revocation logic fails (e.g. GitHub or Discord error). IMO, it's crucial to keep benefit granting logic independent.
> 
> I think we can solve that by setting up a result backend in Dramatiq, and then use a group or pipeline: https://dramatiq.io/reference.html#dramatiq.pipeline We could set up a dedicated task for that case that triggers that pipeline to make sure everything is done in the correct order.



> I understand the problem, but this has quite a big impact on the code and potential performance issues in case of products with loads of benefits (think Lexington Themes). It'll also probably fail the whole process if the benefit revocation logic fails (e.g. GitHub or Discord error). IMO, it's crucial to keep benefit granting logic independent.
> 
> I think we can solve that by setting up a result backend in Dramatiq, and then use a group or pipeline: https://dramatiq.io/reference.html#dramatiq.pipeline We could set up a dedicated task for that case that triggers that pipeline to make sure everything is done in the correct order.

Thank you @frankie567. That was my feeling as well, but I was uncertain if it would become too much of a problem or not. I was also considering creating a task to do this specific thing. Dramatiq pipelines sounds like it could work? I'll look into it a bit more!

PR #9427, opened 2026-02-06
clients/web: pass default organization currency when creating a fixed amount discount

frankie567:


frankie567:
@copilot Please take this to the finish line, there apparently is TypeScript errors.

PR #9428, opened 2026-02-06
Issue #6836 - fix: sidebar in mobile view does not close automatically upon selection

### LINKED ISSUE #6836: dashboard: sidebar in mobile view does not close automatically upon selection
> From [the ticket](https://app.plain.com/workspace/w_01JE9TRRX9KT61D8P2CH77XDQM/thread/th_01K5785PE9TYJ8X28X3KB8T6DZ/)

The sidebar in the mobile view does not automatically close when a user makes the selection, making it a 2-step process to move between the Polar dashboard.

Image

ChetanAsabe:
Fix #6836 

Fixed an issue where the mobile sidebar stayed open after selecting a navigation item.


pieterbeulque:
Sorry for letting this PR go stale.

While the current UX feels broken, it does allow you to immediately go to a subsection of any item in the navigation, while with this you'd always have to open the sidebar twice, so this is also not ideal. I'm not sure what the best fix would be.

PR #9434, opened 2026-02-09
 benefits: Revoke benefits immediately on product change 

### LINKED ISSUE #9410: Fix event ordering in subscription upgrade benefit grant flow

Yopi:
Fix #9410

Today we have an issue where a meter reset will happen, and benefit revocation and
benefit grants will happen after the fact. This means that someone might lose credits
if they are granted 100 credits today, the meter resets, they receive 1000 credits and
get the 100 credits removed. This means that they end up with 900 credits.

If we instead revoke the benefits, reset the meters and then grant the benefits they will
end up with the 1000 credits as expected.

We utilise the Dramatiq groups+pipelines to ensure that the benefits are revoked first,
and then the meters are reset and finally the new benefits are enqueued.

Alternative approach to #9419

### DIFF
--- a/server/polar/benefit/tasks.py
+++ b/server/polar/benefit/tasks.py
@@ -10,11 +10,14 @@ from polar.exceptions import PolarTaskError
 from polar.logging import Logger
 from polar.models.benefit_grant import BenefitGrantScopeArgs
 from polar.product.repository import ProductRepository
+from polar.subscription.repository import SubscriptionRepository
+from polar.subscription.service import subscription as subscription_service
 from polar.worker import (
     AsyncSessionMaker,
     RedisMiddleware,
     TaskPriority,
     actor,
+    enqueue_job,
     get_retries,
 )
 
@@ -90,6 +93,32 @@ async def enqueue_benefits_grants(
 
         await benefit_grant_service.enqueue_benefits_grants(
             session, task, customer, product, member_id=member_id, **resolved_scope
+        )
+
+
+@actor(actor_name="benefit.enqueue_grants", priority=TaskPriority.MEDIUM)
+async def benefit_enqueue_grants(
+    customer_id: uuid.UUID,
+    grant_benefit_ids: list[uuid.UUID],
+    member_id: uuid.UUID | None = None,
+    **scope: Unpack[BenefitGrantScopeArgs],
+) -> None:
+    if subscription_id := scope.get("subscription_id"):
+        async with AsyncSessionMaker() as session:
+            repository = SubscriptionRepository.from_session(session)
+            subscription = await repository.get_by_id(
+                subscription_id, options=repository.get_eager_options()
+            )
+            if subscription:
+                await subscription_service.reset_meters(session, subscription)
+
+    for benefit_id in grant_benefit_ids:
+        enqueue_job(
+            "benefit.grant",
+            customer_id=customer_id,
+            benefit_id=benefit_id,
+            member_id=member_id,
+            **scope,
         )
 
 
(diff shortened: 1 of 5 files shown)


server/polar/worker/_broker.py
```diff
@@ -141,10 +146,17 @@ def get_broker() -> dramatiq.Broker:
...
+    rate_limiter_backend = RateLimiterBackend(url=settings.redis_url)
+
     middleware_list = [
         # Infrastructure & async support
         middleware.ShutdownNotifications(),
         middleware.AsyncIO(),
+        # Results backend for pipeline/group support
+        Results(backend=result_backend, result_ttl=60_000),
+        # Group completion callbacks for orchestrating task sequences
+        GroupCallbacks(rate_limiter_backend),
```
frankie567, line 159: Wondering if we should put those two after the observability middlewares, so we might get their logging.

server/polar/benefit/tasks.py
```diff
@@ -93,6 +96,32 @@ async def enqueue_benefits_grants(
         )
 
 
+@actor(actor_name="benefit.enqueue_grants", priority=TaskPriority.MEDIUM)
```
frankie567, line 99: Not sure if this naming is self-explanatory enough, but I don't really have better to propose 😅

Yopi: We could do `reset_meters_and_enqueue_grants` or `reset_meters_enqueue_grants_after_revocation` for extra clarity? I also think it could be good to make it a bit more differnated than the singular `benefit.enqueue_grant`

server/polar/worker/_broker.py
Yopi: Will do!

server/polar/benefit/tasks.py
frankie567: Yeah, a bit verbose but `reset_meters_and_enqueue_grants` might be clearer.

frankie567:
@Yopi Another thing: shouldn't we remove that call then in `SubscriptionService.create_or_update_from_checkout`?

https://github.com/polarsource/polar/blob/86a42f0727352794c20fe65219fa9ead5014d024/server/polar/subscription/service.py#L632-L633

Yopi:
> @Yopi Another thing: shouldn't we remove that call then in `SubscriptionService.create_or_update_from_checkout`?
> 
> https://github.com/polarsource/polar/blob/86a42f0727352794c20fe65219fa9ead5014d024/server/polar/subscription/service.py#L632-L633

Ah, you're absolutely right. That makes me think if this is putting the abstraction in the wrong place, since the meter reset is on the `Subscription` layer, but this PR moves it into the `Benefit` layer. It's a bit unavoidable since we need it to be sandwiched between revoke and grant. I think it will be correct to just remove that call

frankie567:
> It's a bit unavoidable since we need it to be sandwiched between revoke and grant. I think it will be correct to just remove that call

Yeah, I thought about that as well; it's probably the most straightforward solution.

PR #9455, opened 2026-02-09
Extract name from Stripe's confirmation token

sebastianekstrom:
As a part of simplifying the checkout flow this PR will add backend support for fetching the user name from Stripe's confirmation token when using Apple/Google Pay.

Here's how it works today:

1. The user chooses Apple/Google Pay as a payment service
2. They enter the cardholder name
3. They make the purchase

With this PR (and frontend changes in the `feature/apple-pay-optimization` branch) the flow will instead be:

1. The user chooses Apple/Google Pay as a payment service
2. They make the purchase
3. We fetch the cardholder from Stripe's checkout token

Screenshot 2026-02-05 at 15 29 53

### DIFF
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -1057,6 +1057,31 @@ class CheckoutService:
         ):
             raise PaymentNotReady()
 
+        # For wallet payments (Apple Pay, Google Pay), we hide the customer name field
+        # for better UX and instead extract the name from Stripe's confirmation token.
+        if (
+            checkout.payment_processor == PaymentProcessor.stripe
+            and checkout_confirm.confirmation_token_id is not None
+            and checkout.customer_name is None
+        ):
+            try:
+                confirmation_token = await stripe_service.get_confirmation_token(
+                    checkout_confirm.confirmation_token_id
+                )
+                if (
+                    confirmation_token.payment_method_preview is not None
+                    and confirmation_token.payment_method_preview.billing_details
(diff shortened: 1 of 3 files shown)


server/polar/checkout/service.py
```diff
@@ -1102,6 +1102,31 @@ async def _confirm_inner(
         if len(errors) > 0:
             raise PolarRequestValidationError(errors)
 
+        # For wallet payments (Apple Pay, Google Pay), we hide the customer name field for better UX and
+        # instead extract the name from Stripe's confirmation token
+        if (
```
frankie567: I think we should move that before the `required_fields` check. This way, even if we fail to retrieve the name from the checkout, we'll trigger the validation.

server/tests/checkout/test_service.py
```diff
@@ -4884,6 +4883,199 @@ async def test_payment_not_ready_with_account_setup_complete(
...
+                        "city": "New York",
+                        "state": "US-NY",
+                        "country": "US",
+                    },
+                }
+            ),
+        )
+
+        assert checkout.status == CheckoutStatus.confirmed
+        assert checkout.customer_name is None
```
frankie567: Given my proposal above, those two last tests should expect a validation error to be raised

server/polar/checkout/service.py
sebastianekstrom: Good catch, fixed!

PR #9465, opened 2026-02-10
subscription: We overwrite canceled_at on revocation of subscriptions

Yopi:
This makes the canceled_at be inconsistent with the cancelation event. Since the subscription wasn't actually canceled when it was revoked, it makes more sense to _not_ overwrite the value, if it is already set.

### DIFF
--- a/server/polar/subscription/service.py
+++ b/server/polar/subscription/service.py
@@ -1724,7 +1724,8 @@ class SubscriptionService:
         previous_is_canceled = subscription.canceled
 
         now = utc_now()
-        subscription.canceled_at = now
+        if subscription.canceled_at is None:
+            subscription.canceled_at = now
 
         if customer_reason:
             subscription.customer_cancellation_reason = customer_reason


frankie567:
I'm not sure I get it. The case would be:

* The subscription is first canceled at period end -> we set `canceled_at`
* Then, the subscription is revoked immediately  -> we set again `canceled_at`

To me, it makes sense since this is a new operation happening on the subscription?

Yopi:
> I'm not sure I get it. The case would be:
> 
>     * The subscription is first canceled at period end -> we set `canceled_at`
> 
>     * Then, the subscription is revoked immediately  -> we set again `canceled_at`
> 
> 
> To me, it makes sense since this is a new operation happening on the subscription?

Ahh, sorry. I didn't trace this enough, because I assumed it was:
* The subscription is first canceled at period end -> we set `canceled_at`
* Period end happens -> cancelation is performed and we set it again.

But in that case, I think the mismatch is coming from

* The subscription is first canceled at period end -> we set `canceled_at`
* Then, the subscription is revoked immediately  -> we set again `canceled_at`. However: It is already canceled so we don't generate a new `subscription.canceled` event. 

This causes a mismatch between the timestamp of `subscription.canceled` and the `canceled_at` on the subscription. We also don't send an additional webhook when it actually gets canceled. 

All of that might be intentional though?

frankie567:
Ah ok I see 🤔 

Since it's another manual action from the merchant, I think the actual problem is the lack of event and webhook. IMO, the correct thing to do semantically is indeed to generate those, in which case it makes sense to reset `canceled_at`. What do you think?

Yopi:
> Ah ok I see 🤔
> 
> Since it's another manual action from the merchant, I think the actual problem is the lack of event and webhook. IMO, the correct thing to do semantically is indeed to generate those, in which case it makes sense to reset `canceled_at`. What do you think?

Yeah, no I agree with you that it makes sense to set the `canceled_at`. Maybe we could just mark it as not previously canceled when they do it manually. I'm not sure if it would be correct if we were to _always_ generate those in this flow (i.e. remove the check for if previously canceled).

Yopi:
Alternative https://github.com/polarsource/polar/pull/9473

PR #9511, opened 2026-02-11
Add Hungarian

pieterbeulque:
cc @simonxabris - really appreciate you doing this! If it's not that good in general, just let me know and we can iterate on the prompt or use a different LLM.

### DIFF
--- a/clients/packages/i18n/src/locales/hu.ts
+++ b/clients/packages/i18n/src/locales/hu.ts
@@ -3,7 +3,7 @@ export default {
     footer: {
       poweredBy: 'Működteti',
       merchantOfRecord:
-        'Ezt a megrendelést online viszonteladónk és rekordkereskedőnk, a Polar dolgozza fel, aki a megrendeléssel kapcsolatos kérdéseket és visszaküldéseket is kezeli.',
+        'Ezt a megrendelést online viszonteladónk és hivatalos kereskedőnk, a Polar dolgozza fel, aki a megrendeléssel kapcsolatos kérdéseket és visszaküldéseket is kezeli.',
     },
     form: {
       email: 'E-mail cím',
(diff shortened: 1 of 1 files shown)


clients/packages/i18n/src/locales/hu.ts
```diff
@@ -0,0 +1,200 @@
...
+    },
+  },
+  benefitTypes: {
+    usage: 'Használat',
+    license_keys: 'Licenckulcsok',
+    github_repository: 'GitHub tároló hozzáférés',
+    discord: 'Discord meghívó',
+    downloadables: 'Fájlletöltések',
+    custom: 'Egyedi',
+    meter_credit: 'Mérőegység kreditek',
```
simonxabris: I'm not so sure about this one, it's grammatically correct, but really not natural at all. I've never seen this used like this. Maybe something like `Használat alapú kreditek`, which translates to `Usage based credits` would be better, but I'm not sure if on your side meter credit and usage based can be used interchangeably.

clients/packages/i18n/src/locales/hu.ts
```diff
@@ -0,0 +1,200 @@
+export const hu = {
+  checkout: {
+    footer: {
+      poweredBy: 'Működteti',
+      merchantOfRecord:
+        'Ezt a megrendelést online viszonteladónk és rekordkereskedőnk, a Polar dolgozza fel, aki a megrendeléssel kapcsolatos kérdéseket és visszaküldéseket is kezeli.',
```
simonxabris: So the first part up until `, a Polar...` is not how you'd say merchant of record in hungarian. I looked into it and this would be more in line with how we would say this:

`Ezt a megrendelést online viszonteladónk és hivatalos kereskedőnk, a Polar...` which means more like `...our official merchant/vendor, Polar...`

Disclaimer: If this has any legal ramifications, I'm not sure if my alternative is the exact legal jargon, but I'm sure that what you have isn't either.

clients/packages/i18n/src/locales/hu.ts
pieterbeulque: yes, "Usage based credit" would be a fair alternative, so if that sounds better, let's go with that!

pieterbeulque:
@simonxabris Processed your feedback and merged! Feel free to take another look at it if you have the time. Totally fine if not, appreciate the help! ❤️

PR #9519, opened 2026-02-12
Add configurable retry delay for webhook event delivery

pieterbeulque:
## 📋 Summary

**Related Issue**: Fixes #

This PR adds a configurable retry delay for webhook event delivery, allowing better control over retry timing and reducing potential thundering herd issues.

## 🎯 What

- Added `WEBHOOK_RETRY_INTERVAL_MILLISECONDS` configuration setting (default: 3,500ms)
- Updated webhook retry logic to use the configured delay instead of immediate retries
- Applied the delay to both initial delivery failures and error-based retries

## 🤔 Why

Without a configured delay, webhook retries happen immediately, which can:
- Cause rapid retry storms if the webhook endpoint is temporarily unavailable
- Lead to unnecessary load spikes on both the webhook service and target endpoints
- Reduce the effectiveness of retry logic by not giving endpoints time to recover

Adding a configurable delay allows for exponential backoff strategies and more graceful handling of transient failures.

## 🔧 How

1. Added `WEBHOOK_RETRY_INTERVAL_MILLISECONDS: int = 3_500` to the Settings class in `config.py`
2. Updated `_webhook_event_send()` in `webhook/tasks.py` to pass the delay parameter to both `Retry()` calls:
   - When delivery is not attempted (line 96)
   - When delivery fails with an exception (line 175)

## 🧪 Testing

### Test Instructions

1. Verify webhook retry behavior with the new delay setting
2. Test that the delay is applied correctly in both retry scenarios
3. Confirm that the default 3.5 second delay is reasonable for your use case

## 📝 Additional Notes

The default delay of 3,500 milliseconds (3.5 seconds) can be adjusted via the `WEBHOOK_RETRY_INTERVAL_MILLISECONDS` environment variable to suit different deployment requirements.

## ✅ Pre-submission Checklist

- [ ] All tests pass locally

https://claude.ai/code/session_01VfGnjeZYAS3QQKjySD3SSN

### DIFF
--- a/server/polar/webhook/tasks.py
+++ b/server/polar/webhook/tasks.py
@@ -30,13 +30,9 @@ from .service import webhook as webhook_service
 log: Logger = structlog.get_logger()
 
 
-_WEBHOOK_ORDERING_MAX_RETRIES = 60_000 // settings.WEBHOOK_RETRY_INTERVAL_MILLISECONDS
-_WEBHOOK_TOTAL_MAX_RETRIES = settings.WEBHOOK_MAX_RETRIES + _WEBHOOK_ORDERING_MAX_RETRIES
-
-
 @actor(
     actor_name="webhook_event.send",
-    max_retries=_WEBHOOK_TOTAL_MAX_RETRIES,
+    max_retries=settings.WEBHOOK_MAX_RETRIES,
     queue_name=TaskQueue.WEBHOOKS,
 )
 async def webhook_event_send(webhook_event_id: UUID, redeliver: bool = False) -> None:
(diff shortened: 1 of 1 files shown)


Yopi:
I think this will work. Lets try it and keep an eye on the queue depth + max age

server/polar/webhook/tasks.py
```diff
@@ -172,7 +176,7 @@ async def _webhook_event_send(
             enqueue_job("webhook_event.failed", webhook_event_id=webhook_event_id)
         # Retry
         else:
-            raise Retry() from e
+            raise Retry(delay=settings.WEBHOOK_RETRY_INTERVAL_MILLISECONDS) from e
```
frankie567, line 179: Worth to note that by doing this, we won't have exponential backoff: the delay will always be WEBHOOK_RETRY_INTERVAL_MILLISECONDS. If we pass nothing, Dramatiq automatically computes an exponential backoff delay based on the number of attempts + a jitter: https://github.com/Bogdanp/dramatiq/blob/634277f568c47fcbb8edbfd502328badc42b88bd/dramatiq/middleware/retries.py#L139-L145

pieterbeulque: Yes, that's intentional for when it's not the first event, but we shouldn't do it with networking errors. Jesper & I are still working out the details on Slack 😬

pieterbeulque:
Replaced by #9525

PR #9524, opened 2026-02-12
Add Portuguese (Brazil)

pieterbeulque:
@LukeberryPi would you mind taking a look at `pt-BR.ts` in the diff? If it's a complete miss, let me know and we can tweak the system prompt and do a re-run.


pieterbeulque:
We should actually use these for all Portuguese and maybe look into a `pt-PT` specific override too. Stripe Elements considers both `pt` and `pt-BR` as Portuguese (Brazil) and doesn't have a European / African locale.

`isAcceptedLocale` also isn't working for `pt-BR` yet—it has a `.split('-')`.

pieterbeulque:
Replaced by #9538

PR #9525, opened 2026-02-12
Refactor webhook event ordering with dynamic retry delays

pieterbeulque:
The root cause of the problem is that we have two completely different error paths that were going through the same (exponential back-off) retry logic, causing frequent unnecessary long delays when delivering webhooks.

From a user report:

> Following up this topic, I conducted the initial purchase in the production environment and observed a significant delay of 27 seconds between the subscription.updated and the order.paid events.

On top of that, the first of these two was already delayed 24 seconds from when the webhook send was queued (because another webhook event still had to be delivered). 24 seconds boils down to the 4th attempt in the exponential back-off.

---

In this PR, we opt out of the exponential back-off logic when retrying because of trying to guarantee the webhook delivery order. We attempt to work around unnecessary retries by waiting for {event backlog count} * 300 milliseconds (the p95 response time for webhook endpoints is 243ms).

Then, for webhook delivery failed attempts, we rely on the database state to figure out the retries, no longer retrying on Dramatiq's retry count (since that could include the retries for delivery order). This is still done on a exponential back-off basis, but we need to calculate the delay manually, otherwise it'd take the earlier retries into account.

Finally, we calculate a `max_retries` from these settings to avoid a webhook event going rogue, but it should never hit this case.

### DIFF
--- a/server/polar/config.py
+++ b/server/polar/config.py
@@ -71,8 +71,8 @@ class Settings(BaseSettings):
     PROMETHEUS_REMOTE_WRITE_INTERVAL: Annotated[int, Ge(1)] = 15  # seconds
 
     WEBHOOK_MAX_RETRIES: int = 10
-    WEBHOOK_NOT_LATEST_DELAY_MILLISECONDS: int = 300  # p95 is 236ms
-    WEBHOOK_ORDERING_AGE_LIMIT: timedelta = timedelta(minutes=1)
+    WEBHOOK_FIFO_GUARD_DELAY_MS: int = 300  # p95 is 236ms
+    WEBHOOK_FIFO_GUARD_MAX_AGE: timedelta = timedelta(minutes=1)
     WEBHOOK_EVENT_RETENTION_PERIOD: timedelta = timedelta(days=30)
     WEBHOOK_FAILURE_THRESHOLD: int = 10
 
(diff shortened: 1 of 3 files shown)


server/polar/webhook/tasks.py
```diff
@@ -44,7 +54,7 @@ async def webhook_event_send(webhook_event_id: UUID, redeliver: bool = False) ->
 
 @actor(
     actor_name="webhook_event.send.v2",
```
frankie567: May I ask you to take this opportunity to remove that one? No longer in use ☺️

server/polar/config.py
```diff
@@ -71,6 +71,8 @@ class Settings(BaseSettings):
     PROMETHEUS_REMOTE_WRITE_INTERVAL: Annotated[int, Ge(1)] = 15  # seconds
 
     WEBHOOK_MAX_RETRIES: int = 10
+    WEBHOOK_NOT_LATEST_DELAY_MILLISECONDS: int = 300  # p95 is 236ms
```
frankie567: We need to find a better name for this :D I would suggest `WEBHOOK_ORDERING_DELAY_MILLISECONDS` to match with its brother below, even if not 100% clear yet.

server/polar/webhook/tasks.py
pieterbeulque: It's gone!

server/polar/config.py
pieterbeulque: Yes, good point.

I updated to:

```py
WEBHOOK_FIFO_GUARD_DELAY_MS: int = 300  # p95 is 236ms
WEBHOOK_FIFO_GUARD_MAX_AGE: timedelta = timedelta(minutes=1)
```

which I don't 100% like, but it feels more clear about what it's for.

PR #9539, opened 2026-02-13
Add Korean

pieterbeulque:


hongminpark:
@pieterbeulque 
Looks good to me (as a Korean🇰🇷) 
Thank you so much!! 🙏

pieterbeulque:
@bytaesu would you mind taking a second look at the translations? It's hard for us (Polar team) to verify these translations since it's far from our comfort zone 😁

bytaesu:
> would you mind taking a second look at the translations? It's hard for us (Polar team) to verify these translations since it's far from our comfort zone 😁

Sure!! I'll check today 😄

clients/packages/i18n/src/locales/ko.ts
```diff
@@ -0,0 +1,227 @@
+export default {
+  checkout: {
+    footer: {
+      poweredBy: '제공:',
```
bytaesu: It sounds natural, and in fact, many Korean services use “Powered by” as-is without translating it.

clients/packages/i18n/src/locales/ko.ts
```diff
@@ -0,0 +1,227 @@
...
+        line1: '주소',
+        line2: '상세 주소',
+        postalCode: '우편번호',
+        city: '도시',
+        country: '국가',
+        state: '주',
+        province: '도',
+        stateProvince: '주 / 도',
+      },
+      taxId: '납세자 번호',
```
bytaesu: If this field is intended for businesses, "납세자 번호" -> “사업자등록번호” would be more natural.

+) "사업자등록번호" means "Business Registration Number (BRN)"

clients/packages/i18n/src/locales/ko.ts
```diff
@@ -0,0 +1,227 @@
...
+        label: '청구지 주소',
+        line1: '주소',
+        line2: '상세 주소',
+        postalCode: '우편번호',
+        city: '도시',
+        country: '국가',
+        state: '주',
+        province: '도',
+        stateProvince: '주 / 도',
+      },
```
bytaesu, line 29:
```suggestion
      billingAddress: {
        label: '청구지 주소',
        line1: '주소',
        line2: '상세 주소',
        postalCode: '우편번호',
        city: '시/군/구',
        country: '국가',
        state: '시/도',
        province: '시/도',
        stateProvince: '시/도',
      },
```

Korea doesn't have a state or province concept. I'm not sure how this is reflected in the UI, but this seems the most natural option.

bytaesu: “주” isn't used at all in Korea

pieterbeulque: State & province are only shown when the checkout country is respectively USA or Canada. So, ideally, the translation is what you'd call e.g. the _state_ of Michigan or the _province_ of Québec.

The `stateProvince` key can be shown for other countries that may require a state / region / province / … part to the address.

MikeYi314: "시/도" works as a catch-all, but since "state" and "province" only appear for US/Canada checkouts respectively, it might be worth differentiating:

-> state: "주" (as in 캘리포니아주, 텍사스주. this is the standard Korean translation for US states)
->  province: "주" (퀘벡주, 온타리오주)
-> stateProvince: "시/도" (this one makes sense as a generic fallback for other countries)

Korean users filling in a US address would more naturally associate "주" with a US state than "시/도", which feels like a Korean administrative division (서울특별시, 경기도).

We need "시/도" for `stateProvince` because some cities(시) are not in state or province ("도") 
Ex) Seoul / Incheon does not belong to any state or province.

clients/packages/i18n/src/locales/ko.ts
MikeYi314: @bytaesu is correct. "납세자 번호" would work (in legal sense), but "사업자등록번호” is more widely used.

clients/packages/i18n/src/locales/ko.ts
```diff
@@ -0,0 +1,266 @@
...
+        other: '#개 추가 혜택',
+        _mode: 'plural',
+      },
+      showMoreBenefits: {
+        '=1': '혜택 #개 더 보기',
+        other: '혜택 #개 더 보기',
+        _mode: 'plural',
+      },
+      showLess: '접기',
+      granting: '혜택 제공 중...',
```
MikeYi314: It would be more natural if it was "헤택 적용 중..."
제공 means "providing" but in this sense it's more like applying.

clients/packages/i18n/src/locales/ko.ts
```diff
@@ -0,0 +1,266 @@
...
+        statusRevoked: '취소됨',
+        statusDisabled: '비활성화됨',
+        usage: '사용량',
+        validations: '검증',
+        validatedAt: '검증 일시',
+        neverValidated: '검증된 적 없음',
+        expiryDate: '만료일',
+        noExpiry: '만료 없음',
+        activations: '활성화',
+        activationDeleted: '라이선스 키 활성화 삭제됨',
```
MikeYi314: This should be "라이선스 키 비활성화됨".

"라이선스 키 활성화 삭제됨" => In Korean "활성화" cannot be a noun.
More natural to say "License Key is deactivated", which is "라이선스 키 비활성화됨".

clients/packages/i18n/src/locales/ko.ts
```diff
@@ -0,0 +1,266 @@
...
+        status: '상태',
+        statusGranted: '발급됨',
+        statusRevoked: '취소됨',
+        statusDisabled: '비활성화됨',
+        usage: '사용량',
+        validations: '검증',
+        validatedAt: '검증 일시',
+        neverValidated: '검증된 적 없음',
+        expiryDate: '만료일',
+        noExpiry: '만료 없음',
```
MikeYi314: "만료일 없음" is more natural. 만료 is not a noun. (It's a verb)

clients/packages/i18n/src/locales/ko.ts
```diff
@@ -0,0 +1,266 @@
...
+        activations: '활성화',
+        activationDeleted: '라이선스 키 활성화 삭제됨',
+        activationDeletedDescription: '활성화가 성공적으로 삭제되었습니다',
```
MikeYi314: Same here. We should say 

"라이선스 키가 비활성화되었습니다."

clients/packages/i18n/src/locales/ko.ts
```diff
@@ -0,0 +1,266 @@
...
+        activationDeactivationFailed: '활성화 비활성화 실패',
```
MikeYi314: Same here. We should say :

"라이선스 키 비활성화 실패"

clients/packages/i18n/src/locales/ko.ts
```diff
@@ -0,0 +1,266 @@
...
+    },
+    loading: {
+      processingOrder: '주문 처리 중...',
+      processingPayment: '결제 처리 중',
+      paymentSuccessful: '결제 성공! 상품을 준비하고 있습니다...',
+      confirmationTokenFailed:
+        '확인 토큰 생성에 실패했습니다. 나중에 다시 시도해 주세요.',
+    },
+    cta: {
+      startTrial: '체험 시작',
```
MikeYi314: Let's make it "체험 시작하기", which is consistent to the 3 below (subscribeNow, payNow, getFree)

~~하기 is a noun form of verb in Korean.

clients/packages/i18n/src/locales/ko.ts
```diff
@@ -0,0 +1,266 @@
...
+      day: '일',
+      week: '주',
+      month: '월',
+      year: '년',
+    },
+  },
+  benefitTypes: {
+    custom: '사용자 지정',
+    license_keys: '라이선스 키',
+    github_repository: 'GitHub 저장소 액세스',
```
MikeYi314: It sounds more natural if we say "GitHub 레포 액세스"
Koreans call it "Github Repo" instead of calling it "저장소" which is a direct translation from English to Korean.

clients/packages/i18n/src/locales/ko.ts
MikeYi314: Yeah I think Powered by is better than Korean here. No one says "제공: ". It's a direct translation from English to Korean.

PR #9563, opened 2026-02-13
Apply FIFO guard delay when initially queuing webhook events

pieterbeulque:
When a webhook event is created while earlier events for the same
endpoint are still pending delivery, apply the same delay calculation
used in the task retry path. This avoids unnecessary immediate
processing followed by retry cycles for events that we already know
need to wait.

https://claude.ai/code/session_012G8Jx5kUZqeV1378iUPG3o

### DIFF
--- a/server/polar/webhook/service.py
+++ b/server/polar/webhook/service.py
@@ -752,7 +752,21 @@ class WebhookService:
                 session.add(event_type)
                 events.append(event_type)
                 await session.flush()
-                enqueue_job("webhook_event.send", webhook_event_id=event_type.id)
+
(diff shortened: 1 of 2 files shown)


pieterbeulque:
Theoretically should work, however I'd look into denouncing customer_state.changed webhook triggers first

Yopi (approved):
Cool! I think this is neat, and it probably will cut down on some retries and noise

pieterbeulque:
As discussed on Slack, I think we're setting up a footgun by adding more complexity when queuing versus the more low-hanging fruit of reducing the number of successive similar webhooks, so let's take another stab at that first.

PR #9611, opened 2026-02-18
Add debouncing to customer webhook task

pieterbeulque:
## 📋 Summary

Adds debouncing configuration to the `customer_webhook` task to prevent duplicate webhook events from being processed in rapid succession. Updated our `debounce_key` logic to allow returning null to opt out of the debouncing.

## 🎯 What

- Added `_customer_webhook_debounce_key()` function to generate unique debounce keys based on event type and customer ID
- Updated `customer_webhook` actor decorator with debounce configuration:
  - `debounce_key`: Uses the new key generation function
  - `debounce_min_threshold`: 1 second minimum delay
  - `debounce_max_threshold`: 5 seconds maximum delay

## 🤔 Why

We trigger a lot of `customer_state.changed` webhooks.

## 🔧 How

Implemented debouncing by:
1. Creating a debounce key function that uniquely identifies webhook events by event type and customer ID
2. Configuring the task actor with min/max debounce thresholds to batch rapid events

## 🧪 Testing

- Existing tests pass
- Debouncing behavior is handled by the task framework and will be validated through existing webhook task tests

## ✅ Pre-submission Checklist

https://claude.ai/code/session_013QE2mSwUmuyiziE7eXFyTp

### DIFF
--- a/server/tests/worker/test_debounce.py
+++ b/server/tests/worker/test_debounce.py
@@ -1,22 +1,15 @@
 import uuid
-from collections.abc import Generator
 from unittest.mock import AsyncMock, MagicMock
 
-import dramatiq
 import pytest
-from dramatiq.brokers.stub import StubBroker
 
 from polar.worker._debounce import set_debounce_key
 
 
-@pytest.fixture
-def stub_broker() -> Generator[StubBroker]:
-    broker = StubBroker()
-    broker.actor_options.update(
-        {"debounce_key", "debounce_min_threshold", "debounce_max_threshold"}
-    )
-    yield broker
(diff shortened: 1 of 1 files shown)


server/tests/worker/test_debounce.py
```diff
@@ -0,0 +1,95 @@
+import uuid
```
frankie567, line 1: You're effectively testing the logic of `set_debounce_key`, which is nice. But having the StubBroker, actors machinery is totally unnecessary here. It doesn't test the debounce logic itself.

frankie567 (approved):

PR #9630, opened 2026-02-19
refunds: When a refund happens too fast there is no order

Yopi:
Essentially there is a race condition due to the fact that refunds are happening on a higher prioritized queue than order linking to transactions.

If someone refunds a transaction before it has been linked, there is no order to link the balance.refund event to, and we mess up the metric calculations.

### DIFF
--- a/server/polar/transaction/service/refund.py
+++ b/server/polar/transaction/service/refund.py
@@ -11,7 +11,7 @@ from polar.event.system import BalanceRefundMetadata, SystemEvent, build_system_
 from polar.integrations.stripe.service import stripe as stripe_service
 from polar.kit.math import polar_round
 from polar.logging import Logger
-from polar.models import Customer, Refund, Transaction
+from polar.models import Customer, Order, Refund, Transaction
 from polar.models.refund import RefundStatus
 from polar.models.transaction import TransactionType
 from polar.postgres import AsyncSession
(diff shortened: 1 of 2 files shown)


server/polar/transaction/service/refund.py
```diff
@@ -114,7 +136,7 @@ async def create(self, session: AsyncSession, refund: Refund) -> Transaction:
             payment_user_id=payment_transaction.payment_user_id,
             pledge_id=payment_transaction.pledge_id,
             issue_reward_id=payment_transaction.issue_reward_id,
-            order_id=payment_transaction.order_id,
+            order_id=order.id if order is not None else payment_transaction.order_id,
```
frankie567, line 139: Isn't it simply and always `refund.order_id` then? :D No need to resolve it or fallback to payment_transaction

server/polar/transaction/service/refund.py
```diff
@@ -221,14 +247,13 @@ async def revert(self, session: AsyncSession, refund: Refund) -> Transaction:
...
```
frankie567, line 250: Same comment :D

server/polar/transaction/service/refund.py
Yopi: We need the metadata from the order for the balanece event below

frankie567: Ok, then!

Yopi: But at least it resolves it immediately if the payment transaction has it, so for 99.99% of cases it's a no op

server/polar/transaction/service/refund.py
```diff
@@ -57,6 +57,23 @@ def __init__(self, refund: Refund) -> None:
...
+        refund: Refund,
+        payment_transaction: Transaction,
+    ) -> Order | None:
+        if payment_transaction.order is not None:
+            return payment_transaction.order
+
+        if refund.order_id is None:
+            return None
+
+        statement = select(Order).where(Order.id == refund.order_id)
```
frankie567, line 73: Can we use `OrderRepository.get_by_id`? ☺️

  ➕ small stylistic nitpick: I like to have "private" methods below public ones.

frankie567: Ah, too late 😛

Yopi: Will fix

PR #9648, opened 2026-02-20
Experiment: Move payment disclaimer

sebastianekstrom:


clients/packages/i18n/src/locales/fr.ts
```diff
@@ -26,6 +32,9 @@ export default {
       optional: 'Facultatif',
       apply: 'Valider',
       fieldRequired: 'Ce champ est obligatoire',
+      addBusinessDetails: "Ajouter les informations de l'entreprise",
+      removeBusinessDetails: "Supprimer les informations de l'entreprise",
```
frankie567:
```
      addBusinessDetails: "Ajouter les informations de facturation",
      removeBusinessDetails: "Supprimer les informations de facturation",
```

clients/packages/i18n/src/locales/en.ts
```diff
@@ -4,12 +4,21 @@ export default {
       poweredBy: 'Powered by',
       merchantOfRecord:
         'This order is processed by our online reseller & Merchant of Record, Polar, who also handles order-related inquiries and returns.',
+      mandateSubscriptionTrial:
+        'By clicking "Start trial," you authorize Polar Software, Inc., our online reseller and merchant of record, to charge your selected payment method in the amount shown above at the end of your trial period and on each subsequent billing date until you cancel. You may cancel at any time before the end of your trial to avoid being charged.',
```
pieterbeulque: Should we interpolate `{buttonLabel}` instead of hardcoding `"Start trial"` here so that these don't go out of sync?

clients/packages/i18n/src/locales/en.ts
```diff
@@ -4,12 +4,21 @@ export default {
...
+        'By clicking "Pay now," you authorize Polar Software, Inc., our online reseller and merchant of record, to charge your selected payment method the amount shown above. This is a one-time charge.',
     },
     form: {
       email: 'Email',
       cardholderName: 'Cardholder name',
       purchasingAsBusiness: "I'm purchasing as a business",
+      addBusinessDetails: 'Add business details',
+      removeBusinessDetails: 'Remove business details',
       businessName: 'Business name',
+      billingDetails: 'Billing Details',
```
pieterbeulque: Unsure if nitpick but if the toggle says _Add/remove business details_ shouldn't we title the expanding section _Business details_ too? Or the toggle to _Add/remove billing details_?

clients/packages/checkout/src/components/CheckoutForm.tsx
```diff
@@ -700,55 +704,167 @@ const BaseCheckoutForm = ({
...
+                                        </Button>
+                                      )}
+                                    </div>
+                                  </div>
+                                </FormControl>
+                                <FormMessage />
+                              </FormItem>
+                            )}
+                          />
+                          <CountryPicker
```
pieterbeulque: I think we can remove the duplicated disabled country here. I thought it could be nice by putting the full "business address" in the collapsible section, but we learned doesn't work well with US/CA.

It doesn't add much on its own & can only add confusion as to why it's disabled?

clients/packages/i18n/src/locales/en.ts
pieterbeulque: Or, put differently, basically the same as François' feedback.

pieterbeulque:
Dutch translations are fine!

It's insane how much cleaner this looks by letting the form flow more naturally 😁

clients/packages/i18n/src/locales/en.ts
sebastianekstrom: Good idea!

clients/packages/checkout/src/components/CheckoutForm.tsx
sebastianekstrom: Agreed!

clients/packages/i18n/src/locales/en.ts
sebastianekstrom: Agreed!

PR #9665, opened 2026-02-21
Preserve external_id in customer deletion webhooks

pieterbeulque:
Fall back to the `user_metadata` `__external_id` when `external_id` is cleared.

One reason to do this is that the `external_id` is cleared on a customer before the `customer.deleted` webhook is called, so causes issues if you rely on this to sync your Polar state to your own database.


pieterbeulque:
I thought about that, but doesn't that have different implications? I think the external ID should effectively be 'freed up' again, and in that sense it'd make sense if you `external_id` becomes `None` on the customer. It's just on the payload for the `customer.deleted` webhook that I believe it should still exist.

Or would you not agree? I can see both sides to be fair.

frankie567:
Well, since it's soft-deleted, the customer won't be accessible from Customer API endpoints (hidden from list/get). The only way to have their data again is through a join relationship, e.g. by fetching an old Order linked to that deleted customer. In this case yes, the old `external_id` will appear again.

I think that's fine since the only reason we clear the `external_id` when deleting a customer is when merchants do tests/mistakes and want to start fresh. I don't really think it would be a thing in day-to-day operations.

pieterbeulque:
@frankie567 - that was a lot of LLM usage for a small change, jesus. Should be as simple as this then. Not sure I love the `raw_` prefix.

server/polar/customer/schemas/customer.py
```diff
@@ -123,8 +123,10 @@ class CustomerBase(MetadataOutputMixin, TimestampedSchema, IDSchema):
     id: UUID4 = Field(
         description="The ID of the customer.", examples=[CUSTOMER_ID_EXAMPLE]
     )
-    external_id: str | None = Field(
-        description=_external_id_description, examples=[_external_id_example]
+    raw_external_id: str | None = Field(
```
frankie567: Ah yes, that's a drawback I didn't foresee to use `computed_field` for this use-case 🤔 Maybe a cleaner solution would be to have a `@property` on the Customer model (e.g. `resolved_external_id`) with the `computed_field` logic, and use it as `validation_alias` on the `external_id` field.

pieterbeulque:
@frankie567 - I was running in circles with a computed property or property on the customer, so I moved it back to a validator. If you don't like it, can you maybe take it for a spin to show what you mean? I spent 2+ hours on this in total by now 😕

PR #9709, opened 2026-02-23
server: make TimeoutLockError an operational error

frankie567:
- server: make TimeoutLockError an operational error


frankie567:
Fix #9708

PR #9719, opened 2026-02-23
Prevent non-forever discounts on recurring products when payment unavailable

pieterbeulque:
## 📋 Summary

Adds validation to prevent checkout confirmation when an organization is not payment-ready and has a non-forever discount that makes the checkout free on a recurring product. This ensures organizations cannot start free trials with temporary discounts that would later require payment processing before their account has been reviewed.

## 🎯 What

Added a new validation check in `checkout_service.confirm()` that raises `PaymentNotReady` when:
- The checkout is free due to a discount (`is_payment_required` is False)
- Payment setup is still required (`is_payment_setup_required` is True)
- The discount is not permanent (`duration != DiscountDuration.forever`)
- The organization is not ready for payment

This prevents a scenario where a temporary discount makes a recurring product free, but once the discount expires, the organization would need to process payments without having completed the payment review process.

## 🤔 Why

Organizations that are not payment-ready (e.g., newly created, not yet reviewed) should not be able to start recurring charges. A non-forever discount that makes a checkout free creates a risk: the discount expires and charges begin without the organization having been properly vetted. Forever discounts are safe because they never expire, and one-time products are safe because they have no future charges.

## 🔧 How

- Added validation logic in `_confirm_inner()` method to check discount duration and organization payment readiness
- Allows forever discounts (will never expire) and one-time products (no future charges) to proceed
- Blocks non-forever discounts on recurring products for non-payment-ready organizations
- Added three comprehensive test cases covering the three scenarios:
  1. `test_payment_not_ready_non_forever_discount_recurring` - Validates the new restriction
  2. `test_payment_not_ready_forever_discount_recurring_allowed` - Confirms forever discounts are allowed
  3. `test_payment_not_ready_non_forever_discount_one_time_allowed` - Confirms one-time products are allowed

## 🧪 Testing

- [x] Added three new test cases covering all scenarios
- [x] Tests validate both the restriction and the allowed cases
- [x] Existing tests continue to pass

https://claude.ai/code/session_01Y7wikjWfD5kraAPzgZxPuY

### DIFF
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -1074,28 +1074,21 @@ class CheckoutService:
                 }
             )
 
-        # Check if organization can accept payments (only block paid transactions)
-        if (
-            checkout.is_payment_required
-            and not await organization_service.is_organization_ready_for_payment(
-                session, checkout.organization
-            )
+        # Check if organization can accept payments
+        if not await organization_service.is_organization_ready_for_payment(
+            session, checkout.organization
         ):
-            raise PaymentNotReady()
-
-        # When payments are unavailable, ensure discounts that make a checkout
-        # free are permanent (forever), to avoid starting charges once the
-        # discount expires without the organization having been reviewed.
-        if (
-            not checkout.is_payment_required
-            and checkout.is_payment_setup_required
-            and checkout.discount is not None
-            and checkout.discount.duration != DiscountDuration.forever
-            and not await organization_service.is_organization_ready_for_payment(
-                session, checkout.organization
-            )
-        ):
-            raise PaymentNotReady()
+            if checkout.is_payment_required:
+                raise PaymentNotReady()
+            # When a discount makes a checkout free, ensure it's a permanent
+            # (forever) discount to avoid starting charges once it expires
+            # without the organization having been reviewed.
+            if (
+                checkout.is_payment_setup_required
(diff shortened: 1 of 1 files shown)


server/polar/checkout/service.py
```diff
@@ -1074,14 +1074,21 @@ async def _confirm_inner(
...
+        if not await organization_service.is_organization_ready_for_payment(
+            session, checkout.organization
         ):
-            raise PaymentNotReady()
+            if checkout.is_payment_required:
+                raise PaymentNotReady()
+            # When a discount makes a checkout free, ensure it's a permanent
+            # (forever) discount to avoid starting charges once it expires
+            # without the organization having been reviewed.
+            if (
```
psincraian, line 1086: does this consider 1 time recurring option but applied to orders?

pieterbeulque: Sorry, can you elaborate?

psincraian: If the recurring option of the discount is set to once. will the payment go through when it's applied to a one-time product instead of a subscription?

pieterbeulque: Yep, see test `test_payment_not_ready_non_forever_discount_one_time_allowed`

psincraian: missed that one!

PR #9744, opened 2026-02-24
Handle non-recoverable payment declines by skipping dunning retries

pieterbeulque:
## Summary

When a payment fails with a non-recoverable Stripe decline code, the subscription is marked as `past_due` but automatic dunning retries are skipped. This avoids futile retries for permanently unusable payment methods, letting customers update their payment method instead.

### Unrecoverable decline codes (22)

| Category | Codes |
|---|---|
| Card issues | `expired_card`, `card_not_supported`, `incorrect_number`, `incorrect_cvc`, `invalid_cvc`, `invalid_expiry_year`, `invalid_pin` |
| Issuer blocks | `do_not_honor`, `not_permitted`, `restricted_card`, `stop_payment_order`, `pickup_card` |
| Fraud/security | `fraudulent`, `lost_card`, `stolen_card`, `merchant_blacklist`, `security_violation` |
| Account issues | `invalid_account`, `revocation_of_authorization`, `revocation_of_all_authorizations` |
| Other | `previously_declined_do_not_retry`, `live_mode_test_card` |

### Kept as recoverable (normal dunning)

`insufficient_funds`, `card_velocity_exceeded`, `pin_try_exceeded`, `generic_decline`, `try_again_later`, `processing_error`, `call_issuer`, `authentication_required`, Stripe Radar codes (`highest_risk_level`, `rule`, `elevated_risk_level`, `blocklist`), and payment-method-specific codes (CashApp, Affirm, Link).

## How it works

1. `handle_payment_failure()` fetches the latest payment for the order
2. Checks `Payment.decline_reason` against `UNRECOVERABLE_DECLINE_CODES`
3. If unrecoverable: clears `next_payment_attempt_at` (no retries), marks subscription `past_due`
4. If recoverable: proceeds with normal dunning flow

To make sure subscriptions get recovered when a new payment method is added, I also added logic to retry payments when a new default payment method is added by the customer portal. Previously, the normal dunning cadence would've recovered these, but that's no longer the case, so retrying upon updating payment method felt nice.

### DIFF
--- a/server/polar/order/service.py
+++ b/server/polar/order/service.py
@@ -2094,7 +2094,9 @@ class OrderService:
             )
 
             await repository.update(
-                order, update_dict={"next_payment_attempt_at": utc_now()}
+                order,
+                update_dict={"next_payment_attempt_at": utc_now()},
+                flush=True,
             )
 
             order.subscription.payment_method_id = payment_method.id
--- a/server/polar/models/subscription.py
+++ b/server/polar/models/subscription.py
@@ -15,6 +15,7 @@ from sqlalchemy import (
     String,
     Text,
     Uuid,
+    cast,
     event,
     type_coerce,
 )
@@ -345,11 +346,11 @@ class Subscription(CustomFieldDataMixin, MetadataMixin, RecordModel):
 
     @past_due_deadline.inplace.expression
     @classmethod
-    def _past_due_deadline_expression(cls) -> ColumnElement[datetime | None]:
+    def _past_due_deadline_expression(cls) -> ColumnElement[datetime]:
         total_interval = functools.reduce(
             operator.add, settings.DUNNING_RETRY_INTERVALS
         ) + timedelta(minutes=1)
-        return cls.past_due_at + total_interval
+        return cast(cls.past_due_at + total_interval, TIMESTAMP(timezone=True))
 
     def can_cancel(self, immediately: bool = False) -> bool:
         if not SubscriptionStatus.is_billable(self.status):


pieterbeulque:
@frankie567 - This _should_ work. Do you have any tips on how to manually test this in development? It requires time travel & tinkering with payment methods to allow an initial payment intent & then start failing on cycles…

Also, re: Stripe Radar codes (`highest_risk_level`, `rule`, `elevated_risk_level`, `blocklist`) - maybe these are unrecoverable too?

pieterbeulque:
Actually, I think this can get subscriptions to get stuck in `past_due` and never be revoked.

server/polar/order/service.py
```diff
@@ -1893,11 +1920,58 @@ async def handle_payment_failure(
         if order.subscription is None:
             return order
 
+        # Check if the latest payment has a non-recoverable decline code.
```
frankie567: I think this should not be handled here but rather in `_handle_first_dunning_attempt`, which is the logic triggered after the first payment failure (see next comment)

server/polar/order/service.py
frankie567: Here, you can retrieve the latest payment and check the decline code. If recoverable, set the retry date, otherwise, keep it `None`. The rest of the logic (mark past due, benefit enqueue, etc.) is untouched.

server/polar/order/service.py
```diff
@@ -1893,11 +1920,58 @@ async def handle_payment_failure(
...
+            )
+            return await self._handle_non_recoverable_payment_failure(session, order)
+
         if order.next_payment_attempt_at is None:
             return await self._handle_first_dunning_attempt(session, order)
 
         return await self._handle_consecutive_dunning_attempts(session, order)
 
+    @staticmethod
+    def _is_non_recoverable_decline(payment: Payment) -> bool:
```
frankie567: Let's put that and the `UNRECOVERABLE_DECLINE_CODES` directly on the Payment model in a `@property`. Worth to pre-emptively add a first `if self.processor == PaymentProcessor.stripe` at the beginning to be future-proof.

server/tests/order/test_service.py
```diff
@@ -2800,6 +2800,216 @@ async def test_void_order(
         assert result_order.next_payment_attempt_at is None
         mock_revoke.assert_called_once()
 
+    @freeze_time("2024-01-01 12:00:00")
+    async def test_non_recoverable_decline_code_skips_dunning(
```
frankie567, line 2804: I think two tests are enough: this one, `test_non_recoverable_decline_code_skips_dunning` and `test_recoverable_decline_enters_dunning`. Others are just variants that don't add too much value.

server/polar/order/repository.py
```diff
@@ -103,6 +103,34 @@ async def get_due_dunning_orders(self, *, options: Options = ()) -> Sequence[Ord
...
+            )
+            .options(joinedload(Order.subscription))
+        )
+        orders = await self.get_all(statement)
+        # Filter in Python: past_due_deadline is a computed property
+        return [
+            order
+            for order in orders
+            if order.subscription is not None
+            and order.subscription.past_due_deadline is not None
```
frankie567: Could be worth to have it as a `hybrid_property`. We can ask an LLM to do that in a follow-up task.

server/polar/order/service.py
```diff
@@ -1904,21 +1901,35 @@ async def _handle_first_dunning_attempt(
...
+        latest_payment = await payment_repository.get_latest_for_order(order.id)
+
+        if latest_payment is not None and latest_payment.is_non_recoverable:
+            log.info(
+                "Non-recoverable decline code detected, "
+                "scheduling retry at past_due_deadline",
+                order_id=order.id,
+                decline_reason=latest_payment.decline_reason,
+            )
+            next_retry_date = subscription.past_due_deadline
```
frankie567, line 1928: Why scheduling a retry if the payment is non-recoverable?

server/polar/order/service.py
```diff
@@ -2047,6 +2058,40 @@ async def _calculate_subscription_order_tax(
...
+                order_id=order.id,
+                subscription_id=order.subscription.id,
+                payment_method_id=payment_method.id,
+            )
+
+            await repository.update(
+                order, update_dict={"next_payment_attempt_at": utc_now()}
+            )
+
+            order.subscription.payment_method = payment_method
```
frankie567: Pretty sure you'll get a lazy load error here.

server/polar/order/service.py
```diff
@@ -2047,6 +2058,40 @@ async def _calculate_subscription_order_tax(
...
+            await session.flush()
```
frankie567: Flush looks superfluous here

server/tests/transaction/service/test_payout.py
```diff
@@ -83,8 +83,9 @@ async def test_stripe(
         assert len(transaction.paid_transactions) == 2 + len(
             transaction.account_incurred_transactions
         )
-        assert transaction.paid_transactions[0].id == balance_transaction_1.id
-        assert transaction.paid_transactions[1].id == balance_transaction_2.id
+        paid_transaction_ids = {t.id for t in transaction.paid_transactions}
+        assert balance_transaction_1.id in paid_transaction_ids
+        assert balance_transaction_2.id in paid_transaction_ids
```
frankie567, line 88: Lol, LLM was zealous here :p

pieterbeulque: It was a flaky test!

server/polar/order/service.py
pieterbeulque: Yeah, I should've probably made this more clear with a comment.

How it works now is that we schedule it to be retried at `subscription.past_due_deadline`, so that it'll be picked up by the dunning background job, and that will mark it from `past_due` → `revoked` after the full `past_due` dunning lifecycle (28 days iirc) without retrying the payments in between if we know they'll fail.

frankie567: Ah yes, good call :)

pieterbeulque:
Addressed all feedback!

server/polar/order/repository.py
pieterbeulque: Looking at what Claude with this feedback, looks pretty neat actually, being able to filter in SQL directly instead of in-memory.

frankie567: Yes, I love hybrid properties :D

PR #9766, opened 2026-02-24
Allow 100% off forever to skip payment intent setup

pieterbeulque:
I feel like this is needed to test subscriptions in production, otherwise you still have to use your own credit card details + a 100% off discount to go through the checkout flow. Using your own credit card details (for actual payments) is against our ToS.

I don't really like how it exposes us to dumb edge cases like not having a payment method on file when the discount is removed and all that, but hey…

Your call!

### DIFF
--- a/server/polar/models/checkout.py
+++ b/server/polar/models/checkout.py
@@ -333,6 +333,8 @@ class Checkout(
         if self.is_free_product_price:
             return False
 
+        # PWYW set to `0`, with an extra guard for metered prices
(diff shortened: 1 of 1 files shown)


Yopi:
I think this broke https://github.com/polarsource/polar/blame/main/server/tests/checkout/test_service.py#L5060, but I'm not sure why it didn't in this PR

Yopi:
Ah, because this was based on an older tree

PR #9768, opened 2026-02-25
Add Feature Flag benefit type

pieterbeulque:
Add a simple "Feature Flag" benefit type that merchants can grant when people should have access to something. This will replace our recommendation to use a "Custom" benefit for that.

It's technically the same, except that "Custom" supports a Markdown note, visible to the end customer, and "Feature Flag" supports JSON metadata, invisible to the end customer.

Rationale: I want to use this technical difference to do some future clean-ups to the end-customer journey for pure SaaS models that use only "Feature Flag" and "Meter Credit" benefits, where their benefits can effectively be hidden from emails & customer portal. This'll have Polar interfere less with the customer journey for SaaS-type merchants.

About the implementation, there's a lot of boilerplate code, maybe some of it can be removed, or do you prefer staying explicit?

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -72,6 +72,7 @@
                   "features/benefits/credits",
                   "features/benefits/custom",
                   "features/benefits/discord-access",
+                  "features/benefits/feature-flags",
                   "features/benefits/file-downloads",
                   "features/benefits/github-access",
                   "features/benefits/license-keys"
--- a/docs/features/benefits/custom.mdx
+++ b/docs/features/benefits/custom.mdx
@@ -13,3 +13,11 @@ Secret message only customers can see, e.g [Cal.com](http://Cal.com) link, priva
 
 For custom integrations you can also distinguish benefits granted to customers to offer even more bespoke user benefits.
 
+<Note>
+  Previously, we recommended using a Custom benefit without a note as a way to
+  grant access to software or SaaS features — decoupling entitlements from
+  checking directly on products. We now recommend the
+  [Feature Flag](/features/benefits/feature-flags) benefit for this purpose, as
+  it's purpose-built for feature gating and supports key-value metadata.
+</Note>
+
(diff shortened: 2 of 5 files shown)


clients/packages/i18n/src/locales/fr.ts
```diff
@@ -218,12 +218,13 @@ export default {
...
   benefitTypes: {
+    custom: 'Personnalisé',
     license_keys: 'Clés de licence',
     github_repository: 'Accès au dépôt GitHub',
     discord: 'Invitation Discord',
-    downloadables: 'Fichiers',
-    custom: 'Personnalisé',
-    meter_credit: 'Crédits prépayés',
+    downloadables: 'Téléchargements de fichiers',
+    meter_credit: "Crédits d'utilisation",
```
frankie567: ⚠️ Feels like the LLM overrides translations that we carefully reviewed :/

clients/packages/i18n/src/locales/fr.ts
```diff
@@ -218,12 +218,13 @@ export default {
...
+    feature_flag: 'Drapeau de fonctionnalité',
```
frankie567: Oh my 😄 `Accès à une fonctionnalité`

pieterbeulque: Would just "Feature flag" also work? I feel like almost every language just adopts that from English, but I know the French aren't quick to do so 😁

clients/packages/i18n/src/locales/fr.ts
pieterbeulque: Ah, yes, I changed some values in English (Meter Credit → Meter credit) and that caused the LLM to re-run on those keys. I'll revert.

clients/packages/i18n/src/locales/fr.ts
frankie567: Hard to say 🤔 A Canadian from Quebec would probably yell at us, and I don't think I ever saw "Feature flag" used in the wild.

PR #9830, opened 2026-02-26
tests: 100% forever discount skips payment intent setup

Yopi:
Based on #9766 / d99dd36 we no longer require payment setup if you have a 100% forever discount


pieterbeulque:
I'm calling it but this is going to bite us in the behind when we least expect it 😁

Yopi:
> I'm calling it but this is going to bite us in the behind when we least expect it 😁

Haha not fair to comment that on this PR 🥲

PR #9831, opened 2026-02-26
logs: Scrub logs stored in S3 a bit more heavily

Yopi:
Since we aim to store logs in S3 for a bit longer than the logfire 30 days, I think it would be a good idea to be a bit more heavy handed when it comes to our redactions.

Namely:
```
r"email",
r"user\.?name",
r"full\.?name",
r"first\.?name",
r"last\.?name",
r"phone",
r"address",
r"ip_?address",
r"cookie",
r"^http\.url$",
```


server/polar/logfire.py
```diff
@@ -136,6 +136,18 @@ def configure_logfire(service_name: Literal["server", "worker"]) -> None:
...
+                        r"email",
+                        r"user\.?name",
+                        r"full\.?name",
+                        r"first\.?name",
+                        r"last\.?name",
+                        r"phone",
+                        r"address",
+                        r"ip_?address",
+                        r"cookie",
+                        r"^http\.url$",
```
frankie567, line 149: Why scrubbing the URL? That's interesting info in general :D

Yopi: I was thinking that we get the route from other attributes like `http.url_path`, but the `http.url` contains query arguments that sometimes contains emails. Just tried to put on some belt and suspenders without making this useless for querying. Might be better to keep it in case someone tries to put interesting things in there?

frankie567: Ok, that's a fair point :) Approved 👍

PR #9876, opened 2026-02-27
fe/metrics: Split product dropdown between archived and not

Yopi:
So that all archived products come later

From 
Screenshot 2026-02-27 at 09 38 45

To
Screenshot 2026-02-27 at 09 37 42

### DIFF
--- a/clients/apps/web/src/components/Products/ProductSelect.tsx
+++ b/clients/apps/web/src/components/Products/ProductSelect.tsx
@@ -25,16 +25,18 @@ const ProductsCommandGroup = ({
   groupedProducts,
   productPriceType,
   onSelectProduct,
-  onSelectProductType,
+  onSelectProducts,
   selectedProducts,
   className,
+  isArchived,
 }: {
   groupedProducts: Record<schemas['ProductPriceType'], schemas['Product'][]>
   productPriceType: schemas['ProductPriceType']
   onSelectProduct: (product: schemas['Product']) => void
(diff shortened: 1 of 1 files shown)


clients/apps/web/src/components/Products/ProductSelect.tsx
```diff
@@ -28,13 +28,15 @@ const ProductsCommandGroup = ({
   onSelectProductType,
   selectedProducts,
   className,
+  isArchived,
 }: {
   groupedProducts: Record<schemas['ProductPriceType'], schemas['Product'][]>
   productPriceType: schemas['ProductPriceType']
   onSelectProduct: (product: schemas['Product']) => void
-  onSelectProductType: (productPriceType: schemas['ProductPriceType']) => void
+  onSelectProductType: (products: schemas['Product'][]) => void
```
pieterbeulque: Okay I was confused about these types at first but I realize what you did now. Can we rename the method `onSelectProducts` to be a bit less confusing?

Yopi: Haha, yes definitely

Yopi: But the redesign itself makes sense?

pieterbeulque: It's not the greatest UI ever seen but we can iterate on it later.

Yopi: Better or worse than current? 😄

PR #9891, opened 2026-02-27
feat: Checkout settings

birkjernstrom:
Introducing the concept combined with initial setting (backoffice) to
enforce 3DS on all transactions for an organization.

Allowing us to add nuanced controls in case needed.


server/polar/organization/schemas.py
```diff
@@ -338,6 +348,7 @@ class OrganizationCreate(Schema):
     notification_settings: OrganizationNotificationSettings | None = None
     customer_email_settings: OrganizationCustomerEmailSettings | None = None
     customer_portal_settings: OrganizationCustomerPortalSettings | None = None
+    checkout_settings: OrganizationCheckoutSettings | None = None
```
frankie567: Don't add it here, otherwise, it'll be public in the API.

server/polar/organization/schemas.py
```diff
@@ -365,6 +376,7 @@ class OrganizationUpdate(Schema):
...
```
frankie567: Same comment as above.

server/polar/organization/service.py
```diff
@@ -298,6 +298,9 @@ async def update(
         if update_schema.notification_settings is not None:
             organization.notification_settings = update_schema.notification_settings
 
+        if update_schema.checkout_settings is not None:
+            organization.checkout_settings = update_schema.checkout_settings
```
frankie567: Same problem, we don't want to let users change those settings.

server/migrations/versions/2026-02-26-2226_add_checkout_settings_to_organizations.py
```diff
@@ -0,0 +1,39 @@
...
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "organizations",
+        sa.Column(
+            "checkout_settings",
+            postgresql.JSONB(astext_type=sa.Text()),
+            nullable=False,
+            server_default='{"require_3ds": false}',
```
frankie567: I prefer to not define DB server default and control that on the Python side. Ideally, we would first make a first PR to add that field as nullable + a backfill script.

Once the backfill script is run, merge a second PR with a migration to make it non-nullable and add the logic associated.

PR #9893, opened 2026-02-27
Add min/max seats to checkout sessions

pieterbeulque:


server/polar/checkout/service.py
```diff
@@ -406,12 +406,21 @@ async def create(
                 ) from e
 
         # Validate seats for seat-based pricing
+        min_seats = getattr(checkout_create, "min_seats", None)
+        max_seats = getattr(checkout_create, "max_seats", None)
```
frankie567: Not sure you need `getattr` in this context

PR #9906, opened 2026-02-28
Fix product switcher showing prices in default currency instead of detected currency

pieterbeulque:


pieterbeulque:
Replaced with #9907

PR #9935, opened 2026-03-02
metrics: Fallback to global exchange rate

Yopi:
If an organization doesn't have a lot of orders, or is newly created, there might not be an exchange rate to use that's close in time. In that case lets look at the global exchange rates and use the closest in time.

### DIFF
--- a/server/polar/metrics/queries.py
+++ b/server/polar/metrics/queries.py
@@ -313,8 +313,37 @@ def get_active_subscriptions_cte(
         .group_by(fx_day, fx_currency)
     )
 
+    closest_global_fx_rate = (
+        select(fx_value)
+        .select_from(Transaction)
+        .where(
+            Transaction.type == TransactionType.payment,
+            Transaction.presentment_currency.is_not(None),
+            func.lower(Transaction.presentment_currency)
+            == func.lower(Subscription.currency),
+            Transaction.created_at >= start_timestamp,
+            Transaction.created_at <= end_timestamp,
+        )
+        .order_by(
+            func.abs(
+                func.extract(
+                    "epoch",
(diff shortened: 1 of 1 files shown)


Yopi:
I think this is risk-free but would appreciate two eyes on it @frankie567

Yopi:
> Makes sense! Curiosity: does PostgreSQL evaluate the subquery in COALESCE or is it short-circuited? :D

Had to look it up, but it looks like postgres does short-circuit COALESCE, so it's performant if we have the exchange rate.

```
docker compose exec db psql -U polar -c "EXPLAIN ANALYZE SELECT COALESCE(1, (SELECT 1 FROM pg_sleep(5)));"
                                     QUERY PLAN
------------------------------------------------------------------------------------
 Result  (cost=0.00..0.01 rows=1 width=4) (actual time=0.001..0.001 rows=1 loops=1)
 Planning Time: 0.055 ms
 Execution Time: 0.041 ms
(3 rows)
```

```
docker compose exec db psql -U polar -c "EXPLAIN ANALYZE SELECT COALESCE(NULL::int, (SELECT 1 FROM pg_sleep(5)));"
                                                     QUERY PLAN
---------------------------------------------------------------------------------------------------------------------
 Result  (cost=0.01..0.02 rows=1 width=4) (actual time=5007.795..5007.797 rows=1 loops=1)
   InitPlan 1 (returns $0)
     ->  Function Scan on pg_sleep  (cost=0.00..0.01 rows=1 width=4) (actual time=5007.783..5007.785 rows=1 loops=1)
 Planning Time: 0.091 ms
 Execution Time: 5007.848 ms
(5 rows)
```

PR #9967, opened 2026-03-03
Add external member ID support for member session creation

pieterbeulque:
## 📋 Summary

This PR adds support for creating member sessions using an external member ID in addition to the existing member ID approach. This allows more flexible member session creation workflows.

## 🎯 What

- Split `MemberSessionCreate` schema into a base class and two concrete implementations:
  - `MemberSessionMemberIDCreate`: Creates sessions using internal member ID (existing behavior)
  - `MemberSessionExternalMemberIDCreate`: Creates sessions using external member ID (new)
  - `MemberSessionCreate`: Union type supporting both approaches
- Updated `member_session.create()` service method to handle both member ID and external member ID lookups
- Added comprehensive test coverage for external member ID creation, including success and error cases
- Updated all existing tests to use the appropriate schema type

## 🤔 Why

This change enables more flexible member session creation by allowing clients to identify members using their external ID, which may be more convenient in some integration scenarios. The union type approach maintains backward compatibility while supporting the new functionality.

## 🔧 How

1. **Schema refactoring**: Created a base schema with common fields (`return_url`) and two concrete implementations for different lookup methods
2. **Service logic**: Modified the `create()` method to detect which schema type is being used and apply the appropriate database filter (by `member.id` or `member.external_id`)
3. **Error handling**: Unified error reporting to use the appropriate field name in validation errors
4. **Test coverage**: Added tests for successful external member ID creation and error cases (not found, not accessible)

## 🧪 Testing

- Added 3 new service tests: `test_creates_session_with_external_member_id`, `test_error_external_member_id_not_found`, `test_error_external_member_id_not_accessible`
- Added 2 new endpoint tests: `test_external_member_id_not_found`, `test_valid_external_member_id`
- Updated 6 existing service tests to use `MemberSessionMemberIDCreate` schema
- Updated 1 existing endpoint test to use the new schema
- All existing tests continue to pass with the refactored schema structure

https://claude.ai/code/session_01Xob4Nc6NRro9oszWV33UGD

### DIFF
--- a/server/polar/member_session/schemas.py
+++ b/server/polar/member_session/schemas.py
@@ -8,12 +8,7 @@ from polar.kit.schemas import IDSchema, Schema, TimestampedSchema
 from polar.member.schemas import Member
 
 
-class MemberSessionCreate(Schema):
-    """
-    Schema for creating a member session using a member ID.
-    """
-
-    member_id: UUID4 = Field(description="ID of the member to create a session for.")
+class MemberSessionCreateBase(Schema):
     return_url: Annotated[
         HttpUrl | None,
         Field(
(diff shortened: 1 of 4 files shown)


psincraian:
I removed the endpoint as wasn't used and we will use customer-sessions :)

PR #9985, opened 2026-03-03
emails: Add alternative flow to enqueue email with template

Yopi:
Start with one call-site, but the ambition is to move all render calls to the workers and out from the API layer. Since we are enqueuing the email task anyway, it's better to let the worker handle the rendering, to keep our API latencies low.

### DIFF
--- a/server/emails/README.md
+++ b/server/emails/README.md
@@ -44,7 +44,7 @@ Don't hesitate to reuse existing schemas for common types like `Organization`, `
 3. Create a new TSX file in the `src/emails` directory with the name of the email template, as snake case, e.g. `customer_greetings.tsx`.
 4. Implement the email template using the `react-email` components.
 
-## How to render it from the Python server?
+## How to send it from the Python server?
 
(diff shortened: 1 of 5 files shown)


Yopi:
Not super slow, only ~0.7-1s, but I think it has averse effects on all endpoints when it runs

frankie567:
> it has averse effects on all endpoints when it runs

Oooooh right! It probably blocks the event loop since subprocess is a sync call 🤦 

Another nice improvement could be to use https://anyio.readthedocs.io/en/stable/subprocesses.html

Yopi:
> > it has averse effects on all endpoints when it runs
> 
> Oooooh right! It probably blocks the event loop since subprocess is a sync call 🤦

Exactly!


> Another nice improvement could be to use https://anyio.readthedocs.io/en/stable/subprocesses.html

Oh, that's cool, didn't know that this existed. I'm thinking it still makes sense to push this to the worker, but we should probably use anyio as well!

PR #9989, opened 2026-03-03
multi-currency: Use babel instead of countryinfo

Yopi:
It seems like countryinfo has some problems publishing the latest version, and the version that's published doesn't include certain territories (among them Åland).

Since we already have babel as a dependency, we can get the currency information from there.


Yopi:
@frankie567 not sure if there was another reason for pulling in the separate library?

frankie567:
TBH, I didn't see babel was including this information 😄 So great catch 👏

PR #9995, opened 2026-03-04
feat: Enable 3DS on all organizations below a certain threshold

birkjernstrom:


server/scripts/checkout_require_3ds_backfill.py
```diff
@@ -0,0 +1,57 @@
...
+) -> None:
+    """
+    Enable require_3ds for organizations with next_review_threshold below
+    the specified amount.
+    """
+    subquery = (
+        select(Organization.id)
+        .where(
+            Organization.next_review_threshold < threshold,
+            Organization.checkout_settings["require_3ds"].astext == "false",
```
psincraian, line 32: I see that orgid `677b046a-2482-46a1-98ad-34060cafcdc5` has no checkout setting, the value that it has is: `{}`.

Just in case you should use something like:

```
    or_(
        Organization.checkout_settings["require_3ds"].astext == "false",
        not_(Organization.checkout_settings.has_key("require_3ds")),
    ),

```

PR #10005, opened 2026-03-04
payout: Bake in all fees into the payout fees

Yopi:
If you do a payout, cancelation / reversal of it and a new payout we need to include the fees of the payout that didn't happen into the second payout.

Otherwise we fail the validation when generating the invoice.

### DIFF
--- a/server/polar/payout/service.py
+++ b/server/polar/payout/service.py
@@ -240,17 +240,22 @@ class PayoutService:
             raise InsufficientBalance(account, balance_amount)
 
         try:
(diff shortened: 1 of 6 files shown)


Yopi:
Was a bit difficult to figure out where to place this in the flow to make sense, since we're not creating new transactions for these. They still need to count though, see: https://polar-olw4768.slack.com/archives/C0942KS0BD2/p1772622899234129

Yopi:
Hmm, yeah I think that sounds better. That was my feeling as well, that that the invoice that becomes generated becomes a bit off. I'll do that instead!

PR #10014, opened 2026-03-05
Add dev enable payments script

pieterbeulque:


pieterbeulque:
@copilot can you move this to dev/cli/commands instead of server/scripts?

PR #10017, opened 2026-03-05
Allow creating a subscription programmatically on effectively free seat-based prices

pieterbeulque:
For seat-based products, benefits are granted per-seat (member level), not per-customer. When a merchant wants to offer a free tier that customers can later upgrade to a paid seat-based plan, they previously had to use a separate free product. But that product has no concept of seats: benefits are granted to the customer. Upgrading from that to a paid seat-based product doesn't work cleanly because the subscription transitions from customer-level to member-level benefit grants.

The fix lets merchants subscribe programmatically to a $0/seat product instead, so subscriptions start as seat-based. That way, benefits are granted per-seat from the start, and upgrading to a paid seat-based tier is a straightforward product swap with no change in how benefits are managed.

### DIFF
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -1715,7 +1715,7 @@ class CheckoutService:
             )
 
         for price in subscription.prices:
-            if not price.is_free():
+            if not price.is_free:
                 raise PolarRequestValidationError(
                     [
                         {
--- a/server/polar/models/checkout.py
+++ b/server/polar/models/checkout.py
@@ -319,9 +319,7 @@ class Checkout(
     def is_free_product_price(self) -> bool:
         if self.product_prices is None:
             return False
-        return all(
-            (is_free_price(price) or price.is_free) for price in self.product_prices
-        )
+        return all(is_free_price(price) for price in self.product_prices)
 
     @property
     def has_metered_prices(self) -> bool:
--- a/server/polar/subscription/service.py
+++ b/server/polar/subscription/service.py
@@ -365,7 +365,7 @@ class SubscriptionService:
             default_price := PriceSet.from_product(
                 product, product.organization.default_presentment_currency
             ).get_default_price()
-        ) and not default_price.is_free():
+        ) and not default_price.is_free:
             errors.append(
                 {
                     "type": "value_error",


psincraian:
- When upgrading from free to paid, customers should go through checkout, right? But usually they cannot have 2 subscriptions active per org
- What happens when a product gets updated? For example, before they only had 1 tier with $0/seat but then they update to $10/seat. In theory that should work and not affect existing customers. But the merchant can update the product, and then do a subscription update that will put the merchant on the new pricing. This can cause an issue if there is no payment method attached to the customer

pieterbeulque:
> When upgrading from free to paid, customers should go through checkout, right

That's correct, but I don't understand the remark here. Everything works if they go through checkout for the $0/seat plan, but we can't create the $0/seat subscription programmatically, like we can do for a free subscription.

Ideally, we want to match the current behavior on how to create free subscriptions programmatically but opt in to the seat-based / member model, so that you can have a freemium model without forcing customers through a $0 checkout.

> What happens when a product gets updated?

I'd expect the subscription becomes past due or revoked immediately. That's not a problem introduced by this PR, it should behave the same as when you change the price on a free product (not seat-based).

psincraian:
On the first point, how does the customer move from a free seat-based product to a paid seat-based product? 

Should the merchant call the cancel subscription endpoint and then go to the checkout flow? That will make the benefits unavailable until the checkout is completed.

pieterbeulque:
Free → paid subscriptions are upgraded upon checkout if you pass in the `subscription_id` to the checkout session: https://polar.sh/docs/api-reference/checkouts/create-session#body-subscription-id-one-of-0

psincraian:
Ohh didn't know that! That's really nice 👏 approved for me

frankie567:
@pieterbeulque I think I would be a bit more satisfied if we had a common method `is_free` on the  price models that each variant implements. So `_ProductPriceFree` would return `True` and `ProductPriceSeatUnit` would implement that logic. Others can return `False` for now.

pieterbeulque:
@frankie567 I did that! It's indeed cleaner.

I went back and forth on whether the `is_free_product_price` used to determine whether a payment setup is required should also account for this or not, and then eventually decided against updating it.

This means:

 - Merchants can choose to subscribe customers to a $0/seat plan programmatically, then upgrade them later, opting in to have the member model from the very start
 - A checkout for a $0/seat plan will still ask for credit card details similar to a free trial, instead of allowing to start for free (email only) like a free product

To me, this felt like the right balance…

frankie567 (approved):

PR #9961, opened 2026-03-02
Add subscription renewal and trial conversion reminder emails

pieterbeulque:
## 📋 Summary

Related:  #9818 #7620

Implements automated reminder emails for subscriptions: renewal reminders for long-cycle subscriptions and trial conversion reminders before trials end.

## 🎯 What

- **New Model**: `SubscriptionReminder` to track sent reminders and prevent duplicates
- **Repository Methods**: 
  - `get_subscriptions_needing_renewal_reminder()` - finds active subscriptions with billing cycles > 45 days where renewal is ~7 days away
  - `get_subscriptions_needing_trial_conversion_reminder()` - finds trialing subscriptions about to convert, with 3-day and 1-day reminder windows based on trial length
- **Scheduled Tasks**: 
  - `send_renewal_reminders()` - hourly cron job to identify and queue renewal reminder emails
  - `send_trial_conversion_reminders()` - hourly cron job to identify and queue trial conversion reminder emails
  - Individual task handlers for sending emails and recording reminders
- **Email Templates**: New React email components for renewal and trial conversion reminders
- **Service Methods**: `send_renewal_reminder_email()` and `send_trial_conversion_reminder_email()` in subscription service
- **Organization Settings**: New toggles for `subscription_renewal_reminder` and `subscription_trial_conversion_reminder` in customer email settings
- **Database Migration**: New `subscription_reminders` table with unique constraint on (subscription_id, type, target_date) to prevent duplicate reminders

## 🤔 Why

Subscriptions with long billing cycles (45+ days) benefit from renewal reminders to reduce churn. Trial conversion reminders help users prepare for paid billing and reduce surprise cancellations. The reminder tracking system prevents duplicate emails even if tasks are retried.

## 🔧 How

- Renewal reminders target subscriptions 7 days before `current_period_end` with billing cycles > 45 days (yearly, 2+ month, 7+ week, 45+ day intervals)
- Trial conversion reminders use a 3-day window for trials ≥3 days and 1-day window for shorter trials (1-3 days)
- Both use a 1-hour window around the target date to handle hourly cron execution
- Race condition guards check for existing reminders before sending
- Unique constraint on reminder table prevents database-level duplicates
- Email settings allow organizations to opt-in/out of each reminder type

## 🧪 Testing

- Existing tests pass
- New repository query methods tested via their complex WHERE clauses with proper filtering for billing cycle duration, date windows, and reminder existence
- Task handlers include race condition guards (checking for existing reminders before creation)
[...]

PR #10018, opened 2026-03-05
email: Add a send log (db)

Yopi:
For us to keep track of sent emails, and potentially emails that never succeeded in getting sent.

### DIFF
--- a/server/polar/models/email_log.py
+++ b/server/polar/models/email_log.py
@@ -1,0 +1,35 @@
+from enum import StrEnum
+from typing import Any
+from uuid import UUID
+
+from sqlalchemy import String, Uuid
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.orm import Mapped, mapped_column
+
+from polar.enums import EmailSender
+from polar.kit.db.models import RecordModel
+
+
+class EmailLogStatus(StrEnum):
+    sent = "sent"
+    failed = "failed"
+
+
+class EmailLog(RecordModel):
+    __tablename__ = "email_logs"
+
+    organization_id: Mapped[UUID | None] = mapped_column(
+        Uuid, nullable=True, index=True
+    )
+    status: Mapped[EmailLogStatus] = mapped_column(String, nullable=False, index=True)
+    processor: Mapped[EmailSender] = mapped_column(String, nullable=False)
+    processor_id: Mapped[str | None] = mapped_column(String, nullable=True)
+    to_email_addr: Mapped[str] = mapped_column(String, nullable=False, index=True)
+    from_email_addr: Mapped[str] = mapped_column(String, nullable=False)
+    from_name: Mapped[str] = mapped_column(String, nullable=False)
+    subject: Mapped[str] = mapped_column(String, nullable=False)
+    email_template: Mapped[str | None] = mapped_column(String, nullable=True)
+    email_props: Mapped[dict[str, Any] | None] = mapped_column(
+        JSONB, nullable=False, default=dict
+    )
+    error: Mapped[str | None] = mapped_column(String, nullable=True)
--- a/server/migrations/versions/2026-03-05-1607_add_email_logs_table.py
+++ b/server/migrations/versions/2026-03-05-1607_add_email_logs_table.py
@@ -1,0 +1,84 @@
+"""Add email_logs table
+
+Revision ID: d9f75192e3e7
+Revises: d7d40e0b8721
+Create Date: 2026-03-05 12:34:55.749039
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# Polar Custom Imports
+
+# revision identifiers, used by Alembic.
+revision = "d9f75192e3e7"
+down_revision = "d7d40e0b8721"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
+    op.create_table(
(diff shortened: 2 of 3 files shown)


Yopi:
I'll split the migration and the rest

Yopi:
@pieterbeulque anything you think is missing in terms of things we should store?

pieterbeulque:
I'd add `organization_id` for easier querying of all emails sent for a specific merchant.

Also optional hard FK's (nullable) for `customer_id` (+ `member_id`) for end-customer facing communication and `user_id` (for Polar system notifications) could make it easier to find emails.

pieterbeulque:
In #9961, François had some suggestions: include `processor` and `processor_id` instead of hardcoding Resend in there for future compatibility. + Store all props in a jsonb column

Yopi:
I was thinking the same, and maybe call it `message_log` or something so that we don't necessarily tie it to emails either.

I was thinking that everything we might want to filter on or censor can be nice to have as its own separate column

Yopi:
Went back and forth on it, but `email_logs` felt like the best abstraction. Pulled out `organization_id` and let the rest lie in props.

Yopi:
I think we can go with this. Based on François comment I didn't make any of this strict FKs, but just loose IDs.

I also made us extract the organization id from the props, which I think is decent enough for now but it would be better if we passed it in as an argument to the worker.

server/polar/models/email_log.py
```diff
@@ -0,0 +1,45 @@
...
+    sent = "sent"
+    failed = "failed"
+
+
+class EmailLogType(StrEnum):
+    email_template = "email_template"
+    email_content = "email_content"
+
+
+class EmailLogProcessor(StrEnum):
```
frankie567: Can we reuse existing `EmailSender` enum? We can move it to `polar/enums.py`.

server/polar/models/email_log.py
```diff
@@ -0,0 +1,45 @@
...
+
+from polar.kit.db.models import RecordModel
+
+
+class EmailLogStatus(StrEnum):
+    sent = "sent"
+    failed = "failed"
+
+
+class EmailLogType(StrEnum):
```
frankie567: What is it for?

server/polar/models/email_log.py
```diff
@@ -0,0 +1,45 @@
...
+    type: Mapped[EmailLogType] = mapped_column(String, nullable=False)
+    status: Mapped[EmailLogStatus] = mapped_column(String, nullable=False, index=True)
+    processor: Mapped[EmailLogProcessor] = mapped_column(String, nullable=False)
+    processor_id: Mapped[str | None] = mapped_column(String, nullable=True)
+    to_email_addr: Mapped[str] = mapped_column(String, nullable=False, index=True)
+    from_email_addr: Mapped[str] = mapped_column(String, nullable=False)
+    from_name: Mapped[str] = mapped_column(String, nullable=False)
+    subject: Mapped[str] = mapped_column(String, nullable=False)
+    email_template: Mapped[str | None] = mapped_column(String, nullable=True)
+    message_props: Mapped[dict[str, Any] | None] = mapped_column(
```
frankie567: `email_props` maybe? #pedantic

server/polar/models/email_log.py
```diff
@@ -0,0 +1,45 @@
...
+        JSONB(none_as_null=True), nullable=True, default=None
```
frankie567: I'm not sure we use `none_as_null` option anywhere else, so for consistency we might disable it.

frankie567 (commented):
Nice, it'll be really helpful. Just a few thoughts.

server/polar/models/email_log.py
Yopi: Maybe it's unnecessary, but to differentiate the emails that were sent via a template that is rendered in the worker (where we have the props and content separated) and the ones that are rendered and the worker only gets the html.

server/polar/models/email_log.py
Yopi: Haha I was going back-and-forth on email_log vs message_log, to make this slightly more generic, but I settled on email_log after all. Yes, `email_props` makes sense

server/polar/models/email_log.py
Yopi: We do on tax rate (order and wallet_transaction), but better to do `mapped_column(JSONB, nullable=False, default=dict)` like we do for the others

server/polar/models/email_log.py
Yopi: Yep, that might be nicer

server/polar/models/email_log.py
frankie567: In which case do we only have the HTML?

Yopi: Currently it's `notifications.send` and the "Your organization review is complete" email. We could probably move those over to the templates as well, but I'm not sure

frankie567: Under the hood, they use the React templates, so I guess it's just a matter of moving things around: 

https://github.com/polarsource/polar/blob/2603d83e8344a4676f91bba47e7aa2504a76dc31/server/polar/notifications/notification.py#L33-L43

https://github.com/polarsource/polar/blob/2603d83e8344a4676f91bba47e7aa2504a76dc31/server/polar/organization/tasks.py#L142-L151

Yopi: All right, will do so and skip this one 😄

PR #10030, opened 2026-03-05
email: Add a send log (implementation)

Yopi:
For us to keep track of sent emails, and potentially emails that never succeeded in getting sent.

Follow up from https://github.com/polarsource/polar/pull/10018

### DIFF
--- a/server/tests/email/test_tasks.py
+++ b/server/tests/email/test_tasks.py
@@ -1,0 +1,168 @@
+import pytest
+from pytest_mock import MockerFixture
+from sqlalchemy import select
+
+from polar.email.sender import SendEmailError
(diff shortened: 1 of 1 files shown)


server/polar/email/repository.py
```diff
@@ -0,0 +1,64 @@
...
+from polar.logging import Logger
+from polar.models.email_log import (
+    EmailLog,
+    EmailLogStatus,
+)
+
+log: Logger = structlog.get_logger()
+
+
+def _extract_organization_id(
```
Yopi, line 17: I acknowledge that this isn't a great solution, but I think
1) it will work for now
2) it allows us to roll this out without having to update all call sites of the enqueue task

PR #10042, opened 2026-03-06
server/subscription: require successful payment before proceeding with update

### LINKED ISSUE #5608: Subscription upgrade and failed payment
When a customer (or a seller) upgrades a subscription, we immediately switch to the new plan, issue an invoice and try to pay it immediately using the payment method on file.

If the payment fails, the subscription passes in `past_due` state and benefits are revoked.

This may not be the best behavior — even though IMO there is no good solution for this case. 

Another approach would be to first try the payment and upgrade the plan only when succeeded. Unfortunately, we can't implement this behaviour right now, as we rely on Stripe to manage subscriptions and invoice prorations.

frankie567:
Fix #5608

- server/subscription: require successful payment before proceeding with update

### DIFF
--- a/docs/guides/proration-for-subscription-changes.mdx
+++ b/docs/guides/proration-for-subscription-changes.mdx
@@ -9,7 +9,11 @@ Proration is the adjustment of charges based on the unused portion of the curren
 
 ## Types of Proration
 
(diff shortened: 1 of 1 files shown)


frankie567:
@copilot Update the docs page `proration-for-subscription-changes` to mention the subscription update is now only applied if the payment succeeds. Otherwise, an error is raised and the subscription is unchanged

PR #10099, opened 2026-03-10
seerver/subscription: add subscription_update events

frankie567:
- seerver/subscription: add subscription_update events

### DIFF
--- a/server/polar/enums.py
+++ b/server/polar/enums.py
@@ -51,15 +51,12 @@ class SubscriptionRecurringInterval(StrEnum):
 
 
 class SubscriptionProrationBehavior(StrEnum):
-    invoice = "invoice"  # Invoice immediately
-    prorate = "prorate"  # Add prorations to next invoice
-
-    def to_stripe(self) -> Literal["always_invoice", "create_prorations"]:
-        if self == SubscriptionProrationBehavior.invoice:
-            return "always_invoice"
-        if self == SubscriptionProrationBehavior.prorate:
-            return "create_prorations"
-        raise ValueError(f"Invalid proration behavior: {self}")
+    invoice = "invoice"
+    """Invoice immediately, and add prorations to the invoice."""
+    prorate = "prorate"
+    """Don't invoice immediately, but add prorations to the next invoice."""
+    next_period = "next_period"
+    """Don't invoice immediately, and don't add prorations. The new price will be applied at the start of the next period."""
 
 
 class InvoiceNumbering(StrEnum):
--- a/server/polar/event/system.py
+++ b/server/polar/event/system.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING, Any, Literal, NotRequired, overload
 from sqlalchemy.orm import Mapped
 from sqlalchemy.util.typing import TypedDict
 
+from polar.enums import SubscriptionProrationBehavior
 from polar.kit.address import AddressDict
 from polar.models import Customer, Event, Organization
 from polar.models.benefit import BenefitType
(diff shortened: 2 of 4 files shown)


frankie567:
@Yopi I've added a new system event with new metadata. I see in `server/tinybird/datasources/system_events_by_org.datasource` all the fields from metadata seem to be extracted in their own column. Should I do that too?

Yopi:
> @Yopi I've added a new system event with new metadata. I see in `server/tinybird/datasources/system_events_by_org.datasource` all the fields from metadata seem to be extracted in their own column. Should I do that too?

Yes I think that makes sense. Since columns are ~free in Clickhouse, and a lot faster to search by instead of doing a JSON extract, anything that is standardized makes sense to pull out to its own.

I haven't fixed the automatic deploys of the TB schema yet just so you know. You can deploy locally via `tb --cloud login && tb --cloud deploy` with the Tinybird CLI

frankie567:
@Yopi Thanks! So the process is:

1. Change the schema
2. Deploy it with the CLI

Nothing else? No backfill or some kind of migration? Or logic to extract the metadata JSON to those columns?

Yopi:
For the deployment you just change the schema and deploy it, yes. Tinybird will perform a migration of the historical data, and if needed you add a `FORWARD QUERY` https://www.tinybird.co/docs/forward/test-and-deploy/evolve-data-source#forward-query. In this case I think it depends on if you merge the python changes first or the Tinybird changes first (e.g. if there is any data to backfill). 

They were working on some changes to make adding columns faster, so I think now the deploy will be slower if you do backfill it. It took ~45 minutes last time I did it.

To map the events to the columns you add it into the TinybirdEvent here: https://github.com/polarsource/polar/blob/main/server/polar/integrations/tinybird/service.py#L107, this ensures that the fields end up in the right place.

frankie567:
@Yopi May I ask you for a quick sanity check on the following? :D https://github.com/polarsource/polar/pull/10099/commits/cd9424c4bbd09f9f9627bfced084872a3f8ba02a

Yopi:
Since server/tinybird/datasources/system_events_by_org.datasource is a materialized view you will need to add it to the materialization https://github.com/polarsource/polar/blob/main/server/tinybird/pipes/system_events_by_org_mv.pipe.

Then you will also need to add it to the base datasource https://github.com/polarsource/polar/blob/main/server/tinybird/datasources/events_by_ingested_at.datasource. You should probably also add it to the alternative-events that we have for querying by timestamp https://github.com/polarsource/polar/blob/main/server/tinybird/datasources/events_by_timestamp.datasource + https://github.com/polarsource/polar/blob/main/server/tinybird/pipes/events_by_timestamp_mv.pipe

Otherwise it looks good!

Yopi:
In Tinybird you can see the lineage of the datasources: https://cloud.tinybird.co/aws/us-east-1/polar_sandbox_dedicated/lineage, which helps in seeing which datasources you need to update if you update one later in the pipeline

Screenshot 2026-03-10 at 09 37 26
Screenshot 2026-03-10 at 09 37 34

frankie567:
Ok! Done in f429f6830. Small question though: in the "root" datasource, I see that we give the JSON path of each field from the input event:

https://github.com/polarsource/polar/blob/f429f6830f3d711bd91c85e3c636eafd6a54922d/server/tinybird/datasources/events_by_ingested_at.datasource#L57-L63

If we can do that, why do we make the extraction manually on the Python side? :)

Yopi:
Haha, that's a very valid question. My initial feeling was that we get a bit more control if we massage the data in Python. We are also removing the extracted fields from the `user_metadata` that's stored in Tinybird via the `pop`. It might be wholly unnecessary to do so though, so we could opt to only do the json extraction in Tinybird if we prefer

frankie567:
Ha I see. Yeah popping them from the user_metadata makes sense. All good!

PR #10126, opened 2026-03-10
docs: Add information about how we use Tinybird

Yopi:


Yopi:
@frankie567 anything missing that you would have wanted to know before today? 😄

frankie567:
Looks good! Wondering if we should put it in the handbook versus a dangling README?

Yopi:
My first instinct was to put it next to the code, but the handbook probably makes more sense yes 😄. I'll move it. Thanks!

PR #10207, opened 2026-03-13
Add currencies

pieterbeulque:


pieterbeulque:
@copilot Can you fix the broken test? This PR adds support for CN, so maybe let's check for Russia instead (sanctioned).

PR #10225, opened 2026-03-13
ci: Set up preview env

Yopi:
Sets up backend preview environments on a shared VM.

Vercel frontends are proxied on the same VM to ensure that cookies and CORS doesn't become a problem. The VM is deployed separately (based on the [setup-vm.sh script](https://github.com/polarsource/polar/blob/16eac4c3a593953b8bad7f8d5154b1e0a39fa4f9/infra/preview/setup-vm.sh)). 

The VM is only accessible via the tailnet to keep authorization on the same level as the current preview environments. Vercel can access it via a specific token.

### DIFF
--- a/infra/preview/log-viewer.py
+++ b/infra/preview/log-viewer.py
@@ -67,9 +67,9 @@ class LogHandler(BaseHTTPRequestHandler):
         path = self.path.split("?")[0]
         query = self.path.split("?")[1] if "?" in self.path else ""
 
-        wake_match = re.match(r"/wake/pr-(\d+)", path)
+        wake_match = re.match(r"/wake/pr-(\d+)$", path)
         if wake_match:
-            pr_num = wake_match.group(1)
+            pr_num = int(wake_match.group(1))
             trigger = os.path.join(TRIGGER_DIR, f"pr-{pr_num}.wake")
             try:
                 open(trigger, "w").close()


clients/apps/web/next.config.mjs
```diff
@@ -129,7 +157,24 @@ const nextConfig = {
...
+              source: '/v1/:path*',
+              destination: `${apiUrl}/v1/:path*`,
+            },
+            {
+              source: '/backoffice/:path*',
+              destination: `${apiUrl}/backoffice/:path*`,
+            },
+            {
+              source: '/healthz',
+              destination: `${apiUrl}/healthz`,
```
frankie567, line 174: Maybe stupid question but: why don't we do that rewrite at Caddy level?

frankie567 (approved):
That's super awesome, Jesper. What a huge work 👏 Let's go for it!

Questions:

1 .I don't see where you set some services environment variables, like Stripe. Are there set on the VM directly?
2. Where is the VM hosted?

Yopi:
> Questions:
> 
> 1 .I don't see where you set some services environment variables, like Stripe. Are there set on the VM directly? 

Today I didn't configure it. We need to figure out how to handle the stripe webhooks as well, but I was thinking for most cases this will probably be good enough. The env file used for the backend is set and updated here: https://github.com/polarsource/polar/pull/10225/changes#diff-95c007e25c60db1247627aed8e38823c78b7c84236411d0ac9b9b331534f3458R111-R131

> 2. Where is the VM hosted?

Following some type of least resistance, I spun up a Digitalocean droplet. It might make sense to host it somewhere else

clients/apps/web/next.config.mjs
Yopi: I was toying a bit with making sure that it still worked with the Vercel URL. I went a bit back and forth with this, and I think now it works (https://polar-sandbox-git-yopi-preview-environment-ssh-polar-sh.vercel.app/), but when you login you get redirected anyway

PR #10228, opened 2026-03-13
fix: replace caplog with log mock in dunning order tests

psincraian:
## Summary

- Fixes 5 failing tests in `TestProcessDunningOrder` across `test_service.py` and `test_tasks.py`
- Root cause: `structlog.get_logger()` returns the root logger which has `propagate=False` in the logging config, so pytest's `caplog` fixture never captures the output
- Replaces `caplog.text` assertions with `mocker.patch("polar.order.service.log")` + `log_mock.warning.assert_called_once_with(...)`, matching the existing codebase pattern (e.g. `test_checkout_service.py`, `test_sender.py`)

## Test plan

- [x] All 12 tests in `TestProcessDunningOrder` pass (both files)
- [x] Lint passes

### DIFF
--- a/server/tests/order/test_service.py
+++ b/server/tests/order/test_service.py
@@ -2971,10 +2971,11 @@ class TestProcessDunningOrder:
         save_fixture: SaveFixture,
         customer: Customer,
         product: Product,
-        caplog: pytest.LogCaptureFixture,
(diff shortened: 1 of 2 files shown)


psincraian:
Folding this fix into #10222 instead

PR #10231, opened 2026-03-13
server: use Pydantic AI gateway

frankie567:
- server: use Pydantic AI gateway


psincraian:
Looks good! we should ping Birk to make sure we have auto top up.

psincraian:
also, we need to setup this variable manually inside vercel

PR #10278, opened 2026-03-16
Store name, country, date of birth on users & organizations

pieterbeulque:


pieterbeulque:
Merging as we're only adding optional/nullable fields

PR #10303, opened 2026-03-17
dramatiq: Prevent zombie processes on shutdown

Yopi:
It seems like sometimes when we have invoked shutdown, some threads gets stuck waiting for a coroutine, wherein they are spinning in a polling loop (future.result(timeout=0.1) every 100ms) because the event loop was stopped and their future can never resolve. After 60+ seconds they get a "RuntimeError: Timed out while waiting for coroutine." which will kill them

These threads are then dead / zombiefied and kept alive even after the rest of the application has shut down.

This change ensures that ensures that those pending zombies gets canceled so that the application can shut down properly which otherwise prevents a deploy from happening.

```
Deploy SIGTERM
  → worker.stop(timeout=600000)
    → ShutdownNotifications raises Shutdown in threads
    → join_all waits for threads (some are zombies)
    → join_all returns (timeout or threads done)
    → "Stopping fork process..." ← health server killed
    → middleware cleanup (close HTTPX, Redis, DB, event loop)
    → "Worker has been shut down"
  → but zombie threads still alive, process can't fully exit
  → health endpoint dead, no messages processing
  → Render health check fails for 15 minutes
  → deploy marked as FAILED
````

### DIFF
--- a/server/polar/worker/_asyncio.py
+++ b/server/polar/worker/_asyncio.py
@@ -1,0 +1,47 @@
+import asyncio
+
+import dramatiq
+import structlog
+from dramatiq.asyncio import get_event_loop_thread
+from dramatiq.middleware.asyncio import AsyncIO
+
+from polar.logging import Logger
+
+log: Logger = structlog.get_logger()
+
+
+async def _cancel_pending_tasks() -> None:
+    current = asyncio.current_task()
+    tasks = [t for t in asyncio.all_tasks() if t is not current and not t.done()]
+    if not tasks:
+        return
+    log.info("Cancelling pending asyncio tasks", count=len(tasks))
+    for task in tasks:
+        task.cancel()
+    for task in tasks:
+        try:
+            await task
+        except (asyncio.CancelledError, Exception):
+            pass
+
+
+class GracefulAsyncIOMiddleware(AsyncIO):
+    """Extends dramatiq's AsyncIO middleware to cancel pending tasks before
+    stopping the event loop during shutdown.
+
+    Without this, worker threads stuck in run_coroutine() become zombies
+    after the event loop is stopped — their futures never resolve and they
+    spin until TimeLimit fires (up to 60 seconds), blocking process exit.
+    """
+
+    def after_worker_shutdown(
+        self, broker: dramatiq.Broker, worker: dramatiq.Worker
+    ) -> None:
+        event_loop_thread = get_event_loop_thread()
+        if event_loop_thread is None:
+            return
+
+        if event_loop_thread.loop.is_running():
+            event_loop_thread.run_coroutine(_cancel_pending_tasks())
+
+        super().after_worker_shutdown(broker, worker)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by Yopi** [comment]
I believe that is what we're seeing here: https://logfire-us.pydantic.dev/polar/polar?q=service_name+%3D+%27worker-webhook%27%0AAND+deployment_environment+%3D+%27production%27&since=2026-03-11T14%3A08%3A00.000Z&until=2026-03-11T14%3A22%3A26.000Z

**Comment by Yopi** [comment]
It's the worker threads. In `shutdown.py`: https://github.com/frankie567/dramatiq/blob/63ea404e2204b2d2b3e644053b4b294cb1039ad4/dramatiq/middleware/shutdown.py#L88-L92 it's checked if we should notify or not, which defaults to false. So we never add it to the list of notifications and we never raise `Shutdown`

**Comment by Yopi** [comment]
This of course begs the question if we should just set notify to true instead 😄.

**Comment by frankie567** [comment]
Oh wow! 

> This of course begs the question if we should just set notify to true instead 😄.

Indeed, that's what I was considering 😄 But it can cause some headaches since our jobs are not all idempotent. 

---

I'm curious to reproduce it and see if we can tweak the EventLoopThread to avoid that. We already run on a fork, so :p

**Comment by frankie567** [comment]
What's weird though is that we see the following [stack trace in Logfire](https://logfire-us.pydantic.dev/polar/polar?q=trace_id%3D%27019cdd3d1fc60974df17d44c3f1e6a5e%27+and+span_id%3D%270f5077dbcc99a3a5%27&spanId=0f5077dbcc99a3a5&traceId=019cdd3d1fc60974df17d44c3f1e6a5e&env=-clear-&since=2026-03-11T13%3A42%3A07.064876Z&until=2026-03-11T14%3A42%3A07.064876Z):

```
  File "/app/server/.venv/lib/python3.14/site-packages/dramatiq/asyncio.py", line 172, in run_coroutine
    raise RuntimeError("Timed out while waiting for coroutine.") from e
RuntimeError: Timed out while waiting for coroutine.
```

... which can only happens in the `Interrupt` catch block 🤯 

https://github.com/frankie567/dramatiq/blob/63ea404e2204b2d2b3e644053b4b294cb1039ad4/dramatiq/asyncio.py#L163-L173

EDIT: Ah, it's because TimeLimit has its own manager 🙃 https://github.com/frankie567/dramatiq/blob/63ea404e2204b2d2b3e644053b4b294cb1039ad4/dramatiq/middleware/time_limit.py

frankie567:
I'm really not sure I get it. Tried it locally, and the behaviour I encounter is the following:

* Long coroutine starts
* Process is exited 
* Thread waits for the long coroutine to finish
* TimeLimitExceeded kicks in, exception is raised
* Worker shuts down, no zombie process

So, at worst, the worker takes 60 seconds to shutdown.

Yopi:
I'm also not sure if I get it 😄. I think you're right in that this is not the problem though. I'll go back to the drawing board

PR #10329, opened 2026-03-17
dramatiq: Add AsyncIO middleware wrapper

Yopi:
Adds a watchdog wrapper to check if the event loop is stuck, and if so it dumps the context of all running threads.

This might be able to help us diagnose the issue where sometimes -after- deployment, the event loop gets stuck on a certain number of threads.

### DIFF
--- a/server/polar/worker/_asyncio.py
+++ b/server/polar/worker/_asyncio.py
@@ -1,0 +1,99 @@
+import asyncio
+import faulthandler
+import sys
+import tempfile
+import threading
+
+import dramatiq
+import structlog
+from dramatiq.asyncio import get_event_loop_thread
+from dramatiq.middleware.asyncio import AsyncIO
+
+from polar.logging import Logger
+
+log: Logger = structlog.get_logger()
+
+HEARTBEAT_INTERVAL = 5.0
+HEARTBEAT_TIMEOUT = 15.0
+
+
+class _EventLoopWatchdog(threading.Thread):
+    """Monitor thread that detects when the event loop stops responding.
+
+    Schedules a callback on the event loop every HEARTBEAT_INTERVAL seconds.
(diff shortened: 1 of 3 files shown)


Yopi:
So instead of looking at the stopping threads (#10303), we look at the newly started process. 

I was a bit hasty in my earlier assessment, as it looks like it's actually the newly deployed worker that locks up. The ambition with this PR is to give us a sense or _where_ or _what_ is causing the problems.

Visible here: https://logfire-us.pydantic.dev/polar/polar/explore
```
  SELECT
    start_timestamp,
    CASE service_version
      WHEN 'de8c9f7d04ecc26cfde8b64613dd5a69bc10cd4d' THEN 'NEW'
      WHEN '25a021e598dc3b47e8751fd31c3dafcf1f483cce' THEN 'OLD'
    END AS deploy,
    (
      SELECT MIN(r2.start_timestamp)
      FROM records r2
      WHERE r2.service_name = records.service_name
        AND r2.service_version = records.service_version
        AND r2.message = 'Worker process is ready for action.'
    ) AS deploy_time,
    message
  FROM records
  WHERE service_name = 'worker-webhook' and service_version IN ('de8c9f7d04ecc26cfde8b64613dd5a69bc10cd4d', '25a021e598dc3b47e8751fd31c3dafcf1f483cce')
    AND (
      message LIKE '%Starting event loop%'
      OR message LIKE '%Worker process is ready%'
      OR message LIKE '%Time limit exceeded%'
      OR message LIKE '%Stopping worker%'
      OR message LIKE '%Shutting down%'
      OR message LIKE '%Stopping event loop%'
      OR message LIKE '%Worker has been shut down%'
    )
  ORDER BY start_timestamp
  LIMIT 50
```

PR #10285, opened 2026-03-16
Update Finance → Account page

pieterbeulque:
## 📋 Summary

This PR removes the onboarding questionnaire fields (`intended_use`, `customer_acquisition`, and `future_annual_revenue`) from the organization setup flow and related systems.

## 🎯 What

- Removed the "How will you integrate Polar into your business?" form field from the organization profile settings UI
- Removed the "Main customer acquisition channels" checkbox group from the form
- Removed the "Expected annual revenue" money input field from the form
- Made these fields optional in the `OrganizationDetails` schema (defaulting to `None` or empty list)
- Removed these fields from organization review/evaluation systems
- Removed these fields from backoffice organization display views
- Updated seed data to remove these fields from test organizations
- Updated the appeal review agent prompt to remove references to these fields and added a note about ignoring self-reported revenue projections

## 🤔 Why

These onboarding questionnaire fields are being removed as they are:
1. Self-reported data that can be easily fabricated and are unreliable for decision-making
2. No longer needed in the organization review/approval process
3. Cluttering the organization setup UI without providing sufficient value

The removal simplifies the onboarding experience and reduces reliance on unverifiable self-reported metrics.

## 🔧 How

- Updated the React form component to remove the three form fields and their associated imports (`MoneyInput`, `Checkbox`)
- Changed the schema definitions to make these fields optional with sensible defaults
- Removed field collection from organization review data collectors and evaluation tasks
- Cleaned up backoffice views that displayed these fields
- Updated test data and scripts to remove these fields
- Updated the appeal review agent documentation to clarify that revenue projections should be ignored

## 🧪 Testing

- Existing tests updated to remove these fields from test payloads
- Schema changes maintain backward compatibility by making fields optional
- No new tests needed as this is a removal of functionality

## 📝 Additional Notes

The `OrganizationDetails` TypedDict was updated to use `total=False` to properly reflect that most fields are now optional. This change is backward compatible as the fields are now optional with sensible defaults.

https://claude.ai/code/session_012aYuoMLzEPBhNngCzu8Uzs

PR #10337, opened 2026-03-17
Remove `/v1/accounts/search` as it is no longer used

pieterbeulque:
I believe this should be fine since it's a private API & we're removing the only place it's still used in #10285.


pieterbeulque:
Yeah Github messed up, cherry-picked in #10410

PR #10375, opened 2026-03-18
Fix: Use standard datetime.timedelta instead of sqlalchemy_utils

### LINKED ISSUE #10374: Wrong timedelta import causes incorrect Redis TTL handling
### Description

Incorrect import of `timedelta` in `server/polar/customer/service.py`.

The code imports:
```python
from sqlalchemy_utils.types.range import timedelta
```
instead of using the standard library datetime.timedelta. This is used for Redis cache expiration in get_state().

### Current Behavior

The service uses a non-standard timedelta implementation from sqlalchemy_utils, which is intended for database range types rather than general time calculations.

This may:
- Cause unexpected behavior when calling .total_seconds()
- Produce incorrect Redis TTL values
- Introduce fragile dependency on SQLAlchemy internals

### Expected Behavior

The code should use the standard Python `datetime.timedelta` for time calculations:
```python
from datetime import timedelta
```
This ensures:
- Correct and predictable .total_seconds() behavior
- Proper Redis TTL values
- No dependency on unrelated SQLAlchemy utilities

### Screenshots

N/A

### Environment:

-   Operating System: Windows 11
-   Browser (if applicable): N/A

Jah-yee:
Fixes #10374 - Wrong timedelta import causes incorrect Redis TTL handling

The code was importing timedelta from `sqlalchemy_utils.types.range` which is intended for database range types, not general time calculations. This could cause incorrect `.total_seconds()` behavior and wrong Redis TTL values.

Changed to use standard library `datetime.timedelta` for proper time calculations and Redis cache expiration.

### DIFF
--- a/server/polar/customer/service.py
+++ b/server/polar/customer/service.py
@@ -1,7 +1,7 @@
 import builtins
 import uuid
 from collections.abc import Sequence
-from datetime import datetime
+from datetime import datetime, timedelta
 from typing import Any
 
 import structlog
@@ -9,7 +9,6 @@ from pydantic import TypeAdapter
 from sqlalchemy import UnaryExpression, asc, desc, func, or_
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import joinedload
-from sqlalchemy_utils.types.range import timedelta
 
 from polar.auth.models import AuthSubject
 from polar.benefit.grant.repository import BenefitGrantRepository


abstract-333:
@Jah-yee 
https://github.com/polarsource/polar/blob/8a3ed7e9981aef89d24962d9137178594e8700ea/CONTRIBUTING.md?plain=1#L5-L44

frankie567:
Good catch, thanks to both of you  @Jah-yee and @abstract-333 🙏 

This qualifies as a minor fix for me, so great that you pushed forward 👍

pieterbeulque:
Can you run `uv run task lint` in `server` to make CI/CD pass?

Yopi (approved):

frankie567 (approved):

PR #10388, opened 2026-03-19
Refactor user authentication to fetch from API instead of middleware header

pieterbeulque:
## 📋 Summary

This PR refactors the user authentication flow to fetch user data directly from the API when needed, rather than passing it through middleware headers.

## 🎯 What

- Removed the `x-polar-user` header that was being set in the proxy middleware
- Removed the `schemas` import from `proxy.ts` as it's no longer needed
- Refactored `_getAuthenticatedUser()` in `utils/user.ts` to fetch user data directly from the API via `getServerSideAPI()` instead of reading from middleware headers
- Simplified the authentication check in the proxy to only verify if the user is authenticated (boolean) rather than fetching and storing user data
- Updated tests to reflect the new authentication flow

## 🤔 Why

This change improves the architecture by:
- Reducing coupling between middleware and application code
- Eliminating the need to serialize/deserialize user data through headers
- Making the user data fetching logic more explicit and testable
- Centralizing user data retrieval in the `utils/user.ts` module where it's actually used

## 🔧 How

1. Modified `proxy.ts` to only check if authentication is valid (response.ok) rather than storing user data
2. Removed the header-based user data passing mechanism
3. Updated `_getAuthenticatedUser()` to directly call the API endpoint `/v1/users/me` when needed
4. Updated corresponding tests to remove assertions about the `x-polar-user` header

## 🧪 Testing

- [x] All existing tests pass
- [x] Updated tests to reflect new authentication flow

The existing test suite has been updated to verify:
- Authenticated users can access protected routes
- Unauthenticated access to public routes works
- Login redirects work correctly
- Error handling for unexpected API responses

## 📝 Additional Notes

The refactoring maintains the same authentication behavior while improving code organization and reducing unnecessary data passing through HTTP headers.

https://claude.ai/code/session_015o6Ld2dBEnzUwCFiaRKPf3

### DIFF
--- a/clients/apps/web/src/proxy.test.ts
+++ b/clients/apps/web/src/proxy.test.ts
@@ -7,6 +7,12 @@ vi.mock('./utils/client', () => ({
   createServerSideAPI: vi.fn(),
 }))
 
+vi.mock('next/dist/server/web/spec-extension/adapters/request-cookies', () => ({
+  RequestCookiesAdapter: {
+    seal: (cookies: unknown) => cookies,
+  },
+}))
+
 const nextConfig = {}
 
 describe('proxy matcher configuration', () => {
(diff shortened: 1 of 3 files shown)


pieterbeulque:
Replace with #10389 for now

PR #10398, opened 2026-03-19
cache custom field

Yopi:
As we are calling validate_custom_field_data on every checkout creation,
this will allocate / create a new model for every call, even if there are
no custom fields to validate.

This change does two things:
- It creates an empty custom field schema to validate against, so we don't need to create a new one
- It caches all other custom fields schema in a dict in memory.

Since types are never GCd in python, this should save us some minor memory leakage over time.

Additionally we instantiate the type adapter for notifications so that we can reuse it in `parse_payload` instead of creating a new instance every time.

### DIFF
--- a/server/polar/custom_field/data.py
+++ b/server/polar/custom_field/data.py
@@ -1,6 +1,7 @@
 import datetime
 from collections.abc import Sequence
 from typing import TYPE_CHECKING, Any
+from uuid import UUID
 
 from pydantic import BaseModel, Field, ValidationError, create_model
 from sqlalchemy import event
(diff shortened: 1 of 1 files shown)


Yopi:
```
import gc
import sys
import resource

from pydantic import BaseModel, Field, TypeAdapter, create_model

def get_rss_mb():
    r = resource.getrusage(resource.RUSAGE_SELF)
    return r.ru_maxrss / 1024 / 1024 if sys.platform == 'darwin' else r.ru_maxrss / 1024

def count_types():
    return sum(1 for obj in gc.get_objects() if isinstance(obj, type))

N = 100_000

# --- Test 1: create_model ---
gc.collect()
types_before = count_types()
rss_before = get_rss_mb()

for i in range(N):
    fields = {f'field_{i % 10}': (str, Field(default=None))}
    model = create_model('CustomFieldDataInput', **fields)

gc.collect()
types_after = count_types()
rss_after = get_rss_mb()
print(f'create_model x{N}:  +{types_after - types_before} types,  +{rss_after - rss_before:.0f} MB RSS')

# --- Test 2: TypeAdapter ---
gc.collect()
types_before = count_types()
rss_before = get_rss_mb()

from polar.notifications.notification import Notification as NotificationSchema
for i in range(N):
    ta = TypeAdapter(NotificationSchema)

gc.collect()
types_after = count_types()
rss_after = get_rss_mb()
print(f'TypeAdapter  x{N}:  +{types_after - types_before} types,  +{rss_after - rss_before:.0f} MB RSS')
```

PR #10400, opened 2026-03-19
web: Add max requests to uvicorn

Yopi:
This will restart the web worker after 50k requests (with 10k jitter to ensure that they don't restart at the same time). This will help with memory compaction and trailing memory leaks due to e.g. memory fragmentation.


frankie567:
Good idea, let's go!

PR #10415, opened 2026-03-19
Rephrase payments unavailable as test mode in dashboard

pieterbeulque:
Tinker


pieterbeulque:
@stilla-bot In a separate coding agent, can you create a PR that cleans up `useOrganizationPaymentStatus` and the endpoint that calls based on the fact that we no longer use the `steps` and everything it returns?

PR #10423, opened 2026-03-19
events: Decouple parent and child ingestion

Yopi:
This allows us to decouple the ingestion of a child event and its parent. By storing the pending parent id on events,
we can at a later date resolve this relationship. We write the events to postgres always, but we wait with writing
to Tinybird until we have resolved the relationship.

An orphan child will not show up anywhere, although we could have a list of orphan events in the dashboard. This is
intentional as we want to avoid writing and updating data in Tinybird. Especially since we will need to update both the
root_id, as well as the ancestor_chain. If you are ingesting a grand-grand child and then their parent, you will have to ingest
two more events before you complete the chain and can truly know the events full ancestry.

### DIFF
--- a/server/migrations/versions/2026-04-22-0833_add_pending_parent_external_id_to_events.py
+++ b/server/migrations/versions/2026-04-22-0833_add_pending_parent_external_id_to_events.py
@@ -1,0 +1,38 @@
+"""add pending_parent_external_id to events
+
+Revision ID: f96e4424ec70
+Revises: 9e14393a579f
+Create Date: 2026-03-19 15:21:24.903188
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "f96e4424ec70"
+down_revision = "9e14393a579f"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "events",
+        sa.Column("pending_parent_external_id", sa.String(), nullable=True),
+    )
+    op.create_index(
+        "ix_events_org_pending_parent",
+        "events",
+        ["organization_id", "pending_parent_external_id"],
+        unique=False,
+        postgresql_where=sa.text("pending_parent_external_id IS NOT NULL"),
+    )
+
+
+def downgrade() -> None:
+    op.drop_index(
+        "ix_events_org_pending_parent",
+        table_name="events",
+    )
+    op.drop_column("events", "pending_parent_external_id")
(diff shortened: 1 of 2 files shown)


Yopi:
Before I split this into two, I just wanted to get some feedback on the idea

server/polar/event/service.py
```diff
@@ -745,104 +697,65 @@ async def ingest(
...
-        with logfire.span("insert_batch", event_count=len(events)):
-            event_ids, duplicates_count = await repository.insert_batch(events)
+        event_ids, duplicates_count = await repository.insert_batch(events)
+
+        has_hierarchy = any(
+            event_dict.get("pending_parent_external_id") for event_dict in events
+        )
+
+        if has_hierarchy:
+            resolvable = await repository.find_resolvable_parents(event_ids)
```
frankie567: I'm not sure I'm able to wrap by head around that recursive CTE query, but do you have an idea of how expensive it is? I'm worried that it might add lot of overhead to the ingest endpoint.

server/polar/event/service.py
```diff
@@ -852,8 +765,7 @@ async def ingest(
                 )
                 await self._create_meter_events(session, inserted_events)
 
-        with logfire.span("enqueue_events", event_count=len(event_ids)):
-            enqueue_events(*event_ids)
+        enqueue_events(*ready_ids)
```
frankie567: So, there is a risk that we might have orphan events that we never resolve, right? Not sure if it's a bad thing or not, in any case, feels like it would be user's fault.

frankie567 (commented):
Makes lot of sense, and seems to greatly simplify the ingestion logic (modulo that recursive CTE query which is 🤯 me 😄).

server/polar/event/service.py
Yopi: Exactly, so if they ingest something with a parent_id with never sending in the parent, we will never show it. We'll store it in postgres but not in Tinybird.

We _could_ add a listing of unresolved events, but it feels like an edge case that shouldn't really happen.

server/polar/event/service.py
Yopi: I think if it's not super deep, it wont be very expensive. We don't really put any limitations on the depth of events today. But either way it might make sense to rethink it to make it simpler and more readable

frankie567: Fair enough :) In any case, it would be possible to defer that in the background if needed, right? Let's try it like this anyway :)

Yopi:
@frankie567 I'm splitting this into two now

frankie567 (approved):
So far, so good :D

PR #10448, opened 2026-03-20
seats: Introduce graduated seat tiers

### LINKED ISSUE #7361: Graduated vs. Tiered Seat Pricing
### Feature request

Right now it’s easy to set tiers that cause backward totals (e.g. 20 seats = $500, 21 seats = $420).

Maybe an option to have graduated pricing to avoid the backwardness on the total.
Stripe and LemonSqueezy have it (volume pricing vs graduated pricing).

With marginal pricing, only the seats beyond each threshold get the lower rate (e.g. first 10 at $30, next 10 at $25, seats above 20 at $20).

Yopi:
Fixes #7361

Instead of only allowing to do volume discounts on seats, we can introduce graduate pricing. 

This means that for the first n seats we can have a price ($10), and for the next m seats we can have another price ($15) which means that with volume pricing it would cost (n+m) * $15, but with graduate pricing it would cost n * $10 + m * $15.

Screenshot 2026-03-20 at 12 45 21
Screenshot 2026-03-20 at 12 45 18
Screenshot 2026-03-20 at 12 45 44

### DIFF
--- a/clients/packages/checkout/src/components/CheckoutSeatSelector.tsx
+++ b/clients/packages/checkout/src/components/CheckoutSeatSelector.tsx
@@ -6,7 +6,7 @@ import { formatCurrency } from '@polar-sh/currency'
 import type { AcceptedLocale } from '@polar-sh/i18n'
 import Button from '@polar-sh/ui/components/atoms/Button'
 import Input from '@polar-sh/ui/components/atoms/Input'
-import { useEffect, useMemo, useRef, useState } from 'react'
+import { useEffect, useRef, useState } from 'react'
 import type { ProductCheckoutPublic } from '../guards'
 import { ErrorResponse } from '../providers/CheckoutProvider'
 import MeteredPricesDisplay from './MeteredPricesDisplay'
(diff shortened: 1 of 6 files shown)


server/polar/models/product_price.py
psincraian: we should update this method also

Yopi: Thanks!

Yopi: It's a bit unclear _exactly_ what it should be. I took the average now, I think that should be right

server/polar/models/product_price.py
```diff
@@ -374,12 +380,44 @@ def get_tier_for_seats(self, seats: int) -> SeatTier:
...
-        tier = self.get_tier_for_seats(seats)
-        return tier["price_per_seat"]
+        pricing_type = self.seat_tiers.get("pricing_type", SeatPricingType.volume)
+        match pricing_type:
+            case SeatPricingType.volume:
+                tier = self.get_tier_for_seats(seats)
+                return tier["price_per_seat"]
+            case SeatPricingType.graduated:
+                # Price per seat varies across tiers, so we return the average
+                return self.calculate_amount(seats) // seats
```
pieterbeulque: I took a look at where this surfaces and I do think it's maybe a bit silly to show the average.
It's a "The average human has 1.8 arms" kind of situation.

Outside of `calculate_amount`  (now `calculate_volume`) we only use it to render a single text on checkout. I'd remove the `get_price_per_seat` method completely, move `self.get_tier_for_seats()` into `calculate_volume` and remove the `price_per_seat` property on checkout and fix that separately.

server/polar/product/schemas.py
```diff
@@ -293,6 +294,10 @@ class ProductPriceSeatTiers(Schema):
     - maximum_seats = last tier's max_seats (None for unlimited)
     """
 
+    pricing_type: SeatPricingType = Field(
```
pieterbeulque: Nit-pick but maybe `pricing_type` adds another overload to the concept of a `price type` (which could arguably already be taken for fixed / free / pwyw / … - are these pricing types?)

Maybe `SeatTieringType` or sth, just thinking out loud.

clients/packages/checkout/src/components/CheckoutSeatSelector.tsx
```diff
@@ -175,10 +196,27 @@ const CheckoutSeatSelector = ({
...
+                    {formatCurrency('compact', locale)(
+                      tier.pricePerSeat,
+                      currency,
+                    )}
+                  </span>
+                ))}
+              </span>
+            ) : (
+              <span className="dark:text-polar-500 text-xs text-gray-500">
+                {formatCurrency('compact', locale)(pricePerSeat, currency)} per
```
pieterbeulque: This is the only place we use that `price_per_seat` for on checkout.

We could replace this by finding the current matching tier & using the current tier's `pricePerSeat` instead so that we can remove the `price_per_seat` on checkout completely

server/polar/product/schemas.py
Yopi: That sounds more narrow and descriptive, I agree

clients/packages/checkout/src/components/CheckoutSeatSelector.tsx
Yopi: Fair enough

server/polar/product/schemas.py
Yopi: Opted for `SeatTierType`

smashah:
when is this going live?

Yopi:
> when is this going live?

@smashah it is live. Go into settings and enable seat-based billing if you're not seeing seats on the product pricing page

PR #10525, opened 2026-03-23
i18n: Add Turkish (TR) locale

cenksen:
## Summary

This PR adds a complete Turkish (`tr`) translation for the checkout i18n package.

## Changes

- Added `clients/packages/i18n/src/locales/tr.ts` with full Turkish translations
- Registered `tr` in `SUPPORTED_LOCALES` and `LOCALE_NAMES` in `config.ts`
- Imported and wired `tr` locale in `src/index.ts`

### DIFF
--- a/clients/packages/i18n/src/config.ts
+++ b/clients/packages/i18n/src/config.ts
@@ -10,6 +10,7 @@ export const SUPPORTED_LOCALES = [
   'pt',
   'pt-PT',
   'ko',
+  'tr'
 ] as const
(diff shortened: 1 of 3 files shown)


pieterbeulque:
Thanks for contributing! I've set up the automated translations for Turkish in #10527 (necessary so that we can keep Turkish translations when we make changes / add new strings) instead. I've compared your translations here with the ones auto-generated there and most look the same. Would you mind doing a review of #10527 and then we can continue there? Appreciate your time! 🙇

PR #10527, opened 2026-03-24
Add Turkish

pieterbeulque:


clients/packages/i18n/src/locales/tr.ts
```diff
@@ -0,0 +1,268 @@
...
+      billingAddress: {
+        label: 'Fatura adresi',
+        line1: 'Açık adres',
+        line2: 'Daire veya bina no',
+        postalCode: 'Posta kodu',
+        city: 'İl',
+        country: 'Ülke',
+        state: 'Eyalet',
+        province: 'Bölge',
+        stateProvince: 'Eyalet / Bölge',
```
emrebal98: - `city` can be translated as `Şehir`, which feels more natural. `İl` is also correct, but most people expect to see `Şehir` instead of `İl`.
- `state` does not have a meaningful equivalent in the local context of Türkiye, but a direct translation would be `Eyalet`.
- `province` can be translated as `İlçe`, which represents subdivisions of a city. Direct translation of this word also means city in Turkish, but for address usage mostly we just write `Şehir` and `İlçe`. Since city is a required field it needs to be  `Şehir`.
- For `stateProvince`, we can use `Eyalet / İlçe`. `İlçe` alone is usually sufficient for addresses, but if we want to preserve the original structure, `Eyalet` can remain.

clients/packages/i18n/src/locales/tr.ts
```diff
@@ -0,0 +1,268 @@
...
+        'Artık {product} avantajlarından yararlanabilirsiniz.',
+      failedDescription:
+        'Lütfen tekrar deneyin veya destek ekibiyle iletişime geçin.',
+    },
+    loading: {
+      processingOrder: 'Sipariş işleniyor...',
+      processingPayment: 'Ödeme işleniyor',
+      paymentSuccessful: 'Ödeme başarılı! Ürünleriniz hazırlanıyor...',
+      confirmationTokenFailed:
+        'Onay belirteci oluşturulamadı, lütfen daha sonra tekrar deneyin.',
```
emrebal98: - `anahtarı` would be more suitable than `belirteci`.

emrebal98 (commented):
Other than that, everything looks good. I’m planning to use Polar for my SaaS product, and I’d be happy to help review future Turkish translations as well.

emilwidlund (approved):

PR #10542, opened 2026-03-24
fix: uv sync

psincraian:


psincraian:
@copilot @claude @stilla check what changed. Seems that my Mac is unusable when I run the api task.

Is there any new bug on the updated packages? Has any package been compromised?

PR #10547, opened 2026-03-24
Improve i18n DX

sebastianekstrom:
* Relax the TS checks and use `en.ts` as a fallback
* Add validation script on CI to ensure LLM translations


sebastianekstrom:
Fixed in https://github.com/polarsource/polar/pull/11012

PR #10553, opened 2026-03-24
Always show payments unavailable for denied organizations

pieterbeulque:


pieterbeulque:
@stilla-bot Check the failling tests from the Blacksmith comment and remove the tests that are no longer relevant because of my changes. Fix the remaining, if any.

pieterbeulque:
@stilla-bot run `uv run task lint` in `server`

PR #10579, opened 2026-03-25
feat: make customer email nullable for team customers

psincraian:
## Summary
- Make `Customer.email` nullable in the application layer to support team customers without email
- Team customers (seat-based) can now be created without an email — the owner member's email is used for checkout, notifications, and Stripe interactions
- Add `get_email_recipients()` to resolve notification recipients: for team customers, returns owner/billing manager member emails; for individual customers, returns customer email
- Update checkout confirmation to not require email when a customer is already identified
- Handle nullable email across order/subscription notification flows, customer portal, seat invitations, and Stripe customer sync

## Known issues to fix in follow-up
- Webhook schema validation fails for `subscription.created` because the deprecated `SubscriptionUser.email` field is `str` (non-nullable) — needs to be made `str | None`

## Test plan
- [ ] Create a team customer without email via API
- [ ] Create a checkout session for the team customer — verify owner member email is used
- [ ] Complete checkout — verify Stripe customer uses owner email
- [ ] Verify order/subscription confirmation emails are sent to owner/billing manager members
- [ ] Verify seat invitation emails work for team customers without email
- [ ] Verify individual customer flows are unchanged

### DIFF
--- a/server/polar/customer/schemas/customer.py
+++ b/server/polar/customer/schemas/customer.py
@@ -4,7 +4,7 @@ from typing import Annotated, Literal
 
 from annotated_types import MaxLen
 from fastapi import Path
-from pydantic import UUID4, Discriminator, Field, computed_field, model_validator
+from pydantic import UUID4, Discriminator, Field, Tag, computed_field, model_validator
 from pydantic.aliases import AliasChoices
 
 from polar.config import settings
@@ -119,7 +119,8 @@ def _customer_create_type(v: dict[str, object] | object) -> str:
 
 
 CustomerCreate = Annotated[
-    CustomerIndividualCreate | CustomerTeamCreate,
+    Annotated[CustomerIndividualCreate, Tag("individual")]
+    | Annotated[CustomerTeamCreate, Tag("team")],
     Discriminator(_customer_create_type),
     SetSchemaReference("CustomerCreate"),
 ]


server/polar/backoffice/customers/endpoints.py
```diff
@@ -514,7 +514,7 @@ async def revoke_benefits(
 
         await add_toast(
             request,
-            f"Benefit revocation task enqueued for {customer.email}.",
+            f"Benefit revocation task enqueued for {customer.name or customer.email or 'customer'}.",
```
psincraian: use customer.display name

server/polar/backoffice/customers/endpoints.py
```diff
@@ -527,7 +527,7 @@ async def revoke_benefits(
         with tag.div(classes="flex flex-col gap-4"):
             with tag.p():
                 text(
-                    f"Are you sure you want to revoke all benefits for {customer.email}? "
+                    f"Are you sure you want to revoke all benefits for {customer.name or customer.email or 'this customer'}? "
```
psincraian: use customer.display name

server/polar/checkout/service.py
```diff
@@ -1124,10 +1139,14 @@ async def _confirm_inner(
                             "intent_status": intent.status,
                         }
 
-                # Check for trial abuse
+                # Check for trial abuse (skip for customers without email)
                 if (
                     checkout.trial_end is not None
                     and checkout.organization.prevent_trial_abuse
+                    and (
```
psincraian: if customer.email is None, we should skip trial abuse for customer.type == 'team'

server/polar/trial_redemption/service.py
```diff
@@ -31,11 +36,19 @@ async def create_trial_redemption(
         customer: Customer,
         product: Product | None = None,
         payment_method_fingerprint: str | None = None,
-    ) -> TrialRedemption:
+        checkout_email: str | None = None,
```
psincraian: revert those changes, we don't need it

server/polar/trial_redemption/service.py
```diff
@@ -14,11 +14,16 @@ async def check_trial_already_redeemed(
         customer: Customer,
         product: Product | None = None,
         payment_method_fingerprint: str | None = None,
+        checkout_email: str | None = None,
```
psincraian: revert those changes, for customer.type == 'team' and without email, we shouldn't check for trial redemption. Or return true here

server/polar/subscription/service.py
```diff
@@ -2291,44 +2290,53 @@ async def _send_customer_email(
...
-                    "organization": organization,
-                    "product": product,
-                    "subscription": subscription,
-                    "url": portal_url,
-                    **(extra_context or {}),
-                },
-            }
-        )
+
+        from polar.customer.service import customer as customer_service
```
psincraian: imports should be on the top

server/polar/subscription/service.py
```diff
@@ -2291,44 +2290,53 @@ async def _send_customer_email(
...
+                    },
+                }
+            )
+
+            enqueue_email_template(
+                email,
+                **organization.email_from_reply,
+                to_email_addr=recipient_email,
+                subject=subject,
+            )
```
psincraian: extract this to an inner function

server/polar/webhook/webhooks.py
```diff
@@ -429,7 +434,10 @@ def get_slack_payload(self, target: User | Organization) -> str:
         fields: list[SlackText] = [
             {"type": "mrkdwn", "text": f"*Product*\n{self.data.description}"},
             {"type": "mrkdwn", "text": f"*Amount*\n{amount_display}"},
-            {"type": "mrkdwn", "text": f"*Customer*\n{self.data.customer.email}"},
+            {
+                "type": "mrkdwn",
+                "text": f"*Customer*\n{self.data.customer.name or self.data.customer.email or 'Team Customer'}",
```
psincraian: can we use customer.display_email?

server/polar/webhook/webhooks.py
```diff
@@ -400,7 +400,12 @@ def get_discord_payload(self, target: User | Organization) -> str:
         fields: list[DiscordEmbedField] = [
             {"name": "Product", "value": self.data.description},
             {"name": "Amount", "value": amount_display},
-            {"name": "Customer", "value": self.data.customer.email},
+            {
+                "name": "Customer",
```
psincraian, line 404: can we use customer.display_email?

server/polar/webhook/webhooks.py
```diff
@@ -519,7 +527,12 @@ def get_discord_payload(self, target: User | Organization) -> str:
         fields: list[DiscordEmbedField] = [
             {"name": "Product", "value": self.data.description},
             {"name": "Refunded", "value": amount_display},
-            {"name": "Customer", "value": self.data.customer.email},
+            {
+                "name": "Customer",
```
psincraian, line 531: can we use customer.display_email?

server/polar/webhook/webhooks.py
```diff
@@ -551,7 +564,10 @@ def get_slack_payload(self, target: User | Organization) -> str:
         fields: list[SlackText] = [
             {"type": "mrkdwn", "text": f"*Product*\n{self.data.description}"},
             {"type": "mrkdwn", "text": f"*Refunded*\n{amount_display}"},
-            {"type": "mrkdwn", "text": f"*Customer*\n{self.data.customer.email}"},
+            {
+                "type": "mrkdwn",
+                "text": f"*Customer*\n{self.data.customer.name or self.data.customer.email or 'Team Customer'}",
```
psincraian: can we use customer.display_email?

server/polar/webhook/webhooks.py
```diff
@@ -599,7 +615,12 @@ def get_discord_payload(self, target: User | Organization) -> str:
         fields: list[DiscordEmbedField] = [
             {"name": "Product", "value": self.data.product.name},
             {"name": "Amount", "value": amount_display},
-            {"name": "Customer", "value": self.data.user.email},
+            {
+                "name": "Customer",
```
psincraian, line 619: can we use customer.display_email?

server/polar/webhook/webhooks.py
```diff
@@ -625,7 +646,10 @@ def get_slack_payload(self, target: User | Organization) -> str:
         fields: list[SlackText] = [
             {"type": "mrkdwn", "text": f"*Product*\n{self.data.product.name}"},
             {"type": "mrkdwn", "text": f"*Amount*\n{amount_display}"},
-            {"type": "mrkdwn", "text": f"*Customer*\n{self.data.user.email}"},
+            {
+                "type": "mrkdwn",
+                "text": f"*Customer*\n{self.data.customer.name or self.data.customer.email or 'Team Customer'}",
```
psincraian: can we use customer.display_email?

server/polar/webhook/webhooks.py
```diff
@@ -859,7 +883,12 @@ def _get_discord_fields(
         fields: list[DiscordEmbedField] = [
             {"name": "Product", "value": self.data.product.name},
             {"name": "Amount", "value": amount_display},
-            {"name": "Customer", "value": self.data.user.email},
+            {
+                "name": "Customer",
```
psincraian, line 887: can we use customer.display_email?

server/polar/customer/schemas/customer.py
```diff
@@ -52,28 +53,19 @@
 ]
 
 
-class CustomerCreate(MetadataInputMixin, Schema):
+# --- Create schemas ---
+
+
+class _CustomerCreateBase(MetadataInputMixin, Schema):
```
frankie567: No need of the underscore prefix.

server/polar/customer/schemas/customer.py
```diff
@@ -91,6 +83,52 @@ class CustomerCreate(MetadataInputMixin, Schema):
     )
 
 
+class CustomerCreate(_CustomerCreateBase):
```
frankie567: Let's name it `CustomerIndividualCreate` so we can name the union one `CustomerCreate` properly

server/polar/customer/schemas/customer.py
frankie567: I would name the schema `MemberOwnerCreate` for clarity.

server/polar/customer/schemas/customer.py
```diff
@@ -91,6 +83,52 @@ class CustomerCreate(MetadataInputMixin, Schema):
...
+            "The email address of the team customer. "
+            "Optional for team customers — if omitted, an owner with an email must be provided."
+        ),
+        examples=[_email_example],
+    )
+
+    @model_validator(mode="after")
+    def _require_owner_email_when_no_email(self) -> "CustomerTeamCreate":
+        if self.email is None:
+            if self.owner is None or self.owner.email is None:
```
frankie567: I see that every fields can be nullable in `OwnerCreate`. Is that expected? What result do we expect from an "empty" owner?

server/polar/customer/schemas/customer.py
```diff
@@ -91,6 +83,52 @@ class CustomerCreate(MetadataInputMixin, Schema):
...
+        return self
+
+
+def _customer_create_discriminator(v: dict[str, object] | object) -> str:
+    if isinstance(v, dict):
+        return v.get("type", "individual")  # type: ignore[return-value]
+    return getattr(v, "type", "individual")
+
+
+CustomerCreateInput = Annotated[
```
frankie567: Let's name it `CustomerCreate`.

server/polar/customer/schemas/customer.py
```diff
@@ -91,6 +83,52 @@ class CustomerCreate(MetadataInputMixin, Schema):
...
+    Annotated[CustomerCreate, Tag("individual")]
+    | Annotated[CustomerTeamCreate, Tag("team")],
+    Discriminator(_customer_create_discriminator),
+    SetSchemaReference("CustomerCreateInput"),
```
frankie567: Let's name it `CustomerCreate`.

server/polar/customer/schemas/customer.py
```diff
@@ -91,6 +83,52 @@ class CustomerCreate(MetadataInputMixin, Schema):
...
+def _customer_create_discriminator(v: dict[str, object] | object) -> str:
+    if isinstance(v, dict):
+        return v.get("type", "individual")  # type: ignore[return-value]
+    return getattr(v, "type", "individual")
+
+
+CustomerCreateInput = Annotated[
+    Annotated[CustomerCreate, Tag("individual")]
+    | Annotated[CustomerTeamCreate, Tag("team")],
+    Discriminator(_customer_create_discriminator),
```
frankie567: Since the discrimination logic is on a single field, we can do directly:

```py
CustomerCreate = Annotated[CustomerIndividualCreate | CustomerTeamCreate, Discriminator("type"), SetSchemaReference("CustomerCreate")]
```

The nice thing is that it's then properly documented in the OpenAPI schema.

server/polar/customer/schemas/customer.py
frankie567: Do we plan to migrate those ones so `type` is always present?

server/polar/customer/schemas/customer.py
```diff
@@ -166,19 +222,45 @@ class CustomerBase(MetadataOutputMixin, TimestampedSchema, IDSchema):
...
-            email_hash = hashlib.sha256(self.email.lower().encode()).hexdigest()
-            return f"https://www.gravatar.com/avatar/{email_hash}?d=404"
 
-        return f"https://img.logo.dev/{domain}?size=64&retina=true&token={settings.LOGO_DEV_PUBLISHABLE_KEY}&fallback=404"
+def _customer_type_discriminator(v: dict[str, object] | object) -> str:
+    if isinstance(v, dict):
+        t = v.get("type")
+    else:
+        t = getattr(v, "type", None)
+    return "team" if t == "team" or t == CustomerType.team else "individual"
```
frankie567: I think I would prefer to have that proxy on the Customer model itself. Typically:

```py
class Customer(MetadataMixin, RecordModel):
    # ...
    _type: Mapped[CustomerType | None] = mapped_column(
        String,
        nullable=True,
        default=CustomerType.individual,
    )
    
    # ...

    @property
    def type(self) -> CustomerType:
        return self._type or CustomerType.individual
```

This way, we could have a clean discriminator in the schemas + a non-nullable `type` on `CustomerBase`.

server/polar/customer/schemas/customer.py
```diff
@@ -166,19 +222,45 @@ class CustomerBase(MetadataOutputMixin, TimestampedSchema, IDSchema):
...
 
 
 class Customer(CustomerBase):
```
frankie567: Rename to `CustomerIndividual`

server/polar/customer/schemas/customer.py
```diff
@@ -166,19 +222,45 @@ class CustomerBase(MetadataOutputMixin, TimestampedSchema, IDSchema):
...
     """A customer in an organization."""
+
```
frankie567, line 235: Add the `type` as a literal `Literal[CustomerType.individual]`.

server/polar/customer/schemas/state.py
```diff
@@ -188,3 +187,31 @@ class CustomerState(CustomerBase):
     active_meters: list[CustomerStateMeter] = Field(
         description="The customer's active meters.",
     )
+
+
+class CustomerState(_CustomerStateFields, Customer):
```
frankie567: Rename `CustomerStateIndividual` so we can name the union `CustomerState`

server/polar/customer/schemas/state.py
```diff
@@ -188,3 +187,31 @@ class CustomerState(CustomerBase):
...
+    * Active subscriptions
+    * Granted benefits
+    * Active meters
+    """
+
+
+CustomerStateResponse = Annotated[
+    Annotated[CustomerState, Tag("individual")]
+    | Annotated[CustomerStateTeam, Tag("team")],
+    Discriminator(_customer_type_discriminator),
```
frankie567: Same comment about plain discriminator

server/polar/customer/schemas/customer.py
psincraian: This is to make it compatible with previous versions, there are the following options:

1. No owner specified: an implicit owner is created with the same email as the customer. This only works if the customer has an email, otherwise it's a bad request.
2. Customer and owner specified: by default we take the email from the customer but this can be overwritten.

server/polar/customer/schemas/customer.py
psincraian: Yes! :D since last week type is always filled for customers

server/polar/customer/schemas/customer.py
frankie567: I see. Could we make the email mandatory on `OwnerCreate` then? I feel it would be clearer and less error-prone. And if a user wants to set the same email on the customer and the owner member, it's rather easy to pass the same value.

psincraian: Sounds good! I can do that :+1:

PR #10599, opened 2026-03-26
metrics: Add seat metrics

Yopi:
Adds total, pending and claimed seats as metrics

### DIFF
--- a/clients/apps/web/src/components/Metrics/dashboards/SeatsContent.tsx
+++ b/clients/apps/web/src/components/Metrics/dashboards/SeatsContent.tsx
@@ -18,7 +18,7 @@ const SEAT_METRICS: (keyof schemas['Metrics'])[] = [
   'seat_utilization_rate',
 ]
 
-interface CustomersContentProps {
+interface SeatsContentProps {
   data: ParsedMetricsResponse
   interval: schemas['TimeInterval']
 }
@@ -75,7 +75,7 @@ function MetricSection({
   )
 }
 
-export function CustomersContent({ data, interval }: CustomersContentProps) {
+export function SeatsContent({ data, interval }: SeatsContentProps) {
   return (
     <div className="flex flex-col gap-y-8">
       <MetricSection
--- a/server/polar/metrics/metrics.py
+++ b/server/polar/metrics/metrics.py
@@ -18,7 +18,7 @@ from sqlalchemy import (
 
 from polar.enums import SubscriptionRecurringInterval
 from polar.kit.time_queries import TimeInterval
-from polar.models import Checkout, Subscription
+from polar.models import Checkout, CustomerSeat, Order, Subscription
 from polar.models.checkout import CheckoutStatus
 
 from .queries import MetricQuery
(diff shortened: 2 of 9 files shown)


server/polar/metrics/metrics.py
```diff
@@ -448,6 +448,71 @@ def get_cumulative(cls, periods: Iterable["MetricsPeriod"]) -> float:
...
+
+    @classmethod
+    def get_sql_expression(
+        cls, t: ColumnElement[datetime], i: TimeInterval, now: datetime
+    ) -> ColumnElement[int]:
+        return func.count(CustomerSeat.id).filter(
+            or_(
+                CustomerSeat.claimed_at.is_(None),
+                i.sql_date_trunc(
+                    cast(SQLColumnExpression[datetime], CustomerSeat.claimed_at)
```
psincraian, line 505: I don't get this comparison with claimed_at 🤔

Yopi: If you're looking at the metrics for day 1, 2, 3. If the seat gets claimed on day 2 it will be marked as pending on day 1

psincraian (approved):

frankie567 (approved):
Sounds sensible, but it means we now have metrics that don't rely on Tinybird events?

Yopi:
> Sounds sensible, but it means we now have metrics that don't rely on Tinybird events?

I was sure I wrote a reply here, but it's lost. 

We already have some metrics that are still based on SQL for various reasons. The revenue ones are because we need to look up the FX rate, and some other ones we don't have the data in Tinybird. 

For the seats, I think that maybe we could introduce events where they are missing and generate them via Tinybird, but I didn't feel like it warrants doing that before rolling these out, and then we can do it based on performance.

```
ActiveSubscriptionsMetric
CommittedSubscriptionsMetric
MonthlyRecurringRevenueMetric
CommittedMonthlyRecurringRevenueMetric
AverageRevenuePerUserMetric
CheckoutsMetric
SucceededCheckoutsMetric
ChurnedSubscriptionsMetric
ChurnRateMetric
```

clients/apps/web/src/components/Metrics/dashboards/SeatsContent.tsx
```diff
@@ -0,0 +1,95 @@
...
+      </div>
+    </div>
+  )
+}
+
+export function CustomersContent({ data, interval }: CustomersContentProps) {
+  return (
+    <div className="flex flex-col gap-y-8">
+      <MetricSection
+        title="Customers"
```
pieterbeulque, line 82: My only nit-pick is that this implies 'all customers' but reading the calculation it's only "Customers with seats". I'm not sure how to improve it, I understand what it's trying to do.

pieterbeulque: Maybe the dumb v1 can just be naming this section explicitly "Customers with seats" or "Seat-based customers" and we can fix it later?

pieterbeulque (approved):

clients/apps/web/src/components/Metrics/dashboards/SeatsContent.tsx
Yopi: Ahhh.. You're right. Would it be better if those are more general customer metrics as opposed to seat metrics then?

pieterbeulque: Yeah, that's the alternative, that we tackle this as customer metrics (so showing all customers instead of only seat-based customers) and then seat metrics is a subsection of that.

Given that the dashboard is called "Customers" that may make more sense.

So either widen it up to do that, or rename the dashboard to "Seats" and then show the metrics you currently have (and labeling customers "Customers with seats"). If that makes sense?

Yopi: Lets name it Seats for now

PR #10666, opened 2026-03-30
Handle external customer ID conflicts during checkout confirmation

pieterbeulque:
## 📋 Summary

**Related Issue**: Fixes #

This PR fixes the checkout confirmation flow to properly handle external customer ID conflicts and ensures external IDs are set when linking to existing customers.

## 🎯 What

1. **Fixed external ID assignment**: When confirming a checkout with an external customer ID and linking to an existing customer by email, the external ID is now properly set on the customer (previously it was not being set).

2. **Added conflict detection**: The checkout confirmation now validates that:
   - If a customer already has a different external ID, a `CheckoutCustomerExternalIdMismatch` exception is raised
   - If another customer already owns the external ID being used, a `CheckoutCustomerExternalIdMismatch` exception is raised

3. **Added comprehensive test coverage**: Two new test cases validate the conflict scenarios:
   - `test_existing_email_conflicting_external_id`: Customer exists with a different external ID
   - `test_existing_email_external_id_owned_by_other_customer`: External ID is owned by a different customer

## 🤔 Why

This change prevents data integrity issues where:
- A customer could be linked to multiple external IDs
- An external ID could be associated with multiple customers
- External IDs wouldn't be properly synchronized when confirming checkouts

## 🔧 How

Modified `_create_or_update_customer` in `server/polar/checkout/service.py` to:
1. Check if the customer already has an external ID that conflicts with the checkout's external ID
2. Verify that no other customer owns the external ID before assigning it
3. Assign the external ID to the customer if validation passes

Updated the test in `test_existing_email_external_id_provided` to verify the external ID is now correctly set.

## 🧪 Testing

- [x] All existing tests pass
- [x] Added new tests for conflict scenarios
- [x] Updated existing test assertion to verify external ID is set

The changes are covered by:
- `test_existing_email_external_id_provided`: Verifies external ID is set when linking by email
- `test_existing_email_conflicting_external_id`: Verifies exception on conflicting external ID
- `test_existing_email_external_id_owned_by_other_customer`: Verifies exception when external ID is owned by another customer

https://claude.ai/code/session_01TeYjM9p86prrFUTs923iA4

### DIFF
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -2589,6 +2589,21 @@ class CheckoutService:
         if checkout.locale is not None:
             customer.locale = checkout.locale
 
+        if checkout.external_customer_id is not None and not created:
+            if (
+                customer.external_id is not None
+                and customer.external_id != checkout.external_customer_id
+            ):
+                raise CheckoutCustomerExternalIdMismatch()
+            if customer.external_id is None:
+                existing = await repository.get_by_external_id_and_organization(
+                    checkout.external_customer_id,
+                    checkout.organization.id,
+                )
+                if existing is not None and existing.id != customer.id:
+                    raise CheckoutCustomerExternalIdMismatch()
+                customer.external_id = checkout.external_customer_id
+
         customer.stripe_customer_id = stripe_customer_id
         customer.user_metadata = {
             **customer.user_metadata,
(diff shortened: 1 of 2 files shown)


frankie567:
@pieterbeulque 🛑 This was on purpose. When we match an existing customer, we _don't_ automatically set the given external ID; since we're on a bit of a undefined zone where we're not sure if we should override the external ID or not.

pieterbeulque:
That makes sense! Thanks! Was mostly exploring the issue here, I like using GitHub's diffs view to see what Claude came up with 😁

PR #10670, opened 2026-03-30
logfire: Disable inspect_arguments

Yopi:
Logfire's inspect_arguments feature (enabled by default on Python 3.11+) inspects the caller's stack frame and parses the source AST on every span creation to resolve f-string variables.

  Memray profiling showed this was the single largest allocator: 20.5 GB of allocations across 80k requests, all from ast.parse(). The allocations are short-lived but the volume causes memory fragmentation — RSS grows
  even though the objects are freed.

  With inspect_arguments=False:
  - RSS stays flat at 671 MB (vs 850+ MB growing to 1000+)
  - Throughput improves from 445 to 726 req/s for an arbitrary request that returns 401 locally
  - All observability (spans, traces, scrubbing, attributes) is preserved
  - The only loss is automatic f-string variable resolution, which we don't use

### DIFF
--- a/server/polar/logfire.py
+++ b/server/polar/logfire.py
@@ -169,6 +169,7 @@ def configure_logfire(service_name: Literal["server", "worker"]) -> None:
         environment=settings.ENV,
         service_name=resolved_service_name,
         service_version=os.environ.get("RELEASE_VERSION", "development"),
+        inspect_arguments=False,
         code_source=logfire.CodeSource(
             repository="https://github.com/polarsource/polar",
             revision=os.environ.get("RELEASE_VERSION", "main"),


Yopi:
In logfire this works like the following: 
https://github.com/pydantic/logfire/blob/main/logfire/_internal/formatter.py#L246-L275
```python
def logfire_format_with_magic(
    format_string: str,
    kwargs: dict[str, Any],
    scrubber: BaseScrubber,
    fstring_frame: types.FrameType | None = None,
) -> tuple[str, dict[str, Any], str]:
    # Returns
    # 1. The formatted message.
    # 2. A dictionary of extra attributes to add to the span/log.
    #      These can come from evaluating values in f-strings.
    # 3. The final message template, which may differ from `format_string` if it was an f-string.
    try:
        chunks, extra_attrs, new_template = chunks_formatter.chunks(
            format_string,
            kwargs,
            scrubber=scrubber,
            fstring_frame=fstring_frame,
        )
        return ''.join(chunk['v'] for chunk in chunks), extra_attrs, new_template
    except KnownFormattingError as e:
        warn_formatting(str(e) or str(e.__cause__))
    except FStringAwaitError as e:
        warn_fstring_await(str(e))
    except Exception:
        # This is an unexpected error that likely indicates a bug in our logic.
        # Handle it here so that the span/log still gets created, just without a nice message.
        log_internal_error()

    # Formatting failed, so just use the original format string as the message.
    return format_string, {}, format_string
```
https://github.com/pydantic/logfire/blob/main/tests/test_formatter.py#L69-L89
```python
def test_first():
    v = chunks('{bar}', {'bar': 42})
    assert v == snapshot([{'t': 'arg', 'v': '42'}])
```

PR #10671, opened 2026-03-30
Upgrade production API to pro_plus plan

Yopi:
## Summary
- Upgrades the production API Render plan from `pro` (4 GB) to `pro_plus`
- 4 uvicorn workers each grow to ~1,050 MB RSS from cache warmup, totaling ~4.2 GB which exceeds the `pro` plan's 4 GB limit
- This caused repeated OOM crashes on instances `6d29z` and `9578d` on March 28-29

## Test plan
- [ ] `terraform plan` shows only the plan attribute change
- [ ] After apply, monitor RSS — workers should no longer hit container memory limit


Yopi:
I'll bring this back if we run out of memory again

PR #10678, opened 2026-03-30
Self-serve customer email update

### LINKED ISSUE #9664: Customer Portal: Allow email change
Allow customers to change their email address through the Customer Portal. The new email must be verified before the change takes effect.

>Many customers have requested to change the email address associated with their purchase. It would be awesome if polar would allow customers change their email addresses in the customer portal. It would save us time on customer support inquires.
>
>— [Plain](https://app.plain.com/workspace/w_01JE9TRRX9KT61D8P2CH77XDQM/thread/th_01KFBH2J7PTKP58ADQPA9QRGFK)

## Flow

1. Customer enters a new email address in the Customer Portal
2. We send a verification link to the **new** email address
3. Verification link is valid for **60 minutes**
4. Once verified, the customer's email is updated across their account

The old email should remain active until verification is complete. If the link expires, the change is discarded.

## Open Question: Duplicate Emails

What happens if the new email already belongs to an existing customer?

### Option A: Block it

Reject the change with an error like "This email is already in use."

**Pro:** Simple, no data integrity risks.
**Con:** Leaks information — it confirms that the email is a known customer in our system. Could be abused for enumeration.

### Option B: Block with generic error

Always say something like "Unable to complete this change" regardless of the reason.

**Pro:** Doesn't leak whether the email exists.
**Con:** Bad UX — the customer has no idea what went wrong. We'd need to handle support tickets.

### Option C: Silently proceed, fail on verification

Always send the verification email, even if the target email is already taken. If clicked on a duplicate, show an error at that point (or just don't complete the change).

**Pro:** Doesn't leak email existence at input time. The person clicking the link is the owner of the new email, so telling *them* it's already in use is less of a concern.
**Con:** Slightly more confusing flow if it fails at the verification step.

### Option D: Merge purchases

Merge the two customer records when the email change is verified.

**Pro:** Best UX for customers who genuinely have two accounts.
**Con:** Significant complexity — need to reconcile subscriptions, billing history, payment methods. Edge cases around active subscriptions on both accounts. Probably not worth it.

pieterbeulque:
Fixes #9664

### DIFF
--- a/clients/packages/ui/src/lib/utils.ts
+++ b/clients/packages/ui/src/lib/utils.ts
@@ -1,5 +1,6 @@
 import { twMerge, type ClassNameValue } from 'tailwind-merge'
 
+// Utility for merging Tailwind CSS class names
 export function cn(...inputs: ClassNameValue[]) {
   return twMerge(inputs)
 }
--- a/server/polar/__init__.py
+++ b/server/polar/__init__.py
@@ -1,0 +1,1 @@
+# Merge queue test
--- a/server/polar/backoffice/organizations/endpoints.py
+++ b/server/polar/backoffice/organizations/endpoints.py
@@ -828,7 +828,11 @@ async def remove_member(
         user_email = user.email if user else str(user_id)
 
         # Attempt to remove the member safely
-        await user_organization_service.remove_member_safe(session, user_id, id)
+        await user_organization_service.remove_member_safe(
+            session,
+            user_id=user_id,
+            organization_id=id,
+        )
 
         # Add success toast and redirect
         await add_toast(
--- a/server/polar/backoffice/organizations_v2/endpoints.py
+++ b/server/polar/backoffice/organizations_v2/endpoints.py
@@ -2674,7 +2674,11 @@ async def remove_member(
             user_organization as user_organization_service,
         )
 
-        await user_organization_service.remove_member(session, organization.id, user_id)
+        await user_organization_service.remove_member(
+            session,
+            user_id=user_id,
+            organization_id=organization.id,
+        )
     except Exception as e:
         logger.error("Failed to remove member", error=str(e))
         raise HTTPException(status_code=400, detail=str(e))
--- a/server/polar/integrations/polar/client.py
+++ b/server/polar/integrations/polar/client.py
@@ -11,7 +11,7 @@ from polar.logging import Logger
 
 if TYPE_CHECKING:
     from polar_sdk import Polar as PolarSDK
-    from polar_sdk.models import Customer
+    from polar_sdk.models import Customer, Member
 
 log: Logger = structlog.get_logger()
 
@@ -36,16 +36,15 @@ class PolarSelfClient:
         )
 
     async def create_customer(self, *, external_id: str, email: str, name: str) -> None:
-        from polar_sdk.models import CustomerCreate, CustomerType
+        from polar_sdk.models import CustomerTeamCreate
         from polar_sdk.models.polarerror import PolarError
 
         try:
             await self._sdk.customers.create_async(
-                request=CustomerCreate(
+                request=CustomerTeamCreate(
                     email=email,
                     name=name,
                     external_id=external_id,
-                    type=CustomerType.TEAM,
                 )
             )
         except PolarError as e:
@@ -74,6 +73,13 @@ class PolarSelfClient:
     async def get_customer_by_external_id(self, external_id: str) -> Customer:
         return await self._sdk.customers.get_external_async(external_id=external_id)
 
+    async def get_member_by_external_id(
+        self, *, external_customer_id: str, external_id: str
+    ) -> Member | None:
+        return await self._sdk.members.get_member_by_external_id_async(
+            external_id=external_id,
+        )
+
     async def add_member(
         self, *, customer_id: str, email: str, name: str, external_id: str
     ) -> None:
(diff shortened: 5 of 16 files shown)


server/polar/customer_email_update/repository.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class CustomerEmailVerificationRepository(RepositoryBase[CustomerEmailVerification]):
+    model = CustomerEmailVerification
+
+    async def get_valid_by_token_hash(
+        self, token_hash: str
+    ) -> CustomerEmailVerification | None:
+        statement = (
+            sql.select(CustomerEmailVerification)
```
frankie567: Can use `select` directly imported from `sqlalchemy`

server/polar/customer_email_update/repository.py
```diff
@@ -0,0 +1,44 @@
...
+                    Customer.organization
+                )
+            )
+        )
+        result = await self.session.execute(statement)
+        return result.scalars().unique().one_or_none()
+
+    async def expire_by_customer_id(self, customer_id: UUID) -> None:
+        statement = (
+            sql.update(CustomerEmailVerification)
```
frankie567: Same

server/polar/customer_email_update/repository.py
```diff
@@ -0,0 +1,44 @@
...
+            .options(
+                joinedload(CustomerEmailVerification.customer).joinedload(
+                    Customer.organization
+                )
+            )
+        )
+        result = await self.session.execute(statement)
+        return result.scalars().unique().one_or_none()
+
+    async def expire_by_customer_id(self, customer_id: UUID) -> None:
```
frankie567: Instead of expiring the token, just delete them, we don't really care.

server/polar/customer_email_update/service.py
```diff
@@ -0,0 +1,218 @@
...
+                )
+
+        # Sync to Stripe
+        if customer.stripe_customer_id is not None and customer.email is not None:
+            await stripe_service.update_customer(
+                customer.stripe_customer_id, email=customer.email
+            )
+
+        # Mark the verification record as verified
+        await repository.update(record, update_dict={"verified_at": utc_now()})
```
frankie567: Instead of expiring the token, just delete them, we don't really care.

server/polar/customer_email_update/service.py
```diff
@@ -0,0 +1,218 @@
...
+                        organization_name=organization.name,
+                        new_email=record.email,
+                    )
+                ),
+                to_email_addr=old_email,
+                subject="Your email address has been changed",
+            )
+
+        # Invalidate all existing sessions and create a new one
+        await customer_session_service.delete_customer_sessions(session, customer.id)
```
frankie567: Feels like a bit too much for this PR. Let's keep it simple.

server/polar/customer_portal/endpoints/customer_email_update.py
```diff
@@ -0,0 +1,109 @@
...
+    session: AsyncReadSession = Depends(get_db_read_session),
+) -> None:
+    """Check if an email change verification token is still valid."""
+    valid = await customer_email_update_service.check_token(session, token)
+    if not valid:
+        raise InvalidCustomerEmailUpdate()
+
+
+# No auth required: the verification token serves as proof of email inbox
+# access. A new customer session is returned on success.
```
frankie567: Actually, I think I prefer the contrary approach: check that the current session token matches the customer of that verification request, i.e. `auth_subject.subject.id == verification.customer_id`

server/polar/customer_session/service.py
```diff
@@ -248,10 +248,18 @@ async def get_by_token(
         return result.unique().scalar_one_or_none()
 
     async def delete_expired(self, session: AsyncSession) -> None:
-        statement = delete(CustomerSession).where(
-            CustomerSession.expires_at < utc_now()
-        )
-        await session.execute(statement)
+        from .repository import CustomerSessionRepository
```
frankie567: Claude!

server/polar/customer_portal/endpoints/customer_email_update.py
pieterbeulque: As discussed on Slack, decided to keep this + generating a new customer session token after updating the email address for a smooth request → email → verify → customer portal customer flow.

pieterbeulque:
@frankie567 I believe this one is good to go now, can you take another look?

server/polar/customer_email_update/service.py
```diff
@@ -0,0 +1,216 @@
...
+                await member_repository.update(
+                    member, update_dict={"email": customer.email}
+                )
+                log.info(
+                    "customer_email_update.synced_member_email",
+                    customer_id=customer.id,
+                    member_id=member.id,
+                    old_email=old_email,
+                    new_email=customer.email,
+                )
```
psincraian, line 177: _non blocking_: maybe we should have something like member_service.update_email(customer, new_email) 

so we reuse the same pattern

server/polar/customer_email_update/service.py
```diff
@@ -0,0 +1,216 @@
...
+            subject="Verify your new email address",
+        )
+
+    async def check_token(self, session: AsyncReadSession, token: str) -> bool:
+        token_hash = get_token_hash(token, secret=settings.SECRET)
+        repository = CustomerEmailVerificationRepository.from_session(session)
+        record = await repository.get_valid_by_token_hash(token_hash)
+        return record is not None
+
+    async def verify(
```
psincraian, line 123: I think we should send a webhook to merchants when this happens. Maybe, they want to have this in sync (I usually like on having that).

And, thinking about this. Maybe we should ship this behind a feature flag in the customer portal that's disabled by default on the orgs. If we release this to the general public, it may have sync issues between merchant and polar.

server/polar/customer_session/service.py
```diff
@@ -248,10 +249,14 @@ async def get_by_token(
...
 
     async def delete_expired(self, session: AsyncSession) -> None:
-        statement = delete(CustomerSession).where(
-            CustomerSession.expires_at < utc_now()
-        )
-        await session.execute(statement)
+        repository = CustomerSessionRepository.from_session(session)
+        await repository.delete_expired()
+
+    async def delete_customer_sessions(
```
psincraian: This is not used

server/polar/customer_email_update/service.py
pieterbeulque: We do log a `customer.update` event and send the `customer.update` webhook, so that should be sufficient, no?

I added the customer portal setting (default false), I think that's a good call.

Issue #10709, opened 2026-03-31
Tax-inclusive pricing appears misleading in checkout


resulgg:
### Feature request

<img width="488" height="274" alt="Image" src="https://github.com/user-attachments/assets/6e6ce497-13c2-4ca3-9c4d-225044a824ff" />

Is your feature request related to a problem?

Yes.

The current checkout UI makes it look like taxes are not included in the displayed price, even when the product is actually being sold tax-inclusive.

For example, the main price is shown as $49.99 / yr, and there is also a separate Taxes line. This creates the impression that tax is being added on top of the listed price, even though the displayed price already includes tax.

This can confuse customers and make pricing feel less transparent than it actually is.

Describe the solution you'd like

The checkout UI should make it much clearer when the displayed price is tax-inclusive.

For example:

Show a label such as “Tax included” or “Incl. tax” next to the main price
Clarify in the breakdown that the tax amount is already included in the total
Present the tax line in a way that communicates it is informational only, not an extra charge being added on top

Example:

$49.99 / yr (tax included)
Breakdown:
Subtotal: $41.66
Tax: $8.33
Total: $49.99 / yr

This would make it immediately clear that the customer is seeing the final price already.

Describe alternatives you've considered
Adding a small “Tax included” note below the main price
Renaming the breakdown labels to make the total more explicit
Showing a clearer “Total” row and visually indicating that tax is already part of that amount
Additional context

Right now, the separate Taxes line makes the UI feel similar to a tax-exclusive checkout, even when the pricing is actually tax-inclusive.

A small wording or layout change could make the pricing much clearer and help avoid confusion at checkout.

pieterbeulque:
Thanks for the report! We shipped an improvement that now explicitly states "Taxes (included)" in the price breakdown.

PR #10713, opened 2026-03-31
Add tax included indicator in checkout

sebastianekstrom:
This PR will clarify the checkout pricing if the org has enabled tax inclusive pricing:

* We will subtract the taxes from the subtotal
* We will show `Taxes (included)` in the pricing breakdown (as opposed to only `Taxes`)

Screenshot 2026-04-22 at 12 03 13

### One time purchase, without discount

| Before | After  |
|--------|--------|
| Screenshot 2026-04-22 at 12 02 54 | Screenshot 2026-04-22 at 12 03 13 |

### One time purchase, with discount

| Before | After  |
|--------|--------|
| Screenshot 2026-04-22 at 12 01 25 | Screenshot 2026-04-22 at 11 58 34 |

### Subscription

| Before | After  |
|--------|--------|
| Screenshot 2026-04-22 at 12 00 23 | Screenshot 2026-04-22 at 12 00 06 |

### Org with tax _excluded_ pricing, with a $999 product

| Before | After  |
|--------|--------|
| Screenshot 2026-04-22 at 12 09 18 | Screenshot 2026-04-22 at 12 08 47 |

### DIFF
--- a/clients/packages/i18n/src/locales/en.ts
+++ b/clients/packages/i18n/src/locales/en.ts
@@ -57,6 +57,11 @@ export default {
         _llmContext:
           'Taxes applied to the order. This is VAT or sales tax. Prefer the specific term used in the target locale over a generic taxes (e.g. TVA in French, BTW in Dutch, etc.)',
       },
+      inclTax: {
+        value: 'Taxes (included)',
+        _llmContext:
+          'Label for the tax line in the pricing breakdown when tax is already included in the displayed price. Prefer the specific tax term used in the target locale (e.g. TVA (incluse) in French, BTW (inbegrepen) in Dutch, Moms (inkluderad) in Swedish, etc.)',
+      },
       free: 'Free',
       payWhatYouWant: {
         value: 'Pay what you want',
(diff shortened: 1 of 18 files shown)


pieterbeulque:
This one would be good to go from my end! 

Related to #10709

clients/packages/i18n/src/locales/en.ts
```diff
@@ -57,6 +57,11 @@ export default {
         _llmContext:
           'Taxes applied to the order. This is VAT or sales tax. Prefer the specific term used in the target locale over a generic taxes (e.g. TVA in French, BTW in Dutch, etc.)',
       },
+      inclTax: {
+        value: 'Taxes (included)',
```
pieterbeulque, line 61: I wonder if "Including taxes" reads better or worse

psincraian: In my case I prefer "Taxes (included)"

sebastianekstrom: I kinda like the structure of it now where the left side prefix is what you see on the right side:

* Subtotal | Amount
* Taxes | Amount
* Total | Amount

Using `Including taxes` kinda breaks that:

* Subtotal | Amount
* Including | Amount
* Total | Amount

That said, no strong opinions on this one!

psincraian (approved):

frankie567 (approved):

clients/packages/i18n/src/locales/en.ts
pieterbeulque: 🤝

PR #10728, opened 2026-04-01
payment: Handle Stripe webhook coming before order is stored

Yopi:
When a sync payment (subscription update) fails, the transaction rolls
back and the order is never committed. Stripe still sends a webhook, and resolve_order can't find the order.

We removed a blanket guard that returned None for all sync payments to prevent silently dropped webhooks for *successful* payments where the transaction just hadn't committed yet.

Now we mark the payment intent with metadata on failure, so the webhook side can tell the difference:
- No flag → order is coming, keep retrying
- Flag present → order was discarded, exit cleanly

### DIFF
--- a/server/polar/integrations/stripe/payment.py
+++ b/server/polar/integrations/stripe/payment.py
@@ -159,6 +159,14 @@ async def resolve_order(
             uuid.UUID(order_id), options=order_repository.get_eager_options()
         )
         if order is None:
+            if object.OBJECT_NAME == "payment_intent":
+                updated_intent = await stripe_service.get_payment_intent(object.id)
+                if (
+                    updated_intent.metadata
+                    and updated_intent.metadata.get("polar_failed_sync_payment")
+                    == "true"
+                ):
+                    return None
             raise OrderDoesNotExist(order_id)
         return order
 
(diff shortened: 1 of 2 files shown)


frankie567:
@Yopi Good catch, but then I think we'll hit the other side of the coin. I added this guard initially because if the payment fails, the subscription PATCH is aborted and the order is not saved. Without the guard, the webhook will constantly hit `OrderDoesNotExist` and exhaust its retries.

I think the solution to solve that is to do a similar thing to the trial abuse detection. In this context, we add a metadata to the payment intent to mark it as trial abuse:

https://github.com/polarsource/polar/blob/954019c25de5254176503c670457e3780672969e/server/polar/checkout/service.py#L1166-L1177

On the webhook's side, we can fetch a fresh instance of the intent and detect that metadata:

https://github.com/polarsource/polar/blob/954019c25de5254176503c670457e3780672969e/server/polar/integrations/stripe/payment.py#L204-L213

I think we can apply the same pattern by marking the intent as "failed sync payment", and then early exit `resolve_order`.

Yopi:
Thank you @frankie567, the input I wanted since I saw you added it. 😄 

I updated the PR with your feedback, if you don't mind to have another 👁️ 👁️

PR #10744, opened 2026-04-01
Store member email case-sensitive, compare case-insensitive

pieterbeulque:


server/polar/member/service.py
```diff
@@ -244,7 +244,7 @@ async def create_owner_member(
                     }
                 ]
             )
-        email = raw_email.strip().lower()
+        email = raw_email.strip()
```
frankie567, line 247: I've quickly checked the code paths, but I would say calling `strip` here is superfluous since the value comes from a validated `EmailStrDNS` property.

PR #10837, opened 2026-04-07
onboarding: Separate submit for review to other endpoint

Yopi:
Split save details and submit for review into two separate endpoints. This clarifies the button, and makes sure that submitting for review is explicit

### DIFF
--- a/server/polar/organization/schemas.py
+++ b/server/polar/organization/schemas.py
@@ -486,6 +486,33 @@ class OrganizationUpdate(Schema):
     )
 
 
+class OrganizationReviewSubmissionDetails(Schema):
+    product_description: Annotated[
+        str, StringConstraints(strip_whitespace=True, min_length=30)
+    ]
+
+
+def _empty_review_submission_details_to_dict(value: Any) -> Any:
+    if value is None:
+        return {}
+    return value
+
+
+class OrganizationReviewSubmission(Schema):
+    name: Annotated[str, StringConstraints(strip_whitespace=True, min_length=1)]
+    website: Annotated[str, StringConstraints(min_length=1)]
+    email: EmailStrDNS
+    socials: list[OrganizationSocialLink] = Field(min_length=1)
+    details: Annotated[
+        OrganizationReviewSubmissionDetails,
+        BeforeValidator(_empty_review_submission_details_to_dict),
+    ]
+
+
+class OrganizationReviewSubmissionBody(Schema):
+    body: OrganizationReviewSubmission
+
+
 class OrganizationPaymentStatus(Schema):
     payment_ready: bool = Field(
         description="Whether the organization is ready to accept payments"
--- a/server/polar/organization/service.py
+++ b/server/polar/organization/service.py
@@ -1,3 +1,4 @@
+import builtins
 import uuid
 from collections.abc import Sequence
 from datetime import UTC, datetime, timedelta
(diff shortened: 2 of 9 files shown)


pieterbeulque:
I was about to open a similar issue just now. Thanks for this.

I introduced this issue by moving the `POST /v1/organizations` to step 3. In our initial onboarding, step 2 did the `POST /v1/organizations` and step 3 was a `PATCH /v1/organizations/:id`. However, a `PATCH` that sets the organization details has the side-effect of triggering an organization review.

I merged the two steps in a single API call to avoid triggering a premature organization review. So, another fix could be to decouple the organization review initialization from the `PATCH /v1/organizations` endpoint, then revert the onboarding to a POST on step 2 + a PATCH on step 3, and just rely on the API to validate the slug.

As another option, I was considering maybe adding a private `/validate-slug` endpoint that validates a slug without trying to create the organization.

Do with it whatever you want 😁

Yopi:
@pieterbeulque took your advise, and reverted the POST/PATCH in the organization creation, and added a specific endpoint to submit review which we call from the dashboard

server/polar/organization/service.py
```diff
@@ -401,23 +406,96 @@ async def update(
...
             organization.details = cast(
                 OrganizationDetails, update_schema.details.model_dump()
             )
+
+        organization = await repository.update(organization, update_dict=update_dict)
+
+        await self._after_update(session, organization)
+        return organization
+
+    def _validate_review_submission(
```
frankie567: I would love us to find a way to use a proper Pydantic schema for this rather than manually spinning up validation logic like this

frankie567 (changes requested):

server/polar/organization/service.py
Yopi: Yes, agree. I think this was based on how we do it for the org update where we also do the manual valdation, but this PR has lingered for a bit.

I'll see what we can whip up!

Yopi: Made an attempt!

server/polar/organization/service.py
```diff
@@ -401,23 +407,35 @@ async def update(
...
+
+        await self._after_update(session, organization)
+        return organization
+
+    async def submit_for_review(
+        self, session: AsyncSession, organization: Organization
+    ) -> Organization:
+        try:
+            OrganizationReviewSubmissionBody.model_validate({"body": organization})
+        except PydanticValidationError as e:
```
frankie567, line 422: Not certain, but I think we don't need to explicitly catch and re-raise, it should be handled naturally by FastAPI builtin handlers.

server/polar/organization/schemas.py
```diff
@@ -465,6 +465,33 @@ class OrganizationUpdate(Schema):
...
+def _empty_review_submission_details_to_dict(value: Any) -> Any:
+    if value is None:
+        return {}
+    return value
+
+
+class OrganizationReviewSubmission(Schema):
+    name: Annotated[str, StringConstraints(strip_whitespace=True, min_length=1)]
+    website: Annotated[str, StringConstraints(min_length=1)]
+    email: Annotated[str, StringConstraints(min_length=1)]
```
frankie567: Shouldn't we use `EmailStrDNS` again here?

server/polar/organization/service.py
Yopi: I don't think we've mapped  `PydanticValidationError` in FastAPI. Without it we get:
```
ValidationError: 1 validation error for OrganizationReviewSubmissionBody                                    
body.socials                                                                                                
  List should have at least 1 item after validation, not 0 [type=too_short, input_value=[], input_type=list]
    For further information visit https://errors.pydantic.dev/2.13/v/too_short                              
```

in the API log, and HTTP 500 / Interal Server Error on the API. We do the same here: https://github.com/polarsource/polar/blob/main/server/polar/custom_field/data.py#L96-L101

I am guessing we _don't_ want to automatically map it in case the pydantic validation leaks something internally, or?

server/polar/organization/schemas.py
Yopi: We should!

server/polar/organization/service.py
frankie567: Ok, I just hallucinated it :D

frankie567 (approved):

PR #10951, opened 2026-04-13
Fix committed metrics

sebastianekstrom:
Committed metrics were calculated relative to request time now, which made them sensitive to timezone and the current date. This means the same query could return different values depending on when and where it was requested from.

The fix is to calculate committed metrics relative to each bucket timestamp instead.

Example: For a single subscription at `$500` that starts at `T` and is canceled at `T+2`, historical cMRR should be:

  - `T`: $500
  - `T+1`: $500
  - `T+2`: $0
  - `T+3`: $0

With `timestamp`, looking at the `T+1` bucket always returns `$500`.

With `now`, if you open the dashboard at `T+3` and look back at `T+1`, that old bucket can be recalculated using the current time instead of the bucket’s time, which can incorrectly make it drop to `$0`.

### DIFF
--- a/clients/apps/app/components/Metrics/utils.ts
+++ b/clients/apps/app/components/Metrics/utils.ts
@@ -1,4 +1,4 @@
-import { schemas } from '@polar-sh/client'
+import { getMetricsRangeDates, schemas } from '@polar-sh/client'
 import { formatCurrency } from '@polar-sh/currency'
 import {
   differenceInDays,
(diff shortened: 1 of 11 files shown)


Yopi:
Sorry @sebastianekstrom , I made some other changes you need to rebase ontop of, but I think this makes sense!

Yopi (approved):

Yopi:
@sebastianekstrom ok to go? 😄

PR #11019, opened 2026-04-15
fix: Don't link to deleted benefits from customer page

joebon:
## 📋 Summary

**Related Issue**: Fixes #

## 🧪 Testing

### Test Instructions

1.
2.
3.

## ✅ Pre-submission Checklist

- [ ] All tests pass locally

### DIFF
--- a/server/polar/benefit/strategies/base/schemas.py
+++ b/server/polar/benefit/strategies/base/schemas.py
@@ -56,7 +56,7 @@ class BenefitPublicBase(TimestampedSchema, IDSchema):
         ..., description="Whether the benefit is selectable when creating a product."
     )
     deletable: bool = Field(..., description="Whether the benefit is deletable.")
-    is_deleted: bool = Field(..., description="Whether the benefit is soft-deleted.")
+    is_deleted: bool = Field(..., description="Whether the benefit is deleted.")
     organization_id: UUID4 = Field(
         ..., description="The ID of the organization owning the benefit."
     )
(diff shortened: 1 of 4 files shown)


server/polar/benefit/strategies/base/schemas.py
```diff
@@ -56,6 +56,7 @@ class BenefitPublicBase(TimestampedSchema, IDSchema):
         ..., description="Whether the benefit is selectable when creating a product."
     )
     deletable: bool = Field(..., description="Whether the benefit is deletable.")
+    is_deleted: bool = Field(..., description="Whether the benefit is soft-deleted.")
```
Yopi: It might be nicer to expose deleted_at instead of a boolean, but this also works

Yopi (approved):
🙌

server/polar/benefit/strategies/base/schemas.py
joebon: Hrm, I'm sure you're right but I don't see the point currently and unless there's a clear usage for that timestamp it's cleaner and simpler to consume a bool. IMHO. Enlighten me!

Yopi: More opinion maybe. I think, similar as to a db schema, you might end up wanting or needing that information at some point (listing deleted benefits, etc) and then you’d have both or need to break the API.

Thinking about it now though, I wonder if the deleted benefits should be included in this API response at all? 🤔

PR #11028, opened 2026-04-15
Fix organization appeal not showing after decline

pieterbeulque:


psincraian (approved):

clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/finance/account/AccountPage.tsx
```diff
@@ -40,6 +40,8 @@ export default function ClientPage({
 
   const isDenied = organization.status === 'denied'
 
+  const canAppeal = reviewStatus?.appeal_decision !== 'rejected'
```
Comment (pieterbeulque) [inline clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/finance/account/AccountPage.tsx:43]: `null/undefined` → no appeal yet or appeal still pending
`approved` → will go in the `isActive` case instead.
`rejected` → hard denial, will see the denial screen instead of the `AIValidationResult`

PR #11041, opened 2026-04-16
Prevent infinite toasts from firing

sebastianekstrom:
While we all love a nice buttered toast, an infinite amount of of them is not good for you. This PR fixes that.

https://github.com/user-attachments/assets/960a358a-c06a-4ac7-8b29-6f9d7ea11fe5


sebastianekstrom:
Already fixed in https://github.com/polarsource/polar/pull/11050

PR #11045, opened 2026-04-16
organization: Allow reset proration

Yopi:
We need to accept the reset proration setting, otherwise we fail the validation schema.

However, we still want to only allow it when the organization has the feature flag to enable it. As such, we move the validation from the schema layer to the service layer.

We still want to hide the part of the enum from the OpenAPI definition. We do this through Annotated + WithJsonSchema.


Yopi:
FYI @frankie567

PR #11046, opened 2026-04-16
fix(meter): do not bill metered usage ingested during a trial

psincraian:
## Summary

- `meter.create_billing_entries` now skips events whose ingestion falls within the subscription's trial period, so no billable `BillingEntry` rows exist for trial usage. The meter watermark still advances, so these events are not retried once the trial converts.
- Restores the invariant static prices already obey (nothing written during a trial) for metered pricing.
- Adds an end-to-end regression test covering the full trial lifecycle with metered usage and mid-trial cancellation.

## Problem

Polar's intended behaviour is that trials never produce a charge — neither from static recurring prices nor from metered usage. Static prices upheld this implicitly: the cycle task writes no `BillingEntry` rows while a subscription is in a trial, so at `trial_end` the pending set is empty and order creation silently no-ops.

Metered pricing broke that invariant. The `meter.billing_entries` cron wrote `BillingEntryType.metered` rows live on event ingest, with zero trial awareness. Those rows sat with `order_item_id IS NULL` and got swept into the order created at `trial_end` — via `subscription_cancel` if the customer canceled mid-trial, or via `subscription_cycle_after_trial` if the trial converted — charging the customer for usage accumulated during the trial.

## Fix

Add a guard in `MeterService.create_billing_entries` that skips event-to-`BillingEntry` creation when the event's `ingested_at` precedes `subscription.trial_end`. The existing watermark advance (`last_event = event` set at the top of the loop, before any `continue`) ensures trial events are not retried once the trial ends.

After this change, metered behaves exactly like static: no pending rows during a trial, so the `subscription_cancel` / `subscription_cycle_after_trial` order comes up empty and the existing `NoPendingBillingEntries` swallow handles the no-op.

## Test plan

- [x] New e2e test: `server/tests/e2e/lifecycle/test_trial_metered_cancel.py` drives the full checkout → `setup_intent.succeeded` → `cancel_at_period_end=true` → `events/ingest` → `meter.billing_entries` → scheduler cycle flow. Asserts `stripe_service.create_payment_intent` is never called and no billable `subscription_cancel` order is created.
- [x] Existing meter suite passes (`tests/meter/`, 115 tests)
- [x] Existing billing entry + subscription suites pass (`tests/billing_entry/`, `tests/subscription/`, 205 tests)
- [x] Backend lint passes
- [x] Backend type check passes

## Notes

This PR contains the code fix only. Pending `BillingEntry` rows already written for in-flight trials are not touched by this change; a separate data cleanup is required to soft-delete them before those trials end.

### DIFF
--- a/server/polar/meter/service.py
+++ b/server/polar/meter/service.py
@@ -565,10 +565,18 @@ class MeterService:
             if customer_price is None:
                 continue
 
+            # Polar never charges during a trial: skip trial events so no
+            # billable BillingEntry rows exist. The watermark still advances,
+            # so these events aren't retried once the trial converts.
+            subscription = customer_price.subscription_product_price.subscription
+            if (
+                subscription.trial_end is not None
+                and event.ingested_at < subscription.trial_end
+            ):
+                continue
+
             # Get the paying customer (billing manager) from the subscription
-            paying_customer = (
-                customer_price.subscription_product_price.subscription.customer
-            )
+            paying_customer = subscription.customer
 
             entry = await self._create_subscription_holder_billing_entry(
                 session,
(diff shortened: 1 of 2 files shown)


Yopi (approved):
Makes sense

PR #11049, opened 2026-04-16
perf(seeds): batch Tinybird ingest per org in seeds_load

psincraian:
## Summary

- `uv run task seeds_load` spent 2–3 minutes on Tinybird HTTP round-trips because each seeded customer triggered a separate `POST /v0/events?wait=true` and each call blocked ~4s waiting for materialization.
- Accumulate events per-org (and per-run in `create_single_org_seed`) and flush in one batched call via a new `_flush_tinybird_events` helper, chunked at 2500 events (~5 MB) to stay well under Tinybird's 10 MB payload limit.
- Drops Tinybird calls from O(customers) (~50) to O(orgs) (~10) for a full seed.

## Why

Logs showed a repeating 4s gap between `tinybird.ingest` spans — the pace of the ingest was the seed script's bottleneck. Nothing in seed flow reads from Tinybird between writes, so batching is safe.

## Notes for reviewers

- Events still land in Tinybird before `session.commit()`; DB inserts still happen per-customer, so referential integrity is unchanged.
- `TINYBIRD_FLUSH_CHUNK = 2500` is hoisted to module scope with a rationale comment tied to the client's 10 MB limit.
- Kept sequential flushing (no `asyncio.gather` across chunks) to avoid surprises around Tinybird's write-ordering semantics.
- Did not add `--skip-tinybird` or `wait=false` in this PR — happy to follow up if desired.

## Test plan

- [ ] `uv run task seeds_load` on a clean DB — verify it completes noticeably faster and seeds the same orgs/products/customers.
- [ ] Spot-check the Tinybird `events_by_ingested_at` datasource for the seeded orgs to confirm events still land.
- [ ] `uv run task test server/tests/scripts/test_seeds_load.py`.

### DIFF
--- a/server/scripts/seeds_load.py
+++ b/server/scripts/seeds_load.py
@@ -1,8 +1,10 @@
 import asyncio
 import random
+from collections.abc import Sequence
 from datetime import UTC, datetime, timedelta
 from decimal import Decimal
 from typing import Any, Literal, NotRequired, TypedDict
+from uuid import UUID
 
 import dramatiq
 import typer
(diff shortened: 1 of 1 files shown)


Yopi (approved):
Thank you! I was wondering why it had become so slow all of a sudden.

We _could_ do wait=false for the seeds load, but in general it is not a good idea

PR #11051, opened 2026-04-17
cache: Introduce a metrics cache

Yopi:
We need a short lived metrics cache to alleviate hitting Tinybird too hard with the same query.


frankie567 (approved):

Yopi:
@frankie567 I did some calculations on our memory usage given the current traffic + the short ttl which shouldn't push us very high, but I've also set up an alert on redis memory usage (>600MB) for completeness sake.

PR #11071, opened 2026-04-19
seats: Require checkout secret when assinging seat anonymously

Yopi:
If someone has purchased a product, we previously allowed them to assign a seat only with the checkout id. This also requires the checkout secret, which someone will have if they are in the buy flow.

### DIFF
--- a/server/polar/customer_seat/endpoints.py
+++ b/server/polar/customer_seat/endpoints.py
@@ -1,3 +1,4 @@
+import secrets
 from typing import Annotated, cast
 
 from fastapi import Depends, Query, Request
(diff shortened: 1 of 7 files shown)


server/polar/customer_seat/endpoints.py
```diff
@@ -109,6 +110,14 @@ async def assign_seat(
             ):
                 raise NotPermitted("checkout_client_secret does not match.")
 
+        if checkout.status != CheckoutStatus.succeeded:
+            raise NotPermitted("Seats can only be assigned from a successful checkout.")
+
+        if checkout.is_expired:
```
frankie567, line 116: Redundant with the status check.

frankie567 (changes requested):

server/polar/customer_seat/endpoints.py
psincraian: I checked and the status for `succeded` checkouts don't change. I would like to avoid using this in the frontend for more than 24 hours to avoid problems

frankie567: Ok, why not. That could use a comment then because a succeeded checkout can't semantically "expire" :D

frankie567 (approved):

sebastianekstrom (approved):

PR #11073, opened 2026-04-19
oat: Prevent minting a higher privilege token than used

Yopi:
This disables being able to create a new token (if you have `organization_access_tokens:write`) that has more privileges than the token used.


frankie567 (approved):
Good catch! That's why having an API for creating tokens is a bit weird 😅 Maybe we should explore having more of a "manual" flow with clever prefilled parameters and a redirect, but that's another topic.

Yopi:
Haha yeah, I guess this should be baked into the RBAC as well (cc @joebon)

PR #11077, opened 2026-04-20
tinybird: Pass external + customer ids via readable filter

Yopi:
Even though the queries are already guarded by the readable org id filter,
it's a good idea to pass these through the readable filter as well

### DIFF
--- a/server/polar/event_type/service.py
+++ b/server/polar/event_type/service.py
@@ -2,6 +2,7 @@ from collections.abc import Sequence
 from uuid import UUID
 
 from polar.auth.models import AuthSubject
+from polar.customer.repository import CustomerRepository
 from polar.event.system import SYSTEM_EVENT_LABELS
 from polar.event.tinybird_repository import TinybirdEventRepository
 from polar.event_type.repository import EventTypeRepository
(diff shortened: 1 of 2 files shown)


server/polar/event_type/service.py
```diff
@@ -54,6 +55,22 @@ async def list_with_stats(
         if not organization_ids:
             return [], 0
 
+        customer_repository = CustomerRepository.from_session(session)
+        all_customer_ids: list[UUID] = list(customer_id or [])
+        all_external_ids: list[str] = list(external_customer_id or [])
+        if customer_id is not None:
+            all_external_ids.extend(
```
frankie567, line 62: Won't that actually _add_ the readable external ID to the requested list? Shouldn't it be an intersection?

server/polar/event_type/service.py
```diff
@@ -54,6 +55,22 @@ async def list_with_stats(
...
+                await customer_repository.get_readable_external_ids_by_ids(
```
frankie567, line 63: Shouldn't it be `get_readable_ids`?

Yopi:
| Customer ID | External Customer ID |
| ------------- | ------------- | 
| pol-1 | ext- 1 |
| pol-2 | ext-2 |

| Event ID  | Customer ID | External Customer ID |
| ------------- | ------------- | ---- |
| 1  | pol-1 | null | 
| 2  | null  | ext-1 | 
| 3  | pol2  | ext-2 |

If you search for all events with customer-id `pol-1` I was thinking that you will also expect the external events that are ingested with `ext-1`. Likewise if you search for events by `ext-1` you will expect to get system events with `pol-1`.

So if you specify any customer_id, we lookup the corresponding external ids with those customer ids and add it to the list of search. And vice-versa.

I don't think we can do an intersection, because between internal and external events they will only have one id or the other, not both

frankie567:
Ok, I think I misunderstood the objective then. I thought you wanted to make sure the customer ID was readable by the caller, but actually, it's a matter of being complete when returning results filtered by (external) Customer ID?

Yopi:
> Ok, I think I misunderstood the objective then. I thought you wanted to make sure the customer ID was readable by the caller, but actually, it's a matter of being complete when returning results filtered by (external) Customer ID?

What I describe is how it works currently (and after this change). Since we are already filtering on organization id being readable by the org this should make no change. I just wanted to have this as a defense-in-depth in case we implicitly trust the IDs somewhere else in the code.

So what I wanted to do here is to ensure that 1) the external + internal customer IDs are matching the internal + external customer IDs and 2) that they are readable by the caller

frankie567 (approved):
I think I get it :p

PR #11084, opened 2026-04-20
products: Increase product limit on orders page

Yopi:
Instead of iterating one product at a time, we can get multiple products at the same time.

This speeds up the toggling of selected products a lot when dealing with > 2 products.


frankie567 (approved):
Nice catch, feels like a copy-paste oopsie :p

PR #11090, opened 2026-04-20
fix(organization): gate CREATED → ACTIVE on all onboarding checks

psincraian:
## 📋 Summary

An organization should only transition from `CREATED` to `ACTIVE` once all three gates are satisfied: AI review PASS, Stripe identity verification, and a ready payout account. Today the AI-review task flips orgs to `ACTIVE` on `APPROVE` alone, and the Stripe `account.updated` path that used to provide the other gates was removed in the Apr 8 payout-account refactor.

## 🎯 What

- New `OrganizationService.maybe_activate(session, organization) -> bool`: single idempotent entry point that transitions `CREATED → ACTIVE` only when every gate passes. Delegates to `confirm_organization_reviewed(silent=True)` so the lifecycle (held-balance release, `initially_reviewed_at`, threshold doubling) stays in one place.
- New `OrganizationRepository.get_all_by_payout_account_admin(user_id)` for the identity-webhook fan-out.
- Wired `maybe_activate` into the three trigger points:
  - AI review `APPROVE` — replaces the direct `confirm_organization_reviewed` call in `organization_review/tasks.py`.
  - Stripe `account.updated` — `payout_account.service.update_account_from_stripe` now calls `maybe_activate` for every org linked to the updated payout account.
  - `identity.verification_session.verified` — `user.service.identity_verification_verified` now calls `maybe_activate` for every org whose payout account this user admins.
- New `server/scripts/backfill_stuck_active_orgs.py` (dry-run by default, `--execute` to apply) for orgs already stuck in `CREATED` that meet every gate.

## 🤔 Why

Root cause breakdown:

1. **Pre 2026-04-08** the AI-review APPROVE branch in `organization_review/tasks.py` was a no-op. The transition came from the Stripe `account.updated` webhook, which called the now-deleted `organization_service.update_status_from_stripe_account`.
2. **Commit `5fe86daf7` (Apr 8)** made the APPROVE branch set `status = ACTIVE` directly — but skipped identity verification and payout-account readiness.
3. **Commit `32a91cb17` (Apr 14)** swapped the direct assignment for `confirm_organization_reviewed`, but the premature-transition issue remained.

## 🔧 How

`maybe_activate` checks, in order (short-circuiting cheaply):

1. `organization.status == CREATED`
2. `details_submitted_at` set and `details` non-empty
3. Latest `OrganizationReview.verdict == PASS`
4. `payout_account.is_details_submitted AND is_charges_enabled AND is_payouts_enabled`
5. `payout_account.admin.identity_verification_status == verified`

Each of the three triggers fetches the relevant org(s) and calls `maybe_activate`; whichever completes the last gate performs the transition. `confirm_organization_reviewed(silent=True)` keeps the existing lifecycle (set_status → capabilities, `initially_reviewed_at`, threshold doubling, `organization.reviewed` task).

`payout_account.service` keeps a late import of `organization.service` because `organization.service` imports `payout_account.service` at module top (real cycle). `user.service` can import `organization.service` at module top — no cycle.

## 🧪 Testing

- [x] `uv run ruff check` + `uv run mypy` pass on all six changed files.
- [ ] `uv run task test` — not run yet; reviewers, please exercise the organization-review and webhook test suites.

### Test Instructions

1. **New-org happy path**: create an org, submit details, confirm AI agent returns `APPROVE`. Org should remain `CREATED` until payout account is set up **and** identity is verified. Then it flips to `ACTIVE`.
2. **Order-of-events fuzz**: cover all orderings — AI first + payout last, identity first + AI last, etc. In each case only the final event should activate.
3. **Backfill dry-run**:
   ```
   cd server
   uv run python -m scripts.backfill_stuck_active_orgs
   ```
   Should list ~16 orgs matching the fully-ready predicate. `--execute` activates them via the normal path.

## 📝 Additional Notes

- Agent-run review re-invocations (`POST /organizations/{id}/run-review-agent`, `context="manual"`) still do **not** trigger activation — that endpoint writes to `OrganizationAgentReview`, not to `OrganizationReview`. That's by design and unchanged here.
- Stripe webhook fan-out is sequential inside a single session. Typical cardinality is 1 org per payout account, so the N+1 concern is theoretical today; if we see higher fan-out we can preload and pass it into `maybe_activate`.

## ✅ Pre-submission Checklist

- [x] Code follows project style
- [x] Self-review performed
- [x] Comments kept minimal (one line at the late-import)
- [x] Generates no new warnings
- [ ] Tests added for new `maybe_activate` paths — would welcome reviewer input on the right test layer
- [ ] All tests pass locally
- [x] **AI/LLM Policy**: AI-assisted, lint/mypy verified locally; no runtime execution of the backfill against prod yet

### DIFF
--- a/server/polar/organization/service.py
+++ b/server/polar/organization/service.py
@@ -927,8 +927,7 @@ class OrganizationService:
         admin = payout_account.admin
         if (
             admin is None
-            or admin.identity_verification_status
-            != IdentityVerificationStatus.verified
+            or admin.identity_verification_status != IdentityVerificationStatus.verified
         ):
             return False
 
--- a/server/polar/organization/service.py
+++ b/server/polar/organization/service.py
@@ -867,19 +867,29 @@ class OrganizationService:
     async def maybe_activate(
         self, session: AsyncSession, organization: Organization
     ) -> bool:
-        """Transition CREATED → ACTIVE only when all onboarding gates pass.
+        """Transition → ACTIVE only when every onboarding gate passes.
 
         Gates:
-          1. Details submitted (details + details_submitted_at)
-          2. AI review verdict = PASS
-          3. Payout account: details submitted, charges enabled, payouts enabled
-          4. Payout account admin: identity verified
+          1. Status can currently transition to ACTIVE (not already ACTIVE,
+             not OFFBOARDING).
+          2. Details submitted (details + details_submitted_at).
+          3. Review is approved: verdict PASS, or verdict FAIL with an
+             APPROVED appeal.
+          4. Payout account: details submitted, charges enabled, payouts
+             enabled.
+          5. Payout account admin: identity verified.
 
         Idempotent — safe to call from multiple triggers (AI review, Stripe
-        account.updated, identity verification). Returns True iff the org was
-        transitioned.
+        account.updated, identity verification, appeal approval). Returns True
+        iff the org was transitioned.
         """
-        if organization.status != OrganizationStatus.CREATED:
+        if organization.status not in (
+            OrganizationStatus.CREATED,
+            OrganizationStatus.REVIEW,
+            OrganizationStatus.SNOOZED,
+            OrganizationStatus.DENIED,
+            OrganizationStatus.BLOCKED,
+        ):
             return False
 
         if not organization.details_submitted_at or not organization.details:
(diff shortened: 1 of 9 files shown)


server/polar/organization/service.py
```diff
@@ -864,6 +864,90 @@ async def confirm_organization_reviewed(
...
+
+        payout_account_repository = PayoutAccountRepository.from_session(session)
+        payout_account = await payout_account_repository.get_by_id(
+            organization.payout_account_id,
+            options=(joinedload(PayoutAccount.admin),),
+        )
+        if payout_account is None:
+            return False
+        if not (
+            payout_account.is_details_submitted
```
frankie567: There is a property "is_payout_ready" that abstracts those details which are only relevant for Stripe accounts.

psincraian (commented):
Addressed in f7cc5b50a :)

frankie567 (approved):

PR #11094, opened 2026-04-21
Allow region subtags to switch within primary language IETF for better currency formatting

pieterbeulque:
Helpful context to review this PR:

 - We have _supported locales_: `en`, `nl`, `fr`, …
 - From these, we derive _accepted locales_: `en-US`, `nl-BE`, `fr-FR`, `fr-CA`

Browsers report their support languages in a HTTP header `Accept-Language` in a list like this:

image

would match, for example, `en-CA q=1.0; en-US q=0.8; nl q=0.8; en`.

The regional part of the IETF tag (e.g. CA / US), it what `Intl` relies on to show `US$` and `$` for `usd` / `$` and `CA$` for `cad` for Canada & USA respectively.

When localized checkout is set and the checkout locale isn't explicitly set (i.e. `null`), we fall back to auto-detecting based on the `Accept-Language` header, and everything works as expected. However, we lost this user preference when merchants set a language-only IETF locale (e.g. `en`) or when we the default to `en-US` (when people didn't opt in to localized checkout).

What this PR does: even when the checkout locale is explicitly set, find the first IETF tag matching on the primary language in the `Accept-Language` header, and then override the checkout setting with the regional variant as preferred by the user, mapping `en` or `en-US` to `en-CA`. For the translations themselves this will have no impact, but it helps the `Intl` formatting helpers to use the proper regional variants if relevant.

### DIFF
--- a/clients/apps/web/src/utils/i18n/index.ts
+++ b/clients/apps/web/src/utils/i18n/index.ts
@@ -5,7 +5,7 @@ import {
 } from '@polar-sh/i18n'
 import { headers } from 'next/headers'
 
-function parseAcceptLanguageHeader(
+export function parseAcceptLanguageHeader(
   acceptLanguageHeader: string | null,
 ): { code: string; q: number }[] {
   if (!acceptLanguageHeader) return []
@@ -40,7 +40,7 @@ function getPrimaryLanguageForLocale(locale: string): string {
   return locale.split('-')[0].toLowerCase()
 }
 
-function findMatchingLocaleInAcceptLanguageHeader(
+export function findMatchingLocaleInAcceptLanguageHeader(
   acceptLanguageHeader: string | null,
   targetLocale: AcceptedLocale,
 ): AcceptedLocale | null {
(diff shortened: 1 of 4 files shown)


clients/apps/web/src/utils/i18n/index.ts
```diff
@@ -30,6 +36,31 @@ function getLocaleFromAcceptLanguage(
   return DEFAULT_LOCALE
 }
 
+function getPrimaryLanguageForLocale(locale: string): string {
+  return locale.split('-')[0].toLowerCase()
+}
+
+function findMatchingLocaleInAcceptLanguageHeader(
```
frankie567: Maybe picky, but it would help to have a few unit tests to understand the different inputs/outputs we expect :)

pieterbeulque: Yes, it could definitely benefit from some unit tests.

There you go!

frankie567 (approved):

sebastianekstrom (approved):

PR #11095, opened 2026-04-21
polar: Backfill set created_at

Yopi:
As discussed on Slack, we might want to do this?

For the customer set created_at to the organization created_at .
For the members set it to when the UserOrganization created_at.
For the free subscription set it to organization.created_at.

The events are still created with `now` as the created_at, which might be fine semantically

### DIFF
--- a/server/polar/customer/service.py
+++ b/server/polar/customer/service.py
@@ -1,6 +1,7 @@
 import builtins
 import uuid
 from collections.abc import Sequence
+from datetime import datetime
 from typing import Any
 
 import structlog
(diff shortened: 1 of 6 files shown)


frankie567 (commented):
Makes sense! I would be on the side though of also backdating the events as well. I'm worried we might be confused one day if we look at the events history of a Customer and sees they're completely off with what the actual data says. WDYT?

Yopi:
> Makes sense! I would be on the side though of also backdating the events as well. I'm worried we might be confused one day if we look at the events history of a Customer and sees they're completely off with what the actual data says. WDYT?

I also agree, but was worried it would become a bigger thing. I added support to specify it now, so I think it works. And then we can roll it back from the services later if we want

frankie567 (approved):

PR #11096, opened 2026-04-21
server: remove held balances

frankie567:
Accounts are now automatically created alongside the organization, so there is no need anymore to keep track of held balances.


server/migrations/versions/2026-04-21-0843_remove_heldbalance.py
```diff
@@ -0,0 +1,150 @@
...
+# Polar Custom Imports
+
+# revision identifiers, used by Alembic.
+revision = "532186b5c028"
+down_revision = "346767c5c3fd"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
```
psincraian, line 22: I think this will still cause some errors on Sentry, as we first deploy the migraiton and then the code. Mainly SQLAlchemy tries to map every class to a table

psincraian (approved):
Nice job 👏

server/migrations/versions/2026-04-21-0843_remove_heldbalance.py
frankie567: Apparently it went well :D Since there was no live code that was actually querying them, I think it passed smoothly

PR #10975, opened 2026-04-14
polar: Move away from inline imports

Yopi:
We want to avoid loading the entire SDK, because it doesn't get lazily loaded, but having inline imports in every method does not look great either.

By splitting the client and implementation like this, we can get code that is a bit nicer to read without getting a memory footprint in the cases where we dont want it.

PR #11099, opened 2026-04-21
polar: Track AI onboarding costs

Yopi:
By ingesting an event named ai.organization_review.{submission|setup_complete|threshold|manual|appeal} with the llm tokens and cost, we can start to track the onboarding costs for our customers inside Polar.

### DIFF
--- a/server/polar/integrations/polar/client.py
+++ b/server/polar/integrations/polar/client.py
@@ -231,11 +231,11 @@ class PolarSelfClient:
             except httpx.RequestError as e:
                 _raise_network_error(span, e, "track_event_ingestion")
 
-    async def track_llm_usage(
+    async def track_organization_review_usage(
         self,
         *,
         external_customer_id: str,
-        event_name: str,
+        review_context: str,
         vendor: str,
         model: str,
         input_tokens: int,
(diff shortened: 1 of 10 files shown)


server/polar/integrations/polar/client.py
```diff
@@ -230,6 +231,70 @@ async def track_event_ingestion(
...
+        *,
+        external_customer_id: str,
+        event_name: str,
+        vendor: str,
+        model: str,
+        input_tokens: int,
+        output_tokens: int,
+        cost_usd: Decimal,
+    ) -> None:
+        from polar_sdk.models import (
```
frankie567, line 245: Can't we have top-level imports?

frankie567 (approved):
Amazing!

server/polar/integrations/polar/client.py
Yopi: I have a separate PR here that I would be happy to get your input on regarding the approach: https://github.com/polarsource/polar/pull/10975

pieterbeulque (approved):
Agree with the logic, didn't really look at the code

server/polar/integrations/polar/client.py
Yopi: I'll circle back to your suggestion in #10975, and merge this like this for now

PR #11104, opened 2026-04-21
feat(organization): migrate enforcement reads to capabilities (PR2)

psincraian:
## 📋 Summary

PR2 of the [organization capabilities override plan](https://github.com/polarsource/polar/pull/11061). Migrates the six merchant-facing enforcement gates from `Organization.status` checks to the `capabilities` JSONB column added in PR1, so backoffice admins can flip a single capability without inventing new statuses.

> ⚠️ **Backfill prerequisite — already done.** This PR reads from `Organization.capabilities`. The backfill script [`scripts/backfill_org_capabilities.py`](server/scripts/backfill_org_capabilities.py) (merged earlier on main) has already been executed in production, so every org row has its capabilities populated. The hybrid accessors still defensively fall back to `STATUS_CAPABILITIES[status]` if a row is unexpectedly NULL. The column NOT NULL flip will be a separate small follow-up PR.

## 🎯 What

- **New typed accessors on `Organization`** — `can_accept_payments`, `can_renew_subscriptions`, `can_payout`, `can_refund` (joining the existing `can_authenticate` + new `can_access_dashboard`). All are hybrid properties with both Python and SQL expression forms. A private `_capability()` helper falls back to `STATUS_CAPABILITIES[status]` for any row where the column hasn't been backfilled.
- **`set_status()`** now also writes the legacy `refunds_blocked` shim so it doesn't drift across status transitions until PR5 drops it.
- **Enforcement migrated**:
  - `checkout/service.py:get_by_id` and the `create` early-gate → `can_authenticate` (these were originally `is_blocked()`, i.e. `status == BLOCKED` → `api_access=False`; they're not "can accept payments" gates — creating a checkout object is read-shaped). The actual payment-acceptance check still happens at `/confirm` via `is_organization_ready_for_payment`.
  - `is_organization_ready_for_payment` → `can_accept_payments`
  - Subscription cycle: scheduler SQL filter → `can_renew_subscriptions` + defensive short-circuit in `cycle()`
  - Dunning: `order/repository.py:get_due_dunning_orders` JSONB filter + `process_dunning_order` short-circuit + `trigger_payment` per-trigger gating (renewal vs checkout)
  - Payouts → `can_payout`
  - Refunds → `not can_refund` (with the legacy `refunds_blocked` OR-shim retained until PR5)
- **Routing fixes** — restored `set_status()` calls in `organization_review/tasks.py` and the seed/enable_payments scripts (bypassed in PR1's model-only commit).

## 🤔 Why

Today, capabilities like checkout, payouts, and refunds are coupled to `Organization.status`. To deny one capability or grant one on an otherwise-blocked org, operators have no fine-grained control. PR2 flips the gates to read from the JSONB column, enabling the override UI in PR4. Status transitions still wipe overrides — temporary by design.

## 🔧 How

Per the plan, defaults of `STATUS_CAPABILITIES` (in `models/organization.py`) preserve today's behaviour:
- `checkout_payments` follows `payment_ready_statuses()` (ACTIVE, OFFBOARDING, REVIEW, SNOOZED).
- `subscription_renewals` follows the looser scheduler filter (everything except CREATED/BLOCKED/DENIED) — covers both cycle and dunning, since they're two sides of the same "continue billing" switch.

The legacy `refunds_blocked` column stays as a **write-through shim** until PR5 drops it; `set_status()` mirrors `capabilities["refunds"]` to it on every transition so the two stay in sync.

## 🧪 Testing

- [x] All existing tests pass (`uv run task test_fast` across organization/order/checkout/payout/refund/oauth2/customer_session/checkout_link/customer_portal/organization_access_token/personal_access_token/backoffice — 1451+ passed)
- [x] New tests: `TestCapabilityOverrides`, `TestSetStatusCapabilities`
- [x] Lint + types clean (`uv run task lint && uv run task lint_types`)

### Test instructions

```bash
cd server
uv run task test_fast -- tests/organization tests/order tests/checkout tests/payout tests/refund
```

Manual: in psql, `SELECT status, capabilities FROM organizations LIMIT 20;` — every row's `capabilities` matches `STATUS_CAPABILITIES[status]` (modulo any prior `refunds_blocked` flips, which were migrated by the backfill script).

## Notes for reviewers

- **Out of scope** (separate follow-up PRs): alembic migration to flip `capabilities` to NOT NULL (after this PR has been deployed for a release cycle), API schema exposure (PR3), backoffice override UI (PR4), dropping `refunds_blocked` and `blocked_at` (PR5).
- **Behavioral note on grandfathering**: a pre-cutoff (Aug 4 2025) org stuck in CREATED would previously bypass the `payment_ready_statuses` check and be considered ready. With the capability gate, CREATED defaults to `checkout_payments=False`, so this edge case now returns False. In practice no such orgs exist.


server/polar/organization/service.py
```diff
@@ -1095,22 +1095,15 @@ async def is_organization_ready_for_payment(
         if settings.ENV == Environment.sandbox:
```
psincraian: In another PR, I would like to simplify sandbox and move orgs to active status by default

server/polar/organization/service.py
```diff
@@ -1070,71 +1070,16 @@ async def set_organization_offboarding(
...
-        if not admin or admin.identity_verification_status not in [
-            IdentityVerificationStatus.verified,
-            IdentityVerificationStatus.pending,
-        ]:
-            return False
-
-        return True
+    def is_organization_ready_for_payment(self, organization: Organization) -> bool:
+        """Whether the org can accept payments. Sandbox bypasses the capability."""
+        return settings.is_sandbox() or organization.can_accept_payments
```
psincraian, line 1082: In another PR, I will try to improve sandbox flow to move orgs to activated by default

server/polar/order/repository.py
```diff
@@ -35,6 +35,7 @@
     UserOrganization,
 )
 from polar.models.order import OrderStatus
+from polar.models.organization import Organization as OrgModel
```
frankie567: Actually, you could directly import `Organization` from `polar.models` and remove the import from ` polar.auth.models`. They're exactly the same object, it's just re-exported from `polar.auth.models` for convenience.

frankie567 (approved):
Super clean approach, I really love how it makes the logic much more smooth across the different payment flows 👍

PR #11109, opened 2026-04-21
metrics: Introduce a trial_monthly_recurring_revenue metric

Yopi:
The trial mrr metric will allow a merchant to track the MRR that is currently in trials.

Additionally this PR refactors the `active_subscriptions` based metrics to be calculated in `get_sql_expression`, like all other metrics.

### DIFF
--- a/server/polar/metrics/metrics.py
+++ b/server/polar/metrics/metrics.py
@@ -200,6 +200,28 @@ class TrialMonthlyRecurringRevenueMetric(SQLMetric):
         return cumulative_last(periods, cls.slug)
 
 
+class TrialCommittedMonthlyRecurringRevenueMetric(SQLMetric):
+    slug = "trial_committed_monthly_recurring_revenue"
+    display_name = "Trial Committed Monthly Recurring Revenue"
+    type = MetricType.currency
+    query = MetricQuery.active_subscriptions
+
+    @classmethod
+    def get_sql_expression(
+        cls, t: ColumnElement[datetime], i: TimeInterval, now: datetime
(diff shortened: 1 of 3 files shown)


frankie567 (approved):

server/polar/metrics/metrics.py
```diff
@@ -11,19 +11,52 @@
...
+def _not_in_trial(t: ColumnElement[datetime], i: TimeInterval) -> ColumnElement[bool]:
+    buckets = t.table.c
+    return or_(
+        buckets.trial_end.is_(None),
+        i.sql_date_trunc(cast(SQLColumnExpression[datetime], buckets.trial_end))
+        <= i.sql_date_trunc(t),
+    )
+
+
+def _in_trial(t: ColumnElement[datetime], i: TimeInterval) -> ColumnElement[bool]:
```
pieterbeulque, line 41: If a subscription is still in the trial period but set to cancel at period end, should it be considered in trial revenue?

To be in line with MRR, technically yes. But I'm not sure if that's what you want to see if you look for your trialing revenue.

Yopi: It feels sort of overkill to have `trial_mrr` and `trial_committed_mrr`, but maybe we need to for completeness sake.

I would be more for that than to have different definitions of mrr for trial and non trial

pieterbeulque: I can live with that!

Yopi: Added the committed one as well now

pieterbeulque (approved):

PR #11124, opened 2026-04-21
paginate: add count_subquery helper and lint for .subquery() misuse

psincraian:
## Summary

Follow-up to [#11119](https://github.com/polarsource/polar/pull/11119) — prevent the same class of bug from recurring.

- Extract the count-subquery projection into a named helper `count_subquery(statement)` in `polar.kit.pagination`. Both paginate helpers (`kit.pagination.paginate()` and `RepositoryBase.paginate()`) now go through it.
- Add `scripts/lint_subquery.py`, a small AST walker that flags `select(Model).subquery()` without `.with_only_columns(...)`. Wired into `uv run task lint_check` via a new `lint_subquery` taskipy task, so CI picks it up through the existing lint step in `test_server.yaml` — no new workflow needed.
- Append a rule under "Repository Pattern" in `server/CLAUDE.md` documenting the footgun and pointing at the helper + lint.

## Why

`deferred=True` **does not propagate into `.subquery()`**. The count subquery inside paginate was silently re-materializing every mapped column of the target entity — including the deferred `blocked_at` on `Organization` (the column dropped in #11105, which triggered the sandbox outage fixed in #11119). It was also loading TSVECTOR search vectors and other deferred large-string columns on 9 other models (`Order`, `Customer`, `Product`, `Benefit`, `PayoutAccount`, `Organization`, …) into every paginate count — a real performance tax, not just a latent bug.

`#11119` fixed the two call sites. This PR makes the pattern idiomatic via the helper and prevents future regressions via the lint. The CLAUDE.md rule alone is insufficient — docs aren't in every reader's head (human or agent) when they write a new query.

## Lint scope

- **Flags** the high-confidence inline pattern `select(Model).subquery()` (with optional `.where`/`.order_by` in between).
- **Accepts** `.with_only_columns(...)` anywhere in the chain, `union`/`union_all`/`intersect`, explicit-column selects (`select(col1, col2, ...)`), and the `# noqa: subquery-all-columns` escape hatch.
- **Does not flag** variable-rooted chains (`statement.subquery()`) — static analysis can't determine the root. Those rely on the CLAUDE.md rule and code review.

## Out of scope / follow-up

The rolling-deploy half of the original incident (bundling `op.drop_column` + ORM attribute removal in one PR) is a separate class of bug worth its own plan. `alembic check` already runs in CI but can't catch it because the PR is internally consistent at HEAD. Flagging here; not addressed in this PR.

## Test plan

- [x] `uv run task lint_check` — ruff + format + new subquery lint all green (835 files scanned)
- [x] `uv run task lint_types` — mypy clean
- [x] `pytest -k paginat` — 3 passed, 2 skipped
- [x] `pytest tests/organization/` — 262 passed
- [x] Synthetic verification locally: `select(Organization).subquery()` (inline + chained) is flagged; `# noqa`, `with_only_columns`, and explicit-column variants pass

### End-to-end CI verification

Confirmed on GitHub Actions that the lint fails the build on a real violation and returns to green once removed:

| Commit | Canary | Server: Linters 📝 | Duration |
|---|---|---|---|
| [`63e9a9f`](https://github.com/polarsource/polar/pull/11124/commits/63e9a9fe9) | — | ✅ pass | 2m20s |
| [`c2b445b`](https://github.com/polarsource/polar/pull/11124/commits/c2b445be9) | added `select(Organization).subquery()` in `_lint_canary.py` | ❌ fail | 14s |
| [`24f31dc`](https://github.com/polarsource/polar/pull/11124/commits/24f31dc51) | canary removed | ✅ pass | — |

The canary commits are part of the history but leave the tree unchanged — safe to squash-merge. (015ff51 dropped the lint in response to review; 1e96713 reverted that drop after reconsidering.)


server/scripts/lint_subquery.py
```diff
@@ -0,0 +1,174 @@
+"""Flag `.subquery()` calls that don't project explicit columns.
```
frankie567: Isn't it a bit overkill to have a specific linter script for that purpose? ☺️

psincraian: Fair point — dropped it in 015ff51. Kept the `count_subquery()` helper and the CLAUDE.md rule, which is where the value was.

psincraian: ahh, this Claude Code automatic fix :( I like when it fixes lint but not when answers comments

psincraian: We have some rules that should not be broken, and this one can cause some problems on production for a short period of time. But I think that's worse than having a lint running every time.

It can become a problem if those lints are unrelaible or become too slow. But if the second happen, I think we can find some optimisations.

frankie567: Not 100% convinced. I guess what troubles me the most is that we have dangling script for linting rules, but unfortunately Ruff doesn't support to add custom rules. But well, we can see how it goes.

frankie567 (approved):

PR #11126, opened 2026-04-21
feat(authz): centralized authorization with PolicyGuard

joebon:
## 📋 Summary

Introduce a centralized authorization module (`server/polar/authz/`) that separates authentication (identity) from authorization (permissions). Authorization is enforced declaratively via FastAPI dependency injection — endpoints declare what policy they need, and it's impossible to forget the check.

## 🎯 What

- New `server/polar/authz/` module with policy functions, OrgPolicyGuard DI factory, and typed `Authorize*` dependencies
- Organization endpoints migrated to `Authorize*` DI dependencies (get_account, delete, invite, leave, remove_member)
- Authorization logic removed from `OrganizationService.request_deletion`
- `AccountService.get_by_organization` added for auth-free account lookup
- Handbook documentation for the authorization system
- 25 new tests for policies, org-access resolver, and PolicyGuard behavior

## 🤔 Why

Authorization for organization endpoints is currently ad-hoc — manual `get_admin_user()` calls scattered across endpoints and services, with no single place to audit who can do what. This makes it easy to forget an auth check and hard to reason about access control.

## 🔧 How

**OrgPolicyGuard** — a FastAPI dependency factory that resolves the organization, checks membership, runs a policy function, and returns an `AuthorizedOrganization` containing both the org and auth_subject:

```python
# Before — manual, easy to forget
async def get_account(id: OrganizationID, auth_subject: auth.OrganizationsRead, ...):
    organization = await organization_service.get(session, auth_subject, id)
    if not await account_service.is_user_admin(...):
        raise NotPermitted(...)

# After — declarative, enforced by DI
async def get_account(authorized: AuthorizeFinanceRead, ...):
    account = await account_service.get_by_organization(session, authorized.organization.id)
```

**Policies return denial reasons** — `True` if allowed, a descriptive string if denied (e.g., "Only the account admin can delete an organization with an account"). The string is passed through to the 403 response.

**Typed dependencies** — `AuthorizeFinanceRead`, `AuthorizeMembersManage`, `AuthorizeOrgDelete`, `AuthorizeOrgAccessUser` are pre-built for common patterns.

See `handbook/engineering/backend-development/authorization.mdx` for full documentation.

## 🧪 Testing

- [x] All existing tests pass (`uv run task test` for backend)
- [x] I have added new tests for new functionality
- [x] I have run linting and type checking (`uv run task lint && uv run task lint_types` for backend)

### Test Instructions

1. Run `uv run pytest tests/organization/ tests/authz/ -q` — 285 tests pass
2. Review `server/tests/authz/` — 25 new tests covering:
   - Policy unit tests (admin allowed, non-admin denied with message, no-account orgs, org tokens)
   - Org-access resolver (user with/without orgs, organization subject)
   - PolicyGuard integration (401 anonymous, 404 non-member, 403 non-admin with message, 200 admin)

## 📝 Additional Notes

- This is the first PR in a series — subsequent PRs will migrate repositories to pure data access and clean up `web_read`/`web_write` scopes
- `Account.admin_id` remains as the source of truth for finance access — policies wrap it
- No database migrations, no API response changes

## ✅ Pre-submission Checklist

- [x] All tests pass locally

### DIFF
--- a/handbook/engineering/backend-development/authorization.mdx
+++ b/handbook/engineering/backend-development/authorization.mdx
@@ -1,0 +1,180 @@
+---
+title: "Authorization"
+description: "How authorization works in Polar: authentication vs authorization, PolicyGuard, policies, and org-scoped data access."
+---
+
+## Overview
+
+Authorization in Polar is split into two layers:
+
+- **Authentication** (`Authenticator`): "Who are you?" — validates tokens/sessions, resolves the subject (User, Organization, Customer, Member), extracts token scopes.
+- **Authorization** (`server/polar/authz/`): "Are you allowed to do this?" — checks org membership, evaluates policies, enforces access control.
+
+These are separate concerns. Authentication happens via FastAPI's `Authenticator` dependency. Authorization happens via **PolicyGuard** dependencies that resolve resources and check policies before the endpoint body runs.
+
+## PolicyGuard
+
+A PolicyGuard is a FastAPI dependency factory that combines resource resolution and authorization into a single step. When an endpoint declares a PolicyGuard dependency, the framework:
+
+1. Authenticates the subject
+2. Resolves the resource from the path parameter
+3. Checks that the subject has access to the owning organization
+4. Evaluates the policy function
+5. Returns an `AuthorizedOrganization` (or `AuthorizedAccount`, etc.) containing both the resource and the auth subject
+
+If any step fails, the appropriate HTTP error is raised before the endpoint body executes.
+
+```python
+from polar.authz.dependencies import AuthorizeFinanceRead
+
+@router.get("/{id}/account")
+async def get_account(
+    authorized: AuthorizeFinanceRead,
+    session: AsyncReadSession = Depends(get_db_read_session),
+) -> Account:
+    # authorized.organization is already resolved and access-checked
+    # authorized.auth_subject is the authenticated subject
+    account = await account_service.get_by_organization(
+        session, authorized.organization.id
+    )
+    if account is None:
+        raise ResourceNotFound()
+    return account
+```
+
+### Error behavior
+
+| Scenario | HTTP status | Reason |
+|----------|-------------|--------|
+| No valid credentials | 401 Unauthorized | `Authenticator` rejects the request |
+| Token lacks required scopes | 403 Forbidden | `Authenticator` scope check fails |
+| Resource doesn't exist | 404 Not Found | Resource not found in database |
+| Subject not a member of the org | 404 Not Found | Returns 404 (not 403) to avoid leaking existence |
+| Subject is a member but policy denied | 403 Forbidden | Policy returned a denial reason |
+
+### Guard variants
+
+There are three PolicyGuard variants for different resource types:
+
+| Guard | Resolves by | Use case |
+|-------|-------------|----------|
+| `OrgPolicyGuard` | `{id}` → Organization | Org-scoped endpoints (`/organizations/{id}/...`) |
+| `AccountPolicyGuard` | `{id}` → Account → owning Organization | Account endpoints (`/accounts/{id}`) |
+| `PayoutAccountPolicyGuard` | `{id}` → PayoutAccount → owning Organization | Payout account endpoints (`/payout-accounts/{id}`) |
+
+### Typed dependencies
+
+Pre-built typed dependencies are defined in `server/polar/authz/dependencies.py`:
+
+```python
+AuthorizeFinanceRead    # OrgPolicyGuard + finance.can_read
+AuthorizeFinanceWrite   # OrgPolicyGuard + finance.can_write
+AuthorizeMembersManage  # OrgPolicyGuard + members.can_manage (User-only)
+AuthorizeOrgDelete      # OrgPolicyGuard + organization.can_delete (User-only)
+AuthorizeOrgAccessUser  # OrgPolicyGuard + always allow (User-only)
+AuthorizeAccountRead    # AccountPolicyGuard + finance.can_read
+AuthorizeAccountWrite   # AccountPolicyGuard + finance.can_write
+AuthorizePayoutAccountRead   # PayoutAccountPolicyGuard + finance.can_read
+AuthorizePayoutAccountWrite  # PayoutAccountPolicyGuard + finance.can_write
+```
+
+## Policies
+
+Policies are async functions in `server/polar/authz/policies/`. Each policy receives the database session, auth subject, and organization, and returns either `True` (allowed) or a denial reason string.
+
+```python
+# server/polar/authz/policies/finance.py
+async def can_read(
+    session: AsyncReadSession,
+    auth_subject: AuthSubject[User | Organization],
+    organization: Organization,
+) -> PolicyResult:
+    if is_organization(auth_subject):
+        return True
+    if is_user(auth_subject):
+        if organization.account_id is None:
+            return "Organization has no account"
+        if await account_service.is_user_admin(
+            session, organization.account_id, auth_subject.subject
+        ):
+            return True
+        return "Only the account admin can access financial information"
+    return "Not permitted"
+```
+
+The denial reason string is passed through to the `NotPermitted` exception and appears in the 403 response body.
+
+### Policy files
+
+| File | Functions | Controls access to |
+|------|-----------|-------------------|
+| `policies/finance.py` | `can_read`, `can_write` | Accounts, transactions, payouts |
+| `policies/organization.py` | `can_delete` | Organization deletion |
+| `policies/members.py` | `can_manage` | Inviting, removing, and changing members |
+
+## Org-scoped data access
+
+Repositories are pure data access — they don't check authentication or authorization. Instead, they accept explicit org IDs:
+
+```python
+# Repository — pure data access
+class ProductRepository(RepositoryBase[Product]):
+    def get_by_org_ids_statement(self, org_ids: set[UUID]) -> Select[tuple[Product]]:
+        return self.get_base_statement().where(
+            Product.organization_id.in_(org_ids)
+        )
+```
+
+Services resolve which orgs the subject can access, then pass the IDs to the repository:
+
+```python
+# Service — resolves org access, delegates to repository
+from polar.authz.service import get_accessible_org_ids
+
+class ProductService:
+    async def list(self, session, auth_subject, ...):
+        org_ids = await get_accessible_org_ids(session, auth_subject)
+        repository = ProductRepository.from_session(session)
+        statement = repository.get_by_org_ids_statement(org_ids)
+        ...
+```
+
+`get_accessible_org_ids` returns the set of organization IDs the subject can access:
+- For a **User**: all orgs they're a member of (via `UserOrganization`)
+- For an **Organization** token: just that org's ID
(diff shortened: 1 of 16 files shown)


server/polar/authz/dependencies.py
```diff
@@ -0,0 +1,204 @@
...
+    )
+
+    async def dependency(
+        id: OrganizationID,
+        auth_subject: Annotated[
+            AuthSubject[User | Organization], Depends(_authenticator)
+        ],
+        session: AsyncReadSession = Depends(get_db_read_session),
+    ) -> AuthorizedOrganization:
+        from polar.organization.repository import OrganizationRepository
```
frankie567: Could we import it at top level?

Yopi: Can the imports be moved to the top of the file, or is there a reason why you want them to be inline?

server/polar/authz/dependencies.py
```diff
@@ -0,0 +1,204 @@
...
+
+        org_ids = await get_accessible_org_ids(session, auth_subject)
+
+        repository = OrganizationRepository.from_session(session)
+        organization = await repository.get_by_id(id)
+
+        if organization is None or organization.id not in org_ids:
```
frankie567: Wondering if it would make sense to have that as a single query:

```py
organization = await get_accessible_organization(session, auth_subject, id)
if organization is None:
    raise ResourceNotFound()
````

File: server/polar/authz/dependencies.py (hunk 1 above)
Comment (Yopi) [reply server/polar/authz/dependencies.py]: Ah is it so it's only imported in the paths that they're used?

File: server/polar/authz/dependencies.py [hunk 3]
```diff
@@ -0,0 +1,204 @@
...
+        Scope.organizations_read,
+        Scope.organizations_write,
+    }
+
+    _authenticator = Authenticator(
+        allowed_subjects=_allowed,
+        required_scopes=_scopes,
+    )
+
+    async def dependency(
```
Yopi, line 75: This is a neat solution! I like it

server/polar/authz/dependencies.py
```diff
@@ -0,0 +1,204 @@
...
+        from .policies import organization as org_policy
+
+        return await org_policy.can_delete(session, auth_subject, organization)
+
+    return policy
+
+
+AuthorizeFinanceRead = Annotated[
+    AuthorizedOrganization, Depends(OrgPolicyGuard(_finance_can_read()))
+]
```
Yopi: If you pull out `PolicyResult` into its own file I think you can do 
```
AuthorizeFinanceRead = Annotated[
    AuthorizedOrganization, Depends(OrgPolicyGuard(finance.can_read()))
]
```
Directly without the intermediary private method.

server/polar/organization/endpoints.py
```diff
@@ -127,24 +133,13 @@ async def get(
     tags=[APITag.private],
 )
 async def get_account(
-    id: OrganizationID,
-    auth_subject: auth.OrganizationsRead,
+    authorized: AuthorizeFinanceRead,
```
frankie567: I think we should come up with a better name for `AuthorizedOrganization`, the guards and the argument itself.

Just looking at the endpoint logic, it's quite surprising to see `authorized.organization` or `authorized.auth_subject`.

What you would think of:

```py
@dataclass(frozen=True)
class AuthzContext:
    organization: OrganizationModel
    auth_subject: AuthSubject[User | Organization]
```

```py
from polar import authz # We could re-export the deps from the root module to have a clear namespace

async def get_account(
    authz_context: authz.FinanceRead,
    session: AsyncReadSession = Depends(get_db_read_session),
) -> Account:
    authz_context.organization
    authz_context.auth_subject
```

server/polar/organization/endpoints.py
```diff
@@ -272,13 +266,9 @@ async def delete(
     If deletion cannot proceed immediately (has orders, subscriptions, or
     Stripe deletion fails), a support ticket will be created for manual handling.
     """
-    organization = await organization_service.get(session, auth_subject, id)
-
-    if organization is None:
-        raise ResourceNotFound()
-
+    assert is_user(authorized.auth_subject)
```
frankie567: Would be nice to be able to avoid that by making `AuthorizedOrganization` a generic. Something like:

```py
class AuthorizedOrganization[S: User | Organization]:
    organization: Organization
    auth_subject: S
```

And then the guards would be:

```py
AuthorizeOrgDelete = Annotated[
    AuthorizedOrganization[User],
    Depends(
        OrgPolicyGuard(
            _org_can_delete(),
            allowed_subjects={User},
            required_scopes={Scope.web_write, Scope.organizations_write},
        )
    ),
]
```

Which would then automatically please the type checker.

server/polar/authz/dependencies.py
```diff
@@ -0,0 +1,204 @@
...
+AuthorizeFinanceWrite = Annotated[
+    AuthorizedOrganization, Depends(OrgPolicyGuard(_finance_can_write()))
```
Yopi: This should have required_scopes right?

frankie567 (changes requested):
Great work! Really like this unified approach.

I've a few feedback regarding the API itself, I think we can make it even better :)

Yopi (commented):
Really nice so far!

server/polar/authz/dependencies.py
joebon: Good find! Will fix

server/polar/authz/dependencies.py
joebon: Nopes, no reason - I'll move it

joebon: Totes!

server/polar/authz/dependencies.py
joebon: I'll give it a try! Should I put `PolicyResult` in `__init__.py` or a dedicated module `policy_result`?

server/polar/organization/endpoints.py
joebon: Yeah, I agree the name(s) aren't great.

My idea with the `Authorize*` prefix for all of these (FastAPI DI) dependencies were that their names would immediately signal that this is what actually performs authorization.
Makes sense @frankie567 and is it still equally clear with `authz.FinanceRead`?

I _def_ get your point re: `author_context` but I'm not blown away. I get some `data` / `misc` vibe of it. Ideally we'd just call it `authz`, but that's a conflict with the module itself 🤦.
Is `auth_result` to silly`?

Naming of things…

server/polar/organization/endpoints.py
joebon: Sounds like a nice approach! But what does `AuthorizedOrganization[User]` really imply? An authorized organization that is a user? 🤯

server/polar/authz/dependencies.py
joebon: I'll look into it, thanks for noticing!

server/polar/organization/endpoints.py
frankie567: > My idea with the `Authorize*` prefix for all of these (FastAPI DI) dependencies were that their names would immediately signal that this is what actually performs authorization. Makes sense @frankie567 and is it still equally clear with `authz.FinanceRead`?

Both are okay I would say, but maybe we should avoid to import the module so we could reuse it as an argument name (see below).

> I _def_ get your point re: `author_context` but I'm not blown away. I get some `data` / `misc` vibe of it. Ideally we'd just call it `authz`, but that's a conflict with the module itself 🤦. Is `auth_result` to silly`?
> 
> Naming of things…

Yeah, not 100% convinced with the "context" thing neither 🤔  

`authz` would actually be the most convenient and clear IMO. It's also the module name, but since we're not importing it directly, I can live with that.

server/polar/organization/endpoints.py
frankie567: I think it comes with a rename of `AuthorizedOrganization` (e.g. `AuthzContext`). This way it's clearer:

```py
AuthzContext[User] # an auth context with a User
AuthzContext[User | Organization] # an auth context with either an Org or User
```

joebon: Ah, yes - that makes perfect sense! Fixing

server/polar/authz/dependencies.py
joebon: This is actually mimicking the current behavior, see [this](https://github.com/polarsource/polar/blob/77b5e16837dee92b3e2f928443a22d9433efecb1/server/polar/organization/auth.py#L9-L22).
I agree that the permitted scopes here are too wide, but I'd suggest to not introduce behavioral changes in this PR but rather in a dedicated PR. Agree @Yopi ?

server/polar/authz/dependencies.py
Yopi: I think a dedicated file is fine

server/polar/authz/dependencies.py
Yopi: Yes definitely, let's keep the same functionality

server/polar/authz/dependencies.py
joebon: Added a `types` module and put it there

server/polar/authz/repository.py
```diff
@@ -0,0 +1,47 @@
+from uuid import UUID
+
+from sqlalchemy import select
+
+from polar.auth.models import AuthSubject, Organization, User, is_organization, is_user
+from polar.models import UserOrganization
+from polar.models.organization import Organization as OrganizationModel
```
frankie567: Actually, this is exactly the same as the `Organization` from `polar.auth.models`, it's just re-exported from `polar.auth.models` for convenience. So we could actually just do:

```py
from polar.auth.models import AuthSubject, User, is_organization, is_user
from polar.models import Organization, UserOrganization
```

server/polar/organization/endpoints.py
```diff
@@ -127,24 +132,11 @@ async def get(
     tags=[APITag.private],
 )
 async def get_account(
-    id: OrganizationID,
-    auth_subject: auth.OrganizationsRead,
+    authz: AuthorizeFinanceRead,
```
frankie567, line 135: Yeah, that feels great like this :)

handbook/engineering/backend-development/authorization.mdx
```diff
@@ -0,0 +1,180 @@
...
+5. Returns an `AuthorizedOrganization` (or `AuthorizedAccount`, etc.) containing both the resource and the auth subject
+
+If any step fails, the appropriate HTTP error is raised before the endpoint body executes.
+
+```python
+from polar.authz.dependencies import AuthorizeFinanceRead
+
+@router.get("/{id}/account")
+async def get_account(
+    authorized: AuthorizeFinanceRead,
```
frankie567: We should update to `authz` to enforce this new naming in the doc :)

frankie567 (approved):
A few nitpicks, but yeah, looks really good now :)

handbook/engineering/backend-development/authorization.mdx
joebon: AH, yes! Forgot to update this, thanks

server/polar/authz/repository.py
joebon: Great, thanks - fixing

PR #11146, opened 2026-04-22
refactor: migrate repositories to pure data access

joebon:
## 📋 Summary

Migrate 25 repositories from `get_readable_statement(auth_subject)` to `get_by_org_ids_statement(org_ids)`, removing authorization logic from the repository layer. Services now call `get_accessible_org_ids()` and pass explicit org IDs.

> Depends on #11126

## 🎯 What

- 25 repositories migrated to `get_by_org_ids_statement(org_ids)`
- Services call `get_accessible_org_ids(session, auth_subject)` then pass org IDs to repos
- `get_payload_organization` refactored to use `get_accessible_org_ids` internally
- 7 repos intentionally deferred (order, subscription, product prices, customer portal — handle Customer/Member subjects)

## 🤔 Why

Repositories currently mix data access with authorization — each duplicates a `UserOrganization` subquery. Moving authorization to the service layer means repos are pure data access, and adding new permission checks doesn't require touching 30+ files.

## 🔧 How

Identical transformation applied to every repo:

```python
# Before
statement = repository.get_readable_statement(auth_subject)

# After
statement = repository.get_by_org_ids_statement(org_ids)
```

Best reviewed commit by commit — each commit is one batch of related modules.

## 🧪 Testing

- [x] All existing tests pass
- [x] I have run linting and type checking

### Test Instructions

1. Run `uv run task test_fast` — all tests pass
2. This is a behavior-preserving refactor — existing tests verify correctness

## 📝 Additional Notes

- No database migrations, no API response changes
- Some repos call `get_accessible_org_ids` internally (customer, customer_seat, meter, license_key) — pragmatic trade-off, deferred cleanup

## ✅ Pre-submission Checklist

- [x] All tests pass locally
- [x] **AI/LLM Policy**: If I used AI assistance, I have tested and executed the code locally


server/polar/event/service.py
```diff
@@ -155,7 +156,7 @@ async def list(
         aggregate_fields: Sequence[str] = (),
         cursor_pagination: bool = False,
     ) -> tuple[Sequence[Event], int]:
-        organization_ids = await self._get_readable_organization_ids(
+        organization_ids, org_ids = await self._get_filtered_org_ids(
```
frankie567: Naming of those two is really error-prone, I would say. Maybe: `filtered_orgs, accessible_orgs = ...`

server/polar/organization/resolver.py
```diff
@@ -62,17 +60,26 @@ async def get_payload_organization(
...
+                        "organization_id",
+                    ),
+                    "msg": "Organization not found.",
+                    "type": "value_error",
+                    "input": model.organization_id,
+                }
+            ]
+        )
+
+    from polar.organization.repository import OrganizationRepository
```
frankie567: I think we could have it at top level

frankie567 (changes requested):
Great! I think the diff count talks by itself :) 

Maybe I overthink it but I'm wondering if we could blend some type safety into it.

Imagine we define a dedicated type for accessible org ids:

```py
AccessibleOrganizationId = NewType("AccessibleOrganizationId", UUID)
```

The `get_accessible_org_ids` would return this type:

```py
async def get_accessible_org_ids(
    session: AsyncReadSession,
    auth_subject: AuthSubject[User | Organization],
) -> set[AccessibleOrganizationId]:
    # ...
```

Then, in the repository methods, or any other places where we need it, we could use this type in argument:

```py
def get_by_org_ids_statement(self, org_ids: set[AccessibleOrganizationId]) -> Select[tuple[Benefit]]:
```

This way, we ensure at type checking level that we only pass true organization ID that were vetted by the authorization helpers, and not random list of UUID.

Maybe over engineered, but would like your opinion :)

joebon:
> Great! I think the diff count talks by itself :)
> 
> Maybe I overthink it but I'm wondering if we could blend some type safety into it.
> 
> Imagine we define a dedicated type for accessible org ids:
> 
> ```python
> AccessibleOrganizationId = NewType("AccessibleOrganizationId", UUID)
> ```
> 
> The `get_accessible_org_ids` would return this type:
> 
> ```python
> async def get_accessible_org_ids(
>     session: AsyncReadSession,
>     auth_subject: AuthSubject[User | Organization],
> ) -> set[AccessibleOrganizationId]:
>     # ...
> ```
> 
> Then, in the repository methods, or any other places where we need it, we could use this type in argument:
> 
> ```python
> def get_by_org_ids_statement(self, org_ids: set[AccessibleOrganizationId]) -> Select[tuple[Benefit]]:
> ```
> 
> This way, we ensure at type checking level that we only pass true organization ID that were vetted by the authorization helpers, and not random list of UUID.
> 
> Maybe over engineered, but would like your opinion :)

I love the way you're thinking! And this is a great opportunity to do things like this IMO - reduce the risk of accidental misuse and security incidents.

And I agree with your reasoning re: a dedicated type - it's still possible to misuse/abuse but then you'd need to deliberately instantiate this specific type with non-valid org uuids. Let me give it a spin!

server/tests/checkout/test_service.py
```diff
@@ -417,6 +417,7 @@ async def product_seat_based_with_min_max(
 
 
 @pytest.mark.asyncio
+@pytest.mark.auth
```
frankie567: We don't need to add auth marker on services tests, it's only to authenticate the test HTTP client. Here, we pass directly the subject.

server/tests/subscription/test_service.py
```diff
@@ -187,6 +187,7 @@ def frozen_time() -> Generator[datetime, None]:
 
 
 @pytest.mark.asyncio
+@pytest.mark.auth
```
frankie567: Same as above

frankie567 (approved):
As we would say in French: *Top moumoute ! 👍*

server/tests/checkout/test_service.py
joebon: Ah, great - thanks! Sloppy, will clean up

joebon: Hrm 🤔 , when I removed it a bunch of seemingly unrelated tests started to fail. And asked Claude to analyze it and got this:

> - There it is. The old `get_readable_statement` had an implicit fallthrough: if the subject was neither `is_user` nor `is_organization` (i.e. anonymous), it returned the base statement *with no org filter* — effectively returning all prices. So anonymous subjects could look up any price.                                                                                                                                                                                                                           
>                                                                                                                                                                                                                                                                >  Our new code calls `get_accessible_org_ids` which returns an empty set for anonymous subjects, so `get_statement_by_org_ids(set()) matches nothing, and the price lookup fails.                                                                                
>.                                                                                                                                                                                                                                                                
    So the question is: *should anonymous checkout creation be able to look up prices?* Let me check:                                                                                                                                                             

> - …                                                                                                                                                                                                
> 
> - The service method takes `AuthSubject[User | Organization]` — not anonymous. But let me check the endpoint:                                                                                                                                                   
>                                                                                                                                                                                                                                                             
>  Searched for 6 patterns, read 1 file (ctrl+o to expand)                                                                                                                                                                                                     
>                                                                                                                                                                                                                                                               
> - So the endpoint requires `User | Organization` — never anonymous. The old code's anonymous fallthrough in `get_readable_statement` was dead code that happened to work. Our new code is actually *more correct* — it properly rejects anonymous subjects.         
>                                                                                                                                                                                                                                                               
>  So @frankie567 is right that `@pytest.mark.auth` shouldn't be needed on service tests if the service never receives anonymous subjects in production. But the test fixture parametrizes with an anonymous subject by default (without `@pytest.mark.auth`), which hits the code path that now correctly rejects it.                                                                                                                                                                                                           
>                                                                                                                                                                                                                                                               
>  The real fix: these tests should pass the `auth_subject` explicitly as a fixture parameter rather than relying on the class-level marker. But that's a bigger refactor of the test fixtures. For now, keeping `@pytest.mark.auth` is the pragmatic answer — it ensures the test gets an authenticated subject, which is what the service always receives in production.                                                                                                                                                    

WDYT @frankie567 ?

frankie567: I'm dumb, it's perfectly right to have those markers. Ignore me 🙈

server/polar/authz/types.py
```diff
@@ -1,9 +1,15 @@
...
+from typing import NewType
+from uuid import UUID
 
 from polar.auth.models import AuthSubject, Organization, User
 from polar.models import Organization as OrganizationModel
 from polar.postgres import AsyncReadSession
 
+# A UUID that has been verified as accessible by the current auth subject.
+# Only get_accessible_org_ids should produce these values.
+AccessibleOrganizationID = NewType("AccessibleOrganizationID", UUID)
```
Yopi, line 11: 🖤

Yopi (approved):

PR #11147, opened 2026-04-22
feat(authz): add AccountPolicyGuard and PayoutAccountPolicyGuard

joebon:
## Summary

**Related Issue**: #11126

Add policy guard dependencies for account and payout-account endpoints, and migrate finance repositories to pure data access.

> Depends on #11146

## What

- `AccountPolicyGuard` and `PayoutAccountPolicyGuard` — resolve account/payout-account by ID, find owning organization, check finance policy
- Typed dependencies: `AuthorizeAccountRead`, `AuthorizeAccountWrite`, `AuthorizePayoutAccountRead`, `AuthorizePayoutAccountWrite`
- Account, payout_account, payout repos migrated to `get_statement_by_org_ids` with `AccessibleOrganizationID`
- `get_by_account` and `get_by_payout_account` on `OrganizationRepository` (with blocked-org filtering)
- `_check_policy` helper — deduplicates policy evaluation across all three guard types
- `use_write_session` parameter on all guards — write guards load entities on the write db session, so endpoints use them directly without re-fetching
- Finance-specific required scopes on `AuthorizeFinanceRead`/`AuthorizeFinanceWrite`

## Why

Finance endpoints (`/accounts/{id}`, `/payout-accounts/{id}`) access resources by their own IDs, not org IDs. They need a guard variant that resolves account → owning organization → policy check, similar to the existing `OrgPolicyGuard` but with an extra resolution step.

## How

The new guards follow the same pattern as `OrgPolicyGuard`: authenticate, resolve the resource, find the owning organization, verify membership, evaluate the policy. The `use_write_session` parameter controls whether the guard uses a read-only or write db session, so write endpoints can use the guard's entity directly without re-fetching.

## Checklist


server/polar/authz/dependencies.py
```diff
@@ -72,21 +83,20 @@ def OrgPolicyGuard(
         required_scopes=_scopes,
     )
 
+    _get_session = get_db_session if use_write_session else get_db_read_session
```
frankie567: Yeah, TBH, I'm really not sure we need that. As you mentioned in Slack, it carries over some cognitive burden when using OrgPolicyGuard that's error prone. So I would say let's just use `get_db_session`.

server/polar/account/endpoints.py
```diff
@@ -25,41 +21,26 @@
...
-    session: AsyncReadSession = Depends(get_db_read_session),
-) -> Account:
-    account = await account_service.get(session, auth_subject, id)
-    if account is None:
-        raise ResourceNotFound()
-
-    return account
+    authz: AuthorizeAccountRead,
+) -> AccountSchema:
+    return AccountSchema.model_validate(authz.account)
```
frankie567: You don't need to recast to a Pydantic model, you can directly return an `Account` and FastAPI automatically does the conversion to `AccountSchema` via `response_model` setting.

server/polar/account/endpoints.py
```diff
@@ -25,41 +21,26 @@
...
     session: AsyncSession = Depends(get_db_session),
-) -> Account:
-    account = await account_service.get(session, auth_subject, id)
-    if account is None:
-        raise ResourceNotFound()
-
-    return await account_service.update(session, account, account_update)
+) -> AccountSchema:
+    updated = await account_service.update(session, authz.account, account_update)
+    return AccountSchema.model_validate(updated)
```
frankie567: Same comment as above

server/polar/payout_account/endpoints.py
```diff
@@ -52,48 +54,31 @@ async def create(
...
-    auth_subject: WebUserRead,
-    session: AsyncReadSession = Depends(get_db_read_session),
-) -> PayoutAccount:
-    payout_account = await payout_account_service.get(session, auth_subject, id)
-    if payout_account is None:
-        raise ResourceNotFound()
-    return payout_account
+    authz: AuthorizePayoutAccountRead,
+) -> PayoutAccountSchema:
+    return PayoutAccountSchema.model_validate(authz.payout_account)
```
frankie567: Same thing with Pydantic casting

frankie567 (changes requested):

server/polar/account/endpoints.py
joebon: Thanks! Fixed

server/polar/authz/dependencies.py
joebon: Yeah, I agree - removed

frankie567 (approved):

PR #11149, opened 2026-04-22
events: Decouple parent and child ingestion

Yopi:
This allows us to decouple the ingestion of a child event and its parent. By storing the pending parent id on events, we can at a later date resolve this relationship. We write the events to postgres always, but we wait with writing to Tinybird until we have resolved the relationship.

An orphan child will not show up anywhere, although we could have a list of orphan events in the dashboard. This is intentional as we want to avoid writing and updating data in Tinybird. Especially since we will need to update both the root_id, as well as the ancestor_chain. If you are ingesting a grand-grand child and then their parent, you will have to ingest two more events before you complete the chain and can truly know the events full ancestry.

### DIFF
--- a/server/polar/event/repository.py
+++ b/server/polar/event/repository.py
@@ -11,6 +11,7 @@ from sqlalchemy import (
     ColumnExpressionArgument,
     Numeric,
     Select,
+    String,
     UnaryExpression,
     and_,
     asc,
@@ -23,6 +24,7 @@ from sqlalchemy import (
     or_,
     select,
     text,
+    update,
 )
 from sqlalchemy.dialects.postgresql import aggregate_order_by, insert
 from sqlalchemy.orm import aliased, joinedload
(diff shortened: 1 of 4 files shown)


server/polar/event/service.py
```diff
@@ -923,126 +873,65 @@ async def ingest(
...
             raise PolarRequestValidationError(errors)
 
         repository = EventRepository.from_session(session)
-        with logfire.span("insert_batch", event_count=len(events)):
-            event_ids, duplicates_count = await repository.insert_batch(events)
+        event_ids, duplicates_count = await repository.insert_batch(events)
+
+        has_hierarchy = any(
+            event_dict.get("pending_parent_external_id") for event_dict in events
+        )
```
frankie567: Would it be a valuable optimization to only consider those events when executing the recursive query? Something like:

```py
event_ids_with_hierarchy = [event_id for event_id, event_dict in zip(event_ids, events) if event_dict.get("pending_parent_external_id")] # Assuming event_ids are in the same order as events, didn't check :p 

if event_ids_with_hierarchy:
    resolvable = await repository.find_resolvable_parents(event_ids_with_hierarchy)
    # ...
```

server/polar/event/repository.py
```diff
@@ -111,6 +89,141 @@ async def insert_batch(
 
         return inserted_ids, duplicates_count
 
+    async def find_resolvable_parents(
+        self, inserted_ids: Sequence[UUID]
+    ) -> Sequence[tuple[UUID, UUID, UUID]]:
```
frankie567, line 94: This is still very hard for me to understand that, but I'll trust you :D

frankie567 (approved):

server/polar/event/service.py
Yopi: I think if we narrow it then if you have an unresolved event from previously and you ingest the parent here (i.e. no parent_id) then it would not get resolved since it's not part of the query.

frankie567: I see, but isn't it the same issue with the `has_hierarchy` check?

Yopi: Ah, you're absolutely right. If you don't pass in any parent_id's then existing ones doesn't resolve 😄. Nice catch!

server/polar/event/repository.py
Yopi: I tried to simplify the query, and added the equivalent SQL to make it easier to follow (maybe?)

server/polar/event/service.py
frankie567: > Ah, you're absolutely right.

Jesper Sonnet 4.6 😂

server/polar/event/repository.py
frankie567: Definitely helps, thanks 🙏

server/polar/event/service.py
Yopi: Hahah, only sonnet 🥲

frankie567: Jesper Mythos 😘

PR #11157, opened 2026-04-22
backoffice: per-capability override UI for organizations

psincraian:
## Summary

- Adds a **Capabilities** card to the organization Settings tab in the backoffice so staff can flip an individual capability (`checkout_payments`, `subscription_renewals`, `payouts`, `refunds`, `api_access`, `dashboard_access`) without inventing a new status.
- Each override requires a reason (min 10 chars) that's appended to `internal_notes` with a timestamp and the admin email. Overrides reset to the status default on the next status transition (existing `Organization.set_status` behaviour — no change there).
- Generalises the existing `refunds_blocked` override lever to all six capability families. `refunds_blocked` is kept in sync as a write-through shim.

## Why

Operators today have a single per-field lever (`refunds_blocked`) and otherwise have to move an org's `status` to gate anything — which resets all capabilities at once. This PR unlocks finer control (e.g. pause payouts on an otherwise-ACTIVE org, or re-enable API access on a BLOCKED org pending review) without new statuses and without the collateral damage of a status transition.

## Screenshots

Capabilities card on an ACTIVE org (all six defaults enabled, green dots):

Capabilities card — ACTIVE defaults

Override modal after clicking *Disable* on a row. Current / target / status-default summary plus required reason textarea (min 10 chars):

Override modal

After submitting a reason, the row flips — grey dot, `overridden` pill, button flipped to *Enable*. Other rows untouched:

After the override is applied

## Changes

- `polar/models/organization.py` — `CAPABILITY_METADATA` (labels + descriptions) and `CAPABILITY_NAMES` (frozenset) exported as the single source of truth for valid capability names and human-readable copy.
- `polar/organization/service.py` — new `OrganizationService.set_capability(...)` that validates the name, replaces the `capabilities` JSONB dict (so SQLAlchemy flags it dirty), appends a timestamped `internal_notes` entry via the existing `_append_internal_note` helper, and mirrors `refunds` → `refunds_blocked`. Uses `Organization.get_effective_capabilities()` (added in #11104) rather than inlining the fallback.
- `polar/backoffice/organizations_v2/endpoints.py` — new `GET|POST /organizations/{id}/capabilities/{capability}?value=…` route gated on `get_admin`. `GET` renders the modal; `POST` validates the form and applies. Uses `URL.include_query_params` rather than string concatenation.
- `polar/backoffice/organizations_v2/views/modals/set_capability_modal.py` (new) — modelled on `DeletePayoutAccountModal`. Shows current / target / status-default summary, required reason textarea with inline validation errors, neutral `Enable`/`Disable` submit button (the green/grey status dot on the row carries the state — no color variants on the button).
- `polar/backoffice/organizations_v2/views/sections/settings_section.py` — new Capabilities card positioned between Social Media and Danger Zone. Each of the six rows shows a status dot, label, description, `overridden` pill when the current value diverges from the status default, and the toggle button.

## Test plan

- [x] `uv run task test` — `tests/organization/test_service.py::TestSetCapability` adds 6 service-level cases (flip, internal note, no-op, `refunds_blocked` mirror, unknown-capability rejection, status-transition reset). 159 organization tests pass post-rebase onto #11104.
- [x] `ruff check` and `ruff format --check` clean.
- [x] **End-to-end smoke test in the real backoffice** (playwright-driven): opened an ACTIVE org → Settings → clicked *Disable* on Checkout payments → modal rendered with current / target / status-default summary → submitted a reason → verified DB: `capabilities.checkout_payments == false` and `internal_notes` contains the timestamped `"Capability 'checkout_payments' disabled by bo-admin@polar.test"` entry plus the reason. Other five capabilities unaffected.
- [ ] Manual check on a REVIEW org to confirm the `overridden` pill appears when a capability diverges from the REVIEW default, and disappears again after a status transition.

### DIFF
--- a/server/polar/organization/service.py
+++ b/server/polar/organization/service.py
@@ -30,7 +30,12 @@ from polar.models import (
     User,
     UserOrganization,
 )
-from polar.models.organization import OrganizationDetails, OrganizationStatus
+from polar.models.organization import (
+    CapabilityName,
+    OrganizationCapabilities,
+    OrganizationDetails,
+    OrganizationStatus,
+)
 from polar.models.organization_review import OrganizationReview
 from polar.models.transaction import TransactionType
 from polar.models.user import IdentityVerificationStatus
@@ -1192,6 +1197,44 @@ class OrganizationService:
                 "ai_onboarding_completed_at": datetime.now(UTC),
             },
         )
+        return organization
+
+    async def set_capability(
+        self,
+        session: AsyncSession,
+        organization: Organization,
+        capability: CapabilityName,
+        value: bool,
+        *,
+        reason: str,
+        admin_email: str | None = None,
+    ) -> Organization:
+        """Override a single capability on an organization.
+
+        The override persists until the next status transition — `set_status`
+        resets `capabilities` from `STATUS_CAPABILITIES`.
+        """
+        current: OrganizationCapabilities = dict(  # type: ignore[assignment]
+            organization.get_effective_capabilities()
+        )
+        if current[capability] == value:
+            return organization
+
+        current[capability] = value
+        organization.capabilities = current
+
+        action = "enabled" if value else "disabled"
+        by = f" by {admin_email}" if admin_email else ""
+        _append_internal_note(
+            organization,
+            f"Capability '{capability}' {action}{by}",
+            reason=reason,
+        )
+
+        if capability == "refunds":
+            organization.refunds_blocked = not value
+
+        session.add(organization)
         return organization
 
 
--- a/server/polar/backoffice/organizations_v2/endpoints.py
+++ b/server/polar/backoffice/organizations_v2/endpoints.py
@@ -11,7 +11,7 @@ This module provides a modern, three-column layout with:
 import uuid
 from collections.abc import Sequence
 from datetime import UTC, datetime, timedelta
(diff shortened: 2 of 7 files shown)


server/polar/organization/service.py
```diff
@@ -1194,5 +1199,55 @@ async def mark_ai_onboarding_complete(
         )
         return organization
 
+    async def set_capability(
+        self,
+        session: AsyncSession,
+        organization: Organization,
+        capability: str,
```
frankie567: Why not use `CapabilityName` literal? This would avoid the validation check.

frankie567 (approved):

server/polar/organization/service.py
psincraian: that's a good point

psincraian: Good call — switched the service signature to `CapabilityName` and dropped the runtime check + `PolarRequestValidationError` raise. The endpoint still validates the URL path param against `CAPABILITY_NAMES` and casts to `CapabilityName` before calling the service, so the service now trusts its input. Also tightened `CAPABILITY_METADATA` to `dict[CapabilityName, ...]` — direct indexing in both the modal and settings section, no more `# type: ignore[literal-required]`.

PR #11158, opened 2026-04-22
member: prevent duplicate owner when customer email drifts

psincraian:
## Summary

Fixes a bug where correcting `customer.email` for an individual customer could leave the owner `Member`'s email stale, and the next portal sign-in would auto-create a second owner. Downstream, any call to `get_owner_by_customer_id` then raised `MultipleResultsFound`, surfacing as HTTP 500s to both the customer portal (`/{org}/portal/request` empty) and the backoffice "Generate Portal Link" action ("A network error occurred, Please try again later").

A small number of customers are currently affected in prod; the data cleanup is being handled out of band.

## Why the old sync didn't protect us

The two sync sites — `customer_service.update` and `customer_email_update.verify` — both looked up the owner scoped by the *pre-update* email:

- `customer/service.py` guarded `owner.email.lower() == old_email.lower()` and silently skipped sync on any drift.
- `customer_email_update/service.py` looked up via `get_by_customer_and_email(customer, old_email)` and returned `None` on any drift.

So if `owner.email` ever diverged from `customer.email` (typo at customer creation later corrected, whitespace, unrelated prior sync miss), the update was silently lost, and the next sign-in's email-based member lookup missed the stale owner and called `create_owner_member`, which deduped by `(customer_id, email)` — inserting a duplicate with the new email.

## What changed

- New `MemberService.sync_owner_email(session, customer)`: for individual customers, unconditionally mirrors `customer.email` to the single owner member (case-insensitive diff check, so no flap). Team customers are skipped — they legitimately have members with distinct emails.
- `customer_service.update` and `customer_email_update.verify` now call the helper instead of their own duplicated blocks.
- `MemberService.create_owner_member` now dedupes via `get_owner_by_customer_id` (by `customer_id`) instead of `get_by_customer_and_email` (by `(customer_id, email)`). A customer has at most one owner — drift should never produce a second row, regardless of whether sync was ever missed.
- `get_owner_by_customer_id` kept strict: still raises `MultipleResultsFound` on duplicates so the invariant violation stays loud.

## Test plan

- [x] `uv run pytest tests/member tests/customer/test_service.py tests/customer_email_update` — 177 passed
- [x] `uv run task lint_types` clean
- [ ] After merge + deploy: re-check the affected customers' portal sign-in and backoffice "Generate Portal Link" flows.

### DIFF
--- a/server/polar/customer/service.py
+++ b/server/polar/customer/service.py
@@ -343,7 +343,6 @@ class CustomerService:
     ) -> Customer:
         repository = CustomerRepository.from_session(session)
 
-        old_email = customer.email
         email_changed = False
 
         errors: list[ValidationError] = []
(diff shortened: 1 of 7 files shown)


frankie567 (approved):
Thank you for tackling this!

One thought: would it make sense to have a partial index on Member that enforces there is only a single "owner" for a given customer ID?

psincraian:
@frankie567  I think it's a nice addition 👍 I can do that in the next pr

PR #11164, opened 2026-04-22
Create non-org dependent deep links for the dashboard

sebastianekstrom:
It's often useful to link to specific parts of the dashboard, however to do that we always need to include the org slug otherwise the link will 404. E.g https://polar.sh/dashboard/acme-org/products.

This PR adds the ability to automatically deep link into a specific part of the dashboard without having to specify the org slug, e.g: https://polar.sh/to/dashboard/products which then redirects to https://polar.sh/dashboard/acme-org/products.

Examples:
* `/to/dashboard/products` → `/dashboard/acme-org/products`
* `/to/dashboard/products?foo=bar` → `/dashboard/acme-org/products?foo=bar`
* `/to/dashboard/products/checkout-links/123?foo=bar` → `/dashboard/acme-org/products/checkout-links/123?foo=bar`

### DIFF
--- a/clients/apps/web/src/app/(main)/to/[...path]/page.tsx
+++ b/clients/apps/web/src/app/(main)/to/[...path]/page.tsx
@@ -1,0 +1,38 @@
+import { getServerSideAPI } from '@/utils/client/serverside'
+import { getLastVisitedOrg } from '@/utils/cookies'
+import { getUserOrganizations } from '@/utils/user'
+import { cookies } from 'next/headers'
+import { redirect } from 'next/navigation'
+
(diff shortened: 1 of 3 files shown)


psincraian (approved):
Seems that the onboarding is taken into account! So, good to go 👍

pieterbeulque (commented):
Can I not convince you about doing a `/to/` top-level pathname instead of this magic? It's weird that when a organization slug conflicts with any top-level pathname, the redirect won't work only for them, and I think that can be odd.

sebastianekstrom:
@pieterbeulque fair point! I've updated the PR to use the new `/to/` prefix

PR #11168, opened 2026-04-22
events: Simplify parent resolution and move it to worker

Yopi:
1. Update pending_parent_id to parent_id if it exists
2. Set root_id where we have the full chain

Since we allow parent ingestion to happen after the child we can now move this to the worker to avoid doing heavy work in the ingestion flow.

This re-applies the tests from #11160 but simplifies the SQL and splits it into querying by id and external_id instead of an `OR` which forces postgres into a bitmap scan.

### DIFF
--- a/server/polar/event/repository.py
+++ b/server/polar/event/repository.py
@@ -11,6 +11,7 @@ from sqlalchemy import (
     ColumnExpressionArgument,
     Numeric,
     Select,
+    String,
     UnaryExpression,
     and_,
     asc,
@@ -23,6 +24,7 @@ from sqlalchemy import (
     or_,
     select,
     text,
+    update,
 )
 from sqlalchemy.dialects.postgresql import aggregate_order_by, insert
 from sqlalchemy.orm import aliased, joinedload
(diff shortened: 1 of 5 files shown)


server/polar/event/repository.py
```diff
@@ -111,6 +95,160 @@ async def insert_batch(
...
+            .values(parent_id=parent_1d.c.id, pending_parent_external_id=None)
+            .where(
+                Event.organization_id == parent_1d.c.organization_id,
+                Event.pending_parent_external_id.is_not(None),
+                Event.pending_parent_external_id == cast(parent_1d.c.id, String),
+                parent_1d.c.id.in_(inserted_ids),
+            )
+        )
+
+        walk_up_anchor = select(
```
Yopi: We're back here again, I'll think over the night if we can avoid doing this

frankie567 (approved):

Yopi:
@frankie567 I've gone a few rounds now 😮‍💨 . Do you mind giving me another pair of eyes?

frankie567 (approved):
Fingers crossed 😄

PR #11170, opened 2026-04-22
Surface email validation errors in checkout

sebastianekstrom:
Currently if you enter an invalid email in the checkout the backend will throw an error, but we're not exposing that in the UI. This PR fixes that and adds tests to prevent it from happening again, as well as a test helper.

### DIFF
--- a/clients/packages/checkout/src/providers/CheckoutFormProvider.tsx
+++ b/clients/packages/checkout/src/providers/CheckoutFormProvider.tsx
@@ -83,6 +83,7 @@ export const CheckoutFormProvider = ({
         if (error) {
           switch (error.error) {
             case 'PolarRequestValidationError':
+            case 'RequestValidationError':
               setValidationErrors(error.detail, setError)
               break
             case 'AlreadyActiveSubscriptionError':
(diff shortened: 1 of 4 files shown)


clients/packages/checkout/src/providers/CheckoutFormProvider.tsx
```diff
@@ -114,6 +115,7 @@ export const CheckoutFormProvider = ({
       if (error) {
         switch (error.error) {
           case 'PolarRequestValidationError':
```
pieterbeulque, line 117: I wonder if we can remove the `PolarRequestValidationError` case if that was a quirk from using our external SDK. Does Speakeasy namespace these somehow, @frankie567 ?

pieterbeulque (approved):

clients/packages/checkout/src/providers/CheckoutFormProvider.tsx
frankie567: No, I think it's still valid, we do raise this kind of errors when triggering validation errors.

PR #11179, opened 2026-04-23
refactor(member): atomic ownership transfer via single CASE-based UPDATE

psincraian:
## Summary

Replaces the two-step demote-then-promote ownership transfer with a single CASE-based `UPDATE` via a new `MemberRepository.transfer_ownership` helper. The repository method also refreshes the in-memory role on both instances so the webhook payload stays accurate after the raw UPDATE.

This is a prerequisite for a follow-up PR that adds a partial unique index `members(customer_id) WHERE deleted_at IS NULL AND role = 'owner'`: once that index exists, the previous two-statement transfer would briefly expose two owners between UPDATEs and trip the constraint. Landing this refactor first means the follow-up migration can be deployed safely — old code running against the new index would otherwise break during the deploy window.

Also removes two misnamed tests in `tests/customer_portal/endpoints/test_member.py` whose docstrings claimed they tested "cannot demote/remove only owner" but whose bodies instead created two owners and asserted that demoting/removing one succeeded — a scenario the service-level invariant already forbids. Proper "only owner" coverage is already provided by `test_cannot_modify_self` and `test_cannot_remove_self` in the same file.

## Test plan

- [x] `uv run task lint_types` clean
- [x] `uv run pytest tests/member` — 96/96 passing, including two new tests covering the atomic ownership transfer for both the customer-portal and admin-API paths
- [x] `uv run pytest tests/customer_portal/endpoints/test_member.py` — passes after removing the two broken tests
- [ ] Follow-up PR adds the migration + model `Index` once this lands

### DIFF
--- a/server/polar/member/repository.py
+++ b/server/polar/member/repository.py
@@ -1,7 +1,7 @@
 from collections.abc import Sequence
 from uuid import UUID
 
-from sqlalchemy import Select, func, select
+from sqlalchemy import Select, case, func, select, update
 from sqlalchemy.orm import joinedload
 
 from polar.auth.models import AuthSubject, Organization, User, is_organization, is_user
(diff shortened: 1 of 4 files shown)


frankie567 (approved):
Some tests seem to fail, but otherwise looks great 👍

PR #11188, opened 2026-04-23
docs: add design document for customer-portal receipts

psincraian:
## Summary

- Adds a design document at `handbook/engineering/design-documents/receipts.mdx` for the upcoming customer-portal Receipts feature.
- Captures the problem (the existing invoice flow is gated on billing details, leaving paid customers with no downloadable proof of payment), the proposed solution (one mutable PDF per Order with a succeeded Payment), and the rollout plan (per-org boolean flag plus a backfill script).
- Documents the key design trade-offs: per-Order receipt with stable number, lazy render in a Dramatiq actor with a Redis-locked single-renderer, immutable per-render S3 keys in a new dedicated bucket, hardcoded `RCPT-` number prefix.

No code changes — documentation only.

## Test plan

- [ ] Open the rendered handbook locally and verify the new doc renders correctly.
- [ ] Review the design with the team.

### DIFF
--- a/handbook/engineering/design-documents/receipts.mdx
+++ b/handbook/engineering/design-documents/receipts.mdx
@@ -1,0 +1,145 @@
+<Info>
+  **Status**: Draft
+
+  **Created**: 2026-04-23
+
+  **Last Updated**: 2026-04-23
+</Info>
+
+## Briefing
+
+The customer portal already exposes "Download Invoice" on every paid Order, but invoice generation requires `billing_name` and `billing_address`. Customers who never fill those in (indie devs, wallet payers, anyone topping up balance credit) get nothing. Support sees regular requests for proof of payment, and some customers actually require it for bookkeeping, expense reimbursement, or compliance. They don't need a tax invoice, just confirmation that the charge happened.
+
+A **receipt** is the missing artifact: a PDF proving the payment event without the billing-details gate. It mirrors Stripe's `charge.receipt_url`. The PDF is mutable, so refunds are reflected in the same document over time.
+
+## Goals
+
+- Every Order with a succeeded Payment has a downloadable receipt PDF in the customer portal, regardless of whether billing details are set.
+- The receipt is mutable. Issuing a refund updates the PDF; the customer's URL keeps working and serves the latest content.
+- A stable `receipt_number` is shown on the document and stored on the Order, so it can be referenced from support tools.
+- Polar staff can also fetch a customer's receipt from backoffice. Merchants can fetch it through the dashboard API.
+
+### Non-goals (v0)
+
+- No receipts for Orders without a succeeded Payment (100% discount, balance-credit only, $0 trial). The existing invoice flow still covers those cases when billing details exist.
+- No emailed receipts. Customers download when they want to.
+- No public webhook events for receipts.
+
+## Solution
+
+### Receipt per Order
+
+Each Order gets at most one receipt. Polar Orders have at most one succeeded `Payment`, so per-Order and per-Payment collapse into the same shape. The receipt is allocated when the Order transitions to paid. The PDF itself is rendered lazily on first download and re-rendered whenever the payment state changes (typically a refund).
+
+### Allocation
+
+Allocation runs at the end of `_on_order_paid`, the central paid-transition hook. It checks that the Order has at least one succeeded Payment, then allocates a number from a per-customer counter:
+
+```
+RCPT-{customer.short_id_str}-{NNNN}
+```
+
+The counter, `Customer.receipt_next_number`, mirrors the existing `Customer.invoice_next_number` exactly. We deliberately do **not** reuse `Organization.customer_invoice_prefix`. Production data shows invoice prefixes range from short tokens like `DUY` to overlong values (one is 600 characters, the long tail is essentially company-name dumps). More importantly, since invoices and receipts use independent counters, sharing the prefix would produce identical-looking identifiers for different documents (e.g., invoice `ACME-AB1-0001` and receipt `ACME-AB1-0001`), which is a real support headache. A hardcoded `RCPT` prefix sidesteps both problems. The org identity is shown in the PDF body, not in the identifier. If we later want per-org customization, adding a `customer_receipt_prefix` column is a non-breaking change.
+
+Allocation is wrapped in `session.begin_nested()` with a `FOR UPDATE` plus null-check on the Order. Retried hooks return without burning a number, and any exception is caught at the call site so a bad receipt allocation never fails the paid transition itself.
+
+#### Why receipts allocate later than invoices
+
+Invoices are numbered at Order creation time. `OrganizationService.get_next_invoice_number` is called from four places when constructing a new Order, and the `invoice_number` column is non-null and set at insert.
+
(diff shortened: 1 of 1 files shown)


handbook/engineering/design-documents/receipts.mdx
```diff
@@ -0,0 +1,145 @@
...
+  **Status**: Draft
+
+  **Created**: 2026-04-23
+
+  **Last Updated**: 2026-04-23
+</Info>
+
+## Briefing
+
+The customer portal already exposes "Download Invoice" on every paid Order, but invoice generation requires `billing_name` and `billing_address`. Customers who never fill those in (indie devs, wallet payers, anyone topping up balance credit) get nothing. Support sees regular requests for proof of payment, and some customers actually require it for bookkeeping, expense reimbursement, or compliance. They don't need a tax invoice, just confirmation that the charge happened.
```
frankie567: > Customers who never fill those in (indie devs, wallet payers, anyone topping up balance credit) get nothing.

That's not true :) Invoices are generated automatically and attached to the confirmation email. It shows the billing info we have (so the customer name and country at the bare minimum).

I think the heart of the problem is people who want a proof of payment.

frankie567 (approved):

handbook/engineering/design-documents/receipts.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+A **receipt** is the missing artifact: a PDF proving the payment event without the billing-details gate. It mirrors Stripe's `charge.receipt_url`. The PDF is mutable, so refunds are reflected in the same document over time.
```
pieterbeulque, line 12: This isn't how it should work.

Receipts should match bank transactions (for the end customer).

So if you as the buyer have on April 23 a charge of $29.94  and on May 7 a refund incoming of $29.94, you will need two receipts.

handbook/engineering/design-documents/receipts.mdx
```diff
@@ -0,0 +1,145 @@
...
+{org_id}/{order_id}/{utc_iso8601}.pdf
+```
+
+The path lives in a new dedicated bucket, `polar-customer-receipts` (env-prefixed for sandbox and test). It mirrors Polar's one-bucket-per-purpose convention and follows the existing org-scoped, UUID-named pattern used elsewhere for mutable artifacts. A new `aws_s3_bucket` resource plus a matching IAM policy is roughly 10 lines of Terraform.
+
+Because every render writes to a fresh key, there are no overwrites. That means no S3 versioning is needed, no CDN cache can serve stale content, and old keys remain in S3 as a free audit trail. Objects are uploaded with `Cache-Control: public, max-age=31536000, immutable`. The `Order.receipt_path` column always points at the current key.
+
+### Refund regeneration
+
+When a refund is finalized, the refund hook calls into the receipt service to enqueue a render. The old `receipt_path` keeps serving until the new file is uploaded and the column is swapped. There is no null gap: customers downloading mid-window get the previous PDF, and the next download gets the new one with the Refunds section.
```
pieterbeulque, line 67: Same here

handbook/engineering/design-documents/receipts.mdx
```diff
@@ -0,0 +1,145 @@
...
+| 1 | Schema migration: 4 columns and the partial unique index `CONCURRENTLY` |
+| 2 | New `polar/receipt/` module, Terraform for the new bucket, allocation hook, render actor with Redis lock, refund hook, metrics. `receipts_enabled` defaults to `false` everywhere |
+| 3 | Customer-portal, backoffice, and merchant-dashboard `GET` endpoints, frontend component, enable for one internal org |
+| 4 | Per-org enable via the script and flag flip |
+
+## Open questions
+
+None outstanding for v0. The following are explicitly deferred:
+
+- Receipts for Orders without a succeeded Payment. Could be revisited if support volume justifies it.
```
pieterbeulque, line 142: I believe this is not a thing? What would the receipt be for?

handbook/engineering/design-documents/receipts.mdx
```diff
@@ -0,0 +1,145 @@
...
+- Email delivery.
```
pieterbeulque, line 143: I think it was Isac who raised a good point here: if we generate receipts but we don't give merchants the option to send them automatically, customers will keep asking for it over support instead of actually fixing the problem we're trying to self-serve. So it's worth considering if we should email them, and if so, when and how. In other words, can they be bundled in with an existing email? Theoretically, that should be possible for one-time purchases and subscription starts (as the payment proceeds), for subscription renewals, it's harder.

pieterbeulque (changes requested):

handbook/engineering/design-documents/receipts.mdx
psincraian: That's not right :) 

Stripe [charge.receipt.url](https://docs.stripe.com/api/charges/object#charge_object-receipt_url) is a mutable document that's keep up to date with the latest payment information.

With François, we discovered that when a refund is made the url gets up to date. For the refund one, the original payment is shown and the refunded amount.

[Receipt-2224-6958.pdf](https://github.com/user-attachments/files/27014344/Receipt-2224-6958.pdf)
[Refund-3962-0567.pdf](https://github.com/user-attachments/files/27014349/Refund-3962-0567.pdf)

handbook/engineering/design-documents/receipts.mdx
psincraian: At the beginning, I was thinking to add only a to the order view in the customer portal. So the user can click it if they need to download it.

handbook/engineering/design-documents/receipts.mdx
psincraian: I don't think this would be needed, the receipt will mainly show: "Discount applied. Free order" or similar. I don't think this is needed that's why I deferred it

handbook/engineering/design-documents/receipts.mdx
psincraian: Answered above

handbook/engineering/design-documents/receipts.mdx
psincraian: That's my fault and wrong assumption then. Will fix it

handbook/engineering/design-documents/receipts.mdx
pieterbeulque: But it's still two different documents, right? Surely the receipt.url reference is updated, but I'm pretty sure both  still exist in Stripe's records somewhere somehow.

psincraian: Yes, in the S3 we are gonna store all documents just in case.

The idea is to store them in {orgId}/{orderIt}/{timestamp}.pdf

handbook/engineering/design-documents/receipts.mdx
pieterbeulque: Ah yes, agree on deferring, my bad!

handbook/engineering/design-documents/receipts.mdx
pieterbeulque: All good!

handbook/engineering/design-documents/receipts.mdx
pieterbeulque: Okay! My main concern is that if the invoice is already refunded, but the customer didn't download their initial receipt, if they're then satisfied with only the single receipt mentioning both the payment and the refund.

But if Stripe gets away with it, I guess we can too.

psincraian (commented):
fix: comment

pieterbeulque (approved):

PR #11193, opened 2026-04-23
Display "Back to <org>" button on mobile

### LINKED ISSUE #8711: Customer Portal: Show back link on mobile too
The customer reports that the 'Return to App' button is hidden on mobile devices due to the parent div being set to hidden, and no alternative visible button is available. The button works fine on desktop.

sebastianekstrom:
Display the `Back to <org>` button in the customer portal on mobile, similar to what we do on desktop.

Fixes: https://github.com/polarsource/polar/issues/8711

| Before | After  |
|--------|--------|
| Screenshot 2026-04-23 at 17 12 50 | Screenshot 2026-04-24 at 12 44 46 |

### DIFF
--- a/clients/apps/web/src/app/(main)/[organization]/portal/CustomerPortalLayoutWrapper.tsx
+++ b/clients/apps/web/src/app/(main)/[organization]/portal/CustomerPortalLayoutWrapper.tsx
@@ -1,7 +1,12 @@
 'use client'
 
 import { CustomerPortalProvider } from '@/components/CustomerPortal/CustomerPortalProvider'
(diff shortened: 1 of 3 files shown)


pieterbeulque (approved):

pieterbeulque:
Random thought but I was wondering if it'd make sense to move it closer to the organization avatar (and should we show the name there too maybe?) versus in the navigation sidebar. But that can be a separate change.

sebastianekstrom:
@pieterbeulque I've got you brother ❤️

PR #11196, opened 2026-04-23
server/authz: fix guard logic for payout account

### LINKED ISSUE #11195: Bug: `PayoutAccountPolicyGuard` crashes with `MultipleResultsFound` when payout account is shared across organizations
## Bug

`GET /v1/payout-accounts/{id}` raises `MultipleResultsFound` (HTTP 500) when the requested payout account is linked to more than one organization.

## Root Cause

`PayoutAccountPolicyGuard` in `polar/authz/dependencies.py:255` calls `org_repo.get_by_payout_account(payout_account.id)`, which uses `scalar_one_or_none()`. However, `Organization.payout_account_id` has no `UNIQUE` constraint, so multiple orgs can legitimately share the same payout account — and `get_all_by_payout_account` already exists confirming this is an expected scenario.

**Call chain:**

```
GET /v1/payout-accounts/{id}
  → PayoutAccountPolicyGuard (polar/authz/dependencies.py:255)
    → org_repo.get_by_payout_account(payout_account.id)
      → get_one_or_none() → scalar_one_or_none() 💥
```

**Failing query:**

```sql
SELECT * FROM organizations
WHERE deleted_at IS NULL
  AND payout_account_id = $1
  AND status != 'BLOCKED'
```

Returns multiple rows when orgs share a payout account.

## Fix

Replace `get_by_payout_account` with `get_all_by_payout_account` in the guard, then pick the org whose ID is in the user's accessible org IDs:

```python
orgs = await org_repo.get_all_by_payout_account(payout_account.id)
organization = next((o for o in orgs if o.id in org_ids), None)
```

Introduced in #11147.

<sub>Sent by @frankie567 from [Payout account linking investigation](https://app.stilla.ai/m/memo_01kpxqxtzmesm87c8rcx323pcc).</sub>

frankie567:
Fix #11195

- server/authz: fix guard logic for payout account

### DIFF
--- a/server/polar/authz/dependencies.py
+++ b/server/polar/authz/dependencies.py
@@ -179,7 +179,6 @@ class AuthorizedPayoutAccount:
     """Result of a PayoutAccountPolicyGuard dependency."""
 
     payout_account: "PayoutAccountModel"
-    organization: OrganizationModel
     auth_subject: AuthSubject[User | Organization]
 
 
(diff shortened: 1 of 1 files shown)


frankie567:
@joebon Merging this as a quick fix, but we'll need to circle back on this

PR #11199, opened 2026-04-24
server/authz: proper policy guard for payout accounts

### LINKED ISSUE #11195: Bug: `PayoutAccountPolicyGuard` crashes with `MultipleResultsFound` when payout account is shared across organizations

joebon:
## Summary

- Replace the broken `PayoutAccountPolicyGuard` that used org-based finance policies (assuming 1:1 org-payout account relationship) with dedicated payout account policies checking `admin_id` ownership
- Add `organization.can_manage_payout_account` policy requiring org admin role to set the payout account on an organization — previously any org member could do this
- Move payout account authorization from the service layer (`get_by_id_and_admin`) to the endpoint layer, keeping authz in guards/policies where it belongs
- Remove unused `payout_account_service.get_by_id_and_admin`

Fixes #11195

## Test plan

- [x] Policy unit tests: `can_read`/`can_write` for admin (allowed) and non-admin (denied) — `tests/authz/test_policies.py`
- [x] Payout account endpoint tests: admin access, multi-org admin access, non-admin member denied, cross-org denied, happy-path for delete/onboarding-link/dashboard-link — `tests/payout_account/test_endpoints.py`
- [x] Set payout account endpoint tests: anonymous (401), non-member (404), non-admin member (403), admin with own account (200), admin with other user's account (422), unknown account (422) — `tests/organization/test_endpoints.py`
- [x] mypy passes, linter passes, all 65+ tests green

### DIFF
--- a/server/polar/authz/policies/organization.py
+++ b/server/polar/authz/policies/organization.py
@@ -11,28 +11,41 @@ from polar.models import Organization as OrganizationModel
 from polar.postgres import AsyncReadSession
 
 
+async def _require_account_admin(
+    session: AsyncReadSession,
+    auth_subject: AuthSubject[User | Organization],
+    organization: OrganizationModel,
+    denied_msg: str,
+) -> PolicyResult:
+    """Check that the subject is the account admin of the organization.
+
+    Organizations always have a billing account. Only the account admin
+    (the user whose ``admin_id`` matches the account) is allowed.
+    Organization-token subjects are always permitted.
+    """
+    if is_organization(auth_subject):
+        return True
+    if is_user(auth_subject):
+        if await account_service.is_user_admin(
+            session, organization.account_id, auth_subject.subject
+        ):
+            return True
+        return denied_msg
+    return "Not permitted"
+
+
 async def can_delete(
     session: AsyncReadSession,
     auth_subject: AuthSubject[User | Organization],
     organization: OrganizationModel,
 ) -> PolicyResult:
-    """Can the subject delete this organization?
-
-    - Organization tokens: always allowed (the org is the subject)
-    - Users: if the org has an account, only the account admin can delete.
-      If no account, any member can delete.
-    """
-    if is_organization(auth_subject):
-        return True
-    if is_user(auth_subject):
-        if organization.account_id is None:
-            return True
-        if await account_service.is_user_admin(
-            session, organization.account_id, auth_subject.subject
-        ):
-            return True
-        return "Only the account admin can delete an organization with an account"
-    return "Not permitted"
+    """Can the subject delete this organization?"""
+    return await _require_account_admin(
+        session,
+        auth_subject,
+        organization,
+        "Only the account admin can delete the organization",
+    )
 
 
 async def can_manage_payout_account(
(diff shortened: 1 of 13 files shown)


server/polar/authz/policies/organization.py
```diff
@@ -33,3 +33,27 @@ async def can_delete(
...
+    """Can the subject set or change the payout account for this organization?
+
+    - Organization tokens: always allowed (the org is the subject)
+    - Users: if the org has an account, only the account admin can manage.
+      If no account, any member can manage.
+    """
+    if is_organization(auth_subject):
+        return True
+    if is_user(auth_subject):
+        if organization.account_id is None:
```
frankie567: `account_id` is required, so this is not needed

server/polar/authz/policies/organization.py
```diff
@@ -33,3 +33,27 @@ async def can_delete(
...
+            return True
+        if await account_service.is_user_admin(
```
frankie567: I think there is a confusion between `Account` (the pure financial layer that holds transactions) and the `PayoutAccount` (the model that holds the info about _where_ to transfer the funds). I think we're in interested in the latter here.

frankie567: Or maybe you saw this as a temporary workaround until we have a proper RBAC with Owner/Finance role? In this case, it's fine :)

server/polar/authz/policies/organization.py
```diff
@@ -33,3 +33,27 @@ async def can_delete(
...
+async def can_manage_payout_account(
+    session: AsyncReadSession,
+    auth_subject: AuthSubject[User | Organization],
+    organization: OrganizationModel,
+) -> PolicyResult:
+    """Can the subject set or change the payout account for this organization?
+
+    - Organization tokens: always allowed (the org is the subject)
+    - Users: if the org has an account, only the account admin can manage.
+      If no account, any member can manage.
```
pieterbeulque: I'm probably reading this wrong, but can any member grab the admin status for an organization if there is no payout account linked yet?

server/polar/authz/policies/organization.py
joebon: Ah! Thanks

server/polar/authz/policies/organization.py
joebon: Rather a confusion on my side re: whether organizations could exist without a (billing) account that was pointed out and clarified by @frankie567. Rectified now, thanks for highlighting!

server/polar/authz/policies/organization.py
joebon: No, rather a confusion regarding whether orgs could exist without a (billing) account set. Which you clarified. Thanks!

frankie567: Ok! Side note though: I think `Account.admin_id` can go away once we have the roles in place :)

frankie567 (approved):

server/polar/authz/policies/organization.py
joebon: Yes, exactly!

PR #11200, opened 2026-04-24
feat(backoffice): surface first-review organizations in the list view

psincraian:
## Summary
- Promote the first-review rule to an `Organization.is_first_review` hybrid_property on the model, backed by both a Python and a SQL expression so filters and view logic share one source of truth.
- Use the hybrid_property in the `first_reviews=true` filter on the organizations list endpoint (replacing the inline `WHERE` clause that duplicated the rule).
- In the v2 organizations list view, render a dedicated "First Review" badge in place of the generic Review pill when the org qualifies, keeping the same warning color so the palette stays consistent.
- Drop the misleading `(≤ $10)` suffix from the "First Reviews" filter option, since the rule now accepts either an unset `initially_reviewed_at` or a threshold below the first-review limit.

## Test plan
- [ ] Open the backoffice Organizations page and confirm qualifying orgs show "First Review" while others still show "Review".
- [ ] Toggle the "First Reviews" advanced filter and confirm the list narrows to qualifying orgs only.
- [ ] Spot-check an org with a large threshold but no `initially_reviewed_at` — it should qualify.
- [ ] Spot-check an org with a small threshold and an `initially_reviewed_at` set — it should still qualify.
- [ ] `uv run task lint` and `uv run task lint_types` pass.

### DIFF
--- a/server/polar/backoffice/organizations_v2/endpoints.py
+++ b/server/polar/backoffice/organizations_v2/endpoints.py
@@ -112,8 +112,6 @@ logger = structlog.getLogger(__name__)
 
 REVIEW_TICKET_TITLE = "Ongoing organization review"
 
-FIRST_REVIEWS_MAX_THRESHOLD_CENTS = 1000
-
 
 class DeletePayoutAccountForm(BaseModel):
     reason: str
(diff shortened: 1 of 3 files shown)


frankie567 (approved):
I love hybrid properties :D

PR #11202, opened 2026-04-24
Fix UI overflow issue on home

### LINKED ISSUE #9957: UI overflows on mobile safari
### Description

Frontpage (logged in) filters below "Overview" overflow. 

I can fix it while at it. Not sure how many ppl actually uses mobile web, but regardless.

### Current Behavior

See above.

### Expected Behavior

Buttons do not overflow and the page fits to the screen.

### Screenshots
(left is macbook with safari in mobile size, right is iPhone)

Image

Image

### Environment:

-   Operating System: Tahoe 26.3 (Mac), iOS 26.3 (iPhone) 
-   Browser (if applicable): Safari

sebastianekstrom:
Fixes: https://github.com/polarsource/polar/issues/9957

| Before | After  |
|--------|--------|
| 127 0 0 1_3000_dashboard_admin-org(iPhone 12 Pro) | 127 0 0 1_3000_dashboard_admin-org(iPhone 12 Pro) (3) |

### DIFF
--- a/clients/apps/web/src/components/DashboardOverview/OverviewSection.tsx
+++ b/clients/apps/web/src/components/DashboardOverview/OverviewSection.tsx
@@ -61,7 +61,7 @@ export function OverviewSection({ organization }: OverviewSectionProps) {
         <h2 className="text-2xl font-medium text-gray-900 dark:text-white">
           Overview
         </h2>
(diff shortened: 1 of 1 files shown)


psincraian (approved):
Loved that!

PR #11206, opened 2026-04-24
Chore: Add organization_id column to orders - step 1

joebon:
## Summary

**Related Issue**: N/A

Add a denormalized `organization_id` column to the `orders` table to improve query performance when listing orders by organization. This is Phase A of a two-phase rollout.

## What

- Add nullable `organization_id` column with FK and index to the `orders` table
- Set `organization_id` on all new order creation sites (checkout, subscription, trial, wallet, backoffice import)
- Add a batched backfill script to populate existing orders

## Why

Listing orders by organization currently requires joining through the `customers` table (`Order → Customer → Organization`). A direct `organization_id` on orders eliminates this join for organization-scoped queries.

## How

**Phase A (this PR):**
- Migration adds the column as nullable (safe to run before new code deploys)
- All order creation paths now set `organization_id`
- Backfill script (`scripts/backfill_order_organization_id.py`) populates existing rows from `customers.organization_id`
- Existing queries are unchanged — they still use the customer join (updated in Phase B)

**Phase B (follow-up PR, after backfill completes):**
- Make the column NOT NULL
- Update queries to use `Order.organization_id` directly
- Replace the association proxy with a direct relationship

## Checklist


server/polar/models/order.py
```diff
@@ -175,6 +175,17 @@ class Order(CustomFieldDataMixin, MetadataMixin, RecordModel):
     def customer(cls) -> Mapped["Customer"]:
         return relationship("Customer", lazy="raise")
 
+    organization_id: Mapped[UUID | None] = mapped_column(
+        Uuid,
+        ForeignKey("organizations.id", ondelete="cascade"),
```
frankie567: I know we're not 100% consistent over the codebase for this, but for Orders, I would probably set `ondelete="restrict"` for extra safety :)

server/polar/order/service.py
```diff
@@ -595,6 +595,7 @@ async def _create_order_from_checkout(
                 tax_behavior=checkout.tax_behavior,
                 tax_rate=checkout.tax_rate,
                 invoice_number=invoice_number,
+                organization_id=organization.id,
```
frankie567, line 598: We'll need to remember to set those as pure ORM object (i.e. `organization=organization) in PR2 once we get rid of the association proxy

frankie567 (approved):
Just a nitpick on the ON DELETE CONSTRAINT, but otherwise all good :)

server/polar/models/order.py
joebon: Great that you bring it up, I wanted to double check this. Orders might very well be important for audits. But do we ever hard delete organizations? Otherwise this is of less importance

joebon: In this case it's cheap to play it extra safe, I'll fix

frankie567: Normally, we never do indeed. But who knows, a `DELETE` statement is easy to spawn :p

joebon: fixed

server/migrations/versions/2026-04-24-1352_add_organization_id_to_orders.py
```diff
@@ -0,0 +1,47 @@
...
+def upgrade() -> None:
+    op.add_column(
+        "orders",
+        sa.Column("organization_id", sa.Uuid(), nullable=True),
+    )
+    op.create_index(
+        op.f("ix_orders_organization_id"),
+        "orders",
+        ["organization_id"],
+        unique=False,
```
Yopi: You might want to add `postgresql_concurrently=True,` here to prevent the order table from being locked

Yopi (approved):

Issue #11129, opened 2026-04-21
Seat changes override scheduled subscription changes


pieterbeulque:
Tangentially related to #11128 

As discovered by @joebon, if there's a scheduled subscription change from product A → B with proration_behavior "next_period", this gets dropped if the merchant or customer changes seats on a subscription.

Ideally, both are compatible, but it's hairy on how this should work. From a customer perspective, you definitely expect your new seats to become available immediately, regardless of a scheduled upgrade from A → B.

PR #11209, opened 2026-04-24
fix(subscription): clear stale pending seats update on no-op

psincraian:
## Summary

`SubscriptionService.update_seats` early-returned when the requested seat count matched the current count, leaving any next-period `SubscriptionUpdate` untouched. A flow that set a pending seat change and later attempted to revert to the current count by re-issuing the original value left a stale pending update in place — so the subscription silently changed at the next billing cycle.

The fix soft-deletes the pending `SubscriptionUpdate` on the no-op path when it carries a seats change, using the already-eager-loaded `subscription.pending_update` relationship to avoid an extra query.

## Test plan

- [x] New regression test `TestUpdateSeats::test_next_period_revert_to_current_clears_pending_update` reproduces the reported flow (3 → pending=1 → pending=2 → revert to 3) and asserts the pending update is cleared
- [x] All existing `TestUpdateSeats` tests pass (34 tests)
- [x] Backend lint and type checks pass

### DIFF
--- a/server/polar/subscription/service.py
+++ b/server/polar/subscription/service.py
@@ -1405,7 +1405,18 @@ class SubscriptionService:
         old_seats = subscription.seats or 1
         old_amount = subscription.amount
 
+        subscription_update_repository = SubscriptionUpdateRepository.from_session(
+            session
+        )
+
         if old_seats == seats:
+            # No-op for the seat count itself, but a stale next-period
+            # SubscriptionUpdate would otherwise survive — clear it so the
+            # subscription doesn't change at the next billing cycle.
+            pending = subscription.pending_update
+            if pending is not None and pending.seats is not None:
+                await subscription_update_repository.soft_delete(pending)
+                subscription.pending_update = None
             return subscription
 
         event = await event_service.create_event(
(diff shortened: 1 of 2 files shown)


server/polar/subscription/service.py
```diff
@@ -1405,7 +1405,18 @@ async def update_seats(
...
+        subscription_update_repository = SubscriptionUpdateRepository.from_session(
+            session
+        )
+
         if old_seats == seats:
+            # No-op for the seat count itself, but a stale next-period
+            # SubscriptionUpdate would otherwise survive — clear it so the
+            # subscription doesn't change at the next billing cycle.
+            pending = subscription.pending_update
+            if pending is not None and pending.seats is not None:
```
psincraian, line 1417: @frankie567 I checked and there cannot be seat updates and product updates on PendingUpdates. So I assume this is fine

frankie567: Yes, the last one would override the previous, so fine.

frankie567 (approved):
Good catch, thanks 👍

server/polar/subscription/service.py
pieterbeulque: #11129 👀

PR #11218, opened 2026-04-27
backoffice: Add rate limit group admin

Yopi:
So we don't need to use SQL to update/change rate limit groups

### DIFF
--- a/server/polar/backoffice/organizations_v2/endpoints.py
+++ b/server/polar/backoffice/organizations_v2/endpoints.py
@@ -39,6 +39,7 @@ from polar.backoffice.organizations.forms import (
     UpdateOrganizationDetailsForm,
     UpdateOrganizationInternalNotesForm,
     UpdateOrganizationSocialsForm,
+    UpdateRateLimitGroupForm,
 )
 from polar.backoffice.organizations.orders_import import orders_import_sse
 from polar.enums import PayoutAccountType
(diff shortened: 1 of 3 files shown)


server/polar/backoffice/organizations_v2/endpoints.py
```diff
@@ -2159,6 +2159,111 @@ async def edit_order_settings(
...
+    ),
+    RateLimitGroup.elevated: (
+        "Elevated",
+        "Higher quota for trusted organizations with heavy usage",
+    ),
+    RateLimitGroup.restricted: (
+        "Restricted",
+        "Reduced limits, used for abuse mitigation",
+    ),
+    RateLimitGroup.web: (
```
frankie567: This one is not intended to be affected to an organization

server/polar/backoffice/organizations_v2/endpoints.py
```diff
@@ -2159,6 +2159,111 @@ async def edit_order_settings(
...
+                    organization_id=organization_id,
+                )
+            ),
+            hx_target="#modal",
+            classes="space-y-4",
+        ):
+            with tag.div(classes="space-y-3"):
+                for group in RateLimitGroup:
+                    label, desc = RATE_LIMIT_GROUP_DESCRIPTIONS[group]
+                    with tag.label(
```
frankie567: Nobody uses my super form builder classes 😢 

https://github.com/polarsource/polar/blob/64d09b98b1d5c592dd4de97d34484865076f5990/server/polar/backoffice/orders/endpoints.py#L738-L750

frankie567 (changes requested):
We need to remove the web option :)

server/polar/backoffice/organizations_v2/endpoints.py
Yopi: Looks nicer!

frankie567 (approved):

PR #11219, opened 2026-04-27
feat(receipts): provision S3 bucket and IAM for customer receipts

psincraian:
## Summary

PR 2 of the customer-receipts series ([design doc](handbook/engineering/design-documents/receipts.mdx)). Provisions infrastructure only — no application code reads the new env var yet, so the bucket sits empty.

- New `aws_s3_bucket "customer_receipts"` in the `s3_buckets` module, mirroring `customer_invoices`. Private bucket, no public access policy.
- New IAM policy in the `application_access` module granting the application user `PutObject` / `GetObject` (and the same read variants used by sibling buckets) on the receipts bucket, plus user attachment.
- New `customer_receipts_bucket_name` field on the `render_service` module's `aws_s3_config`, exposed as `POLAR_S3_CUSTOMER_RECEIPTS_BUCKET_NAME`.
- Per-env values wired through production, sandbox, and test (`polar-customer-receipts`, `polar-sandbox-customer-receipts`, `polar-test-customer-receipts`).

## Why safe to merge independently

No code path reads `POLAR_S3_CUSTOMER_RECEIPTS_BUCKET_NAME` yet. The bucket exists empty; the IAM grants are inert until later PRs in the series land. Forward-only.

## Test plan

- [x] `terraform fmt -recursive terraform/` — clean
- [x] `terraform validate` — passes for `production`, `sandbox`, `test`
- [ ] `terraform plan` against staging shows only the new resources (bucket + IAM policy + user attachment + env var)
- [ ] After apply: bucket exists, IAM user can write to it

### DIFF
--- a/terraform/modules/application_access/main.tf
+++ b/terraform/modules/application_access/main.tf
@@ -42,7 +42,7 @@ data "aws_iam_policy_document" "customer_invoices" {
 
 data "aws_iam_policy_document" "customer_receipts" {
   statement {
-    sid = "VisualEditor0"
+    sid = "Terraform"
     actions = [
       "s3:PutObject",
       "s3:GetObjectAttributes",
--- a/terraform/modules/application_access/main.tf
+++ b/terraform/modules/application_access/main.tf
@@ -42,7 +42,6 @@ data "aws_iam_policy_document" "customer_invoices" {
 
 data "aws_iam_policy_document" "customer_receipts" {
   statement {
-    sid = "Terraform"
     actions = [
       "s3:PutObject",
       "s3:GetObjectAttributes",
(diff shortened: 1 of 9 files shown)


terraform/modules/application_access/main.tf
```diff
@@ -39,6 +40,20 @@ data "aws_iam_policy_document" "customer_invoices" {
   }
 }
 
+data "aws_iam_policy_document" "customer_receipts" {
+  statement {
+    sid = "VisualEditor0"
```
Yopi: Very minor but: I've been thinking that it make make sense to either omit it or call it something like "Terraform" or "1" since it's not actually created via the visual editor here

Yopi (approved):
Nice!

terraform/modules/application_access/main.tf
psincraian: good point, fixint that

PR #11224, opened 2026-04-27
polar: Buffer self event ingestion

Yopi:
Instead of mapping every `/ingest` call to a self ingest call (i.e. 1:1 or n:1 events in Polar), we have a cron that runs and checks all organizations that have ingested events the past 5 minutes, and logs those as an event.

This will give us a spike every 5 minutes, but it will be a lot less than what we were seeing since we will be rolling up the events.

For the past 24 hours the difference would've been ~58% fewer tasks running in this solution vs the original solution.

### DIFF
--- a/server/polar/event/repository.py
+++ b/server/polar/event/repository.py
@@ -216,6 +216,37 @@ class EventRepository(RepositoryBase[Event], RepositoryIDMixin[Event, UUID]):
 
         return newly_rooted
 
+    async def get_latest_polar_self_ingestion_timestamp(
+        self, organization_id: UUID
+    ) -> datetime | None:
+        statement = select(func.max(Event.timestamp)).where(
+            Event.organization_id == organization_id,
+            Event.name == "event_ingestion",
+        )
+        result = await self.session.execute(statement)
+        return result.scalar_one_or_none()
+
+    async def count_user_events_by_organization(
+        self,
+        *,
+        after: datetime | None,
+        until: datetime,
+        exclude_organization_id: UUID,
+    ) -> dict[UUID, int]:
+        statement = (
+            select(Event.organization_id, func.count(Event.id))
+            .where(
+                Event.source == EventSource.user,
+                Event.organization_id != exclude_organization_id,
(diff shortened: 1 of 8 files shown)


frankie567 (approved):
That would definitely work :) 

May I suggest another approach, just to discuss :) We could leverage the task debounce mechanism. More precisely:

`polar_self.track_event_ingestion` task would just take `organization_id`, and be debounced on that. When we record an event ingestion, we would simply enqueue `enqueue_job("polar_self.track_event_ingestion", organization_id)`. And then, inside this task, do a similar thing as your proposed implementation with a count based on the last flush.  Quite similar, but it allows to optimize when nothing happens and reduce the time before we see the event.

Might be a darling, though.

Yopi:
> That would definitely work :)
> 
> May I suggest another approach, just to discuss :) We could leverage the task debounce mechanism. More precisely:
> 
> `polar_self.track_event_ingestion` task would just take `organization_id`, and be debounced on that. When we record an event ingestion, we would simply enqueue `enqueue_job("polar_self.track_event_ingestion", organization_id)`. And then, inside this task, do a similar thing as your proposed implementation with a count based on the last flush. Quite similar, but it allows to optimize when nothing happens and reduce the time before we see the event.
> 
> Might be a darling, though.

That's a nice idea. The two downsides that I see is: one would be that we we fill the queue with these tasks since we have:
1. organization: `events.ingested`
2. `polar_self.track_event_ingestion`
3. polar: `events.ingested`
4. polar: `tinybird.ingest`

But I guess with the debounce we would still only ever have at most 1 `polar_self.track_event_ingestion` per organization waiting, so it shouldn't grow more than events calls * 3 at most.

The second would be that if you spread your traffic enough we will record more events on the Polar org than we would here. I.e. here we are capping the most amount of events ingested per organizatoin to be 288 per day. At 5s queue depth we would theoretically be able to ingest 17280 events per org per day, so it's a question if we should optimize for storage or latency here

frankie567:
That's fair concern :) Let's start with the cron then

PR #11225, opened 2026-04-27
refactor(authz): migrate custom_field to PolicyGuard pattern

joebon:
## Summary

Reference implementation of the authorization model described in [`handbook/engineering/backend-development/authorization.mdx`](handbook/engineering/backend-development/authorization.mdx) for a module that previously relied on a scope-only `Authenticator` dep. Drafted to evaluate the shape before migrating other modules.

**Goal:** by the time an endpoint body executes, every authorization decision is already made. Endpoint handlers contain business logic only — no `if resource is None: raise 404`, no membership checks, no `get_payload_organization` calls, no auth subject inspection inside services.

## What

All five `custom_field` endpoints (list, create, get, update, delete) now resolve authorization in the dependency layer.

**One new factory in `polar/authz/dependencies.py`:**
- `ResourcePolicyGuard[R]` — for per-id endpoints. Resolves the resource, verifies its owning org is accessible, runs a policy, returns `AuthorizedResource[R]` (resource + organization + auth_subject). Generalizes the existing `AccountPolicyGuard` / `PayoutAccountPolicyGuard` shape (~70 lines).

**Inline deps in `polar/custom_field/auth.py`** for the list and create shapes — no factory, no wrapper dataclass:
- list dep returns `set[AccessibleOrganizationID]` directly.
- create dep takes the request body, resolves the target organization via `get_payload_organization`, runs the create policy, returns `Organization`.

If a second module's list or create dep ends up with the same shape we'll extract a factory then.

**`polar/authz/policies/custom_field.py`** — `can_read` / `can_write` returning `True` (any org member; matches existing behavior).

**`polar/custom_field/repository.py`** (new) — pure data access: `get_by_org_ids_statement(set[AccessibleOrganizationID])`, `apply_query_filter`, `get_by_organization_id_and_slug`, plus `rename_slug_in_attached_data` and `detach_from_all` (the JSONB rename and association-table cleanup that previously sat in the service as raw `session.execute(...)` calls).

**`polar/custom_field/service.py`** rewritten: `list` takes `set[AccessibleOrganizationID]`, `create` takes `(body, organization)`, `update`/`delete` take `CustomField` (already resolved by the guard). No more `auth_subject` parameter, no more `get_accessible_org_ids` call, no more `get_payload_organization` call inside the service. 264 → 134 lines.

**`polar/custom_field/auth.py`** — typed deps: `AuthorizeCustomFieldList`, `AuthorizeCustomFieldCreate`, `AuthorizeCustomFieldRead`, `AuthorizeCustomFieldWrite`.

**`polar/custom_field/endpoints.py`** — every handler is business logic only:

```python
@router.get("/{id}", ...)
async def get(authz: AuthorizeCustomFieldRead) -> CustomField:
    return authz.resource

@router.post("/", ...)
async def create(
    custom_field_create: CustomFieldCreate,
    organization: AuthorizeCustomFieldCreate,
    session,
) -> CustomField:
    return await custom_field_service.create(session, custom_field_create, organization)
```

## Why

`Authenticator` is currently doing both authentication and authorization in most modules — token-type gate plus scope gate, with no policy layer. The handbook describes a clean two-layer model (`Authenticator` for authn, `PolicyGuard` for authz, with policies in `polar/authz/policies/` and repositories taking org IDs rather than `auth_subject`), but the code only matches it for finance, members, organization, and payout-account.

Most modules diverge in two ways:
1. Per-id endpoints fetch the resource inside the service and check `if None: raise 404` — authz happening inside the handler.
2. List/create do `get_accessible_org_ids` and `get_payload_organization` inside the service — also authz happening after the body has started executing.

This PR is the smallest end-to-end migration we can do for a non-finance module to:
1. validate that the handbook's pattern is ergonomic for all five endpoint shapes,
2. show what `OrgPolicyGuard` / `AccountPolicyGuard` / `PayoutAccountPolicyGuard` need to be generalized into (answer: a single `ResourcePolicyGuard`; list and create stay inline until we have a second example),
3. give us a template the rest of the modules can be migrated against opportunistically.

## How

- `ResourcePolicyGuard` takes an explicit `get_organization_id` callable rather than a Protocol on `Mapped[UUID]` — mypy doesn't reliably understand the SQLAlchemy descriptor unwrap, and a one-line lambda at the call site is clearer.
- The create dep inlines body parsing + org resolution + policy check (~12 lines). FastAPI parses the body once and passes it to both the dep and the endpoint, so the endpoint declares `custom_field_create: CustomFieldCreate` normally.
- The pre-tool `check-patterns` hook caught a `session.execute(...)` left in the service mid-edit and forced both queries (JSONB slug rename, association-table delete) into named repository methods.
- All 29 existing custom_field tests pass without modification — external contract is identical.

## Open questions for review

1. **Trivial-policy boilerplate.** 22 lines of `return True` per simple module is friction. Accept it (explicit), or add a shared `default_policy.allow_org_members`?
2. **Should `AccountPolicyGuard` / `PayoutAccountPolicyGuard` collapse into `ResourcePolicyGuard`?** They could; left for a follow-up.
3. **Naming.** `AuthorizeCustomFieldRead` (handbook) vs `CustomFieldRead` (existing module convention). I went with `Authorize…` for all four custom_field deps to be uniform.

## Checklist

- [x] New functionality is covered by tests (existing 29 tests cover the contract; no behavior change)

### DIFF
--- a/server/polar/authz/dependencies.py
+++ b/server/polar/authz/dependencies.py
@@ -14,6 +14,10 @@ from polar.models import Organization as OrganizationModel
 from polar.models import PayoutAccount as PayoutAccountModel
 from polar.models.account import Account as AccountModel
 from polar.organization.repository import OrganizationRepository
+from polar.organization.resolver import (
+    OrganizationIDModel,
+    get_payload_organization,
+)
 from polar.organization.schemas import OrganizationID
 from polar.payout_account.repository import PayoutAccountRepository
 from polar.postgres import AsyncSession, get_db_session
@@ -22,7 +26,7 @@ from .policies import finance, members
 from .policies import organization as org_policy
 from .policies import payout_account as pa_policy
 from .service import get_accessible_org_ids, get_accessible_organization
-from .types import PolicyFn, PolicyResult
+from .types import AccessibleOrganizationID, PolicyFn, PolicyResult
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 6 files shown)


joebon:
Switching the reference module to `customer` — see follow-up branch `authz/customers`. Customer has more endpoints (incl. external_id variants) and exercises the pattern more thoroughly.

PR #11228, opened 2026-04-27
checkout/test: Attempt to fix flaky test

Yopi:
The "fetches the checkout and renders children once it resolves" test asserted on ctxRef.current after waitFor saw the testid, but ctxRef is set in a post-commit useEffect, so the assertion could fire before the effect ran. Moved that one assertion inside waitFor; the GET call assertion is upstream of the testid appearing, so it can stay outside.

~waitFor can resolve before the useEffect has run, which causes the test to fail (see: https://github.com/polarsource/polar/actions/runs/24991162646/attempts/1?pr=10837)~

~This assigns the ref immediately which should avoid the race condition~


sebastianekstrom (approved):
❤️

PR #11229, opened 2026-04-27
refactor(authz): resolve customer authorization at dep layer

joebon:
## Summary

Reference implementation of the authorization model described in [`handbook/engineering/backend-development/authorization.mdx`](handbook/engineering/backend-development/authorization.mdx), applied to the customer module.

**Goal:** by the time an endpoint handler executes, every authorization decision is already made. No `if resource is None: raise 404`, no membership checks, no `get_payload_organization` calls, no auth subject inspection inside services.

## What

All 11 `/v1/customers` endpoints (list, export, create, get, get_external, get_state, get_state_external, update, update_external, delete, delete_external) now resolve authorization in the dependency layer.

**`polar/customer/auth.py`** — 6 typed deps + small helpers:
- `AuthorizeCustomerList` → `set[AccessibleOrganizationID]` (auth + resolve accessible orgs).
- `AuthorizeCustomerCreate` → `Organization` (auth + parse body + `get_payload_organization`).
- `AuthorizeCustomerRead` / `AuthorizeCustomerWrite` → `Customer` resolved by id with org-membership filter.
- `AuthorizeCustomerExternalRead` / `AuthorizeCustomerExternalWrite` → `Customer` resolved by external_id.

The per-resource deps lean on the existing `CustomerRepository.get_readable_by_id` / `get_readable_by_external_id` methods — single-query auth-aware lookups, no extra `get_accessible_organization` round trip.

**`polar/customer/service.py`** — `list` takes `set[AccessibleOrganizationID]`; `create` takes the resolved `Organization`; `get` / `get_external` removed (deps own resource resolution); `update` / `delete` already took a resolved `Customer`.

**`polar/customer/endpoints.py`** — every handler is business logic only:

```python
@router.get("/{id}", ...)
async def get(customer: AuthorizeCustomerRead) -> Customer:
    return customer

@router.post("/", ...)
async def create(
    customer_create: CustomerCreate,
    organization: AuthorizeCustomerCreate,
    session,
) -> Customer:
    return await customer_service.create(session, customer_create, organization)
```

**Scripts** (`seeds_load.py`, `loadtest_setup.py`, `backfill_polar_self_customers.py`) — pass the `organization` they already have in scope instead of an `auth_subject`.

**`tests/customer/test_service.py`** — list tests resolve `org_ids` first; create tests pass the `organization` fixture; the `not_accessible_organization` create test exercises `get_payload_organization` directly (since that validation moved out of the service).

## Why

`Authenticator` is currently doing both authentication and authorization in most modules — token-type gate plus scope gate, with no policy layer. The handbook describes a clean two-layer model with all authorization resolved before the handler runs, but the code only matches it for finance, members, organization, and payout-account.

Customer is a meaningful reference because it has all four endpoint shapes (list, create, per-id, per-external-id) and a non-trivial service. If the pattern is ergonomic for customer, it'll work for everything else.

This PR is the **second attempt** at a reference module migration. The first (#11225, custom_field — closed) added `ResourcePolicyGuard`, `AuthorizedResource`, `OrgListGuard`, `AuthorizedList`, `AuthorizedCreate`, and `authorize_create_payload` — too many abstractions for a single example. This PR adds **zero** new infrastructure to `polar/authz/dependencies.py` and demonstrates the pattern purely through module-internal changes. Abstractions can be extracted once a second module shows the same shape.

## How

- Per-id and per-external-id deps are a thin wrapper over `get_accessible_org_ids` + the existing repo methods. ~10 lines per dep, 2 small `_resolve_by_*` helpers shared across read/write variants.
- The create dep takes the request body as a parameter (FastAPI parses the body once and passes it to both the dep and the endpoint), so the endpoint declares `customer_create: CustomerCreate` normally and the dep returns just the resolved `Organization`.
- 107 existing customer tests pass without modification of the endpoint contract. Service-level tests were updated mechanically to reflect the new signatures.

## Open questions for review

1. **Naming.** `AuthorizeCustomerRead` (handbook style) vs `CustomerRead` (existing module convention). I went with `Authorize…` for all six deps — they're qualitatively different from the old scope-only deps, so the rename signals that.
2. **`AuthorizeCustomerList` / `AuthorizeCustomerCreate` shapes.** The list dep returns `set[AccessibleOrganizationID]`; the create dep returns `Organization`. Both work, but they're not symmetric with `AuthorizeCustomerRead` (which returns `Customer`). If we ever want a uniform `AuthorizedX` wrapper, this is where the inconsistency would surface.
3. **No policy file.** Customer's access semantics are "any org member can read/write" — pure org-membership. There's no policy function to call; the membership filter in the repo IS the policy. Modules with non-trivial policies (finance, etc.) would still need a policy file + a guard that runs it.
4. **The "can org tokens write?" question.** Allowed subjects for all six deps are `{User, Organization}`, matching the existing scope-only deps. No change in who can call which endpoint.

## Checklist

- [x] New functionality is covered by tests (107 existing tests cover the contract; behavior unchanged)

### DIFF
--- a/server/polar/auth/routing.py
+++ b/server/polar/auth/routing.py
@@ -8,6 +8,48 @@ from fastapi.routing import APIRoute
 from polar.auth.dependencies import _Authenticator
 
 
+def _find_authenticator(
+    callable_: Callable[..., typing.Any],
+    seen: set[int] | None = None,
+) -> _Authenticator | None:
+    """Walk a callable's annotated dependencies to find a nested _Authenticator.
+
+    The endpoint may declare ``Depends(_Authenticator)`` directly, or wrap it
+    behind one or more module-level dependency functions (the policy-guard
+    pattern). We recurse until we find the underlying _Authenticator so the
+    OpenAPI annotations (allowed subjects, required scopes) are surfaced
+    regardless of how the dependency tree is structured.
+    """
+    seen = seen if seen is not None else set()
+    if id(callable_) in seen:
+        return None
+    seen.add(id(callable_))
+
+    try:
+        hints = typing.get_type_hints(callable_, include_extras=True)
+    except (NameError, TypeError):
+        return None
+
+    for param in hints.values():
+        if typing.get_origin(param) is not typing.Annotated:
+            continue
+
+        metadata = param.__metadata__
+        if len(metadata) == 0 or not isinstance(metadata[0], Depends):
+            continue
+
+        dependency = metadata[0].dependency
+        if isinstance(dependency, _Authenticator):
+            return dependency
+
+        if dependency is not None:
+            nested = _find_authenticator(dependency, seen)
+            if nested is not None:
+                return nested
+
+    return None
+
+
 class DocumentedAuthSubjectAPIRoute(APIRoute):
     """
     A subclass of `APIRoute` that automatically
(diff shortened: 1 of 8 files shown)


joebon:
Closing — concluded the customer module on main already satisfies the principle. Repo is pure data access (takes `org_ids`), service's `get_accessible_org_ids` call is a one-liner, endpoint orchestration is thin. Moving that call from service to endpoint is deck-chair shuffling without a load-bearing benefit. The handbook's pattern earns its keep on modules with non-trivial authz (finance is already there); trivial-policy modules like customer don't need the migration.

PR #11230, opened 2026-04-27
Pre select country in onboarding based on Vercel header

sebastianekstrom:
Pre select the country dropdown in onboarding, using [x-vercel-ip-country](https://vercel.com/docs/headers/request-headers#x-vercel-ip-country), to reduce friction 👉


pieterbeulque (approved):
😗

PR #11232, opened 2026-04-27
file download: Handle non-ASCII characters in filename header

Yopi:
By setting the filename*=UTF-8 header we can preserve the original name. We fallback to a normalized version for the legacy filename header.

### DIFF
--- a/server/polar/integrations/aws/s3/schemas.py
+++ b/server/polar/integrations/aws/s3/schemas.py
@@ -1,7 +1,9 @@
 import base64
 import hashlib
+import unicodedata
 from datetime import datetime
 from typing import Any, Self
+from urllib.parse import quote
 
 from pydantic import UUID4, computed_field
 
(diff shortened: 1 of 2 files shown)


server/polar/integrations/aws/s3/schemas.py
```diff
@@ -11,7 +13,15 @@
...
-    return f'attachment; filename="{filename}"'
+    if filename.isascii():
+        return f'attachment; filename="{filename}"'
+    ascii_fallback = (
+        unicodedata.normalize("NFKD", filename)
+        .encode("ascii", "ignore")
+        .decode("ascii")
+    )
+    encoded = quote(filename, safe="")
+    return f"attachment; filename=\"{ascii_fallback}\"; filename*=UTF-8''{encoded}"
```
frankie567, line 24: TIL you can provide alternative filename encodings :O

frankie567 (approved):

server/polar/integrations/aws/s3/schemas.py
Yopi: HTTP headers live in the 90s otherwise 😄 

https://datatracker.ietf.org/doc/html/rfc6266#section-4.3

PR #11233, opened 2026-04-27
payout: Restore Ghana payout account

Yopi:


frankie567 (approved):

pieterbeulque:
Thanks for picking this up. ~You'll probably need to regenerate the OpenAPI client too for the enum value to be picked up, no?~

(huh, it's still in there?)

PR #11234, opened 2026-04-27
Chore: Add organization_id column to orders - step 2

joebon:
## Summary

Phase B follow-up to #11206. Makes `orders.organization_id` NOT NULL, replaces the `AssociationProxy` with a real relationship, and switches queries to filter on `Order.organization_id` directly instead of joining through `Customer`.

## What

- **Migration**: `orders.organization_id` set to NOT NULL using the safe `NOT VALID` → `VALIDATE` → `SET NOT NULL` pattern (per the precedent in `2026-04-09-1625_make_organization_account_id_non_.py`) to avoid a long ACCESS EXCLUSIVE lock.
- **Model**: dropped the `AssociationProxy` (`order.organization` previously walked through `customer.organization`); replaced with `relationship("Organization", lazy="raise")`.
- **Queries**: `OrderRepository.get_readable_statement`, `get_statement_by_org_ids`, `get_due_dunning_orders`, and `get_eager_options` no longer join `Customer` to filter/load the org. Eager loads in the customer portal repo, backoffice, customer_seat repo, plain integration, and one backfill script switched from `Order.customer.organization` to `Order.organization`.
- **Order creation**: all 5 sites now assign `organization=organization` (ORM object) rather than `organization_id=organization.id` — addresses [@frankie567's comment](https://github.com/polarsource/polar/pull/11206#discussion_r-) on PR1.
- A few `order.organization.id` accesses simplified to `order.organization_id` to avoid triggering the now-strict `lazy="raise"`.

## Why

Direct `Order.organization_id` filtering eliminates the `Customer` join for organization-scoped queries — the original motivation behind PR1.

## Deploy notes

The migration's `VALIDATE CONSTRAINT` step requires zero NULL rows in `orders.organization_id`. Backfill (`scripts/backfill_order_organization_id.py`) has been run against production and verified — 0 NULL rows remain. Old code (post-PR1) already populates `organization_id` on every Order creation, so concurrent writes during the deploy window are safe.

## Checklist

- [x] This PR addresses a single concern
- [x] Existing test coverage exercises the touched paths (`uv run task test_fast` — 3843 passed, no cartesian-product warnings)
- [x] No unrelated changes

### DIFF
--- a/server/migrations/versions/2026-04-27-1530_make_orders_organization_id_non_nullable.py
+++ b/server/migrations/versions/2026-04-27-1530_make_orders_organization_id_non_nullable.py
@@ -1,0 +1,63 @@
+"""Make orders.organization_id non nullable
+
+Revision ID: b07286a34be3
+Revises: f3a2b1c4d5e6
+Create Date: 2026-04-27 09:40:38.765100
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# Polar Custom Imports
+
+# revision identifiers, used by Alembic.
+revision = "b07286a34be3"
+down_revision = "f3a2b1c4d5e6"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
+    # Use the NOT VALID / VALIDATE pattern to avoid a full-table ACCESS EXCLUSIVE
+    # lock when setting NOT NULL.
+    #
+    # Step 1: add the check constraint as NOT VALID — takes only SHARE UPDATE
+    #         EXCLUSIVE, so concurrent reads and writes are unblocked.
+    op.execute(
+        """
+        ALTER TABLE orders
+            ADD CONSTRAINT orders_organization_id_not_null
+            CHECK (organization_id IS NOT NULL) NOT VALID
+        """
+    )
+
+    # Step 2: validate the constraint — also SHARE UPDATE EXCLUSIVE.  Postgres
+    #         scans the table for violations here, but concurrent DML is still
+    #         allowed (it must simply not introduce new NULLs).
+    op.execute(
+        """
+        ALTER TABLE orders
+            VALIDATE CONSTRAINT orders_organization_id_not_null
+        """
+    )
+
+    # Step 3: set NOT NULL — because a validated check constraint already proves
+    #         no NULLs exist, Postgres 12+ skips the full table scan and holds
+    #         ACCESS EXCLUSIVE for only a very short time.
+    op.alter_column(
+        "orders", "organization_id", existing_type=sa.UUID(), nullable=False
+    )
+
+    # Step 4: the check constraint is now redundant — the NOT NULL column
+    #         constraint takes over.
+    op.execute(
+        """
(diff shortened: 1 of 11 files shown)


server/migrations/versions/2026-04-27-1530_make_orders_organization_id_non_nullable.py
```diff
@@ -0,0 +1,63 @@
...
+    # Step 1: add the check constraint as NOT VALID — takes only SHARE UPDATE
+    #         EXCLUSIVE, so concurrent reads and writes are unblocked.
+    op.execute(
+        """
+        ALTER TABLE orders
+            ADD CONSTRAINT orders_organization_id_not_null
+            CHECK (organization_id IS NOT NULL) NOT VALID
+        """
+    )
+
```
frankie567, line 48: Shouldn't we add an `UPDATE orders` here? So we backfill the few orders that may have appeared between the backfill and this migration

joebon: There shouldn't be any, because we updated all create sites to set `organization_id` in the previous PR. I checked before lunch and there at least then 0 rows with `NULL` values in sandbox and production.

joebon: And checked just now again: **0** in both sandbox and production:
```
SELECT COUNT(*) FROM orders WHERE organization_id IS NULL;
```

frankie567: Nice! I would still add it though so local environments are backfilled even if they forget to run the script :)

joebon: Ah, good point! Fixing

frankie567 (approved):

Yopi (approved):
Looks good!

server/migrations/versions/2026-04-27-1530_make_orders_organization_id_non_nullable.py
```diff
@@ -0,0 +1,77 @@
...
+# Polar Custom Imports
+
+# revision identifiers, used by Alembic.
+revision = "b07286a34be3"
+down_revision = "f3a2b1c4d5e6"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
```
psincraian, line 21: Recently I have been using lock_timeout = '2s'; or similar because in the past I found myself for the alter waiting to be picked for the transition lock

psincraian (approved):
Aside that comment, this looks good! 

Maybe the next suggestion is to have the different big migrations NOT NULL or NoT valid in different migrations. So they have their own transitions

server/migrations/versions/2026-04-27-1530_make_orders_organization_id_non_nullable.py
joebon: Ah, thanks!

PR #11237, opened 2026-04-27
Fix payout and payout accounts access

frankie567:
- server/payout_account: check access based on admin_id instead of organization
- server/payout: check accessible payouts by Account instead of PayoutAccount

### DIFF
--- a/server/polar/payout_account/service.py
+++ b/server/polar/payout_account/service.py
@@ -83,8 +83,7 @@ class PayoutAccountService:
         auth_subject: AuthSubject[User],
     ) -> Sequence[PayoutAccount]:
         repository = PayoutAccountRepository.from_session(session)
-        org_ids = await get_accessible_org_ids(session, auth_subject)
-        statement = repository.get_statement_by_org_ids(org_ids)
+        statement = repository.get_statement_by_user(auth_subject.subject)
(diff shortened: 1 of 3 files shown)


server/polar/payout_account/service.py
```diff
@@ -95,7 +94,7 @@ async def get(
     ) -> PayoutAccount | None:
         repository = PayoutAccountRepository.from_session(session)
         org_ids = await get_accessible_org_ids(session, auth_subject)
-        statement = repository.get_statement_by_org_ids(org_ids).where(
+        statement = repository.get_statement_by_user(auth_subject.subject).where(
```
Yopi, line 97: What if the auth_subject.subject is an org?

Yopi (approved):
Just a small question, not sure if it can happen here

server/polar/payout_account/service.py
frankie567: It can't, because we use the WebUser dependencies: https://github.com/polarsource/polar/blob/a37a1ef426d893f788205419f16966aaee76e4f5/server/polar/payout_account/endpoints.py#L30

It should also be enforced at type level :)

Yopi: Ah! Makes sense. Thanks!

PR #11239, opened 2026-04-27
Fix PostHog event tracking in sign up flows

sebastianekstrom:
Moved `posthog.init()` from a `useEffect` inside `<PolarPostHogProvider />` to a module-level block since child `useEffect`'s run before parent ones, so child `posthog.capture()` calls were firing before init and could get silently dropped.


pieterbeulque (approved):
It's always `useEffect` 🙅

PR #11244, opened 2026-04-28
docs(handbook): focus authorization page on architecture, not catalog

joebon:
## Summary

Trims the `engineering/backend-development/authorization.mdx` page so that it describes how authorization is *organized* rather than maintaining an exhaustive registry of every typed dependency and policy file. Detailed catalogs in handbooks tend to rot — the source of truth lives in `server/polar/authz/`.

## What

- **Typed dependencies section**: replaces the 9-line bullet list of every `Authorize*` dependency with three illustrative examples + a brief description of the always-allow variants and a pointer to `server/polar/authz/dependencies.py` for the full list.
- **Policy files section**: replaces the per-file table with a one-paragraph description of how policies are organized (one file per resource area, `can_*` verbs, new areas get new files).

The architectural content — the auth/authz split, the five-step PolicyGuard flow, the error-behavior table, the three guard variants, the org-scoped data access pattern, scopes, and the "adding a new policy-guarded endpoint" walkthrough — is unchanged.

## Why

Two problems with the previous version:

1. **The catalog was already out of date.** It didn't mention `payout_account` policies or `AuthorizeOrgManagePayoutAccount`, both of which exist in code, and it incorrectly described `AuthorizePayoutAccountRead/Write` as delegating to `finance.can_read/write`.
2. **Even if we caught it up, it would drift again.** The handbook should explain *why* the system is shaped the way it is so that someone reading it can navigate the source confidently — not duplicate the source.

## How

Targeted edits to two subsections only. No restructuring, no new content, no removed concepts — just less detail in the two places that were trying to enumerate things.

## Checklist

- [ ] New functionality is covered by tests *(N/A — docs-only)*
- [ ] Linting and type checking pass *(N/A — docs-only)*

### DIFF
--- a/handbook/engineering/backend-development/authorization.mdx
+++ b/handbook/engineering/backend-development/authorization.mdx
@@ -64,19 +64,17 @@ There are three PolicyGuard variants for different resource types:
 
 ### Typed dependencies
 
-Pre-built typed dependencies are defined in `server/polar/authz/dependencies.py`:
+Each guard is paired with a policy and exposed as a named dependency in `server/polar/authz/dependencies.py`. Endpoints depend on these names rather than wiring guards and policies themselves. For example:
 
 ```python
 AuthorizeFinanceRead    # OrgPolicyGuard + finance.can_read
-AuthorizeFinanceWrite   # OrgPolicyGuard + finance.can_write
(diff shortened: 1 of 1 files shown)


frankie567 (approved):
Good call!

PR #11245, opened 2026-04-28
Filter out web scopes from organization access token form

pieterbeulque:
## Summary

**Related Issue**: 

Removes `web:read` and `web:write` scopes from the available scopes list in the organization access token creation/update form.

## What

Modified the `sortedScopes` computation in `AccessTokenForm` to filter out the `web:read` and `web:write` scopes before sorting the remaining scopes alphabetically.

## Why

The `web:read` and `web:write` scopes should not be available for selection when creating or updating organization access tokens. This change prevents these scopes from appearing in the form's scope options.

## How

Added a `.filter()` call in the scope processing chain to exclude the two web-related scopes before the `.sort()` operation.

## Checklist

https://claude.ai/code/session_01RL942U9rZwg1KhC5du2JoN

### DIFF
--- a/clients/apps/web/src/components/Settings/OrganizationAccessTokensSettings.tsx
+++ b/clients/apps/web/src/components/Settings/OrganizationAccessTokensSettings.tsx
@@ -52,9 +52,9 @@ const AccessTokenForm = ({ update }: { update?: boolean }) => {
     AccessTokenCreate | AccessTokenUpdate
   >()
 
-  const sortedScopes = Array.from(enums.availableScopeValues).sort((a, b) =>
-    a.localeCompare(b),
-  )
+  const sortedScopes = Array.from(enums.availableScopeValues)
+    .filter((scope) => scope !== 'web:read' && scope !== 'web:write')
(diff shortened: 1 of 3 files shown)


joebon (approved):
This PR should reasonably solve the issue, but I'd consider it more as a band aid than a cure.
I'd prefer if we fix the schema in the backend instead, in [server/polar/organization_access_token/schemas.py](https://github.com/polarsource/polar/blob/main/server/polar/organization_access_token/schemas.py#L11), like this:

```
…
from polar.auth.scope import Scope, SCOPES_SUPPORTED
…
AvailableScope = StrEnum(  # type: ignore
      "AvailableScope",                                                                                                                                                                                                                                       
      {Scope(v).name: v for v in SCOPES_SUPPORTED},
)
…
```

Then it should automatically propagate to the frontend through the generated client.

frankie567:
Yep, I agree with Joel here that the schema should actually reflect that. That will also enforce it at server validation level.

frankie567 (approved):
Mucho good!

PR #11247, opened 2026-04-28
product: Treat legacy recurring products as recurring

Yopi:
The is_recurring_product needs to handle legacy recurring products, which do not have a recurring_interval. Otherwise it's impossible to switch from a legacy to non-legacy recurring product.

### DIFF
--- a/server/polar/subscription/service.py
+++ b/server/polar/subscription/service.py
@@ -1124,7 +1124,10 @@ class SubscriptionService:
         assert organization is not None
 
         assert is_recurring_product(product)
-        assert is_recurring_product(previous_product)
+        # We are checking for product.is_recurring instead of is_recurring_product
+        # because legacy products will have is_recurring but not be of type RecurringProduct
+        # which is_recurring_product asserts.
+        assert previous_product.is_recurring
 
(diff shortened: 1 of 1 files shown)


Yopi:
@frankie567 I don't know if this is the appropriate move -- or if we should have the dual check in the upgrade path instead. If it was intentional that we only want to be looking at non-legacy products here?

frankie567 (commented):
It's a bit dangerous because that guard type casts the argument to `RecurringProduct`, which guarantees `recurring_interval` is not None, which would break with a legacy product.

In which code path this causes an issue?

Yopi:
When you have a subscription to a legacy product and try to change it to a new product you're not allowed to since you're going from a "non recurring" to a recurring product

Yopi:
We could change https://github.com/polarsource/polar/blob/main/server/polar/subscription/service.py#L1127 to be `assert previous_product.is_recurring` instead

frankie567:
Yes, I would prefer that + with a comment that we do that to account for legacy recurring products

frankie567 (approved):

PR #11249, opened 2026-04-28
openapi: Fix Discount OpenAPI specification

Yopi:
The previous generation specification was incorrect / or lead the implementer into the wrong direction in regards to what was allowed for discounts.

It's not allowed to have basis_points with type fixed.

By splitting `DiscountCreate` into two schemas that's discriminated on type, and moving the validation in one step we can ensure that the OpenAPI spec becomes less confusing.

See https://app.plain.com/workspace/w_01JE9TRRX9KT61D8P2CH77XDQM/thread/th_01KQ32VWCHB81CR48BHSTN31MF/


frankie567 (approved):
I guess what I wanted to do initially was to have two discriminator: one on type, and one on duration. But it's a bit hard to express in OpenAPI. So, let's do this.

PR #11250, opened 2026-04-28
onboarding: Show errors on the fields that they are corresponding to

Yopi:
Use the same names for the fields as on the API and the error reporting should work out of the box


frankie567 (approved):
:bufo-yes-yes-yes:

sebastianekstrom (approved):

PR #11254, opened 2026-04-28
feat(receipts): allocate receipt_number on paid orders

psincraian:
## Summary

PR3 of the customer-receipts feature split (design doc: [handbook/engineering/design-documents/receipts.mdx](handbook/engineering/design-documents/receipts.mdx)). Adds the dark, flag-gated allocation step that produces a stable per-customer receipt identifier when an order becomes paid.

- New `polar/receipt/` module with a `ReceiptService.allocate(session, order)` entry point.
- `OrderRepository.lock_for_receipt_allocation` issues `SELECT receipt_number FROM orders WHERE id = ? FOR UPDATE`, returning the scalar value so callers see a fresh, lock-protected number rather than a stale identity-mapped instance.
- `CustomerRepository.increment_receipt_next_number` mirrors the existing `increment_invoice_next_number` pattern (atomic `UPDATE … RETURNING`).
- `_on_order_paid` calls `receipt_service.allocate` directly. Failures propagate so retry machinery (Stripe webhook retries / Dramatiq actor retries) drives recovery instead of a custom error envelope.

Idempotent under concurrent allocates on the same order: the row-level lock plus the post-lock null-check prevents counter waste in the rare race.

## Why safe to merge

The whole path is gated on `Organization.is_receipts_enabled` (read from `feature_settings["receipts_enabled"]`), which is `false` everywhere. Until the flag is flipped per-organization, `allocate` short-circuits before doing any work. No production behavior change.

## Test plan

- [x] Unit tests cover flag-off, no-succeeded-payment, allocation, idempotent re-call, and counter advance across orders (`server/tests/receipt/test_service.py`).
- [x] Backend lint + ruff format clean.
- [x] Backend type check (`mypy`) clean.
- [x] Existing 112 order-service tests still pass.
- [x] End-to-end run against the local dev database with the running stack: verified flag off → no allocation, flag on → `RCPT-XXXXXXXXXX-0001`, idempotent re-call → no double-write, second order → `0002` with counter advancing cleanly. Test transaction was rolled back; dev DB unchanged.

### DIFF
--- a/server/polar/receipt/service.py
+++ b/server/polar/receipt/service.py
@@ -1,0 +1,47 @@
+from polar.customer.repository import CustomerRepository
+from polar.kit.db.postgres import AsyncSession
+from polar.models import Order
+from polar.order.repository import OrderRepository
+from polar.payment.repository import PaymentRepository
+
+RECEIPT_NUMBER_PREFIX = "RCPT"
+
+
+class ReceiptService:
+    async def allocate(self, session: AsyncSession, order: Order) -> Order:
+        """Allocate ``receipt_number`` for ``order``.
+
+        Idempotent under concurrent calls: the row-level lock plus the
+        post-lock null-check prevents double-allocation.
+        """
+        if not order.organization.is_receipts_enabled:
+            return order
+
+        if order.receipt_number is not None:
+            return order
+
+        payment_repository = PaymentRepository.from_session(session)
+        succeeded_payment = await payment_repository.get_succeeded_by_order(order.id)
+        if succeeded_payment is None:
+            return order
+
+        order_repository = OrderRepository.from_session(session)
+        current_number = await order_repository.lock_for_receipt_allocation(order.id)
+        # Another transaction may have allocated between the null-check above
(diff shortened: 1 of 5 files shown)


server/polar/receipt/service.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class ReceiptService:
+    async def allocate(self, session: AsyncSession, order: Order) -> Order:
+        """Allocate ``receipt_number`` for ``order``.
+
+        Idempotent under concurrent calls: the row-level lock plus the
+        post-lock null-check prevents double-allocation.
+        """
+        if not order.organization.is_receipts_enabled:
```
frankie567, line 17: Do we really need an option to disable receipts?

frankie567 (approved):

server/polar/receipt/service.py
psincraian: this will be mainly used for the rolling phase, the problem is that I want to backfill the receipt number for every order and organization, and I would like to do this on phases and gradually to avoid errors or locks

PR #11258, opened 2026-04-28
tax: Add Georgia for validating tax ID

Yopi:


frankie567 (approved):
Try to find if there was at least some check digits or something, but doesn't look like it.

Yopi:
> Try to find if there was at least some check digits or something, but doesn't look like it.

No I couldn't find anything either, and as I understand it Stripe just accepts everything that is valid shape and let the business take care of it. 

I can do a sweep of the other ones we are missing for validation

Issue #11210, opened 2026-04-24
Tax Breakdown


frankie567:
# Problem

Currently, we're only able to handle one tax rate and one taxability reason per order/checkout. While it works most of the time, it becomes an issue with jurisdictions that add up several type of taxes; typically for the US where counties might add something on top of the tax state.

While we get the final amount correct, the breakdown shown in the invoice is then incorrect:

<img width="526" height="288" alt="Image" src="https://github.com/user-attachments/assets/9b83d879-6b2c-4845-9f9c-ca1995a8fe7d" />

In this example, New York has 4% tax (so 0.40$), but there is also another 4% for a county that doesn't appear here.

# Current implementation

On Order, Checkout and Wallet Transaction, we store `tax_rate` and `taxability_reason`, which corresponds to the information of a single tax line.

# Proposal

Streamline those two fields into a single one, `tax_breakdown`, a JSONB list where each item contains the tax rate and reason information. For example:

```json
[
    {
        "rate_type": "percentage",
        "basis_points": 400,
        "display_name": "NYC Tax",
        "country": "US",
        "state": "NY",
        "amount": 40,
        "taxability_reason": "standard_rated"
    },
    {
        "rate_type": "percentage",
        "basis_points": 400,
        "display_name": "County Tax",
        "country": "US",
        "state": "NY",
        "amount": 40,
        "taxability_reason": "standard_rated"
    },
]
```

It would then be possible to correctly display the breakdown on the invoice.

### Phase 1

- Implement `TaxBreakdown` type, which is a combination of of `TaxRate` and `TaxabilityReason`. Note: `amount` is now mandatory since we should reflect the amount of that line.
- Add `tax_breakdown` column on `Order`, `Checkout` and `WalletTransaction`.
- Implement a backfill script for existing orders, checkouts and wallet transactions. Since we only have one tax rate right now, we can take a shortcut and consider the `tax_amount` as the `amount` for this single item.
- Make tax calculation logic return this tax breakdown list instead of tax rate + taxability reason; and save it to `Order`, `Checkout` and `WalletTransaction`.
- Adapt the invoice generator so it shows the tax breakdown properly

### Phase 2

- Drop `tax_rate` and `taxability_reason` columns.

PR #11259, opened 2026-04-28
server/tax: return and record tax breakdown

frankie567:
Phase 2 of #11210: make sure to parse and record the full breakdown of taxes.

In Phase 3, we'll make sure to use that properly in the generated invoices.

### DIFF
--- a/server/polar/tax/calculation/numeral.py
+++ b/server/polar/tax/calculation/numeral.py
@@ -10,6 +10,7 @@ from opentelemetry import trace
 from polar.config import settings
 from polar.enums import TaxBehavior
 from polar.kit.address import Address
+from polar.kit.math import polar_round
 from polar.logging import Logger
 
 from ..tax_id import TaxID
@@ -17,11 +18,11 @@ from .base import (
     CalculationExpiredError,
     InvalidTaxIDError,
     TaxabilityReason,
+    TaxBreakdownItem,
     TaxCalculation,
     TaxCalculationLogicalError,
     TaxCalculationTechnicalError,
     TaxCode,
-    TaxRate,
     TaxRecordError,
     TaxServiceProtocol,
 )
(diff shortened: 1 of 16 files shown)


server/polar/tax/calculation/numeral.py
```diff
@@ -262,24 +286,26 @@ async def calculate(
...
+            jurisdictions,
+            country=country,
+            state=state,
+            amount=amount,
+            customer_exempt=customer_exempt,
         )
-
-        note = tax_jurisdiction["note"].lower()
-        taxability_reason = TaxabilityReason.from_numeral(note, customer_exempt)
+        assert sum(item["amount"] for item in tax_breakdown) == total_tax_amount
```
psincraian, line 307: Mmm I'm worried that this can fail. Mainly because on the item["amount
"] we are computing the value using `polar_round` and we are comparing our computed amount against numeral. How will it work when there are fractional cents in here? For exaple:

Jurisdiction | Rate | amount * rate | polar_rount |
----|---|---|--|
State | 4.5% | 4.5 | 5
County | 2.5% | 2.5 | 2

server/polar/tax/calculation/numeral.py
```diff
@@ -102,38 +103,60 @@ def to_numeral_tax_id(tax_id: TaxID) -> NumeralTaxId:
...
-        basis_points=basis_points,
-        amount=amount,
-        amount_currency=amount_currency,
-        display_name=display_name,
+        assert jurisdiction["tax_rate"] is not None
+        rate = jurisdiction["tax_rate"]
+        item_amount = polar_round(amount * rate)
+
+    return TaxBreakdownItem(
+        rate_type="fixed" if is_fixed else "percentage",
```
psincraian, line 130: I think we should align the team on literals vs enums

psincraian (commented):
I miss some tests on the comment that I post

server/polar/tax/calculation/numeral.py
frankie567: Normally, this rounding is the standard in those type of calculations. So, I'm quite confident this will work. Added a test to show how we parse the Numeral API.

Anyway, I'll also upgrade the Numeral API version in a follow-up PR, which embeds the breakdown amounts directly, so this won't be necessary.

server/polar/tax/calculation/numeral.py
frankie567: I would say to use enums in most cases, especially when it's output in the API. In this case, since this is a rather private thing, I thought a literal was enough.

psincraian (approved):
LGTM

PR #11261, opened 2026-04-28
metrics: Add trial mrr + trial cmrr to subscriptions dashboard

Yopi:


clients/apps/web/src/utils/metrics.ts
```diff
@@ -273,6 +273,18 @@ export const METRIC_GROUPS: MetricGroup[] = [
...
+        slug: 'trial_monthly_recurring_revenue',
+        display_name: 'Trial MRR',
+        description:
+          'Normalized monthly revenue from subscriptions currently in a trial period.',
+      },
+      {
+        slug: 'trial_committed_monthly_recurring_revenue',
+        display_name: 'Trial Committed MRR',
+        description:
+          'Trial MRR from subscriptions still within a committed billing period.',
```
pieterbeulque: This description feels a bit… odd. (It does on all committed metrics, not only this one, so we can iterate separately).

clients/apps/web/src/utils/metrics.ts
```diff
@@ -273,6 +273,18 @@ export const METRIC_GROUPS: MetricGroup[] = [
...
       },
+      {
+        slug: 'trial_monthly_recurring_revenue',
+        display_name: 'Trial MRR',
+        description:
+          'Normalized monthly revenue from subscriptions currently in a trial period.',
+      },
+      {
+        slug: 'trial_committed_monthly_recurring_revenue',
+        display_name: 'Trial Committed MRR',
```
pieterbeulque, line 284: I'm not sure if we should render both on a dashboard or if we should be smarter about it.

`trial_monthly_recurring_revenue` → "Trialing MRR"
`trial_committed_monthly_recurring_revenue` → "Converting Trialing MRR"

or, if we don't want to be smart, I think it should be `Committed Trialing MRR`, not `Trialing Committed`. Quite the nit-pick 😅

pieterbeulque: That is, if my understanding of `trial_committed_monthly_recurring_revenue` is correct (i.e. `cancel_at_period_end` is not set)

pieterbeulque (approved):

clients/apps/web/src/utils/metrics.ts
Yopi: Is the name what you're not sure about or the fact that we show both charts? I added both to keep the consistency with the non trial MRRs but I'm also not sure about the value with it

pieterbeulque: Both! If we were to show only one, I'd show `trial_committed_monthly_recurring_revenue` and call it "Trialing MRR". If we want to show both for consistency, then I'd relabel them a bit.

Yopi: "Trialing MRR" and "Trialing MRR not churned"? 😄

Yopi: Updated the names + descriptions

clients/apps/web/src/utils/metrics.ts
```diff
@@ -273,6 +273,17 @@ export const METRIC_GROUPS: MetricGroup[] = [
         description:
           'MRR from subscriptions still within a committed billing period.',
       },
+      {
+        slug: 'trial_monthly_recurring_revenue',
+        display_name: 'Trial MRR Excluding Canceled Trials',
```
pieterbeulque: Isn't this the one including the canceled trials 😁

Yopi: Of course 🤦

Yopi:
@frankie567 could I have an eye on this from you as well?

frankie567 (approved):

PR #11263, opened 2026-04-28
chore: Remove unused feature flags

Yopi:


frankie567:
Is there a risk to lock the table for a long time in production? Maybe we should run a backfill script first

Yopi:
> Is there a risk to lock the table for a long time in production? Maybe we should run a backfill script first

Good point! Could run the batched update maybe? To do it incrementally

frankie567:
Yes, that's what is was thinking :)

Yopi:
Changed it to an ad-hoc script instead

frankie567 (approved):

PR #11264, opened 2026-04-28
metrics: Use feature flag to combine MRR + trial MRR into MRR

Yopi:
Allow the organization to use a feature flag to combine MRR and trial MRR into the MRR metric.

This switches MRR + cMRR to be MetaMetrics, to optionally combine a base MRR metric with the trial metrics.

### DIFF
--- a/clients/apps/web/src/components/Metrics/dashboards/metrics-config.ts
+++ b/clients/apps/web/src/components/Metrics/dashboards/metrics-config.ts
@@ -19,6 +19,8 @@ export function isValidMetricType(value: string): value is MetricType {
 const SUBSCRIPTION_METRICS: (keyof schemas['Metrics'])[] = [
   'monthly_recurring_revenue',
   'committed_monthly_recurring_revenue',
+  'trial_monthly_recurring_revenue',
+  'trial_committed_monthly_recurring_revenue',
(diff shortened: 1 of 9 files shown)


pieterbeulque (changes requested):
I don't think we should do this. I get where it's coming from, we did a poor job changing our definition of MRR and communicating that (i.e. not at all), but adding a flag to return to the "old" numbers isn't the right call.

My main concern is that it opens a lot of other numbers to interpretation that this flag doesn't cover. For example, if you include trialing MRR under MRR, do we then also consider a trialing user an active user? Is a canceled trial now churned revenue? How can we show trial conversion or new MRR movements, if the trial value is part of MRR from day one?

It's worth noting that excluding trials is the standard:

 - [Stripe](https://support.stripe.com/questions/impact-of-subscription-trial-periods-on-monthly-recurring-revenue-%28mrr%29-in-billing)
 - [Chargebee](https://www.chargebee.com/resources/glossaries/what-is-mrr/)
 - [Baremetrics](https://baremetrics.com/academy/6-mistakes-companies-make-when-calculating-mrr)

As with all toggles, I'd prefer for us to take a stance and, in this case, own our earlier mistake.

PR #11265, opened 2026-04-28
ci: Add a check to see if openapi generate changes

Yopi:
We want to warn the developer if they made changes to the backend but forgot to regenerate the openapi specification.


pieterbeulque (approved):
This will lead to cleaner diffs! I don't like it when an unrelated change slips into my PR's, of when my changes clutter someone else's PR.

psincraian (approved):
Love that!

PR #11267, opened 2026-04-28
openapi: Regenerate spec

Yopi:


psincraian:
@Yopi is this right? seems that there are 14 lines that have changed

Yopi:
> @Yopi is this right? seems that there are 14 lines that have changed

I didn't check too carefully, but this is what I got when I generated from the latest main

Yopi:
The discount stuff is correct, I did that earlier today

psincraian (approved):

PR #11270, opened 2026-04-28
Add country-specific minimum payout amounts

pieterbeulque:
## Summary

[Stripe docs for reference](https://docs.stripe.com/global-payouts/send-money)

Implement country-specific minimum payout amounts for Stripe payouts, which vary by recipient country and currency. This ensures compliance with Stripe's per-country payout minimums.

## What

- Added `ACCOUNT_PAYOUT_MINIMUM_BALANCE_PER_PAYOUT_COUNTRY` configuration mapping ISO 3166-1 alpha-2 country codes to USD-equivalent minimum payout amounts (in cents)
- Refactored `get_minimum_payout_for_currency()` to `get_minimum_payout()` to accept an optional `country` parameter
- Updated the method to return the maximum of currency-based and country-based minimums
- Updated payout service (`estimate()` and `create()` methods) to pass country information when checking minimum payout amounts
- Updated frontend text to reflect that minimum withdrawal amounts vary by location

## Why

Stripe enforces per-country minimum payout amounts in the recipient's local currency. The previous implementation only considered currency-based minimums (e.g., $10 USD default), which could be insufficient for certain countries. For example, Panama requires a $50 USD minimum, which is higher than the default. This change ensures payouts respect all applicable minimums.

## How

1. Added a new configuration dictionary with country-specific minimums, converted from local currencies to USD using current FX rates and rounded up to the nearest $5 for safety margin
2. Extended the minimum payout calculation logic to consider both currency and country
3. Updated all call sites in the payout service to pass the payout account's country
4. Updated tests to verify country-specific minimums are enforced correctly
5. Updated UI text to be more generic since minimums now vary by location

## Checklist

- [x] This PR addresses a single concern (implementing country-specific payout minimums)
- [x] New functionality is covered by tests (parametrized test cases added for country-specific scenarios)

https://claude.ai/code/session_013PFnuei5WneNUKfHyFhXoy

### DIFF
--- a/docs/features/finance/payouts.mdx
+++ b/docs/features/finance/payouts.mdx
@@ -45,7 +45,7 @@ The minimum balance required to issue a payout varies based out the payout curre
 | COP      | $50.00              |
 | Other currencies | $10.00 (default) |
 
-A handful of countries also enforce a higher minimum on top of the per-currency floor above, because Stripe applies their local minimum even when payouts are denominated in USD or another currency:
+A handful of countries also enforce a higher minimum:
 
 | Country | Minimum Balance (USD) |
 |---------|----------------------|
--- a/docs/features/finance/payouts.mdx
+++ b/docs/features/finance/payouts.mdx
@@ -45,6 +45,14 @@ The minimum balance required to issue a payout varies based out the payout curre
 | COP      | $50.00              |
 | Other currencies | $10.00 (default) |
 
+A handful of countries also enforce a higher minimum on top of the per-currency floor above, because Stripe applies their local minimum even when payouts are denominated in USD or another currency:
+
+| Country | Minimum Balance (USD) |
+|---------|----------------------|
+| Bahamas      | $30.00 |
+| El Salvador  | $30.00 |
+| Panama       | $50.00 |
+
 ## Manual Withdrawal
 
 We require this to be done manually since:
--- a/server/polar/config.py
+++ b/server/polar/config.py
@@ -397,35 +397,18 @@ class Settings(BaseSettings):
         "usd": _DEFAULT_ACCOUNT_PAYOUT_MINIMUM_BALANCE,
     }
 
-    # Stripe enforces per-country minimum payout amounts (in the recipient's
-    # local currency). Since we trigger payouts in USD, these have been
-    # converted using current FX rates and rounded up to the next multiple
-    # of $5 USD to give us headroom against FX fluctuations. Values are in
-    # USD cents and indexed by ISO 3166-1 alpha-2 country code. Only
-    # countries whose USD-equivalent minimum is meaningful (i.e., greater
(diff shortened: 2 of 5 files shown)


server/polar/config.py
```diff
@@ -396,6 +396,37 @@ def POLAR_SELF_ENABLED(self) -> bool:
...
     }
+
+    # Stripe enforces per-country minimum payout amounts (in the recipient's
+    # local currency). Since we trigger payouts in USD, these have been
+    # converted using current FX rates and rounded up to the next multiple
+    # of $5 USD to give us headroom against FX fluctuations. Values are in
+    # USD cents and indexed by ISO 3166-1 alpha-2 country code. Only
+    # countries whose USD-equivalent minimum is meaningful (i.e., greater
+    # than the default $10 minimum) are listed here. See:
+    # https://docs.stripe.com/payouts/minimum-and-maximum-payout-amounts
```
pieterbeulque:
```suggestion
    # https://docs.stripe.com/global-payouts/send-money
```

psincraian (approved):
LGTM!

We also have this in the docs https://polar.sh/docs/features/finance/payouts

server/polar/config.py
```diff
@@ -553,10 +584,18 @@ def generate_backoffice_url(self, path: str) -> str:
     def stripe_descriptor_suffix_max_length(self) -> int:
         return 22 - len("* ") - len(self.STRIPE_STATEMENT_DESCRIPTOR)
 
-    def get_minimum_payout_for_currency(self, currency: str) -> int:
-        return self.ACCOUNT_PAYOUT_MINIMUM_BALANCE_PER_PAYOUT_CURRENCY.get(
+    def get_minimum_payout(self, currency: str, country: str | None = None) -> int:
```
frankie567: Country should be a required argument, there is no situation where we won't have it

frankie567 (changes requested):

server/polar/config.py
pieterbeulque: Done!

pieterbeulque:
I cross-referenced with the minimums that we already set based on payout currency, and I think we can do with these three country exceptions for now. That felt better just to reduce complexity.

docs/features/finance/payouts.mdx
```diff
@@ -45,6 +45,14 @@ The minimum balance required to issue a payout varies based out the payout curre
 | COP      | $50.00              |
 | Other currencies | $10.00 (default) |
 
+A handful of countries also enforce a higher minimum on top of the per-currency floor above, because Stripe applies their local minimum even when payouts are denominated in USD or another currency:
```
pieterbeulque:
```suggestion
A handful of countries also enforce a higher minimum:
```

frankie567 (approved):

PR #11148, opened 2026-04-22
refactor(auth): remove web_read/web_write scope dependency

joebon:
## Summary

**Related Issue**: #11126

Remove `web_read`/`web_write` from all authorization scope checks and enforce web-session-only access on `WebUser*` dependencies.

> Depends on #11147

## What

- `web_read`/`web_write` removed from all 33 `auth.py` `required_scopes` sets and from `SCOPES_SUPPORTED`, display names
- `RESERVED_SCOPES` removed — `web_read`/`web_write` filtered via `_LEGACY_SCOPES` instead
- `WebUserRead`/`WebUserWrite`/`WebUserOrAnonymous` enforce web-session-only access via `is_web_session()` — API tokens get 403
- `is_web_session()` added — checks `isinstance(session, UserSession)`
- `READ_ONLY_SCOPES` added for impersonation sessions (replaces runtime `:read` suffix filtering)
- Web sessions created with all scopes instead of `{web_read, web_write}`
- `EventTypeWrite` scope changed from `web_write` to `events_write`
- `web_write` removed from `OrgPolicyGuard` defaults and `AuthorizeMembersManage`/`AuthorizeOrgDelete`/`AuthorizeOrgManagePayoutAccount`
- Finance guards use finance-specific scopes instead of defaults
- Legacy session upgrade: exact `{web_read, web_write}` → all scopes in middleware (safe to remove after 2026-05-22)

## Why

`web_read`/`web_write` were never authorization scopes — they were a proxy for "is this a web session?" This conflated authentication method with authorization, and required every `auth.py` to include them as fallbacks for web sessions to pass scope checks.

## How

- `WebUserRead`/`WebUserWrite` now wrap the authenticator with an `is_web_session()` check that raises `NotPermitted` for API tokens
- `web_read`/`web_write` kept in `Scope` enum only for legacy session deserialization (safe to remove after 2026-05-22)
- Legacy sessions with exactly `{web_read, web_write}` transparently upgraded in middleware

## Checklist

- [x] This PR addresses a single concern (one bug fix, one feature, one refactor)
- [x] The diff is reasonably sized and easy to review
- [x] New functionality is covered by tests
- [x] Linting and type checking pass (`uv run task lint && uv run task lint_types`)
- [x] No unrelated changes or drive-by fixes are included

PR #11303, opened 2026-04-29
auth: split web-user authentication and authorization

joebon:
## Summary

Realigns user-personal endpoint dependencies with the two-layer model in the [authorization handbook](https://handbook.polar.sh/engineering/backend-development/authorization), and **closes a 2-day-old gap from #11148** where impersonation sessions could perform writes against user-personal endpoints (e.g. `PATCH /users/me`, `DELETE /personal_access_tokens/{id}`, OAuth-link callbacks). What started as a naming/architectural cleanup turned into a meaningful security fix once the impersonation impact surfaced during review.

## What

Five commits, each addressing a discrete concern:

1. **Split web-user authentication and authorization.** Collapsed `WebUserRead` / `WebUserWrite` (identical aliases) into a single `WebUserSession` authentication primitive, and moved user-personal authorization into `server/polar/authz/dependencies.py` alongside the existing `OrgPolicyGuard` family.
2. **Gate user-personal endpoints with explicit scopes.** Replaced the unparameterized authorization wrapper with scope-gated aliases — initially `AuthorizeUser{Read,Write}` and `AuthorizePayouts{Read,Write}`, plus `AuthorizeOrganizationsRead` (later removed). Each user-personal endpoint declares the scope it actually needs; impersonation is automatically blocked from writes via `READ_ONLY_SCOPES`.
3. **Test the impersonation read-only invariant.** Added `tests/authz/test_user_personal_dependencies.py` (contract tests on `WebUserAuthorizer`, including a `READ_ONLY_SCOPES`-vs-`_write`-scope invariant test) plus impersonation gates on `user`/`feedback`/`payout_account` endpoint tests.
4. **Address review findings.** Migrated `DELETE /users/me` (last user of the legacy `UserWrite`), upgraded `user_repositories` to `Write` (token-refresh side effect), moved `org_stream` to `OrgPolicyGuard` (org-scoped, not user-personal), and removed the now-unused `AuthorizeOrganizationsRead`.
5. **Split aliases into `Web*` and non-`Web*` prefix families.** The previous `AuthorizeUser*` aliases all routed through `WebUserSession`, which broke the mobile app's `DELETE /v1/users/me` (calls with an OAuth2 PKCE bearer token). Final split:
   - `AuthorizeWeb{User,Payouts}{Read,Write}` — User **via web session** only. Rejects API tokens. Used by ~all browser/dashboard flows.
   - `Authorize{User}{Read,Write}` — Any User subject (web/PAT/OAuth2) with the appropriate scope. Used by `DELETE /v1/users/me`.

The handbook section "User-personal endpoints" documents the pattern, the two prefixes, and how the scope gate enforces impersonation read-only.

## Why

Two problems with the previous state:

1. **Architectural confusion.** The handbook splits authentication ("who are you?") and authorization ("are you allowed?"), with `Authenticator` for the former and `PolicyGuard` for the latter. `WebUserRead`/`WebUserWrite` blurred that — both lived in `auth/`, were behaviorally identical (no scope check, no policy), and the Read/Write distinction implied an enforcement gradient that didn't exist. Reading the handbook then the codebase, you'd expect a guard from `authz/` for the user-personal cases and find none.
2. **Impersonation could write.** Backoffice impersonation creates a user session with only `READ_ONLY_SCOPES` (the `_read` set). Resource endpoints already block writes via their org-scoped guards' `required_scopes`. Until #11148 (2026-04-27) the same gate existed for user-personal endpoints via the `web_read`/`web_write` scopes. #11148 removed those scopes intentionally — but the user-personal endpoints didn't get a replacement, so impersonators could `PATCH /users/me`, link OAuth accounts, delete PATs, etc. The handbook gap and the scope gap had the same root cause.

The fix makes both problems go away with the same mechanism: each endpoint declares an appropriate `_read` or `_write` scope, and the property "impersonation is read-only" emerges from a system invariant — `READ_ONLY_SCOPES` contains only `_read`s, so any `_write` declaration is automatically off-limits to impersonation. No new column, no impersonation-specific detection, no per-endpoint guard logic.

## How

`WebUserAuthorizer` is a small factory that depends on `WebUserSession` and checks the auth subject has at least one of the required scopes — same shape as the layered design used by `OrgPolicyGuard`. Each typed alias is one line. `AuthorizeUser*` (the non-`Web*` family) wraps `Authenticator` directly, since it doesn't need the web-session restriction.

### Per-caller scope behavior — no regression for any legitimate caller

| Caller | What changed |
|---|---|
| Web session (browser dashboard) | **Nothing.** Web sessions carry `set(Scope)` since #11148, so they satisfy every new check. |
| Mobile app (OAuth2 PKCE token on `DELETE /v1/users/me`) | **Nothing.** Still goes through `Authenticator(allowed_subjects={User}, required_scopes={user_write})` — same contract as the deleted legacy `UserWrite`. |
| PAT / OAT / OAuth2 token hitting any `AuthorizeWeb*` endpoint | **Nothing.** Was rejected pre-PR by `is_web_session()`; still rejected by `WebUserSession`. |
| Impersonation session (`READ_ONLY_SCOPES`) | **Blocked from writes** — this is the fix. Reads still pass. |

### Two intentional widenings worth flagging

1. **`user_write` is now a broader scope semantically.** Pre-PR, it gated only `DELETE /v1/users/me` and was described as "Delete your user account." Now it gates ~17 user-personal write endpoints (profile patch, email update, OAuth links, PAT delete, OAuth2 consent, integration callbacks, etc.) and is described as "Manage your user account." Same scope value, expanded meaning. The alternative would have been new per-resource scopes (`personal_access_tokens_write`, `oauth2_consent_write`, etc.); we deliberately chose to fold into `user_write` rather than expand the scope surface, and rewrote the description so any pre-existing OAuth2 grant displays accurate consent.
2. **`GET /stream/organizations/{id}`** now also accepts Organization access tokens (OATs). Pre-PR it was web-session-only via `WebUserRead`. The endpoint takes an org `{id}` path param and is functionally org-scoped, so we moved it to `AuthorizeOrgAccess` (the `OrgPolicyGuard` always-allow variant), which the first review pass identified as the architecturally honest home. Scope requirements (`organizations_read` | `organizations_write`) are unchanged; the *subject set* widened.

### No new scopes added to the `Scope` enum

This PR only re-uses pre-existing scopes (`user_read`, `user_write`, `payouts_read`, `payouts_write`, `organizations_read`, `organizations_write`).

## Testing

- **Unit (`tests/authz/test_user_personal_dependencies.py`)** — direct contract tests on `WebUserAuthorizer`: scope match passes, missing scope raises, read aliases accept either `_read` or `_write`, write aliases reject `_read`-only. Includes a `READ_ONLY_SCOPES` invariant test asserting it contains no `_write` scopes — which is the system-level guarantee underpinning impersonation read-only. If a future scope addition violates this, every user-personal write endpoint silently becomes reachable by impersonation, so the test fails loudly.
- **Integration (`tests/{user,feedback,payout_account}/test_endpoints.py`)** — sessions parameterized with `READ_ONLY_SCOPES` are accepted on reads (`GET /users/me`, `GET /payout-accounts/`) and rejected with 403 on writes (`PATCH /users/me`, `DELETE /users/me`, `POST /feedbacks/`, `POST /payout-accounts/`).

The baseline test fixture (`AuthSubjectFixture`) defaults to `set(Scope)`, which is why missing-scope regressions slipped through #11148: every authed test session had every scope. The new tests deliberately exercise the constrained-scope case.

## Checklist


frankie567 (commented):
There is one place where enforcing read-only web scope was useful: impersonation. Ideally, we would like to avoid an admin impersonating a user accidentally making destructive changes like updating an email.

But maybe then it could be handled by the policy?

joebon:
> There is one place where enforcing read-only web scope was useful: impersonation. Ideally, we would like to avoid an admin impersonating a user accidentally making destructive changes like updating an email.

That's a very good point @frankie567, thanks for bringing it up! This flaw actually already exists since #11148, so I increased the scope of this PR instead to fix it. Will also check if we're missing tests to verify this.

frankie567 (approved):

frankie567 (approved):
Yep, feels really good with those user read/write scopes 👍

PR #11304, opened 2026-04-29
feat(receipts): add receipt PDF renderer

psincraian:
## Summary

First half of the receipts feature ([design doc](https://handbook.polar.sh/engineering/design-documents/receipts)). Pure rendering layer — no DB, S3, workers, or HTTP endpoints. Those land in a follow-up PR that calls into this one.

- New `Receipt` Pydantic model + `ReceiptGenerator` that subclasses `InvoiceGenerator`. Adds:
  - Heading shows Invoice number, Receipt number, Date paid (falls back to Date of issue when no payment).
  - Bill-to block falls back to customer email when billing name/address are missing — the whole point of receipts is they don't gate on billing details.
  - Items table inherited from the invoice.
  - Totals rows: Subtotal / Tax / Total / Applied balance / To be paid (inherited) plus new **Amount paid** and **Amount refunded**.
  - **Payment history** table at the bottom (method + last4, date, amount, receipt number).
  - **Refunds** section with Date + Amount columns and a Total refunded line.
  - Red **FULLY REFUNDED** banner at the top of page 1 when cumulative refunds equal the order total.
  - Footer: receipt number, generated-at timestamp, page number.
- New `render_receipt_pdf(receipt)` runs the generator in a subprocess (`python -m polar.receipt.render`) with `asyncio.wait_for` on both `communicate` and the post-`kill` `wait`. The invoice renderer lacks both — receipts get the safer version.
- `polar/invoice/generator.py` refactored: `generate()` is now a sequence of calls to `_render_title`, `_render_heading_items`, `_render_addresses` (which delegates to `_render_seller_block` and `_render_customer_block`), `_render_items_table`, `_render_totals_table`, `_render_notes`. Pure refactor, no behavior change for invoices. ReceiptGenerator only overrides what differs (`_render_customer_block` for the email fallback, plus its own `header`, `footer`, post-`generate()` payment history + refunds sections).

## Examples

| Scenario | What it shows |
|---|---|
| [1_paid.pdf](https://github.com/user-attachments/files/27199159/1_paid.pdf) | Plain paid order — heading rows (Invoice number / Receipt number / Date paid), Bill-to with address + email, items table, Subtotal / Total / **Amount paid**, Payment history. |
| [2_partial_refund.pdf](https://github.com/user-attachments/files/27199176/2_partial_refund.pdf) | Partial refund — adds the **Refunds** section (Date, Amount), a **Total refunded** line, and an **Amount refunded** row in the totals. 
| [3_fully_refunded.pdf](https://github.com/user-attachments/files/27199186/3_fully_refunded.pdf) | Two refunds summing to the order total — the red **FULLY REFUNDED** banner appears at the top of page 1. |
| [4_with_applied_balance.pdf](https://github.com/user-attachments/files/27199189/4_with_applied_balance.pdf) | Order paid partly with wallet credit — totals show **Applied balance** and **To be paid** (inherited from the invoice) alongside **Amount paid**. |
| [5_no_billing_details.pdf](https://github.com/user-attachments/files/27199190/5_no_billing_details.pdf) | Order with no `billing_name` / `billing_address` — Bill-to falls back to just the customer email. |

## Test plan
- [x] `uv run task lint` — passes (Ruff reformatted two test files)
- [x] `uv run task lint_types` — passes
- [x] `uv run pytest tests/receipt/` — 19 tests pass (generator: PDF magic bytes, fully-refunded property edges, heading items in both branches, totals-row presence/absence per state, payment-method formatting; render: subprocess success / failure / timeout)
- [ ] Eyeball the 5 sample PDFs to confirm layout matches the table above

### DIFF
--- a/server/polar/receipt/generator.py
+++ b/server/polar/receipt/generator.py
@@ -1,0 +1,351 @@
+from datetime import datetime
+from typing import TYPE_CHECKING, Any, Self
+
+from fpdf.enums import Align, TableBordersLayout, XPos, YPos
+from fpdf.fonts import FontFace
+from pydantic import BaseModel, Field
+
+from polar.config import settings
+from polar.invoice.generator import (
+    Invoice,
+    InvoiceGenerator,
+    InvoiceHeadingItem,
+    InvoiceItem,
+    InvoiceTotalsItem,
+    format_date,
+)
+from polar.kit.address import Address
+from polar.kit.currency import format_currency
+from polar.kit.utils import utc_now
+
+if TYPE_CHECKING:
+    from polar.models import Order, Payment, Refund
+
+
+class ReceiptRefund(BaseModel):
+    date: datetime
+    amount: int
+    tax_amount: int
+
+
+class ReceiptPayment(BaseModel):
+    date: datetime
+    method: str
+    method_metadata: dict[str, Any] = Field(default_factory=dict)
+    amount: int
+    currency: str
+    processor_id: str | None = None
+
+    @property
+    def display_method(self) -> str:
+        """Format the payment method as 'Visa — 4242' style."""
+        brand = self.method_metadata.get("brand")
+        last4 = self.method_metadata.get("last4")
+        if self.method == "card" and brand and last4:
+            return f"{brand.title()} — {last4}"
+        if last4:
+            return f"{self.method.title()} — {last4}"
+        return self.method.replace("_", " ").title()
+
+
+class Receipt(Invoice):
+    # The whole point of receipts is they don't gate on billing details —
+    # widen the parent's required fields. The generator falls back to the
+    # customer email when these are missing.
+    customer_name: str | None = None  # type: ignore[assignment]
+    customer_address: Address | None = None  # type: ignore[assignment]
+
+    invoice_number: str | None = None
+    paid_at: datetime | None = None
+    customer_email: str | None = None
+    payments: list[ReceiptPayment] = []
+    refunds: list[ReceiptRefund] = []
+    rendered_at: datetime = Field(default_factory=utc_now)
+
+    @property
+    def fully_refunded(self) -> bool:
+        order_total = self.net_amount + self.tax_amount
+        if order_total <= 0:
+            return False
+        return self.total_refunded >= order_total
+
+    @property
+    def heading_items(self) -> list[InvoiceHeadingItem]:
+        items: list[InvoiceHeadingItem] = []
+        if self.invoice_number is not None:
+            items.append(
+                InvoiceHeadingItem(label="Invoice number", value=self.invoice_number)
+            )
+        items.append(InvoiceHeadingItem(label="Receipt number", value=self.number))
+        if self.paid_at is not None:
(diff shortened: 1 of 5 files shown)


server/polar/receipt/generator.py
```diff
@@ -0,0 +1,351 @@
...
+            return f"{self.method.title()} — {last4}"
+        return self.method.replace("_", " ").title()
+
+
+class Receipt(Invoice):
+    # The whole point of receipts is they don't gate on billing details —
+    # widen the parent's required fields. The generator falls back to the
+    # customer email when these are missing.
+    customer_name: str | None = None  # type: ignore[assignment]
+    customer_address: Address | None = None  # type: ignore[assignment]
```
frankie567: For simplicity since we're only working from Orders right now, I think we can avoid to override that and consider we have it in any case

server/polar/receipt/generator.py
```diff
@@ -0,0 +1,351 @@
...
+    """Generate a receipt PDF, mirroring the invoice layout with payment history
+    and an optional Refunds section."""
+
+    refund_stamp_color: tuple[int, int, int] = (192, 57, 43)
+
+    data: Receipt
+
+    def header(self) -> None:
+        super().header()
+        if self.page_no() == 1 and self.data.fully_refunded:
```
frankie567: This feels like a darling to me, the information is already present in the body. Besides, it's a bit shifted:

Image

server/polar/receipt/generator.py
```diff
@@ -0,0 +1,351 @@
...
+                InvoiceTotalsItem(
+                    label="Amount refunded",
+                    amount=-self.total_refunded,
+                    currency=self.currency,
+                )
+            )
+        return items
+
+    @classmethod
+    def build_for_order(
```
frankie567: `from_order`

server/polar/receipt/render.py
```diff
@@ -0,0 +1,74 @@
...
+    if process.returncode != 0:
+        error = stderr.decode("utf-8").strip() or "unknown receipt renderer error"
+        raise ReceiptRenderError(f"Receipt renderer failed: {error}")
+    return stdout
+
+
+def main() -> int:
+    try:
+        payload = ReceiptRenderRequest.model_validate_json(sys.stdin.buffer.read())
+        generator = ReceiptGenerator(payload.receipt, heading_title="RECEIPT")
```
frankie567: `Receipt` (not all capitalized), for consistency with invoices

server/polar/receipt/generator.py
```diff
@@ -0,0 +1,351 @@
...
+        self.cell(self.epw / 3, 10, f"Page {self.page_no()} of {{nb}}", align=Align.R)
+
+    def generate(self) -> None:
+        super().generate()
+        if self.data.payments:
+            self._render_payment_history_section()
+        if self.data.refunds:
+            self._render_refunds_section()
+
+    def _render_customer_block(self) -> float:
```
frankie567: I don't think we need to overload that. If we work from an Order, we'll always have at least the Address country and the email address.

frankie567 (changes requested):

frankie567 (approved):

PR #11308, opened 2026-04-29
openapi: Use python tzdata for timezones

Yopi:
Instead of relying on the operating system timezones.

This will help us decrease spread when someone in the team upgrades their OS, but also between local and CI environments.


frankie567 (changes requested):
Good catch, I think then we can remvove the following in the Docker build:

https://github.com/polarsource/polar/blob/8d6159f4ac8302b06f14333fc2cc932aab37d5cf/server/Dockerfile#L71-L72

frankie567 (approved):

PR #11313, opened 2026-04-29
invoices: CJK font rendering in browser reader

Yopi:
Currently it the PDFs render glyphs correctly in PDF readers that support it (e.g. Preview). With this change it should also work in other readers such as PDF.js in Firefox

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -44,7 +44,40 @@ RUN --mount=type=secret,id=IPINFO_ACCESS_TOKEN \
   curl -fsSL "https://ipinfo.io/data/free/country_asn.mmdb?token=${TOKEN}" -o /data/country_asn.mmdb && \
   echo "${EXPECTED_SHA256}  /data/country_asn.mmdb" | sha256sum -c -
 
-# Stage 4: Final production image
+# Stage 4: Download per-script Noto Sans CJK .ttf fonts with checksum validation.
+# URLs are pinned to versioned (v39/v40/v56) gstatic.com paths returned by
+# https://fonts.googleapis.com/css2; those paths are immutable for a given
+# version. We use TrueType (glyf) builds because PDF.js (Firefox) does not
+# render glyphs from the CFF/OTF subsets that ship in fonts-noto-cjk.
+FROM alpine:3.21 AS download-cjk-fonts
+
+RUN apk add --no-cache curl
+
+RUN <<EOF
+set -eu
+mkdir /fonts
+cd /fonts
+curl -fsSLo NotoSansTC-Regular.ttf 'https://fonts.gstatic.com/s/notosanstc/v39/-nFuOG829Oofr2wohFbTp9ifNAn722rq0MXz76Cy_Co.ttf'
+curl -fsSLo NotoSansTC-Bold.ttf    'https://fonts.gstatic.com/s/notosanstc/v39/-nFuOG829Oofr2wohFbTp9ifNAn722rq0MXz70e1_Co.ttf'
+curl -fsSLo NotoSansSC-Regular.ttf 'https://fonts.gstatic.com/s/notosanssc/v40/k3kCo84MPvpLmixcA63oeAL7Iqp5IZJF9bmaG9_FnYw.ttf'
+curl -fsSLo NotoSansSC-Bold.ttf    'https://fonts.gstatic.com/s/notosanssc/v40/k3kCo84MPvpLmixcA63oeAL7Iqp5IZJF9bmaGzjCnYw.ttf'
+curl -fsSLo NotoSansJP-Regular.ttf 'https://fonts.gstatic.com/s/notosansjp/v56/-F6jfjtqLzI2JPCgQBnw7HFyzSD-AsregP8VFBEj75s.ttf'
+curl -fsSLo NotoSansJP-Bold.ttf    'https://fonts.gstatic.com/s/notosansjp/v56/-F6jfjtqLzI2JPCgQBnw7HFyzSD-AsregP8VFPYk75s.ttf'
+curl -fsSLo NotoSansKR-Regular.ttf 'https://fonts.gstatic.com/s/notosanskr/v39/PbyxFmXiEBPT4ITbgNA5Cgms3VYcOA-vvnIzzuoyeLQ.ttf'
+curl -fsSLo NotoSansKR-Bold.ttf    'https://fonts.gstatic.com/s/notosanskr/v39/PbyxFmXiEBPT4ITbgNA5Cgms3VYcOA-vvnIzzg01eLQ.ttf'
+sha256sum -c <<CHECKSUMS
+619662a0583f38311e92666927e5edbfd30f2a1fbe8593685660bd11bdd46a10  NotoSansTC-Regular.ttf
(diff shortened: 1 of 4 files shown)


server/polar/invoice/generator.py
```diff
@@ -319,13 +357,18 @@ def __init__(
...
-            if cjk_bold_font_file is not None:
-                self.add_font(self.cjk_font_name, fname=cjk_bold_font_file, style="B")
-            fallback_fonts.append(self.cjk_font_name)
+        if self.has_cjk_fallback_fonts():
+            for script, (regular, bold) in self.cjk_script_font_files.items():
+                family = self.cjk_font_name_for_script(script)
+                self.add_font(family, fname=regular)
+                self.add_font(family, fname=bold, style="B")
+
+            customer_script = self.resolve_cjk_script(data.customer_address.country)
```
frankie567: So we rely on the customer's country to know if we embed alternative scripts, right?

I'm wondering then if we could hit a case where the product bought has CJK characters, but the customer is actually from the US. It looks to me the previous version was "safer" since we always embedded them, right?

server/Dockerfile
```diff
@@ -44,7 +44,40 @@ RUN --mount=type=secret,id=IPINFO_ACCESS_TOKEN \
   curl -fsSL "https://ipinfo.io/data/free/country_asn.mmdb?token=${TOKEN}" -o /data/country_asn.mmdb && \
   echo "${EXPECTED_SHA256}  /data/country_asn.mmdb" | sha256sum -c -
 
-# Stage 4: Final production image
+# Stage 4: Download per-script Noto Sans CJK .ttf fonts with checksum validation.
+# URLs are pinned to versioned (v39/v40/v56) gstatic.com paths returned by
+# https://fonts.googleapis.com/css2; those paths are immutable for a given
+# version. We use TrueType (glyf) builds because PDF.js (Firefox) does not
+# render glyphs from the CFF/OTF subsets that ship in fonts-noto-cjk.
```
frankie567, line 51: So, using TTF is the actual fix for the problem at hand? Or does the logic change above has to do with it as well?

server/polar/invoice/generator.py
Yopi: Yes, I agree it's not great because it will most likely hit a corner case somewhere

server/Dockerfile
Yopi: So I'm not an expert on fonts here, but as I understand it:
Since CJK fonts are big, you can use subfonts to only load a part of the font. What we have in main today defaults to the first subfont.

The first commit in this PR attempted to do something similar to the third commit, and load the correct subfont based on the customer address.

The third commit fully separates the sub fonts into font files and decides which font to use based on the address.

I generated these 4 alternative PDFs, and I think that answers your question that yes. TTF is the actual fix
[cjk_invoice_otf.pdf](https://github.com/user-attachments/files/27203663/cjk_invoice_otf.pdf)
[cjk_invoice_ttc_collection.pdf](https://github.com/user-attachments/files/27203665/cjk_invoice_ttc_collection.pdf)
[cjk_invoice_ttc.pdf](https://github.com/user-attachments/files/27203667/cjk_invoice_ttc.pdf)
[cjk_invoice_ttf.pdf](https://github.com/user-attachments/files/27203668/cjk_invoice_ttf.pdf)

Let me go back to the drawing board for a bit

frankie567: Thanks :) Side note: I noticed in your samples the digits on the line items are slightly different than totals:
Capture d’écran 2026-04-29 à 15 57 53

Yopi: New examples, fixed the wrong font for the amounts as well

[01_jp_customer.pdf](https://github.com/user-attachments/files/27235388/01_jp_customer.pdf)
[02_kr_customer.pdf](https://github.com/user-attachments/files/27235390/02_kr_customer.pdf)
[03_sc_customer.pdf](https://github.com/user-attachments/files/27235391/03_sc_customer.pdf)
[04_tc_customer.pdf](https://github.com/user-attachments/files/27235392/04_tc_customer.pdf)
[05_jp_in_tw_edge_case.pdf](https://github.com/user-attachments/files/27235393/05_jp_in_tw_edge_case.pdf)
[06_tw_country_fallback.pdf](https://github.com/user-attachments/files/27235394/06_tw_country_fallback.pdf)
[07_ascii_no_cjk.pdf](https://github.com/user-attachments/files/27235395/07_ascii_no_cjk.pdf)
[08_mixed_cjk.pdf](https://github.com/user-attachments/files/27235396/08_mixed_cjk.pdf)

server/polar/invoice/generator.py
```diff
@@ -195,6 +196,7 @@ def from_order(cls, order: "Order") -> Self:
             customer_name=order.billing_name,
             customer_additional_info=order.tax_id[0] if order.tax_id else None,
             customer_address=order.billing_address,
+            customer_locale=order.customer.locale,
```
frankie567, line 224: We might need to port that to `receipt/generator.py` as well

server/polar/invoice/generator.py
```diff
@@ -319,13 +376,20 @@ def __init__(
...
+
+            customer_script = self.cjk_script_from_locale(
+                data.customer_locale
+            ) or self.resolve_cjk_script(data.customer_address.country)
+            customer_family = self.cjk_font_name_for_script(customer_script)
+            fallback_fonts.append(customer_family)
+            for script in self.cjk_script_font_files:
+                if script != customer_script:
+                    fallback_fonts.append(self.cjk_font_name_for_script(script))
         self.set_fallback_fonts(fallback_fonts, exact_match=False)
```
frankie567, line 413: Had a hard time parsing this :D Could we add some comments... Something like:

server/polar/invoice/generator.py
```diff
@@ -319,13 +376,20 @@ def __init__(
...
         fallback_fonts = [self.hebrew_font_name, self.arabic_font_name]
-        cjk_regular_font_file = self.resolve_font_file(self.cjk_regular_font_files)
-        if cjk_regular_font_file is not None:
-            self.add_font(self.cjk_font_name, fname=cjk_regular_font_file)
-            cjk_bold_font_file = self.resolve_font_file(self.cjk_bold_font_files)
-            if cjk_bold_font_file is not None:
-                self.add_font(self.cjk_font_name, fname=cjk_bold_font_file, style="B")
-            fallback_fonts.append(self.cjk_font_name)
+        if self.has_cjk_fallback_fonts():
+            for script, (regular, bold) in self.cjk_script_font_files.items():
```
frankie567: `# Load fonts into context`

server/polar/invoice/generator.py
```diff
@@ -319,13 +376,20 @@ def __init__(
...
+                family = self.cjk_font_name_for_script(script)
+                self.add_font(family, fname=regular)
+                self.add_font(family, fname=bold, style="B")
+
+            customer_script = self.cjk_script_from_locale(
```
frankie567: `# Set fallback fonts by priority: customer locale/country first, and then all other scripts.`

server/polar/invoice/generator.py
Yopi: Definitely!

server/polar/invoice/generator.py
Yopi: Good idea

frankie567 (approved):
太棒了，做得好！

PR #11324, opened 2026-04-29
feat(receipts): add receipt download endpoints

psincraian:
## Summary

PR5 of the receipts feature ([design doc](https://handbook.polar.sh/engineering/design-documents/receipts)). Adds the GET endpoints that hand back a presigned URL to download an order's receipt PDF, the `receipt_number` field on the `Order` schema, and the receipt link on the backoffice order detail page.

- `GET /v1/orders/{id}/receipt` — auth: `orders_read`
- `GET /v1/customer-portal/orders/{id}/receipt` — auth: customer portal session
- Backoffice: receipt number + PDF link on the order detail view

## Behavior

| Condition | Response |
|---|---|
| `order.receipt_path` set | `200 {url}` (presigned S3 URL, 10-minute TTL) |
| `order.receipt_number` set, `order.receipt_path` not set | `202` (no body); a `receipt.render` job is enqueued |
| `order.receipt_number` not set, order missing, or caller unauthorized | `404` |

The actor that consumes `receipt.render` is in a separate PR — it uses a Redis lock keyed by `order_id` to dedup repeated enqueues from polling clients.

## How

Receipt resolution mirrors the invoice flow:

- Endpoints call `order_service.get_order_receipt` / `customer_order_service.get_order_receipt`.
- Those delegate to `receipt_service.get_pdf_url_or_status`, which either generates a presigned URL or enqueues a render and returns `None` (the 202 signal).
- A new `ReceiptNotAllocated` exception (404) is raised when `receipt_number` is unset — matches the existing `InvoiceDoesNotExist` shape.

Other surface:
- Adds `S3_CUSTOMER_RECEIPTS_BUCKET_NAME` setting (default `polar-customer-receipts`).
- Adds `Order.receipt_filename` property (mirrors `invoice_filename`).
- Adds `receipt_number: str | None` to the public `Order` schema; the frontend uses this to gate the download button.

## Why safe to merge before the render pipeline

In production today, every order has `receipt_number = NULL`:

- The allocation step (PR3) is gated on `Organization.receipts_enabled`, which defaults to `false`.
- No org has the flag enabled yet — that's an operational step run per-cohort after a backfill.

So all three endpoints return `404`, the customer portal button stays hidden (`order.receipt_number != null` is the gate), and the backoffice column shows "Not generated". Once the render pipeline ships and the operator flips the flag, the same endpoints start returning `202` then `200`.

## Test plan

- [x] `uv run task lint` and `uv run task lint_types` pass
- [x] Unit and integration tests cover 401 / 404 / 202 / 200 for both order endpoints, the schema field appears on the `Order` resource, and the receipt service returns `None` when `receipt_path` is null
- [x] Manual end-to-end against the local dev stack: 401 on anonymous, 404 with no `receipt_number`, 202 after setting `receipt_number` (verified `receipt.render` job is enqueued), 200 with a presigned URL after setting `receipt_path`

### DIFF
--- a/server/polar/receipt/service.py
+++ b/server/polar/receipt/service.py
@@ -1,4 +1,5 @@
 import asyncio
+from datetime import datetime
 
 from polar.config import settings
 from polar.customer.repository import CustomerRepository
@@ -10,6 +11,7 @@ from polar.models import Order
 from polar.order.repository import OrderRepository
 from polar.payment.repository import PaymentRepository
 from polar.refund.repository import RefundRepository
+from polar.worker import enqueue_job
 
 from .generator import Receipt
 from .render import render_receipt_pdf
@@ -107,6 +109,18 @@ class ReceiptService:
                 order, update_dict={"receipt_path": key}
             )
             return order
+
+    async def get_pdf_url_or_status(self, order: Order) -> tuple[str, datetime] | None:
+        if order.receipt_path is None:
+            enqueue_job("receipt.render", order_id=order.id)
+            return None
+
+        s3 = S3Service(settings.S3_CUSTOMER_RECEIPTS_BUCKET_NAME)
+        return s3.generate_presigned_download_url(
+            path=order.receipt_path,
+            filename=order.receipt_filename,
+            mime_type="application/pdf",
+        )
 
 
 receipt = ReceiptService()
--- a/server/polar/backoffice/orders/endpoints.py
+++ b/server/polar/backoffice/orders/endpoints.py
@@ -86,6 +86,29 @@ class InvoicePDFItem(description_list.DescriptionListItem[Order]):
         return None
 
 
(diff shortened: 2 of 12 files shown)


server/polar/receipt/service.py
```diff
@@ -108,5 +111,20 @@ async def generate_order_receipt(
...
+            return None
+
+        s3 = S3Service(settings.S3_CUSTOMER_RECEIPTS_BUCKET_NAME)
+        return s3.generate_presigned_download_url(
+            path=order.receipt_path,
+            filename=order.receipt_filename,
+            mime_type="application/pdf",
+        )
+
+    def enqueue_render(self, order_id: uuid.UUID) -> None:
```
frankie567: We probably don't need such a simpler wrapper :)

frankie567 (approved):

PR #11326, opened 2026-04-29
events: Move back to page based pagination

Yopi:
Switch out the show more / infinite pagination to the previous page based pagination. Since Tinybird can count faster than postgres we can go back.


frankie567 (approved):
DO you plan to remove it from the API as well?

Yopi:
> DO you plan to remove it from the API as well?

Yes 😄. Since it was added as hidden, I can remove it after this given that it all works fine

PR #11327, opened 2026-04-29
server/order: check minimum payable amount per currency instead of hard-coded 50

frankie567:
- server/order: check minimum payable amount per currency instead of hard-coded 50
- server: refactor currency minimums

- Unified functions to check minimums and maximums per currency
- Fix a case where we didn't check the right currency minimum when generating a subscription order, which could happen with very small prorations


pieterbeulque:
@frankie567 Is the core issue here that the USD-converted amount should also always be more than $0.50?

frankie567:
> @frankie567 Is the core issue here that the USD-converted amount should also always be more than $0.50?

Exactly. We add enforcements for that when creating a price, but we missed it when trying to pay a too small order, which could happen with very small prorations.

pieterbeulque (approved):
Just wondering if minimum price per currency should then be adapted for most currencies. For example, $0.50 AUD is $0.30 USD?

frankie567:
Yes, good point. I should check if they all convert to at least 0.50 USD 🤔

frankie567:
@pieterbeulque I've completed the minimums for all the currencies so they at least convert to >0.50 USD

PR #11328, opened 2026-04-29
webhook: Speed up deliveries endpoint

Yopi:
By splitting the join into two separate queries we can avoid doing the join on every row and just query filter out the deliveries corresponding to the correct webhook endpoint.


frankie567 (approved):
Always a surprise to me that two queries are faster than a single one, even with an expensive JOIN 😄

PR #11332, opened 2026-04-29
Improve refund merchant experience

pieterbeulque:
Related issue: https://github.com/polarsource/feedback/issues/10

### DIFF
--- a/server/polar/order/schemas.py
+++ b/server/polar/order/schemas.py
@@ -130,6 +130,28 @@ class OrderBase(TimestampedSchema, IDSchema):
     def from_balance_amount(self) -> SkipJsonSchema[int]:
         return self.applied_balance_amount
 
+    @computed_field(
+        description=(
+            "Amount in cents that can still be refunded (net, before taxes). "
+            "Accounts for any applied customer balance and previous refunds."
+        ),
+        examples=[9000],
+    )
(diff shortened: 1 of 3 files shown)


server/polar/order/schemas.py
```diff
@@ -130,6 +130,28 @@ def amount(self) -> SkipJsonSchema[int]:
     def from_balance_amount(self) -> SkipJsonSchema[int]:
         return self.applied_balance_amount
 
+    @computed_field(
```
frankie567, line 133: Both of those fields are available as a property on the model itself, so we can directly define them as regular field without repeating the logic.

frankie567 (changes requested):
Great move! That looks super nice! Just a small implementation detail to fix :)

frankie567 (approved):

server/polar/order/schemas.py
frankie567: @stilla-bot What I suggest should be safe since we don't re-validate anymore saved webhook payload. That said, there are lot of remnants (like AliasChoices) across the codebase from a period where we did that.

It's a separate effort to simplify that, so please open an issue about it. Ideally, look for the places where there are those specific workarounds for stored webhook payloads. Assign it to me.

PR #11338, opened 2026-04-30
polar: Make customer backfill handle large amount of existing customers

Yopi:
Otherwise the script crashes


frankie567 (approved):
That works!

Tip: you can also leverage the `.stream` method from the repository to iterate on large batches while saving your memory.

PR #11339, opened 2026-04-30
members: Split ownership transfer into two steps

Yopi:
Downgrade first and then upgrade, instead of swapping.

In a large table the planner picks an Index Scan on `members_pkey` for the bulk UPDATE used during ownership transfer, visiting rows in id-ASC order. If the new owner's id sorts before the current owner's, the new row is updated to `role='owner'` before the current owner is demoted, tripping the partial unique index.

In small tables the planner picks Seq/Bitmap scan (heap order) and hides the bug. We disable seq/bitmap scans in the test to force the prod plan.

### DIFF
--- a/server/polar/member/repository.py
+++ b/server/polar/member/repository.py
@@ -1,7 +1,7 @@
 from collections.abc import Sequence
 from uuid import UUID
 
-from sqlalchemy import Select, case, func, select, update
+from sqlalchemy import Select, func, select, update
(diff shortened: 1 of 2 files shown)


pieterbeulque (approved):

psincraian (approved):
This is really interesting! Thanks for digging

frankie567 (approved):
Well, TIL! Great job!

PR #11340, opened 2026-04-30
openapi: Fail action if there is a diff

Yopi:
I think we've shown that it works now, but it also seems like we're ignoring it 😄


psincraian (approved):
The guy that forgot that needs to get fired! 👀

pieterbeulque (approved):
Ah, that's why things were still showing up!

PR #11349, opened 2026-04-30
feat(receipts): add receipt download button to customer portal

psincraian:
## Summary

Adds the customer-portal Download Receipt button.

**Server**: emits `order.receipt_generated` on the customer + org SSE channels after `receipt_path` is updated, mirroring `order.invoice_generated`.

**Frontend**: `DownloadReceipt` component on click — initial GET, then either:
- `200` → open the presigned URL,
- `202` → wait for the SSE event (30s timeout), then re-fetch and open,
- `4xx`/`5xx` → Try again / Contact support fallback.

The SSE listener is per-click (not mount-time like invoice's). Receipts also re-render on refunds; a mount-time listener would auto-open tabs the customer didn't ask for.

**Layout**: `OrderDownloadActions` toolbar:
- Desktop: receipt button + kebab dropdown (Download / Edit Invoice), right-aligned, sized to content.
- Mobile: full-width stacked buttons.

Driven by `useIsMobile` so exactly one `DownloadInvoicePortal` mounts per breakpoint — avoids duplicate SSE subscriptions on `order.invoice_generated` that would otherwise open two browser tabs.

Gated on `order.receipt_number != null`. Today every prod order has `receipt_number = NULL`, so the button stays hidden until per-org backfill + flag flip.

## Test plan

- [x] `pnpm lint`, `pnpm typecheck`, `uv run task lint`, `uv run task lint_types` pass
- [x] e2e on a real Stripe-paid order: 202 → `receipt.render` worker → `eventstream_publish` → SSE event delivered to the customer channel → frontend re-fetches → 200 with valid PDF
- [x] Negative case: `receipt_number = null` → button hidden, existing invoice flow unchanged
- [x] Visual states (idle / loading / failed) on desktop and mobile
- [x] New `test_publishes_eventstream` asserts the publish call shape

### DIFF
--- a/clients/apps/web/src/components/CustomerPortal/CustomerPortalOrder.tsx
+++ b/clients/apps/web/src/components/CustomerPortal/CustomerPortalOrder.tsx
@@ -8,7 +8,7 @@ import { Status } from '@polar-sh/ui/components/atoms/Status'
 import { ThemingPresetProps } from '@polar-sh/ui/hooks/theming'
 import { useMemo, useState } from 'react'
 import { twMerge } from 'tailwind-merge'
-import { DownloadInvoicePortal } from '../Orders/DownloadInvoice'
+import { OrderDownloadActions } from '../Orders/OrderDownloadActions'
 import { DetailRow } from '../Shared/DetailRow'
 import { CustomerPortalGrants } from './CustomerPortalGrants'
 import { OrderPaymentRetryModal } from './OrderPaymentRetryModal'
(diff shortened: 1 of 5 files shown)


pieterbeulque (commented):
Is it weird that we use SSE for invoice_generated but we use polling for the receipt?

psincraian:
Yeah, I saw that but I didn't want to overengineer at the beggining.  I think pollling in those cases is more than enough. But, I agree that's nicer to use the same way in both placer

frankie567 (approved):

PR #11350, opened 2026-04-30
Add RFC for RBAC structure and rollout plan

joebon:


handbook/engineering/design-documents/rbac.mdx
```diff
@@ -0,0 +1,180 @@
...
+## Summary
+
+Introduce role-based access control for users on a Polar organization. Two roles ship in this iteration — `admin` and `member` — replacing today's implicit binary where the lone "admin" capability (Finance pages and payouts) is gated by whoever happens to own the payout account at Stripe.
+
+The change is mostly additive on top of the existing two-layer authorization architecture. Role lives on the user–organization membership and is consulted by policies. Scopes remain a per-token capability filter and do not encode role.
+
+## Goals
+
+- Introduce an `OrganizationRole` (`admin`, `member`) attached to organization membership.
+- Decouple "admin" from payout-account ownership. The payout-account owner remains a real, separate concept (legal owner of the Stripe account); it stops being read as an authorization signal.
```
psincraian: I have some doubts here. What happens if the admin of a current org, the admin of the payout account, is removed as a member?

I think we don't allow that now but with this we will allow it and we will have an inconsistent state.

Maybe @frankie567 can provide his vision.

joebon: Yeah, that's a good catch @psincraian. Perhaps that's an extra business rule then: a user who is the admin/owner to the payout account connected to an org is not allowed to be removed from the organization. Or even to be demoted from an `admin`?

psincraian: It's one solution. But then, this validation should also be when adding a new payout account or when switching payout accounts for an org. In the last case, both members should be admins. 

But this can get a bit more complex, as the same payout account can be shared across different orgs 🙈

joebon: Yeah. I'm currently thinking that perhaps we should instead:
- Rename the current organization concept/"role" admin to `Owner` and define it as the owner of the payout account. Then consequently
  - there can and must be one and only one `Owner` for a given organization
  - switching `Owner` for an org remains a support/backoffice task, at least until we've figured out a self-serve flow for that
- Introduce RBAC including two new explicit roles: `admin` and `member`.
  - An `Owner` can do everything that `admins` can do
  
What do you think @pieterbeulque ?

handbook/engineering/design-documents/rbac.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+Remove the legacy `is_admin` field from the API. Stop maintaining the role-side-effect on payout-account-owner changes; payout-account ownership and Polar role become independent concerns from this point.
+
+The dashboard auto-deploys on every merge but does not currently force-refresh stale tabs, so a tab from phase 3 keeps consuming `is_admin` until the user reloads. Two acceptable orderings for phase 5:
+
+1. **Time-based** — wait long enough after phase 4 that weekly-active sessions have refreshed at least once. Cheapest path; small residual risk for very long-lived idle tabs.
+2. **Bundle-mismatch prompt** — ship a "new version available — please refresh" prompt as part of phase 4. Phase 5 can then ship immediately because any stale tab self-heals.
+3. **Accept the break** — ship phase 5 immediately after phase 4 and accept that any dashboard tab still on the previous bundle will see errors until the user reloads. Cheapest in engineering effort; the cost is a brief degraded experience for users with long-lived tabs.
+
+Default to (3) — the user-visible cost is small and short-lived, and it avoids both the wait and the engineering investment in a bundle-mismatch prompt.
```
frankie567, line 237: Agree with 3, don't need to overthink that too much IMO. I think we introduce this kind of breaking changes every week without even thinking about it :D

handbook/engineering/design-documents/rbac.mdx
frankie567: I think that's the main question to solve here. The way I see it:

- The owner of the payout account (currently `admin_id` should be definitely kept. Only the owner of the payout account can read it from `/v1/payout-accounts/` and access the Stripe dashboard.
- An Organization admin should be able to read transactions, read and create payouts, even if they're not the owner of the payout account

The two main questions are:

- Who should be able to _switch_ the Payout Account of an Organization? Probably admins should be able to, with the limitation they can only assign a Payout Account they own (since they can't read others)
- What happens if the owner of a Payout Account is removed from an Organization? Maybe the simplest solution is to forbid it: if we detect that the current payout account belongs to the member we try to remove, simply reject the operation and prompt to switch Payout Account first.

Switching Payout Account owner would still possible but via a support request + backoffice.

joebon: Yeah, I think I agree with all @frankie567 !

Also need to sanity check the analogy for `Account.admin_id`... 🥁

pieterbeulque: Yes, if you're coming at it from a legal / KYC / KYB angle, the payout account owner is the beneficiary of the money earned; Ultimate Beneficial Owner in legal terms.

I agree this should be a special role and cannot be removed from an organization. Changing it, I'm fine with that being a support task until we can't handle the support burden anymore.

pieterbeulque: Alternatively, if an admin goes through KYC and connects a payout account, the current owner (UBO) could possibly be allowed to assign a new admin as the new owner from the UI, that'd be great and it solves for the most common use case, I believe.

joebon: Update with clarification that we're switching from currently checking `Account.admin_id` to checking the role of a user. Sorry for introducing the concept confusion here

frankie567: > Also need to sanity check the analogy for Account.admin_id... 🥁

`Account.admin_id` should definitely be removed; `Account` is now only a utility entity that'll always live under an Organization, so it shouldn't be treated specifically.

joebon: It doesn't look trivial to drop `Account.admin_id` entirely though @frankie567 , but I'd be happy to be enlightened!
Who can be assigned an `Account.admin` looks like it's guarded by `IdentityVerificationStatus.verified` which is backed by a Stripe KYC if I'm not mistaken. Should we limit who can become admins to users with `IdentityVerificationStatus.verified` then instead?

Furthermore such a change would lead to changes in:
- recipients of notification emails (verification status updates etc) - should we send emails to all admins then instead?
- **Polar for Polar** - we're populating `MemberRole.owner` based on `Account.admin_id`. How do we handle that change?

frankie567: > Who can be assigned an `Account.admin` looks like it's guarded by `IdentityVerificationStatus.verified` which is backed by a Stripe KYC if I'm not mistaken. Should we limit who can become admins to users with `IdentityVerificationStatus.verified` then instead?

Well, admin change process will be completely replaced by this role management. Identity verification lives on the user, so we may require it quite easily yes.

> * recipients of notification emails (verification status updates etc) - should we send emails to all admins then instead?

Not sure which emails? 

> * **Polar for Polar** - we're populating `MemberRole.owner` based on `Account.admin_id`. How do we handle that change?

Then we probably need an Owner role as well :)

joebon: Sorry, I dug deeper on the emails and it seems it's only the "Your review is complete" email that's currently sent to account admins today

handbook/engineering/design-documents/rbac.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+## Out of Scope (Follow-Ups)
+
+### Invitation / accept-link flow
+
+Today, inviting a user adds them to the organization immediately. A proper invitation flow (email accept-link, expiring tokens) is a separate, additive feature that does not depend on this design.
```
pieterbeulque, line 253: Should I move this back into our regular planning, or are you planning to do this as a natural follow-up? It doesn't feel super critical, it's just weird to not have it.

handbook/engineering/design-documents/rbac.mdx
joebon: But yeah, given **KYC**/**KYB** and **UBO** it seems to me that we need to retain a singular `Owner`, at least short term.

handbook/engineering/design-documents/rbac.mdx
joebon: I think it makes total sense to do/have it, I've just suggested to break it out from iteration #1. So a natural next step according to me once the plan according to this RFC is acted on

pieterbeulque (approved):
Looks good from a product angle!

handbook/engineering/design-documents/rbac.mdx
```diff
@@ -0,0 +1,252 @@
...
+- **Membership** — the link between a user and an organization. Today it carries no role; this design adds one.
+- **`Account`** — Polar's internal financial container for an organization. Holds the credit balance, fee structure, billing identity (name, address) for invoices and receipts, and a single `admin_id`. One per org. Distinct from the Stripe-side payout account.
+- **`Account.admin_id`** — today, the singular authorization signal for org-level admin actions: member management, organization deletion, payout-account selection, Finance and payout reads/writes. This design moves all those gates onto role-permission checks. After the cutover, `Account.admin_id` is no longer consulted directly by policies; instead, it is the source of truth for who carries the `owner` role on their membership (the two are kept aligned via dual-write). Its real-world meaning is unchanged: legal/billing primary contact, Stripe-identity-verified, recipient of identity-verification notifications, owner-side of the Polar-for-Polar relationship.
```
frankie567: I still don't understand why we want to keep this. To me, `owner` should definitely replace the `Account.admin_id`. The fact that it holds authorization signal is a code legacy problem, but doesn't carry any KYC/KYB information.

IMO, once we have the Owner role in place, `Account.admin_id` should be removed so we have only once source of truth.

joebon: Thanks for your persistence on this!
This is/was just me being overly defensive - happy to include a step to remove `Account.admin_id`. I'll amend

joebon: @frankie567 now updated to also include removal of `Account.admin_id`

handbook/engineering/design-documents/rbac.mdx
```diff
@@ -0,0 +1,268 @@
...
+
+## Goals
+
+- Introduce an `OrganizationRole` enum (`owner`, `admin`, `member`) attached to organization membership.
+- Move org-level admin authorization off direct `Account.admin_id` checks and onto **role permissions** evaluated at policy time.
+- **Decommission `Account.admin_id`.** Once the role-based ownership has been validated in production, drop the column; the role table becomes the sole source of truth for ownership, the KYC/identity-verification gate moves onto role transitions into `owner`, and reads that previously joined `Account.admin_id` query the role instead.
+- Allow multiple admins per organization (alongside the singular `owner`).
+- Enforce that an organization always has at least one user with admin capability — i.e. `role ∈ {owner, admin}`. The singular `owner` always satisfies this.
+- Forbid removing the `owner` from the organization; ownership transfer remains a support/backoffice task in this iteration.
```
psincraian, line 22: Maybe it's implicit, but I think we should provide the tools on the backoffice to make this happen.

joebon: AFAIK this is already possible in backoffice, under the _Team_ tab on the org view. So the work will include migrating that flow to "change org/accout owner" i.e. switch from updating `Account.admin_id` to switching which user has the `owner` role.

psincraian (approved):
Then, I think this looks good! 

Nice job @joebon 👏

frankie567 (approved):
Perfect 🙌

PR #11257, opened 2026-04-28
feat: auto-decide organization appeals with the review agent

psincraian:
## Summary

Replaces the manual Plain ticket appeal flow with an AI auto-decision driven by the existing review agent.

When a denied org submits an appeal:
1. `submit_appeal` enqueues `organization_review.appeal_submitted` (no Plain ticket).
2. The new `review_appeal` actor runs the review agent with a new `ReviewContext.APPEAL` preamble. The merchant's appeal text and the original merchant-facing denial message are both included in the prompt so the AI can judge whether the appeal directly addresses the stated concern.
3. On `APPROVE` → `approve_appeal()` → `maybe_activate()` (payout/identity gates still apply).
4. On `DENY` → `deny_appeal()`. The merchant sees a non-final denial banner that points them at support — given that some share of denied appeals are overridden in support, the copy avoids overpromising finality.

## Frontend

- `AppealForm` polls every 3s while pending, with a 120s timeout fallback (so a stalled worker doesn't poll forever).
- Loader2 spinner inline next to the title while under review.
- Dropped the `prevReviewStatus` shadowing pattern in favor of derivation from `appealMutation.isSuccess` + server data.
- Removed a legacy banner short-circuit in `AccountPage.tsx` that was making the new copy unreachable once an appeal was rejected.

## Prompt scoping

Updated `analyzer.py` so prior denials of *other* orgs by the same admin are treated as a risk signal **only at the THRESHOLD review stage** — not at submission or appeal. A fresh organization deserves a fresh look.

## Screenshots

The four states of the appeal flow on the merchant dashboard:
- `01-denied.png` — initial denial with appeal CTA
- `02-under-review.png` — pending state with loading spinner
- `03-approved.png` — approved
- `04-denied.png` — rejected with non-final copy that points to support

_(uploading inline in this description)_

## Migration

- Denied orgs that **never appealed** pre-merge: get the new flow when they appeal — clean migration.
- Already-decided appeals: render correctly with the new banner copy.
- **In-flight appeals at deploy time** (Plain ticket exists, no decision yet): no AI task is enqueued retroactively. The existing `approve_appeal` / `deny_appeal` service methods are unchanged, so the backoffice / Plain flow still resolves them. The frontend pending copy ("AI is reviewing — about a minute") will be slightly inaccurate for these for up to 120s before falling back to "Appeal still processing".

## Test plan

- [x] Backend: `uv run task lint`, `uv run task lint_types`
[...]

PR #11354, opened 2026-04-30
refactor: remove dead Plain appeal-thread code

psincraian:
## Summary
- Remove `PlainService.create_appeal_review_thread` and the `NoUserFoundError` class — both dead code with no callers.
- Drop the now-unused `OrganizationReview` import in the Plain service.

Organization appeals are auto-decided by the AI review agent (#11257), which approves or denies in place without creating a Plain ticket. The thread-creation method was orphaned by that change and can be safely removed.

## Test plan
- [x] `uv run task lint` passes
- [x] `uv run task lint_types` passes
- [ ] No production behavior change expected (method had no callers)

### DIFF
--- a/server/polar/integrations/plain/service.py
+++ b/server/polar/integrations/plain/service.py
@@ -51,7 +51,6 @@ from polar.models import (
     User,
     UserOrganization,
 )
-from polar.models.organization_review import OrganizationReview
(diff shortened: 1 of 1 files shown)


frankie567 (approved):
Only deleted lines, immediate approval 😂

PR #11355, opened 2026-04-30
feat(receipts): add backfill script for historical orders

psincraian:
## Summary

Adds a one-shot operator-invoked script that retroactively allocates `receipt_number` for paid orders that pre-date the receipts feature, then flips `Organization.feature_settings.receipts_enabled` once allocation is complete. Closes the operational gap for the historical-orders cohort rollout.

The flag flips last per organization, so customers don't see a Download Receipt button for any order until every historical order in that org has a number — no mid-rollout 404 window.

## Changes

- **`scripts/backfill_receipts.py`** (new) — typer CLI with `--all`, `--execute`, `--batch-size`, `--sleep-seconds`. Default is dry-run (matches peer-script convention). Single-org mode takes a UUID argument; `--all` iterates every candidate org. Per-batch commit, configurable inter-batch sleep, per-org session isolation so a failure on one org doesn't poison the rest.
- **`OrderRepository`** — adds `list_unallocated_paid_for_organization` (composite-keyset paginated, ordered by `(customer_id, id)` so the bulk allocator clusters customers per batch), `list_organizations_with_unallocated_paid_orders` (sorted by `total_balance` ascending — small balance first), `bulk_set_receipt_numbers` (single `UPDATE … FROM (VALUES …)` per customer-batch). Plus a private `_has_succeeded_payment()` helper extracting the EXISTS-on-payments semi-join used by the eligibility predicates — this is what excludes free / fully-discounted paid-status orders that have no payment row.
- **`CustomerRepository.increment_receipt_next_number_by(count)`** — atomically reserves a contiguous range of `count` receipt numbers in one `UPDATE … RETURNING`. Mirrors the existing `increment_receipt_next_number` (+1 path) used at runtime.
- **`receipt_service.bulk_allocate_for_customer(session, customer, orders)`** — backfill-only allocation that bypasses the `is_receipts_enabled` gate and the per-row `SELECT FOR UPDATE` lock used by the runtime `allocate`. Safe while the flag is off (no concurrent allocator). Doesn't reflect new values back to in-memory ORM instances on purpose — would mark orders dirty and re-trigger per-row UPDATEs at commit time.

## Why bulk per-customer

Per-batch round-trips drop from `2 + 3·N` (one query per order × three queries) to `1 + 2·C` (page query + per-unique-customer counter UPDATE + per-unique-customer bulk orders UPDATE). The page query is sorted by `(customer_id, id)` so each customer's orders cluster within the same batch, minimizing `C`. The `UPDATE … FROM (VALUES …)` form (instead of a 500-branch `CASE`) keeps parsed SQL small and uses a hash join instead of per-row branch evaluation.

## Test plan

- [x] Unit tests for `bulk_allocate_for_customer` — sequential numbering, idempotency on already-allocated, empty-input
- [x] Existing runtime `allocate` tests still pass (no behavior change to the runtime path)
- [x] `uv run task lint` — clean
- [x] `uv run task lint_types` — clean
- [x] End-to-end smoke against local Docker stack: created 3 orgs with `total_balance` 100 / 10k / 10M, each with paid orders + a free order; `--all --execute` processed them in balance-ascending order, allocated 6 receipts, skipped the 3 free orders, flipped all 3 flags. Idempotent on re-run (0 candidates).

## Operational notes

- **Dry-run default** — `python -m scripts.backfill_receipts --all` previews without mutating; pass `--execute` to actually allocate.
- **Cohort rollout** — start single-org with `python -m scripts.backfill_receipts <org_id> --execute`. Monitor metrics for 24-48h before moving to `--all`.
- **Throttling** — `--sleep-seconds` (default `0.1`) eases DB load between batches. Set higher during peak hours; set `0` for fast offline runs.


frankie567 (commented):
That adds a lot of code in the actual repo/services. Will we need those outside the context of the backfill script? Worried that this could become dead code.

psincraian:
Agreed, let me move those to the script itself

psincraian:
Done in 85f11d174 — moved `bulk_allocate_for_customer`, the page query, the balance-ordered org listing, the bulk-update helper, and the per-customer counter increment all into `scripts/backfill_receipts.py`. The production repos/services are back to upstream main; only the runtime allocate path is touched (which already existed). Net PR diff is now just the new script file: +403 lines, 1 file.

frankie567 (approved):

PR #11357, opened 2026-04-30
server: remove tax_rate and taxability_reason columns

frankie567:
Final step of #11210


server/migrations/versions/2026-04-30-1203_remove_tax_rate_and_taxability_reason.py
```diff
@@ -0,0 +1,228 @@
...
+# revision identifiers, used by Alembic.
+revision = "8847acc26477"
+down_revision = "b07286a34be3"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    # Backfill tax_breakdown from tax_rate and taxability_reason before dropping columns
```
frankie567, line 24: Backfill added for extra safety and be nice with local dev environments, but I ran the backfill script in sandbox and prod, so it should be a quick no-op.

psincraian: the local dev it's a nice feature

psincraian (approved):

PR #11360, opened 2026-04-30
ci: run openapi client check on merge_group events

psincraian:
The `Check generated client` job is a required status check, but the workflow only triggered on `pull_request` events. Merge queue commits never produced a run, so the queue stalled waiting for a status that would never report.

- Add `merge_group:` to the workflow triggers.
- Fall back to `github.sha` for checkout when there is no PR head sha.
- Skip PR-only steps (manual base merge, PR comment) on `merge_group` runs.

Existing in-flight queue items won't retroactively pick up the new trigger — re-queue or emergency-merge them after this lands.


frankie567 (approved):

Yopi:
🙈 thanks @psincraian!

PR #11364, opened 2026-04-30
Add Ghana in sign-up country check too

pieterbeulque:
Another follow-up from #11233


frankie567 (commented):
Wouldn't it be less error prone to use the generated enum from the OpenAPI schema:

https://github.com/polarsource/polar/blob/b74143fa6be4b12365e0f642decf977987026d20/clients/packages/client/src/v1.ts#L53992-L53993

We only need to re-export it in `clients/packages/client/src/enums.ts` and import it from there.

pieterbeulque:
@frankie567 Yes, very much, thank you! PR updated

frankie567 (approved):

PR #11365, opened 2026-04-30
Add daily payout limit enforcement (once per 24 hours)

pieterbeulque:
## Summary

**Related Issue**: this replaces https://github.com/polarsource/feedback/issues/115

Implements a daily payout limit to prevent organizations from requesting multiple payouts within a 24-hour period.

## What

- Added `DailyPayoutLimitReached` exception class to handle cases where a payout is requested too soon after the previous one
- Added `get_latest_by_account()` method to `PayoutRepository` to retrieve the most recent payout for an account
- Modified `PayoutService.create()` to check if a payout was created within the last 24 hours and raise `DailyPayoutLimitReached` if so
- Added comprehensive test coverage for both scenarios: payout within 24 hours (blocked) and payout after 24 hours (allowed)

## Why

This change prevents abuse and ensures payouts are spaced out appropriately. Organizations can only request one payout per day, which helps manage payment processing and reduces operational overhead.

## How

The implementation:
1. Queries the database for the latest payout associated with an account
2. Compares the payout's creation timestamp against the current time
3. Raises a 400 error if a payout was created within the last 24 hours
4. The check is performed within the existing lock mechanism to ensure thread safety

## Checklist

- [x] This PR addresses a single concern (one feature: daily payout limit)
- [x] New functionality is covered by tests (`test_recent_payout_within_24h` and `test_previous_payout_older_than_24h`)

https://claude.ai/code/session_01HMS1Ev2sPXJhBpU5hc59FH


frankie567 (commented):
Logic makes sense but I'm wondering if we should do the extra mile and have:

* Default 24 hours limit at config level
* Specific column at account level to be able to override it

I already foresee the need from support/risk to either increase or decrease that limit for a given merchant.

pieterbeulque:
@frankie567 - Good callout. Should've written a quick spec first 😁 

I updated this PR to add a config and then a property on the organization that proxies that config. We can do that for now, and then in a separate PR I can follow up with the actual database migration and wire that up to the organization property?

server/polar/config.py
```diff
@@ -361,6 +361,7 @@ def POLAR_SELF_ENABLED(self) -> bool:
     API_PAGINATION_MAX_LIMIT: int = 100
 
     ACCOUNT_PAYOUT_DELAY: timedelta = timedelta(seconds=1)
+    ACCOUNT_PAYOUT_INTERVAL: timedelta = timedelta(hours=24)
```
pieterbeulque: I wasn't 100% sure about the naming & placement of this config here.

frankie567: Maybe `ACCOUNT_DEFAULT_PAYOUT_INTERVAL` to stress that it's the global default

frankie567 (approved):

frankie567:
> We can do that for now, and then in a separate PR I can follow up with the actual database migration and wire that up to the organization property?

Great plan!

frankie567 (approved):

pieterbeulque:
@frankie567 Pushed some additional changes so that this had a better merchant UX than just throwing an error, in line with the other checks (i..e not under review, payout account configured correctly):

server/polar/models/organization.py
```diff
@@ -748,6 +748,10 @@ def customer_portal_subscription_update_plan(self) -> bool:
     def checkout_require_3ds(self) -> bool:
         return self.checkout_settings.get("require_3ds", False)
 
+    @property
+    def payout_interval(self) -> timedelta:
```
frankie567: Just struck me that this should be at `Account` level. Will help when we'll add the specific setting.

clients/apps/web/src/components/Payouts/WithdrawModal.tsx
```diff
@@ -122,6 +133,21 @@ const WithdrawModal: React.FC<WithdrawModalProps> = ({
             </div>
           )}
 
+          {canWithdraw && isPayoutIntervalLimited && nextPayoutAt && (
+            <div className="flex flex-col gap-6">
+              <p>
+                You&apos;ve recently requested a withdrawal. The next one can be
```
frankie567: "a payout"

server/polar/organization/endpoints.py
```diff
@@ -143,13 +144,18 @@ async def get(
...
     if account is None:
         raise ResourceNotFound()
 
-    return account
+    next_payout_at = await payout_service.get_next_payout_at(
+        session, account, authz.organization
+    )
+    return AccountSchema.model_validate(account).model_copy(
+        update={"next_payout_at": next_payout_at}
+    )
```
frankie567: I don't really like that we have to manually tweak the schema to make work. What would you think of exposing `payout_interval` on the Account schema, fetch the latest payout using the dedicated API in the UI and make the calculation there?

frankie567 (changes requested):

server/polar/organization/endpoints.py
pieterbeulque: Yes, that's cleaner, updated now.

frankie567 (approved):

PR #11370, opened 2026-05-04
server/tax/calculation/numeral: handle zip state mismatch error

frankie567:
- server/tax/calculation/numeral: handle zip state mismatch error


psincraian (approved):
LGTM!

But I think we should start thinking on localization for the checkout errors also

PR #11372, opened 2026-05-04
polar: Increase the query timeout for the polar event ingestion task temporarily

Yopi:
Instead of doing something complicated for the backfill, I was thinking that we can just let it run to completion once, and then revert this PR

### DIFF
--- a/server/polar/integrations/polar/tasks.py
+++ b/server/polar/integrations/polar/tasks.py
@@ -6,9 +6,14 @@ from dramatiq import Retry
 
 from polar.config import settings
 from polar.event.repository import EventRepository
+from polar.kit.db.postgres import (
+    create_async_engine as _create_async_engine,
+)
+from polar.kit.db.postgres import (
+    create_async_sessionmaker,
+)
 from polar.kit.utils import utc_now
 from polar.worker import (
-    AsyncSessionMaker,
     CronTrigger,
     TaskPriority,
     actor,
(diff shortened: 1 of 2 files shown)


server/polar/integrations/polar/tasks.py
```diff
@@ -115,6 +118,17 @@ async def track_event_ingestion() -> None:
         last_flush = await repository.get_latest_polar_self_ingestion_timestamp(
             self_organization_id
         )
+        raw_conn = await session.connection()
+        asyncpg_conn = (await raw_conn.get_raw_connection()).driver_connection
+        assert asyncpg_conn is not None
+        asyncpg_conn._config = asyncpg_conn._config._replace(
+            command_timeout=_TRACK_EVENT_INGESTION_QUERY_TIMEOUT_SECONDS
+        )
```
frankie567: Could be worth to just open a new low-level engine instead of mangling the session :)

https://github.com/polarsource/polar/blob/ca358e89a5353cd854687f338a30abc7c3b5c49b/server/polar/kit/db/postgres.py#L47-L72

frankie567 (changes requested):

server/polar/integrations/polar/tasks.py
Yopi: Haha, I had the same thought but was thinking it was more complicated. I can do that instead

frankie567 (approved):

Yopi:
It managed to run without this 😄

PR #11374, opened 2026-05-04
feat(receipts): mention receipt download in customer emails

psincraian:
## Summary

- Add a conditional sentence to five customer emails (order confirmation, subscription confirmation, subscription renewal, post-trial cycle, final invoice) pointing customers at their receipt in the customer portal.
- Renders only when `order.receipt_number` is set, so emails are unchanged for orgs without receipts enabled.
- Standardize `subscription_cycled_after_trial`'s invoice line to match the other four templates.

## Test plan

- [ ] Preview each email template in the React Email dev server with and without `receipt_number`.
- [ ] Send a test order confirmation against a sandbox org with `receipts_enabled = true` and verify the link in the email opens the customer portal.

### DIFF
--- a/server/emails/src/emails/index.ts
+++ b/server/emails/src/emails/index.ts
@@ -1,34 +1,34 @@
-import { CustomerEmailChangedNotification } from "./customer_email_changed_notification";
-import { CustomerEmailUpdateVerification } from "./customer_email_update_verification";
-import { CustomerSessionCode } from "./customer_session_code";
-import { EmailUpdate } from "./email_update";
-import { LoginCode } from "./login_code";
-import { NotificationCreateAccount } from "./notification_create_account";
-import { NotificationCreditsGranted } from "./notification_credits_granted";
-import { NotificationNewSale } from "./notification_new_sale";
-import { NotificationNewSubscription } from "./notification_new_subscription";
-import { OAuth2LeakedClient } from "./oauth2_leaked_client";
-import { OAuth2LeakedToken } from "./oauth2_leaked_token";
-import { OrderConfirmation } from "./order_confirmation";
-import { OrganizationAccessTokenLeaked } from "./organization_access_token_leaked";
-import { OrganizationAccountUnlink } from "./organization_account_unlink";
-import { OrganizationInvite } from "./organization_invite";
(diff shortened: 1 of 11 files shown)


server/emails/src/emails/index.ts
```diff
@@ -1,34 +1,34 @@
...
-import { SubscriptionCycledAfterTrial } from "./subscription_cycled_after_trial";
-import { SubscriptionFinalInvoice } from "./subscription_final_invoice";
-import { SubscriptionPastDue } from "./subscription_past_due";
-import { SubscriptionRenewalReminder } from "./subscription_renewal_reminder";
-import { SubscriptionRevoked } from "./subscription_revoked";
-import { SubscriptionTrialConversionReminder } from "./subscription_trial_conversion_reminder";
-import { SubscriptionUncanceled } from "./subscription_uncanceled";
-import { SubscriptionUpdated } from "./subscription_updated";
-import { WebhookEndpointDisabled } from "./webhook_endpoint_disabled";
+import { CustomerEmailChangedNotification } from './customer_email_changed_notification'
```
pieterbeulque, line 1: @sebastianekstrom - Do you have tips on how to enforce coding style on the `server/emails` directory too? I also bump into this whenever I hit save in my IDE.

server/emails/src/emails/order_confirmation.tsx
```diff
@@ -22,6 +22,16 @@ export function OrderConfirmation({
         Thank you for purchasing{' '}
         <span className="font-medium">{order.description}</span>. Your invoice
         is attached.
+        {order.receipt_number && (
+          <>
+            {' '}
+            Your receipt is in the{' '}
```
pieterbeulque: Can we do "You can find your receipt in the [Customer Portal](#)" instead?

frankie567 (approved):

pieterbeulque (approved):
Exciting!

server/emails/src/emails/index.ts
psincraian: yep 😢 would be nice to have the same format. I think I will also enable jesper pr to check the openapi.ts file

PR #11359, opened 2026-04-30
refactor(member): drop redundant session param from MemberRepository methods

psincraian:
## Summary

- `MemberRepository` methods previously took an explicit `session: AsyncSession` / `AsyncReadSession` parameter even though `RepositoryBase.from_session(session)` already stores it on `self.session`. Normalized 6 methods (`get_owner_by_customer_id`, `get_by_customer_and_email`, `list_by_customer`, `transfer_ownership`, `list_by_email_and_organization`, `list_by_customers`) to use `self.session` and updated all callers.
- Behavior-preserving: every call site already constructed the repository with the same session it then passed to the method.
- Read methods now use the `get_one_or_none` / `get_all` base helpers where applicable; writes (`transfer_ownership`) and joinedload-free reads use `self.session` directly.
- Documented the convention in `server/CLAUDE.md` and added a worktree-setup note to root `CLAUDE.md`.

The `MemberRepository` API now matches the convention used by sibling repositories (`CustomerRepository`, `OrderRepository`, etc.).

## Test plan

- [x] `uv run task lint` passes
- [x] `uv run task lint_types` passes (mypy clean across 1291 source files)
- [x] `uv run pytest tests/customer tests/member tests/customer_session tests/customer_seat tests/checkout tests/organization tests/benefit tests/customer_portal tests/customer_email_update -p no:logfire` passes (1472+ tests)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11376, opened 2026-05-04
refactor(organization): drop redundant session param from get_admin_user

psincraian:
## Summary

`OrganizationRepository.get_admin_user` took `session: AsyncReadSession` as a parameter even though the repository was already constructed via `from_session(session)` and holds it as `self.session`. Switched to `self.session`, dropping the redundant arg from all 13 call sites.

Also removed a duplicate `get_admin_user` call in the v2 backoffice org detail view that was re-fetching the admin user for the team section after it had already been loaded earlier in the same request.

Mirrors the recent `MemberRepository` refactor in #11359.

## Test plan

- [x] `uv run task lint`
- [x] `uv run task lint_types`
- [x] `uv run task test_fast`

### DIFF
--- a/server/polar/backoffice/organizations/endpoints.py
+++ b/server/polar/backoffice/organizations/endpoints.py
@@ -901,7 +901,7 @@ async def create_plain_thread(
         if not organization:
             raise HTTPException(status_code=404)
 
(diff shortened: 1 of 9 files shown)


frankie567 (approved):
This was disseminated everywhere :D

PR #11378, opened 2026-05-04
feat: add cancel button for pending subscription updates (fix #11128)

### LINKED ISSUE #11128: Scheduled subscription changes can't be undone cleanly
I believe there's no way to cancel a scheduled subscription change (without billing side effects). Every path that clears a pending update also emits proration billing entries, creates a $0 order, or queues line items for the next invoice.

## Merchant Dashboard

@joebon added this neat UI which is helpful, but read-only. I tried changing the plan again to the current subscription plan, but that's not available.

Image

Image

## Customer Portal

Same story on the Customer Portal. You can see the pending change, but can't do anything about it.

This might be expected, since we also don't allow a custom proration behavior to be chosen by the customer.

## API

Let's say a subscription change was scheduled with `PATCH /v1/subscriptions/{id}` with `{ product_id: B, proration_behavior: "next_period" }`.

To revert, one can try:

- `PATCH { product_id: A, proration_behavior: "invoice" }` → pending update cleared, but a $0 order is created with credit + debit line items.
- `PATCH { product_id: A, proration_behavior: "prorate" }` → pending update cleared, but $0 proration entries are persisted and will appear as line items on the next invoice.
 - `PATCH { product_id: A, proration_behavior: "next_period" }` → pending update is not cleared; it is overwritten to a no-op change, and `subscription.pending_update` remains non-null until the next cycle.

I think the latter is an okay DX, but it should actually clear the pending update for a cleaner state, maybe by calling `soft_delete_unapplied_by_subscription_id`? Otherwise, you end up with an unnecessary pending update:

Image

frankie567:
## Summary

**Related Issue**: #11128

Both Customer Portal and Merchant Dashboard show pending subscription changes in read-only UIs. This adds a way to cancel a scheduled change from the UI.

## What

- Add Cancel buttons next to pending update sections in both Customer Portal and Merchant Dashboard
- Add confirmation modals before clearing pending updates
- Add mutation hooks for both Customer Portal (`useCustomerClearPendingSubscriptionUpdate`) and Merchant Dashboard (`useClearPendingSubscriptionUpdate`)

## Why

Users and merchants need the ability to cancel pending subscription updates (product changes, seat changes) before they take effect. Previously, pending changes could only be cancelled via API.

## How

- Created mutation hooks that PATCH the subscription with `pending_update: null` using existing API endpoints:
  - Customer Portal: `PATCH /v1/customer-portal/subscriptions/{id}` with `CustomerSubscriptionUpdateClear` schema
  - Merchant Dashboard: `PATCH /v1/subscriptions/{id}` with `SubscriptionUpdateClear` schema
- Added Cancel buttons with `variant="secondary"` and `size="sm"` in the header of pending update sections
- Added ConfirmModal with context-aware copy:
  - Customer Portal: "Your subscription will remain unchanged on the next billing cycle. Are you sure you want to cancel this pending update?"
  - Merchant Dashboard: "The customer's subscription will remain unchanged on the next billing cycle. Are you sure you want to cancel this pending update?"
- Updated cache invalidation to refresh data after clearing

## Screenshots

Subscription | frankie567 | Polar

Subscription | frankie567 | Polar 2

Customer Portal | frankie567 | Polar 3

Customer Portal | frankie567 | Polar 4

## Checklist

### DIFF
--- a/clients/apps/web/src/components/CustomerPortal/CustomerPortalSubscription.tsx
+++ b/clients/apps/web/src/components/CustomerPortal/CustomerPortalSubscription.tsx
@@ -1,6 +1,7 @@
 'use client'
 
 import {
+  useCustomerClearPendingSubscriptionUpdate,
   useCustomerCancelSubscription,
   useCustomerOrders,
   usePortalAuthenticatedUser,
@@ -10,7 +11,9 @@ import { Client, schemas } from '@polar-sh/client'
 import { formatCurrency } from '@polar-sh/currency'
 import Button from '@polar-sh/ui/components/atoms/Button'
 import { DataTable } from '@polar-sh/ui/components/atoms/DataTable'
+import { useState } from 'react'
 import FormattedDateTime from '@polar-sh/ui/components/atoms/FormattedDateTime'
+import { ConfirmModal } from '../Modal/ConfirmModal'
 import { useModal } from '../Modal/useModal'
 import { DownloadInvoicePortal } from '../Orders/DownloadInvoice'
 import AmountLabel from '../Shared/AmountLabel'
@@ -37,6 +40,9 @@ const CustomerPortalSubscription = ({
     isShown: cancelModalIsShown,
   } = useModal()
 
+  const [showClearPendingUpdateModal, setShowClearPendingUpdateModal] =
+    useState(false)
+
   // Get authenticated user to check billing permissions
   const { data: authenticatedUser } = usePortalAuthenticatedUser(api)
   const canManageBilling = hasBillingPermission(authenticatedUser)
@@ -48,6 +54,7 @@ const CustomerPortalSubscription = ({
   })
 
   const cancelSubscription = useCustomerCancelSubscription(api)
+  const clearPendingUpdate = useCustomerClearPendingSubscriptionUpdate(api)
 
   const pendingUpdate = subscription.pending_update
   const pendingProduct = products.find(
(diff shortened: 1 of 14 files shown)


clients/apps/web/src/components/CustomerPortal/CustomerPortalSubscription.tsx
```diff
@@ -138,7 +145,17 @@ const CustomerPortalSubscription = ({
...
-          <h3>Pending Update</h3>
+          <div className="flex flex-row items-center justify-between">
+            <h3>Pending Update</h3>
+            <Button
+              variant="secondary"
+              size="sm"
+              onClick={() => setShowClearPendingUpdateModal(true)}
+              loading={clearPendingUpdate.isPending}
+            >
+              Cancel
```
pieterbeulque: My only feedback would be to change this to not use the word "cancel". I'm afraid people will see this and assume it cancels their subscription and then be annoyed if we renew anyway.

How does "Discard" or "Discard change" feel? "Undo change"?

I think "Discard" is both technically correct and unambiguous, as you can't discard a subscription.

frankie567: "Discard" feels a bit weird to me. What about "Cancel scheduled change"? A bit verbose, but extra clear.

pieterbeulque: I agree. I like "Cancel scheduled change", verbose but clear feels like the way to go. Thank you!

frankie567: Here you go: 
Capture d’écran 2026-05-04 à 16 21 50

pieterbeulque (approved):

PR #11379, opened 2026-05-04
feat(organization): release slug on self-deletion

joebon:
## Summary

When a merchant self-deletes their organization, the original slug is now released for reuse by a new organization. Previous slugs are archived on the org for support visibility.

Tracks [polarsource/feedback#134](https://github.com/polarsource/feedback/issues/134) — the most common organization-related support load is merchants who typo their slug, self-delete the org, then can't recreate with the same slug. Pre-deletion checks already require no paid orders and no paid active subscriptions, so no end-customer communication is tied to the slug at delete time.

This is the lightweight subset of [#119](https://github.com/polarsource/feedback/issues/119); the full quarantine / cooldown design is **out of scope**.

## How it works

- New `slug_history` JSONB column on `organizations` (`[]` default).
- In `soft_delete_organization`, the current slug is appended to `slug_history` as `{"slug": ..., "deleted_at": ...}` and the live slug is rewritten to `__deleted__-{old_slug}-{id}`.
- The tombstone prefix uses underscores, which the user-facing slug validator (`_validate_slug` in `polar/kit/schemas.py`) rejects, so it lives in a namespace that can never collide with a user-creatable slug.
- The unique constraint on `slug` stays unconditional — tombstones are unique by org id, and `OrganizationRepository.slug_exists(include_deleted=True)` correctly returns `False` for the freed slug since the soft-deleted row no longer holds it.
- Backoffice navigates orgs by id, so freeing the slug doesn't break admin links. The backoffice settings section now also shows a "Previous slugs" row when `slug_history` is non-empty.

## Why a tombstone instead of a partial unique index

A partial index `WHERE deleted_at IS NULL` would also work, but:
- It splits "constraint scope" from "lookup scope" as a long-term invariant (`slug_exists` mirrors the constraint today).
- It needs a constraint migration on a hot table.
- The tombstone clearly marks the row as a deletion sentinel.

## Out of scope (follow-ups)

- Tracking slug history on backoffice-driven slug edits (wants the actor recorded — different audit shape).
- Full slug quarantine / cooldown from #119.
- Latent issue: `customer_portal/service/organization.py` looks up orgs by slug without a `deleted_at IS NULL` filter. With this change the symptom self-resolves for the deletion path, but it's worth a separate cleanup.

## Test plan

- [x] `uv run task lint_types` clean (1292 files)
- [x] `uv run pytest tests/organization/` (225 passed) — flipped `test_anonymizes_pii_preserves_slug` to assert the tombstone + history, added `test_releases_slug_for_reuse`, added `test_appends_to_existing_slug_history`
- [x] `uv run pytest tests/customer_portal/ tests/checkout/test_service.py` (524 passed) — no downstream regressions
- [x] Manual end-to-end: create org with slug `acme-test`, delete it from the dashboard, verify in psql that `slug = __deleted__-acme-test-<uuid>` and `slug_history = [{"slug": "acme-test", ...}]`, then create a new org with slug `acme-test` and confirm it succeeds. Backoffice org detail should render the previous slug under "Previous slugs".

### DIFF
--- a/server/migrations/versions/2026-05-04-1432_add_slug_history_to_organizations.py
+++ b/server/migrations/versions/2026-05-04-1432_add_slug_history_to_organizations.py
@@ -1,0 +1,35 @@
+"""add slug_history to organizations
+
+Revision ID: 199179085ab4
+Revises: 8847acc26477
+Create Date: 2026-05-04 14:32:41.589638
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# Polar Custom Imports
+
+# revision identifiers, used by Alembic.
+revision = "199179085ab4"
+down_revision = "8847acc26477"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "organizations",
+        sa.Column(
+            "slug_history",
+            postgresql.JSONB(astext_type=sa.Text()),
+            server_default="[]",
+            nullable=False,
+        ),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("organizations", "slug_history")
--- a/server/polar/organization/service.py
+++ b/server/polar/organization/service.py
@@ -614,15 +614,23 @@ class OrganizationService:
         session: AsyncSession,
         organization: Organization,
     ) -> Organization:
-        """Soft-delete an organization, preserving the slug for backoffice links.
+        """Soft-delete an organization, releasing its slug for reuse.
 
-        Anonymizes PII fields (except slug) and sets deleted_at timestamp.
+        Anonymizes PII fields, archives the previous slug to ``slug_history``,
+        and rewrites the live slug to a tombstone so a new organization can
+        claim the original.
         """
         repository = OrganizationRepository.from_session(session)
 
-        update_dict: dict[str, Any] = {}
+        now = datetime.now(UTC)
+        update_dict: dict[str, Any] = {
+            "slug_history": [
+                *organization.slug_history,
+                {"slug": organization.slug, "deleted_at": now.isoformat()},
+            ],
+            "slug": f"__deleted__-{organization.slug}-{organization.id}",
+        }
 
-        # Anonymize PII fields but NOT slug (to keep backoffice links working)
         pii_fields = ["name", "website", "customer_invoice_prefix"]
         github_fields = ["bio", "company", "blog", "location", "twitter_username"]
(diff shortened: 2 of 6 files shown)


server/migrations/versions/2026-05-04-1432_add_slug_history_to_organizations.py
```diff
@@ -0,0 +1,35 @@
...
+revision = "199179085ab4"
+down_revision = "8847acc26477"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "organizations",
+        sa.Column(
```
joebon, line 25: Intentionally KISSing this until we have a more robust model and structure for audit trails.

pieterbeulque: I was thinking of something similar. I just had one idea that, instead of just a plain array, we could store it as a key-value dict where the key is the slug and the value is the timestamp of when it was changed.

That way it's backwards compatible with implementing a quarantine period, and regardless of that, it's helpful as a poor man's audit log.

server/polar/organization/service.py
```diff
@@ -614,15 +614,23 @@ async def soft_delete_organization(
         session: AsyncSession,
         organization: Organization,
     ) -> Organization:
-        """Soft-delete an organization, preserving the slug for backoffice links.
```
pieterbeulque, line 617: This comment was wrong, right?

server/migrations/versions/2026-05-04-1432_add_slug_history_to_organizations.py
joebon: The array actually holds objects, not plain strings/slugs. The only entry to this array in code yet is the deletion, which adds an object with keys `slug` and `deleted_at`. So if I understand your suggestion correctly we are actually already aligned, with the exception that this approach allows for adding `updated_at`/`changed_at` instead of `deleted_at` to also imply what action was taken (change instead of delete).

server/polar/organization/service.py
joebon: Yeah, back office links use organization id rather than slug, so works fine for deleted orgs (with tombstoned slugs)

server/migrations/versions/2026-05-04-1432_add_slug_history_to_organizations.py
pieterbeulque: Ah, yes, I misread that `deleted_at`. That's even better then! 🫡

frankie567 (approved):

PR #11383, opened 2026-05-04
ci: serialize sandbox and production deploys per environment

psincraian:
## Summary

Two main-branch workflow runs could interleave their deploy jobs in the same environment, letting an older image overwrite a newer one (e.g. PR1's prod backend landing after PR2's). The job-level concurrency inside `deploy-environment.yml` only protected individual jobs, not the overall ordering.

This adds a `concurrency` group on `deploy-sandbox` and `deploy-production` so only one run per environment is in flight at a time. `cancel-in-progress: false` keeps in-flight deploys from being interrupted.

## Test plan

- [ ] Merge two PRs to `main` in quick succession; the second `deploy-production` waits for the first.
- [ ] A single-PR deploy still completes end-to-end.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -121,6 +121,9 @@ jobs:
 
   deploy-sandbox:
     name: "Deploy to Sandbox 🧪"
+    concurrency:
+      group: deploy-sandbox
+      cancel-in-progress: false
     needs: [changes, build]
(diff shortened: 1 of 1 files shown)


emilwidlund (approved):

Yopi (approved):

psincraian:
we will add the serialize in the full workflow for now

PR #11384, opened 2026-05-04
Exclude feature flag benefits from order confirmation emails

pieterbeulque:
## Summary

**Related Issue**: 

Filters out feature flag benefits from order confirmation emails to prevent exposing internal feature flag configuration to customers.

## What

Modified the `send_confirmation_email` method in `order/service.py` to exclude benefits of type `BenefitType.feature_flag` when constructing the product data sent in order confirmation emails. Feature flag benefits are now filtered out before the email is enqueued.

Added a test case `test_excludes_feature_flag_benefits` to verify that feature flag benefits are excluded while other benefit types (e.g., custom benefits) are included in the confirmation email.

## Why

Feature flag benefits are internal implementation details used to gate features for customers. They should not be exposed in customer-facing communications like order confirmation emails. This change ensures that only relevant, customer-facing benefits are included in the email.

## How

1. Created a `ProductEmail` instance from the order's product
2. Filtered the benefits list to exclude any benefits with `type == BenefitType.feature_flag`
3. Passed the filtered product to the email template instead of the original product
4. Added comprehensive test coverage to verify the filtering behavior

## Checklist

- [x] This PR addresses a single concern (filtering feature flag benefits from emails)
- [x] New functionality is covered by tests (`test_excludes_feature_flag_benefits`)
- [ ] Linting and type checking pass (to be verified)

https://claude.ai/code/session_013Tfj1n16R2G65B96b268TW

### DIFF
--- a/server/polar/email/schemas.py
+++ b/server/polar/email/schemas.py
@@ -3,8 +3,9 @@ import sys
 from enum import StrEnum
 from typing import Annotated, Literal
 
-from pydantic import BaseModel, Discriminator, TypeAdapter
+from pydantic import BaseModel, Discriminator, TypeAdapter, field_validator
 
+from polar.models.benefit import BenefitType
 from polar.notifications.notification import (
     MaintainerAccountCreditsGrantedNotificationPayload,
     MaintainerCreateAccountNotificationPayload,
@@ -56,6 +57,11 @@ class SubscriptionEmail(SubscriptionBase): ...
 
 class ProductEmail(ProductBase):
     benefits: BenefitList
+
+    @field_validator("benefits", mode="after")
+    @classmethod
+    def filter_feature_flag_benefits(cls, benefits: BenefitList) -> BenefitList:
+        return [b for b in benefits if b.type != BenefitType.feature_flag]
 
 
 class OrderEmail(OrderBase):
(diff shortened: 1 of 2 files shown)


Yopi (approved):

server/polar/order/service.py
```diff
@@ -1663,6 +1664,14 @@ async def send_confirmation_email(
                     return
 
         product = order.product
+        email_product: ProductEmail | None = None
+        if product is not None:
+            email_product = ProductEmail.model_validate(product)
+            email_product.benefits = [
```
frankie567: You know me, so here is what I would recommend :p

```py
class ProductEmail(ProductBase):
    benefits: BenefitList

    @field_validator("benefits", mode="after")
    @classmethod
    def filter_feature_benefits(cls, benefits: BenefitList) -> BenefitList:
        return [b for b in benefits if b.type != BenefitType.feature_flag]
```

I think then no change are needed at all in `order/service.py` :)

pieterbeulque: We should put this in a skill.md file somewhere!

pieterbeulque:
After discussing with Jesper, will come with a proper fix.

PR #11386, opened 2026-05-04
login code: Prevent code race

Yopi:
Migrate login code to repository and set `with_for_update` to ensure that it can't be used in two simultaneous sessions


frankie567 (commented):
Good catch! Could we do the same for customer session code:

https://github.com/polarsource/polar/blob/d52738ab2f712e67f60cd6a0096f1ec8b5a61948/server/polar/customer_portal/service/customer_session.py#L244-L246

frankie567 (approved):

PR #11393, opened 2026-05-04
fix(receipt): allow rendering without billing_name or billing_address

psincraian:
## Summary

- The `receipt.render` Dramatiq task was crashing with `AssertionError` on orders that were missing `billing_name` or `billing_address`. With receipts enabled by default for new organizations, paid orders whose customer hadn't filled in billing details started failing the render job — retrying up to 20 times each before escalating.
- `Receipt.from_order` now falls back through `order.billing_name → order.customer.name → order.customer.email` for the displayed customer name. The PDF renderer skips the address line when `customer_address` is absent. Only `receipt_number` remains a hard requirement.
- Deduped the email when it's already used as the customer name (the email line was previously printed twice in the worst-case fallback).

## Test plan

- [x] Unit + integration tests in `server/tests/receipt/` cover the three fallback levels and the missing-address case.
- [x] Manually rendered the four edge-case PDFs (full / email-only name / no address / email + no address) — layouts hold up.
- [x] `uv run task lint` passes.
- [x] `uv run task lint_types` passes.
- [x] `uv run pytest tests/receipt/ tests/invoice/` — 63 passed.

### DIFF
--- a/server/polar/invoice/generator.py
+++ b/server/polar/invoice/generator.py
@@ -71,7 +71,7 @@ class Invoice(BaseModel):
     seller_address: Address
     seller_additional_info: str | None = None
     customer_name: str
-    customer_address: Address
+    customer_address: Address | None = None
     customer_additional_info: str | None = None
     customer_locale: str | None = None
     subtotal_amount: int
@@ -394,7 +394,9 @@ class InvoiceGenerator(FPDF):
         # customer locale/country first, and then all other scripts.
         customer_script = self.cjk_script_from_locale(
             data.customer_locale
-        ) or self.resolve_cjk_script(data.customer_address.country)
+        ) or self.resolve_cjk_script(
+            data.customer_address.country if data.customer_address else None
+        )
 
         fallback_fonts = [
             family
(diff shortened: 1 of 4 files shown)


frankie567:
I don't understand in which context we can have a paid Order with a missing billing address. In production the following query returns no result:

```sql
SELECT * FROM orders
JOIN payments ON payments.order_id = orders.id AND payments.status = 'succeeded'
WHERE orders.billing_address = '{}' OR orders.billing_address IS NULL;
```

Can you give more details about the context of the issue?

psincraian:
@frankie567 it's because of this issue: https://polar-sh.sentry.io/issues/7459239965/?alert_rule_id=16426328&alert_type=issue&project=4505046565519360&referrer=slack

I saw that we have orders without billing name, and the beginning I assumed that we always used the card name in there, but I was wrong 🤔

frankie567:
Hm, yeah, my bad there is quite a bunch in both cases:

```sql
SELECT * FROM orders
JOIN payments ON payments.order_id = orders.id AND payments.status = 'succeeded'
WHERE orders.billing_address = 'null' OR orders.billing_name IS NULL
ORDER BY orders.created_at DESC;
```

Some are legacy orders (`stripe_invoice_id` filled), and some are from very old subscriptions that we may have migrated too.

frankie567 (approved):

PR #11397, opened 2026-05-05
fix: pass external_customer_id to member-by-external-id SDK calls

psincraian:
## Summary
- Forwards `external_customer_id` to the member-by-external-id SDK calls in the self client wrapper, which is now required server-side to uniquely scope a member.
- Bumps `polar-sdk` 0.31.3 → 0.31.4 so the new query param is exposed on both `get_member_by_external_id_async` and `delete_member_by_external_id_async`.

Fixes SERVER-4D7.

## Test plan
- [ ] `polar_self.remove_member` worker task no longer 422s


Yopi (approved):
Nice catch

frankie567 (approved):

PR #11408, opened 2026-05-05
fix(checkout): close embed overlay after success

pieterbeulque:
When a checkout's `success_url` pointed to an internal Polar URL (the default when the merchant didn't supply one), the confirmation hook called `router.push` inside the embed iframe to navigate to the customer portal — falling back to the portal sign-in page when no `customer_session_token` was present. The fixed-position overlay stayed mounted, so the customer was stranded on a sign-in screen with no way back to the merchant's site short of a hard reload.

- In `useCheckoutConfirmedRedirect`, return after posting the `success` message when running in embed mode; the parent window owns post-success navigation.
- In the embed script's default `success` listener, close the overlay when `redirect` is false (previously it only re-enabled `closable` and left the iframe up indefinitely).

### DIFF
--- a/clients/packages/checkout/src/embed.ts
+++ b/clients/packages/checkout/src/embed.ts
@@ -362,9 +362,7 @@ class EmbedCheckout {
   /**
    * Default listener for the `success` event.
    *
-   * Redirects the parent window to the `successURL` when `redirect` is `true`,
-   * otherwise closes the embedded checkout so the customer is returned to the
-   * merchant's page.
+   * This listener will redirect the parent window to the `successURL` if `redirect` is set to `true`.
    */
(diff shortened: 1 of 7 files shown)


clients/packages/checkout/src/embed.ts
```diff
@@ -373,6 +375,8 @@ class EmbedCheckout {
     this.closable = true
     if (event.detail.redirect) {
       window.location.href = event.detail.successURL
+    } else {
+      this.close()
```
pieterbeulque: I think this is a breaking change, let me cook a bit more.

pieterbeulque:
Not the correct approach

PR #11410, opened 2026-05-05
feat: share infra across `dev docker` instances

psincraian:
## Summary

- Splits `dev docker` into a single shared infra stack (postgres, redis, minio, tinybird; optional prometheus/grafana) and per-worktree app stacks (api, worker, web). Multiple worktrees coexist on one infra stack — each instance auto-bootstraps its own postgres database, Redis index, and S3 bucket pair on first start.
- Service-aware commands (`logs`, `exec`, `shell`, `restart`) auto-route by service name, so the user never needs to know which compose project a service lives in. `dev docker up` brings up shared infra (if needed) then this instance's app stack; `--all` on `down`/`cleanup` extends to the shared stack.
- Replaces the old per-instance-everything model — running N worktrees no longer pays for N copies of postgres/redis/tinybird/minio. The Tinybird container in particular goes from one cold build per worktree to one cold build per machine.

## Why this is safe

Postgres and Redis isolation is straightforward (per-instance database name and DB index). MinIO uses per-instance bucket names. Tinybird is the interesting one: every endpoint pipe filters by ` + "`organization_id`" + ` and each instance has its own postgres database, so org UUIDs across instances are guaranteed disjoint. Cross-instance leakage is impossible even though events from all instances coexist in the same workspace.

## Test plan

- [x] ` + "`dev docker up`" + ` from a clean worktree (no env files, no images, no volumes) — auto-creates network, starts shared infra, bootstraps the per-instance DB/buckets, runs migrations, loads seeds, serves API and web on offset ports
- [x] Second worktree's ` + "`dev docker up`" + ` joins the same shared infra without re-creating it; isolated DB/Redis-index/buckets confirmed
- [x] End-to-end Stripe checkout via Playwright on two instances in parallel — both produce ` + "`paid`" + ` orders in their own postgres databases
- [x] ` + "`dev docker logs db`" + ` / ` + "`exec db psql ...`" + ` / ` + "`exec redis redis-cli ...`" + ` auto-route to the shared project; ` + "`logs api`" + ` / ` + "`exec api uv run alembic current`" + ` route to the per-instance project
- [x] ` + "`dev docker down --all`" + ` and ` + "`dev docker cleanup --all`" + ` correctly stop / wipe the shared stack (with confirmation)
- [x] Lint clean (` + "`ruff check`" + `, ` + "`ruff format --check`" + `)


dev/cli/commands/docker.py
```diff
@@ -155,77 +218,161 @@ def _ensure_server_env() -> None:
...
+            "--status",
+            "running",
+        ],
+        capture_output=True,
+        text=True,
+    )
+    return result.returncode == 0 and bool(result.stdout.strip())
+
+
+def _ensure_shared_running() -> None:
```
joebon: It doesn't look like this one is called anywhere, so let's remove

dev/cli/commands/docker.py
```diff
@@ -453,7 +733,9 @@ def docker_set_instance(
         _write_stored_instance(instance)
         offset = instance * 100
         console.print(f"[green]Stored instance {instance} in .env.docker[/green]")
-        console.print(f"[dim]Ports: API={8000 + offset}, Web={3000 + offset}, DB={5432 + offset}[/dim]")
+        console.print(
+            f"[dim]Ports: API={8000 + offset}, Web={3000 + offset}, DB={db_name(instance)}[/dim]"
```
joebon: Perhaps still clarify the DB port in this print and then output the db name in a separate print?

dev/cli/commands/docker.py
```diff
@@ -16,12 +20,59 @@
...
 
-VALID_SERVICES = ["api", "worker", "web", "db", "redis", "minio", "minio-setup", "tinybird", "prometheus", "grafana"]
+SHARED_PROJECT_NAME = "polar-shared"
+SHARED_NETWORK_NAME = "polar-shared"
+
+# Per-instance naming. Single source of truth — referenced both here and in
+# docker-compose.dev.yml (via env interpolation) and dev/docker/scripts/startup.sh.
+# Redis only has 16 DBs by default, so the index wraps; instance numbers stay
+# small in practice.
+REDIS_DB_COUNT = 16
```
joebon: Hrm, but instance numbers are produced (in `_detect_instance`) from a MD5 hash (unless run through conductor) projected more or less uniformly on the range [1,99] (i.e. not sequential and increasing) so the probability of a collision when wrapping on 16 increases pretty fast even with not so many concurrent work trees.
I'd drop this and just map redis_db directly to the instance number. Increasing the number redis databases for a server is cheap as I understand it

dev/docker/docker-compose.shared.yml
```diff
@@ -0,0 +1,150 @@
...
+    networks:
+      - polar-shared
+    healthcheck:
+      test: ["CMD", "pg_isready", "-q", "-d", "polar", "-U", "polar"]
+      timeout: 40s
+      interval: 2s
+      retries: 20
+
+  redis:
+    image: redis:alpine
```
joebon, line 35:
```suggestion
    image: redis:alpine
    command: ["redis-server", "--databases", "100"]
```

Set an explicit command which also changes the number of databases from 16 to 100 to solve the instance number mapping an possible collisions

joebon (changes requested):
Very nice! Some small suggested tweaks though to avoid collisions and dead code IMHO

dev/cli/commands/docker.py
psincraian: yep, that's what happens when you go through different iterations 🥲

dev/cli/commands/docker.py
psincraian: I would change that soon to be sequential :) but I agree with the suggestion for now

joebon (approved):
brilliant!

PR #11412, opened 2026-05-05
polar: Add webhook handler for Polar events

Yopi:
When we get a webhook that a subscription has been updated, we cross reference to ensure that the organization exists, and look up the benefits granted to the customer (organization). We set the platform fee according to the benefits.


server/polar/account/service.py
```diff
@@ -89,6 +89,23 @@ async def update(
             account, update_dict=account_update.model_dump(exclude_unset=True)
         )
 
+    async def set_platform_fee(
+        self,
+        session: AsyncSession,
+        account: Account,
+        *,
+        take_rate: int | None,
+        flat_fee_in_cents: int | None,
```
frankie567: May I be a bit picky and suggest to call those parameters `fee_percent` and `fee_fixed` for consistency? :D

server/polar/integrations/polar/tasks.py
```diff
@@ -147,3 +155,33 @@ async def track_organization_review_usage(
         output_tokens=output_tokens,
         cost_usd=Decimal(cost_usd),
     )
+
+
+@actor(actor_name="polar_self.webhook.subscription.active", priority=TaskPriority.LOW)
```
frankie567: From what I see, we don't really do anything special with the subscription data.

Usually, our recommendation to merchants is to listen to the customer state or benefit grants eventsrather than trying to listen to the more complex subscription events.

server/polar/account/service.py
```diff
@@ -89,6 +89,23 @@ async def update(
...
+    ) -> Account:
+        repository = AccountRepository.from_session(session)
+        return await repository.update(
+            account,
+            update_dict={
+                "_platform_fee_percent": take_rate,
```
frankie567: Side note to make sure we're aligned here: this value is expressed in basis points (1/10_000), so 4% is 400.

Yopi: Yes, aligned with how we store these 😄

server/polar/account/service.py
Yopi: You definitely may 😄

server/polar/integrations/polar/tasks.py
Yopi: If that's what we recommend, then I suppose we should just list to the benefit grant / revoke and use the data there directly, instead of going back to the API?

frankie567: Yes. If we want to be a bit more future proof, maybe the customer state is a good alternative: we'll get active subscriptions and granted benefits directly in the payload.

Yopi:
@frankie567 I think this is good to go now

server/polar/integrations/polar/client.py
```diff
@@ -188,6 +188,38 @@ async def remove_member(
...
+        from polar_sdk.models.polarerror import PolarError
+
+        with logfire.span(
+            "polar.list_customer_benefit_grants", customer_id=customer_id
+        ) as span:
+            grants: list[BenefitGrant] = []
+            page = 1
+            try:
+                while True:
+                    response = await self._sdk.benefit_grants.list_async(
```
frankie567: Tip: I think Speakeasy havs some kind of iterator pattern:  https://github.com/polarsource/polar-python/tree/main/docs/sdks/benefitgrants#list

server/polar/integrations/polar/service.py
```diff
@@ -104,5 +136,180 @@ def enqueue_track_organization_review_usage(
...
+        return (
+            self._parse_int_metadata(metadata, "fee_percent", benefit_id),
+            self._parse_int_metadata(metadata, "fee_fixed", benefit_id),
+        )
+
+    def _parse_int_metadata(
+        self, metadata: dict[str, Any], field: str, benefit_id: str
+    ) -> int:
+        value = metadata.get(field)
+        if not isinstance(value, str):
```
frankie567: I think we could have an `int` directly, it's allowed by our schema. 

https://github.com/polarsource/polar/blob/6ede0c94b5827b9672287f8d92635d74a8716796/server/polar/kit/metadata.py#L35

But I think the confusion comes from the fact that metadata set from the dashboard is necessarily saved as a string (we don't try to cast automatically). The risk is that if someday we create benefits programmatically with our API, we might hit this condition.

server/polar/integrations/polar/service.py
```diff
@@ -104,5 +136,180 @@ def enqueue_track_organization_review_usage(
...
+
+    def _parse_bool_metadata(
+        self, metadata: dict[str, Any], field: str, benefit_id: str
+    ) -> bool:
+        value = metadata.get(field)
+        if value == "true":
+            return True
+        if value == "false":
+            return False
+        raise SupportBenefitError(
```
frankie567, line 286: Same comment here, we may have true booleans :)

server/polar/integrations/polar/client.py
Yopi: Thanks!

server/polar/integrations/polar/service.py
Yopi: Hmm, I was thinking about this as well, but I thought it was all saved as strings. I also tried to verify it but it looks like I did a poor job. Lets do both variants then, as I'm not sure what will happen if you modify a value that's stored as a number in the GUI

frankie567 (approved):

Yopi:
❤️

PR #11416, opened 2026-05-05
emails: drop organization_reviewed and organization_under_review

pieterbeulque:
The `under_review` email was already wired up but never sent, only its template/schema existed.

The `reviewed` email was sent on initial review but is no longer wanted. Remove both, along with the now-empty `organization.reviewed` task (which was only responsible for sending the email).

The `reviewed` email didn't always go out at the correct time with the changes we made to our process, so I'd rather silence all communication for now and then redo it later.


frankie567 (approved):

psincraian (approved):
Love these PR!

PR #11417, opened 2026-05-05
Refresh app session

sebastianekstrom:
This PR adds an OAuth refresh-token flow for the app so access tokens are renewed proactively and on `401`'s, while persisting refresh tokens alongside access tokens in session storage.

It also routes previously raw API calls through the authenticated client so refresh applies consistently.

**How to test**
1. Run the app against a local backend, guide is in [clients/apps/app/auth/oauthConfig.ts](https://github.com/polarsource/polar/pull/11417/changes#diff-2a4b33757c57992bea73605f1393eaf569ccb66bb78de481855d0abecc10d044)
2. Hard code the token to have a 60 second TTL in [server/polar/oauth2/constants](https://github.com/polarsource/polar/blob/main/server/polar/oauth2/constants.py#L25-L29)
3. Start the app
4. Log in
5. Wait ~40 seconds
6. Navigate to a sub page that makes a network call
7. Your token should now refresh

### DIFF
--- a/clients/apps/app/auth/refresher.test.ts
+++ b/clients/apps/app/auth/refresher.test.ts
@@ -1,4 +1,5 @@
 import type { refreshAsync } from 'expo-auth-session'
+import { TokenError } from 'expo-auth-session'
 import type * as RefresherModule from './refresher'
 
 const mockRefreshAsync = jest.fn<
@@ -6,9 +7,15 @@ const mockRefreshAsync = jest.fn<
   Parameters<typeof refreshAsync>
 >()
 
-jest.mock('expo-auth-session', () => ({
-  refreshAsync: mockRefreshAsync,
-}))
+jest.mock('expo-auth-session', () => {
+  const actual = jest.requireActual<typeof import('expo-auth-session')>(
+    'expo-auth-session',
+  )
+  return {
+    ...actual,
+    refreshAsync: mockRefreshAsync,
+  }
+})
 
 let configureRefresher: (typeof RefresherModule)['configureRefresher']
 let hasRefreshToken: (typeof RefresherModule)['hasRefreshToken']
@@ -164,8 +171,12 @@ describe('isAccessTokenStale', () => {
 
     it('does not record a lifetime when configured with an expired token (cold start path)', () => {
       configure({ expiresAt: NOW - 5_000 })
-
       configure({ expiresAt: NOW + 30_000 })
+
+      expect(isAccessTokenStale()).toBe(false)
+
+      jest.setSystemTime(NOW + 20_001)
+      expect(isAccessTokenStale()).toBe(true)
     })
   })
 })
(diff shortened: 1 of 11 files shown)


emilwidlund (approved):

clients/apps/app/providers/SessionProvider.tsx
joebon, line 1: From my man Claude Shakespeare:
> `SessionProvider.tsx:1275` only writes the widget storage when `accessToken` is truthy — it doesn't clear it when `accessToken` becomes `null`. Pre-existing, not introduced here, but the widget will keep rendering with a stale token after logout. Worth a follow-up.

clients/apps/app/auth/refresher.ts
joebon, line 1: Claude:
> `refresher.ts:854-859` — a transient network blip during refresh logs the user out completely. Consider distinguishing transport errors from `invalid_grant` (e.g. only clear on 4xx from the token endpoint). On a phone with flaky connectivity, an aggressive logout is worse than letting the next request retry.

clients/apps/app/providers/PolarClientProvider.tsx
joebon, line 1: From claude:
> `PolarClientProvider.tsx:1166` constructs the client with `session` baked into the default `Authorization` header. After a successful refresh, `state.expiresAt` is updated synchronously inside `refresher.ts`, but the polar client's default header isn't updated until the `setSession` → state change → `useMemo` rerun completes. In that window, `onRequest` sees a fresh token (so it returns `undefined` and doesn't rewrite the header), and the request goes out with the old bearer token. The 401 path will then refresh again (or share an in-flight one).                                                                                                                                                                                                                          
> 
> Mitigation: have `onRequest` also rewrite the header whenever the request's current `Authorization` doesn't match `state.accessToken`. Cheap and removes the spurious 401. Not a blocker — dedup keeps it from snowballing.

clients/apps/app/auth/refresher.test.ts
joebon, line 1: Let me tell you what my pal Claude came up with:
> Several `it(...)` blocks call setup but assert nothing, so they pass regardless of behavior:                                                                                                                                                                  
>                               
>   - `refresher.test.ts:639-645` — 'updates state so subsequent isAccessTokenStale uses the new lifetime' — no expect.                                                                                                                                           
>   - `refresher.test.ts:560-564` — 'does not record a lifetime when configured with an expired token (cold start path)' — no expect.
>   - `refresher.test.ts:717-724` — 'estimates lifetime from expiresAt on the first configure call' — only `setSystemTime` calls.                                                                                                                                   
>   - `refresher.test.ts:726-743` — 'does not overwrite a known lifetime on subsequent configure calls' — has one `expect` at the end but it doesn't actually assert what the title claims (it would pass even if the lifetime were overwritten).

joebon (approved):
Added some Claude noise/comments which are non-blockers IMO

sebastianekstrom:
All comments adressed! Merging 🚢

PR #11420, opened 2026-05-05
checkout: filter blocked/deleted orgs in get_by_client_secret

Yopi:


frankie567 (changes requested):
For consistency with `get_by_id`, I think we should rather make the check at the method level and raise a 403:

https://github.com/polarsource/polar/blob/0bb81bd21a68c7117592aa48653eb1fafdbf32c0/server/polar/checkout/service.py#L290-L311

Yopi:
> For consistency with `get_by_id`, I think we should rather make the check at the method level and raise a 403:
> 
> https://github.com/polarsource/polar/blob/0bb81bd21a68c7117592aa48653eb1fafdbf32c0/server/polar/checkout/service.py#L290-L311

Makes sense, will do

server/polar/checkout/repository.py
```diff
@@ -37,7 +37,10 @@ async def get_by_client_secret(
     ) -> Checkout | None:
         statement = (
             self.get_base_statement()
-            .where(Checkout.client_secret == client_secret)
+            .join(Organization, onclause=Organization.id == Checkout.organization_id)
```
frankie567: I think the join is no longer needed then :)

frankie567 (approved):

PR #11421, opened 2026-05-05
authz: Exclude blocked organizations from user org ids

Yopi:
Detail: https://app.detail.dev/org_ecfbc7cf-6d21-4ce1-8c61-cda1d650ccf7/bugs/bug_847ae7ad-006e-484f-a9fc-8da618889a9a

### DIFF
--- a/server/polar/authz/repository.py
+++ b/server/polar/authz/repository.py
@@ -13,7 +13,7 @@ class AuthzRepository:
         self.session = session
 
     async def get_user_org_ids(self, user_id: UUID) -> set[UUID]:
-        """Get all organization IDs a user is a member of."""
+        """Get all organization IDs a user is a member of that are accessible."""
         stmt = (
             select(UserOrganization.organization_id)
             .join(Organization, UserOrganization.organization_id == Organization.id)
@@ -21,6 +21,7 @@ class AuthzRepository:
                 UserOrganization.user_id == user_id,
                 UserOrganization.is_deleted.is_(False),
                 Organization.is_deleted.is_(False),
+                Organization.can_authenticate,
             )
         )
         result = await self.session.scalars(stmt)
--- a/server/tests/authz/test_service.py
+++ b/server/tests/authz/test_service.py
@@ -5,6 +5,7 @@ import pytest
 from polar.auth.models import AuthSubject
 from polar.authz.service import get_accessible_org_ids, get_accessible_organization
 from polar.models import Organization, User
+from polar.models.organization import OrganizationStatus
 from polar.models.user_organization import UserOrganization
 from polar.postgres import AsyncSession
 from tests.fixtures.auth import AuthSubjectFixture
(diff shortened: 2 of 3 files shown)


Yopi:
Ah sorry @joebon, didn't see that you also started working on these. This collides with your fix in https://github.com/polarsource/polar/pull/11405, since they are not handled explicitly downstream in this flow 😅

joebon:
Ah, sorry - haven't gotten around to update the status on the deleted orgs-bug. But blocked orgs was not fixed by my PR, so that's still up for grabs

joebon:
@Yopi I think there will be some test failures due to response mismatches (`404` vs `403`)

server/polar/authz/repository.py
```diff
@@ -21,6 +21,7 @@ async def get_user_org_ids(self, user_id: UUID) -> set[UUID]:
                 UserOrganization.user_id == user_id,
                 UserOrganization.is_deleted.is_(False),
                 Organization.is_deleted.is_(False),
+                Organization.status != OrganizationStatus.BLOCKED,
```
frankie567: The recommended way to check that now is to use `Organization.can_authenticate` (9ac233bff17c62843255849b28aab9d22dbbc25d)

frankie567 (changes requested):

server/polar/authz/repository.py
Yopi: Makes sense! I'm a bit unsure though since this is what Petru did in the follow up PR https://github.com/polarsource/polar/pull/11025/changes#diff-64d7a3016f1048dae709cf4a5f87da70d374063e77bcb344c928462b899a04f1R40-R63

frankie567: Hmm, right. Any recommendation, @psincraian?

server/polar/checkout/service.py
```diff
@@ -294,7 +294,9 @@ async def get_by_id(
         id: uuid.UUID,
     ) -> Checkout | None:
         repository = CheckoutRepository.from_session(session)
-        org_ids = await get_accessible_org_ids(session, auth_subject)
+        org_ids = await get_accessible_org_ids(
+            session, auth_subject, include_blocked=True
```
frankie567: The `include_blocked` is there to let it pass at this stage and then fail later with a 403, right?

Yopi: Yes, correct

Yopi: To keep the single GET or POST the same as they are today

joebon:
@Yopi @frankie567 do we actually need to maintain the same `403` responses? Would it be an API contract breach if we changed those to `404`s instead?
Semantically I think I could argue that both those response codes are valid.
But clearly the code becomes a bit messier just for backwards compatibility here.

frankie567:
> @Yopi @frankie567 do we actually need to maintain the same `403` responses? Would it be an API contract breach if we changed those to `404`s instead? Semantically I think I could argue that both those response codes are valid. But clearly the code becomes a bit messier just for backwards compatibility here.

Yeah, TBH, I don't really care what we return to blocked organizations 😄 So, if it makes the code simpler, I would say we can return a 404.

Yopi:
> > @Yopi @frankie567 do we actually need to maintain the same `403` responses? Would it be an API contract breach if we changed those to `404`s instead? Semantically I think I could argue that both those response codes are valid. But clearly the code becomes a bit messier just for backwards compatibility here.
> 
> Yeah, TBH, I don't really care what we return to blocked organizations 😄 So, if it makes the code simpler, I would say we can return a 404.

I was going back and forward on it as well. I was thinking if there was a point in allowing the repository to return both, but maybe not then. I can clean it up again

joebon (approved):
Such clean, very clarity!

frankie567 (approved):
Clean! Let's go 🚀

PR #11422, opened 2026-05-05
oat: validate scopes for all callers when creating tokens

joebon:
## Summary

- `_validate_scopes_within_caller` was exempting all User subjects from scope validation, which let OAuth2 tokens and PATs (which authenticate as a User with limited scopes) create Organization Access Tokens with scopes the caller didn't have.
- The exemption was based on the assumption that User subjects always have full scopes (true only for new web sessions). It silently bypassed validation for OAuth2/PAT/impersonation sessions.
- Removed the early-return entirely. Web sessions don't need a special case: new ones already get `set(Scope)` so validation is a no-op, and the limited ones (e.g. `{web_read}` impersonation) *should* be checked.

Detail bug: `bug_01c5dbf9-e101-4c60-bfdb-9d23ff6a63c0`
Introduced in #11073.

## Test plan

- [x] `uv run pytest tests/organization_access_token/` (9 passed, including new `TestCreateScopeValidation.test_pat_caller_cannot_mint_broader_scope`)
- [x] `uv run task lint_types`
- [x] `uv run task lint`
- [ ] Manual: a web user with all scopes can still create OATs with any subset of scopes
- [ ] Manual: a PAT/OAuth2 caller with limited scopes gets a 422 when requesting broader scopes

### DIFF
--- a/server/polar/organization_access_token/service.py
+++ b/server/polar/organization_access_token/service.py
@@ -6,7 +6,7 @@ from uuid import UUID
 import structlog
 from sqlalchemy import UnaryExpression, asc, desc
 
-from polar.auth.models import AuthSubject, Organization, is_organization, is_user
+from polar.auth.models import AuthSubject, Organization, is_user
 from polar.auth.scope import Scope
 from polar.authz.service import get_accessible_org_ids
 from polar.config import settings
@@ -165,8 +165,6 @@ class OrganizationAccessTokenService:
         auth_subject: AuthSubject[User | Organization],
         requested_scopes: Sequence[str],
     ) -> None:
-        if not is_organization(auth_subject):
-            return
         caller_scopes = auth_subject.scopes
         requested = {Scope(s) for s in requested_scopes}
         excess = requested - caller_scopes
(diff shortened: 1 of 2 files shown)


frankie567 (approved):
Great!

Beyond that, I think we should question the existence of an API accessible with a token to generate other tokens. It feels like an endless source of security holes. @pieterbeulque, I remember you added it to help with Lovable, but do you think it's strictly necessary?

joebon:
> Great!
> 
> Beyond that, I think we should question the existence of an API accessible with a token to generate other tokens. It feels like an endless source of security holes. @pieterbeulque, I remember you added it to help with Lovable, but do you think it's strictly necessary?

That's a really good point @frankie567 , but how would we otherwise output organization tokens that are produced in the dashboard? The UI is only an API consumer after all, authenticated by a (session) token. Do you mean that we should limit the audience of this endpoint to only web sessions?

frankie567:
@joebon Yes, that's what I mean. It was like this before https://github.com/polarsource/polar/pull/9056

joebon:
Cool @frankie567 ! I'll merge this and add a GitHub issue so we don't forget about it.

PR #11445, opened 2026-05-06
fix: show all organization members in backoffice v2 team section

psincraian:
Removes the hardcoded 10-member limit on the team section of the backoffice v2 organization detail view, so the full members list is rendered.


frankie567 (approved):
Curious this was here!

Issue #11399, opened 2026-05-05
Tighten the schema


joebon:
**Scope.** Make the new role column non-null. Schema-only.

**Acceptance criteria**

- [ ] Migration: `ALTER TABLE user_organizations ALTER COLUMN role SET NOT NULL`, shipped after Phase 1 has fully deployed and every row has a value.

Cross-table consistency between `role` and `Account.admin_id` continues to rest on the service-layer validation shipped in Phase 1 (Postgres CHECK can't span tables declaratively, and the column is being dropped in Phase 6 anyway, so a trigger isn't worth the trouble).

**References.** [RFC §Phase 2](https://handbook.polar.sh/engineering/design-documents/rbac#phase-2--tighten-the-schema), parent #6646. Depends on #11398.

Issue #11398, opened 2026-05-05
Additive schema, dual-write


joebon:
**Scope.** Add the role attribute and backfill from `Account.admin_id`, set up dual-write on existing ownership-mutation paths, and ship the permission infrastructure in code without yet consuming it.

**Acceptance criteria**

- [ ] `OrganizationRole` enum (`owner`, `admin`, `member`) added.
- [ ] `role` column added to `UserOrganization` with default `member`.
- [ ] Backfill: every `Account.admin_id` user is `owner` on their membership; everyone else is `member`.
- [ ] Service-layer dual-write: org creation and the backoffice `change_admin` flow swap roles (previous owner → `admin`, new admin → `owner`) in the same transaction as the `Account.admin_id` mutation.
- [ ] Service-layer validation rejects writes that would put a non-`Account.admin_id` user into `owner` or move that user out of `owner`.
- [ ] `role → permissions` and `scope → implied_permissions` tables landed in code; fine-grained permission strings (`organizations:edit_settings`, `organizations:delete`, `organizations:manage_payout_account`, `members:invite`, `members:remove`, `members:set_role`) defined.
- [ ] Tables are not yet consumed by any policy; authorization still reads `Account.admin_id`.

**References.** [RFC §Phase 1](https://handbook.polar.sh/engineering/design-documents/rbac#phase-1--additive-schema-dual-write), parent #6646.

PR #11451, opened 2026-05-06
feat(rbac): add OrganizationRole, dual-write, validation, and backfill

joebon:
## Summary

First PR of the RBAC rollout. Lands the role data shape, the dual-write that keeps it aligned with `Account.admin_id`, the validation guard, the dormant permission tables, and the manual backfill script — without yet flipping any authorization policy off `Account.admin_id`.

Closes:
- #11398 (additive schema, dual-write)
- #11399 (NOT NULL — folded into the same migration since `server_default='member'` keeps inserts well-formed and Phase 2's tightening had no other work).

## What's in this PR

- **Schema**: `OrganizationRole` enum (`owner`, `admin`, `member`), `role` column on `user_organizations` (NOT NULL, `server_default='member'`).
- **Dual-write**:
  - Org creation assigns `owner` to the creator (via `add_user(role=...)`).
  - `account_service.change_admin` swaps roles atomically with the `Account.admin_id` mutation: previous owner demoted to `admin` if they currently carry `owner`; new admin promoted to `owner`. The conditional demote tolerates pre-backfill state (`member`).
- **Validation**: `user_organization_service.set_role` rejects assigning `owner` to a non-`Account.admin_id` user (`InvalidOwnerRoleAssignment`) and rejects moving a user off `owner` directly (`OwnerRoleCannotBeRemoved`). Internal flows that swap roles atomically with an `Account.admin_id` mutation bypass `set_role` by design.
- **Permission vocabulary** (`server/polar/auth/permission.py`): `OrganizationPermission` strings, `SCOPE_IMPLIED_PERMISSIONS` (split `organizations:write` and `members:write` into fine-grained permissions), `ROLE_PERMISSIONS` (admin and owner share the same set in this iteration; member is read of member list only). Not consumed by any policy yet.
- **Backfill script**: `server/scripts/backfill_organization_roles.py`, idempotent, run manually post-deploy.

## Tests

- 14 new tests in `tests/user_organization/test_service.py` covering `set_role` + the validation guards.
- 2 new tests in `tests/account/test_service.py` covering the `change_admin` role swap (post-backfill and pre-backfill cases).
- Existing org-creation test extended to assert the creator gets `OrganizationRole.owner`.
- 351 tests pass across `tests/organization`, `tests/account`, `tests/user_organization`. Lint + mypy clean.

## Operational notes

After deploy, run the backfill manually:

```bash
uv run python -m scripts.backfill_organization_roles --dry-run  # preview
uv run python -m scripts.backfill_organization_roles            # apply
```

Idempotent — re-runs skip rows already at `owner`.

Authorization still reads `Account.admin_id`; the cutover to permission-based checks is a follow-up.

## RFC

[handbook/engineering/design-documents/rbac](https://handbook.polar.sh/engineering/design-documents/rbac)


server/polar/user_organization/service.py
```diff
@@ -109,6 +133,58 @@ async def get_by_user_and_org(
         res = await session.execute(stmt)
         return res.scalars().unique().one_or_none()
 
+    async def set_role(
```
joebon, line 137: Added but not used in this PR (or anywhere else yet). Will be used in follow-up PRs

server/tests/user_organization/test_service.py
```diff
@@ -246,3 +249,110 @@ async def test_list_by_org_excludes_deleted(
...
+        account: Account,
+        organization: Organization,
+        user_second: User,
+    ) -> None:
+        relation = UserOrganization(
+            user_id=user_second.id, organization_id=organization.id
+        )
+        await save_fixture(relation)
+
+        result = await user_organization_service.set_role(
```
joebon, line 269: The exception - we _do_ use `set_role` in this PR, but only here in tests 😄

server/migrations/versions/2026-05-06-1015_add_role_column_to_user_organizations.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def upgrade() -> None:
+    op.execute("SET LOCAL lock_timeout = '2s'")
+    op.add_column(
+        "user_organizations",
+        sa.Column(
+            "role",
+            sa.String(),
+            server_default="member",
```
frankie567: I usually avoid to use `server_default`, as they may hit by surprise compared to a pure ORM default. In this case, I think we can simply create the column as nullable, run an `UPDATE` query and alter the column again to make it non-nullable. There are currently 83k rows in this table, so it should be fairly quick to run in a migration without the need of a backfill step.

server/polar/user_organization/service.py
```diff
@@ -109,6 +133,58 @@ async def get_by_user_and_org(
...
+          `Account.admin_id`.
+        - A user that currently carries `owner` cannot be moved off it
+          directly; ownership transfer flows through `Account.admin_id`
+          mutations (today: backoffice `change_admin`).
+
+        Internal trusted flows that swap roles atomically with an
+        `Account.admin_id` change (`account_service.change_admin`) bypass
+        this method by design.
+        """
+        from polar.account.repository import AccountRepository
```
frankie567: Can we move that inline import?

server/polar/auth/permission.py
```diff
@@ -0,0 +1,124 @@
...
+    },
+    Scope.members_write: {
+        OrganizationPermission.members_invite,
+        OrganizationPermission.members_remove,
+        OrganizationPermission.members_set_role,
+    },
+}
+
+
+def implied_permissions(scopes: set[Scope]) -> set[OrganizationPermission]:
```
frankie567: I'm not sure I understand this (I know this was in the RFC, but probably misunderstood) ☺️ Why do we need to bridge the API scopes and the role permissions?

In my understanding, there are two stages to check if a token is allowed to perform an action. For example, the action "Delete Organization" requires the scope `organizations:write` and the permission `organizations:delete`. Then the flow would be:

1. Does the token has the scope `organizations:write`?
    * No -> 403 ❌
    * Yes -> Continue ⬇️
2. Does the authenticated subject has a role with the permission `organizations:delete`? 
    * No -> 403 ❌
    * Yes -> Proceed

Said another way, it acts as two consecutive gates.

That's why I don't understand why we need to have a mapping like this. Could you clarify? :)

joebon: Hah! That's a good point, I got lost 😅 . Originally I was aiming to reuse scope constants as permission constants (to not duplicate things) but that mission derailed. I'll remove and cleanup

server/polar/user_organization/service.py
joebon: Totes!

server/migrations/versions/2026-05-06-1015_add_role_column_to_user_organizations.py
joebon: Sure thing! I'll change the migration to do these steps in order:
1) Add role column with `NULL`s
2) `UPDATE` all rows to set `role='member'`
3) `UPDATE` all rows which should be `owner`, like the backfill script does
4) Change role to `NOT NULL`

And I'll remove the backfill script as it won't be needed

server/polar/auth/permission.py
joebon: Fixed

server/polar/user_organization/service.py
joebon: Fixed

server/migrations/versions/2026-05-06-1015_add_role_column_to_user_organizations.py
joebon: Fixed

frankie567 (approved):
Beautiful :)

PR #11452, opened 2026-05-06
refine invoice actions layout

emilwidlund:
Moving away from dropdown for invoice actions.

image

image


psincraian (approved):
This looks nice!

PR #11454, opened 2026-05-06
auth: Only check the relevant auth

Yopi:
Based on the prefix of the key, we can know which type of key it is.

As I understand it, the auth prefixes were introduced:
2024-03-26 - oauth2 prefix
2024-06-18 - gets polar_pat_ prefix
2025-01-02 - checksum is added
2025-02-04 - organization auth tokens are added with polar_oat_ from day one

Since there are no pats older than 2024-06-19 in use, this should be fine to merge.

### DIFF
--- a/server/polar/auth/middlewares.py
+++ b/server/polar/auth/middlewares.py
@@ -23,12 +23,18 @@ from polar.models import (
     UserSession,
 )
 from polar.models.member_session import MEMBER_SESSION_TOKEN_PREFIX
-from polar.oauth2.constants import is_registration_token_prefix
+from polar.oauth2.constants import is_access_token_prefix, is_registration_token_prefix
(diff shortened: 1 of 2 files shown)


frankie567 (approved):
Ah yes, nice! I remember we talked about that when you joined, but decided to wait a bit because of live tokens. Good thing that we can move forward with it now :)

Yopi:
> Ah yes, nice! I remember we talked about that when you joined, but decided to wait a bit because of live tokens. Good thing that we can move forward with it now :)

You might be thinking of the discussion of checking the checksums of the tokens https://github.com/polarsource/polar/pull/8499. I reached out to all of the organizations that have older tokens than that today, so I think we can do that as a next step. This one should be fine to do independently of it, as the token prefixes are older

PR #11459, opened 2026-05-06
subscriptions: Re-anchor on trial conversion

### LINKED ISSUE #11450: Trial-to-paid transition can lead to too long first paid period
The first paid period after a non-monthly-aligned trial spans (trial_end → trial_end + interval) snapped to anchor_day, which can be almost two full intervals long.

## Claude Code walkthrough

1. Trial end is computed without anchor alignment
`server/polar/checkout/service.py:2257` calls `TrialInterval.get_end`, which is a pure `relativedelta` add (`server/polar/kit/trial.py:20-29). For a 7-day trial on a monthly subscription, this gives trial_end = started_at + 7 days = 2026-05-05. Trial end day (5) ≠ anchor day (28). This part is fine on its own.

2. Subscription creation stores anchor_day but uses trial_end as period end
server/polar/subscription/service.py:556-585:

```py
current_period_start = utc_now()                   # 2026-04-28
trial_end = checkout.trial_end                     # 2026-05-05
if trial_end is not None:
    current_period_end = trial_end                 # 2026-05-05
…
subscription.anchor_day = current_period_start.day # 28
```

So while the subscription is trialing, `current_period_end` (=`trial_end`) is intentionally not aligned to `anchor_day`. That's fine — but it means the next `cycle()` call will receive an unaligned `d`.

3. `cycle()` feeds an unaligned `d` into `get_next_period`, which assumes alignment
`server/polar/subscription/service.py:712-721`:

```py
current_period_end = subscription.current_period_end           # 2026-05-05 (trial_end)
subscription.current_period_start = current_period_end         # 2026-05-05
subscription.current_period_end = (
    subscription.recurring_interval.get_next_period(
        current_period_end,            # 2026-05-05  ← NOT anchor-aligned
        subscription.anchor_day,       # 28
        subscription.recurring_interval_count,  # 1
    )
)
```

`server/polar/enums.py:73-78` (for monthly intervals):

```
next = d + relativedelta(months=leap)              # 2026-05-05 + 1mo = 2026-06-05
if next.day != anchor_day:                         # 5 ≠ 28 → True
    _, max_month_day = calendar.monthrange(...)    # June has 30 days
    next = next.replace(day=min(anchor_day, ...))  # 2026-06-28
return next                                        # ← 54 days from 2026-05-05
```

`get_next_period` was designed for the active-cycling case where `d` is already on the anchor day (or its month-end clamp). When `d.day != anchor_day`, it adds a full interval and then jumps to anchor day, effectively "skipping" the anchor day that exists between d and d + 1 interval.

## Proposed fix

I _think_ we should re-anchor the subscription to the trial conversion date so that you get the 30 day billing period after the 7 day trial period, but that may be breaking how it worked before. Ideally, our fix should align with our previous behavior.

Yopi:
Fix #11450

Before 323a537a4, we used return d + relativedelta(months=leap) in get_next_period. After it we introduced the anchor_day to get the correct date to snap to. However for trials this meant that we went too far into the future.

E.g. on trial_start (subscription anchor day) 2024-04-28, with a 7 day trial we converted the trial on 2024-05-05. The get_next_period would then return 2024-06-28, making the first period 1 month + 23 days, instead of 1 month.

This change resets the anchor day to the trial conversion day, which is the same as the behaviour as it was before the anchor day was introduced. With the addition that it will stay the same (i.e. if the anchor day is 31 it will stay 31).

### DIFF
--- a/server/polar/subscription/service.py
+++ b/server/polar/subscription/service.py
@@ -712,6 +712,8 @@ class SubscriptionService:
             if update_cycle_dates and not pending_update_changed_interval:
                 current_period_end = subscription.current_period_end
                 subscription.current_period_start = current_period_end
+                if previous_status == SubscriptionStatus.trialing:
+                    subscription.anchor_day = current_period_end.day
                 subscription.current_period_end = (
                     subscription.recurring_interval.get_next_period(
                         current_period_end,
--- a/server/tests/subscription/test_service.py
+++ b/server/tests/subscription/test_service.py
@@ -1314,6 +1314,38 @@ class TestCycle:
 
         enqueue_email_mock.assert_not_called()
 
+    @freeze_time("2024-04-28 12:00:00")
+    async def test_trial_end_rebases_anchor_to_trial_end_day(
+        self,
+        session: AsyncSession,
+        enqueue_job_mock: MagicMock,
+        enqueue_email_mock: MagicMock,
+        save_fixture: SaveFixture,
+        product: Product,
+        customer: Customer,
+    ) -> None:
+        subscription = await create_trialing_subscription(
+            save_fixture,
+            product=product,
+            customer=customer,
+            trial_interval=TrialInterval.day,
+            trial_interval_count=7,
+            scheduler_locked_at=utc_now(),
+        )
+        assert subscription.anchor_day == 28
+        assert subscription.trial_start == datetime(2024, 4, 28, 12, 0, 0, tzinfo=UTC)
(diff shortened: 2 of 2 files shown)


pieterbeulque (approved):
The logic as described in the test makes sense to me!

frankie567 (approved):
Really nice catch @pieterbeulque and @Yopi 🙏 Sorry to have messed up here

PR #11463, opened 2026-05-06
events: subscription past_due + recovered

Yopi:
Adds an event for subscriptions that are past due and one for subscriptions that later recover.

These events will make it easier for us to answer questions on how likely it is that a subscription recovers from having been past_due

### DIFF
--- a/clients/packages/client/src/v1.ts
+++ b/clients/packages/client/src/v1.ts
@@ -27680,6 +27680,98 @@ export interface components {
       meter: components['schemas']['Meter']
     }
     /**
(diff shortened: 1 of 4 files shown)


pieterbeulque:
>These events will make it easier for us to answer questions on how likely it is that a subscription recovers from having been past_due

oooh

frankie567 (approved):
You're definitely the smartest man in this room 🙂

Yopi:
> You're definitely the smartest man in this room 🙂

Hahah, that is not how I've been feeling this week 😂

PR #11466, opened 2026-05-06
rate-limit: short-circuit blocked callers around auth + DB

joebon:
## Summary

**Related Issue**: #

Move rate-limit checks ahead of auth so a 429 short-circuits before the request pays for a DB session and the full auth resolution stack. A high-volume customer hammering the API was driving CPU spikes despite every response being a 429, because the rate limiter ran *after* `AuthSubjectMiddleware`.

This replaces an earlier sketch in this branch (a separate fast-path middleware with its own block-state in Redis) with the lighter-weight approach Frankie suggested in review: have the existing `RateLimitMiddleware` itself run before auth, and feed it the right identity via a cached `tokenHash → rate_limit_key` mapping written by `AuthSubjectMiddleware` on successful auth.

## What

- `RateLimitMiddleware` is now positioned outside `AsyncSessionMiddleware`, `FlushEnqueuedWorkerJobsMiddleware`, and `AuthSubjectMiddleware`. A 429 returns without acquiring a DB session or running auth.
- `_authenticate` no longer reads `scope["state"]["auth_subject"]` (auth hasn't run yet at that point). Instead it parses the bearer token from raw scope headers, hashes it (blake2b), and looks up `rl:ident:<hash>` in Redis. On miss, it falls back to client IP. With no client info, it falls back to the literal `"anonymous"`.
- `AuthSubjectMiddleware` seeds the cache after successful non-anonymous auth, mapping the token hash to `(rate_limit_user, rate_limit_group)` with a 5-minute TTL.
- A single `rate-limit` Redis client is created at app build time, shared between `RateLimitMiddleware` and `AuthSubjectMiddleware`, and closed in the lifespan shutdown.

## Why

Production CPU spike: a single customer was generating enough rate-limited traffic that the per-request work *before* the 429 (auth subject resolution does up to 5 sequential token lookups, plus sentry/logfire setup and a worker enqueue for token usage tracking) was driving real cost. Returning 429 cheaply turns the bad client into a near-no-op.

## How

- Counters stay where they are: `RateLimitMiddleware`'s own `(zone, identity)` counters remain the single source of truth — no separate block-state to maintain or invalidate.
- Cold start: the very first request from a new token misses the cache, falls through to IP-based rate limiting, completes auth, and seeds the mapping. Every subsequent request from that token is identified by token hash without consulting the DB.
- Token revocation: a revoked token will keep being rate-limited under its old identity for up to the TTL, but `AuthSubjectMiddleware` still rejects with 401 — there's no way to use a revoked token to abuse fresh quota.
- Anonymous traffic and tokens that fail to authenticate keep getting IP-based rate limiting, same as before.

## Checklist

- [x] New functionality is covered by tests (`tests/test_rate_limit.py`, 14 tests)

### DIFF
--- a/server/polar/auth/middlewares.py
+++ b/server/polar/auth/middlewares.py
@@ -40,7 +40,7 @@ from polar.personal_access_token.service import (
     personal_access_token as personal_access_token_service,
 )
 from polar.postgres import AsyncSession
-from polar.rate_limit import write_cached_identity
+from polar.rate_limit import clear_cached_identity, write_cached_identity
 from polar.redis import Redis
 from polar.sentry import set_sentry_user
 from polar.worker import enqueue_job
@@ -204,6 +204,9 @@ class AuthSubjectMiddleware:
         try:
             auth_subject = await get_auth_subject(request, session)
         except OAuth2Error as e:
+            token = get_bearer_token(request)
+            if token is not None:
+                await clear_cached_identity(self.redis, token)
             response = await oauth2_error_exception_handler(request, e)
             return await response(scope, receive, send)
 
@@ -213,7 +216,7 @@ class AuthSubjectMiddleware:
             token = get_bearer_token(request)
             if token is not None:
                 await write_cached_identity(
-                    self.redis, token.encode("ascii"), auth_subject.rate_limit_key
+                    self.redis, token, auth_subject.rate_limit_key
                 )
 
         with logfire.set_baggage(**auth_subject.log_context):
--- a/server/polar/auth/middlewares.py
+++ b/server/polar/auth/middlewares.py
@@ -40,6 +40,8 @@ from polar.personal_access_token.service import (
     personal_access_token as personal_access_token_service,
 )
 from polar.postgres import AsyncSession
+from polar.rate_limit import write_cached_identity
+from polar.redis import Redis
 from polar.sentry import set_sentry_user
 from polar.worker import enqueue_job
 
@@ -187,8 +189,9 @@ async def get_auth_subject(
 
 
 class AuthSubjectMiddleware:
-    def __init__(self, app: ASGIApp) -> None:
+    def __init__(self, app: ASGIApp, redis: Redis) -> None:
         self.app = app
+        self.redis = redis
 
     async def __call__(self, scope: ASGIScope, receive: Receive, send: Send) -> None:
         if scope["type"] != "http":
@@ -205,6 +208,13 @@ class AuthSubjectMiddleware:
             return await response(scope, receive, send)
 
         scope["state"]["auth_subject"] = auth_subject
+
+        if not isinstance(auth_subject.subject, Anonymous):
+            token = get_bearer_token(request)
+            if token is not None:
+                await write_cached_identity(
+                    self.redis, token.encode("ascii"), auth_subject.rate_limit_key
+                )
 
         with logfire.set_baggage(**auth_subject.log_context):
             log.info("Authenticated subject", **auth_subject.log_context)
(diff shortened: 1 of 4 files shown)


frankie567 (commented):
I like the idea but I'm wondering if we could have a slightly different approach that would avoid us to duplicate the rate limit data in Redis.

Roughly maybe we could:

* In `AuthSubjectMiddleware`, after the subject is identified, write a key mapping a hash of the token (similar to what you did) with the rate limit key.
* In `_authenticate` of `rate_limit.py`, read the key from that mapping. If unset, fallback to the IP
* That way, we could move the RateLimitMiddleware _before_ the `AuthSubjectMiddleware`.

Not sure if it's better. What annoys me in all cases is that we have to manually parse the Authorization header. Thoughts? :)

Yopi:
I couldn't add the comment, but the response should be `application/json`, right?

joebon:
> I like the idea but I'm wondering if we could have a slightly different approach that would avoid us to duplicate the rate limit data in Redis.
> 
> Roughly maybe we could:
> 
> * In `AuthSubjectMiddleware`, after the subject is identified, write a key mapping a hash of the token (similar to what you did) with the rate limit key.
> * In `_authenticate` of `rate_limit.py`, read the key from that mapping. If unset, fallback to the IP
> * That way, we could move the RateLimitMiddleware _before_ the `AuthSubjectMiddleware`.
> 
> Not sure if it's better. What annoys me in all cases is that we have to manually parse the Authorization header. Thoughts? :)


Good idea, a lot cleaner than having N number of different rate limiting middlewares sprinkled everywhere.

And yes, I agree re: auth header parsing. We're in a middleware so it is what it is to some extent but I'll see what I can do about it.

joebon:
Is this closer to what you had in mind @frankie567 ?

server/polar/auth/middlewares.py
```diff
@@ -201,11 +204,21 @@ async def __call__(self, scope: ASGIScope, receive: Receive, send: Send) -> None
...
 
         scope["state"]["auth_subject"] = auth_subject
 
+        if not isinstance(auth_subject.subject, Anonymous):
+            token = get_bearer_token(request)
+            if token is not None:
+                await write_cached_identity(
+                    self.redis, token, auth_subject.rate_limit_key
+                )
+
```
frankie567, line 229: The only missing part here is the web session case, which has generous rate limits compared to anonymous users.

This web session is read from a cookie, so we won't match a bearer token. I guess we could also simply check for the presence of that cookie?

server/polar/rate_limit.py
```diff
@@ -75,8 +135,12 @@ def get_middleware(app: ASGIApp) -> RateLimitMiddleware:
         case _:
             rules = {}
     return RateLimitMiddleware(
-        app, _authenticate, RedisBackend(create_redis("rate-limit")), rules
+        app, partial(_authenticate, redis=redis), RedisBackend(redis), rules
```
frankie567, line 166: I remember being frustrated this middleware expects their redis instance upfront instead of being able to fetch it from the current scope. We would have been able then to use the lifespan redis instead of maintaining a separate one. But it would require to rewrite it, so well, it is how it is :D

frankie567 (commented):
Yep, looks great like this! Just the case of web sessions we need to account for and that should be good to go :)

server/polar/auth/middlewares.py
joebon: Great catch! Added a suggestion, wdyt?

frankie567: Awesome!

frankie567 (approved):

PR #11467, opened 2026-05-06
rate-limit: share + close the rate-limit Redis client via lifespan

joebon:
## Summary

**Related Issue**: #

Stacked on top of #11466. Closes the Redis-client lifecycle leak flagged in that PR's review by sharing one client between both rate-limit middlewares and disposing it in the lifespan shutdown.

## What

- `rate_limit.get_middleware` now takes a `redis: Redis` parameter instead of calling `create_redis("rate-limit")` internally.
- `create_app()` creates one `rate_limit_redis` client and passes it to both `rate_limit.get_middleware` (post-auth limiter) and `RateLimitFastPathMiddleware`.
- The client is stashed on `app.state.rate_limit_redis` so the lifespan shutdown can close it alongside the existing `app` Redis client.

## Why

Before this change, both rate-limit middlewares each constructed their own `create_redis("rate-limit")` client at `create_app()` time and never closed them on shutdown — two leaked connection pools per app lifecycle. Annoying in production (slow shutdown, pool exhaustion under reload loops), noisy in tests/dev with auto-reload.

The fix is the "construct once, close in lifespan" pattern the existing `redis = create_redis("app")` already uses (`server/polar/app.py:154,183`).

## How

One Redis client is enough — both middlewares connect with the same `client_name="<env>.rate-limit"` and share the same key namespace, so a single connection pool is correct. `get_middleware` was the awkward outlier (factory hiding its Redis dependency); refactoring it to take the client as a parameter mirrors how `RateLimitFastPathMiddleware` is already wired and keeps the lifecycle visible at the call site.

Lifespan shutdown closes it via `getattr(app.state, "rate_limit_redis", None)` — testing builds skip the rate-limit middlewares entirely, so the attribute may be absent.

## Checklist

### DIFF
--- a/server/polar/app.py
+++ b/server/polar/app.py
@@ -181,6 +181,9 @@ async def lifespan(app: FastAPI) -> AsyncIterator[State]:
     stop_remote_write_pusher()
 
     await redis.close(True)
+    rate_limit_redis = getattr(app.state, "rate_limit_redis", None)
+    if rate_limit_redis is not None:
+        await rate_limit_redis.close(True)
     await async_engine.dispose()
     if async_read_engine is not async_engine:
         await async_read_engine.dispose()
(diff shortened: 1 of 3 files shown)


joebon:
Superseded by the rewrite of #11466 — the new approach already creates one shared rate-limit Redis client at app build time and closes it in the lifespan shutdown, so this PR has nothing left to add.

PR #11472, opened 2026-05-06
ConfirmModal: trim whitespace when matching confirm prompt

pieterbeulque:
Customers were getting locked out of deletion flows when the source
identifier (e.g., organization slug) had trailing whitespace they didn't
type. Trim both sides of the comparison so the match is edit-proof.


Yopi (approved):
Makes sense

PR #11487, opened 2026-05-07
chore: Dedicated queue for PDF generation jobs step 2

joebon:
# Dedicated queue for PDF generation jobs

## Why

Currently, if enough pdf generation jobs (such as for instance `order.invoice`) is processed concurrently the worker machine runs out of memory.

## What

Move `order.invoice` and `receipt.render` jobs to a dedicated queue `invoices_and_receipts`, consumed by dedicated workers dimensioned for high(er) memory usage but still to not OOM.

## How

Multi-step release plan
1. ~Update existing workers to start consuming a new queue~ **Done**
2. Duplicate `order.invoice` => `order.invoice.v2` and `receipt.render` => `receipt.render.v2` and enqueue the new variants to the new queue **(this step)**
3. Create a new worker in sandbox that consumes only `invoices_and_receipts`
4. Update the regular worker in sandbox so it stops consuming the new queue
5. Verify this solves OOM issues in sandbox
6. Create a new worker in production that consumes only `invoices_and_receipts`
7. Update the low priority worker in production so it stops consuming the new queue
8. Verify this solves OOM issues in production
9. Rename jobs back to `order.invoice` and `receipt.render` (keeping `.v2` around a while to drain in queue/flight jobs)
10. Remove `.v2` entirely

### DIFF
--- a/server/polar/order/tasks.py
+++ b/server/polar/order/tasks.py
@@ -19,6 +19,7 @@ from polar.worker import (
     AsyncSessionMaker,
     CronTrigger,
     TaskPriority,
+    TaskQueue,
     actor,
     can_retry,
     enqueue_job,
@@ -212,7 +213,11 @@ async def order_confirmation_email(order_id: uuid.UUID) -> None:
         await order_service.send_confirmation_email(session, order)
 
 
-@actor(actor_name="order.invoice", priority=TaskPriority.LOW)
+@actor(
+    actor_name="order.invoice",
+    priority=TaskPriority.LOW,
+    queue_name=TaskQueue.INVOICES_AND_RECEIPTS,
+)
 async def order_invoice(order_id: uuid.UUID) -> None:
     async with AsyncSessionMaker() as session:
         repository = OrderRepository.from_session(session)
--- a/server/polar/order/tasks.py
+++ b/server/polar/order/tasks.py
@@ -213,12 +213,7 @@ async def order_confirmation_email(order_id: uuid.UUID) -> None:
         await order_service.send_confirmation_email(session, order)
 
 
-@actor(
-    actor_name="order.invoice",
-    priority=TaskPriority.LOW,
-    queue_name=TaskQueue.INVOICES_AND_RECEIPTS,
-)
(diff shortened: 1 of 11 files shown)


server/polar/order/tasks.py
```diff
@@ -212,7 +213,11 @@ async def order_confirmation_email(order_id: uuid.UUID) -> None:
         await order_service.send_confirmation_email(session, order)
 
 
-@actor(actor_name="order.invoice", priority=TaskPriority.LOW)
+@actor(
+    actor_name="order.invoice",
+    priority=TaskPriority.LOW,
+    queue_name=TaskQueue.INVOICES_AND_RECEIPTS,
+)
```
frankie567: This will break in-flight `order.invoice` since the worker will try to look for an `order.invoice` job on the default queue and won't find it.

We need to duplicate the job and call e.g. `order.invoice.v2` for the new queue, and start to enqueue this one in the code base.

After this is deployed, we can wait a few hours, remove the old one and swap v2.

server/polar/receipt/tasks.py
```diff
@@ -33,6 +34,7 @@ def __init__(self, order_id: uuid.UUID) -> None:
 @actor(
     actor_name="receipt.render",
     priority=TaskPriority.LOW,
+    queue_name=TaskQueue.INVOICES_AND_RECEIPTS,
```
frankie567: Same comment :)

frankie567 (changes requested):

server/polar/order/tasks.py
joebon: Ah, crap! Thanks, you're right. I'll amend. This process just got even longer...

joebon (commented):
Addressed in 57f0d8d0a — added v2 actors (`order.invoice.v2`, `receipt.render.v2`) on the new queue alongside the originals, and switched all enqueue call sites + the in-actor retry to v2. v1 actors stay on the default queue to drain in-flight messages. Follow-up PR will remove v1 once `low_priority` is empty of these jobs.

psincraian (approved):

frankie567 (approved):

Issue #11424, opened 2026-05-05
Payout transaction delay


frankie567:
RFC for https://github.com/polarsource/feedback/issues/136

We want to implement a payout transaction delay at Order level, i.e. an Order that just got paid shouldn't be immediately available for payout.

The proposal is to add a configurable delay at Account level to determine the minimum age an Order's payment is available for payout.

## Proposal

1. Introduce a `payout_transaction_delay` property on Account, with a sane default (7 days?)
2. Introduce logic to exclude Account transactions that are more recent than that delay
3. Change `/v1/transactions/summary` and `/v1/payouts/estimate` so they output the balance available for payout based on this logic
4. Change payout creation to only include available payout balance
5. Tweak UI to show the balance available for payout, and mark the "waiting" transactions in the table
6. Add backoffice feature to change the payout transaction delay value

## Future expansions

With that approach, it'll be fairly easy in the future to add a per-transaction override by adding a `payout_delay` property at the transactions level — and leverage the same mechanics to exclude it from payouts.

frankie567:
@pieterbeulque Do you think I correctly covered the need here? :)

pieterbeulque:
Yep, I believe that's it! Ideally, `/v1/payouts/estimate` returns both balances, the one you can withdraw and the one that's still "frozen", so that you have the option to show both in the UI.

frankie567:
> Yep, I believe that's it! Ideally, `/v1/payouts/estimate` returns both balances, the one you can withdraw and the one that's still "frozen", so that you have the option to show both in the UI.

Yep, that's what I thought :)

pieterbeulque:
How granular are we storing the delay? Could it be hours too or is it N days?

frankie567:
> How granular are we storing the delay? Could it be hours too or is it N days?

Stored as a SQL interval / Python timedelta, so we can go up to the second 😄

PR #11494, opened 2026-05-07
Payout Delays logic

frankie567:
Part of #11424

- Logic for locking payout of transactions that are more recent than the configured Account delay
- API to return the Account delay and the current available balance
- UI to show the available balance, configured delay and state on each transaction (see video below). **Feedback welcome 🙏**

<div>
    <a href="https://www.loom.com/share/b4c3892281f043beac7994f1ee941609">
      <p>UI Review: Payout Delays and Statuses - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/b4c3892281f043beac7994f1ee941609">
      
    </a>
  </div>

### DIFF
--- a/clients/apps/web/src/components/Payouts/AccountBalance.tsx
+++ b/clients/apps/web/src/components/Payouts/AccountBalance.tsx
@@ -5,6 +5,13 @@ import { Modal } from '@/components/Modal'
 import { schemas } from '@polar-sh/client'
 import { formatCurrency } from '@polar-sh/currency'
 import Button from '@polar-sh/ui/components/atoms/Button'
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipTrigger,
+} from '@polar-sh/ui/components/ui/tooltip'
+import InfoOutlined from '@mui/icons-material/InfoOutlined'
+import { ISODuration } from '@/utils/duration'
 import React, { useCallback } from 'react'
 import { useModal } from '../Modal/useModal'
 import { Well, WellContent, WellFooter, WellHeader } from '../Shared/Well'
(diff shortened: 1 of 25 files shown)


pieterbeulque:
Some design feedback:

I'd flip the numbers as it feels more in line with bank apps and other dashboards.

**Balance**
$[Available balance]
Total balance: $[Total balance] `(i)`

Let's make the font-size on the total balance the same as the "Minimum withdrawal amounts apply" line.

Then for the table, I think the badge is great, but would use a neutral color (blue) instead of orange, as it implies something may be wrong (e.g. a past due payment) while the money is already settled. From the same lens, I'd call the label something like "Upcoming" or "Soon available" (or, from a bank term, "Clearing"? though I don't like it that much) instead of "Waiting" (feels like something can go wrong still)

pieterbeulque:
Three more:

- Maybe rename the card "Available balance" for extra clarity, and linking "Available" there to the badge in the table is helpful
- Right-align the "Status" column
- Make sure the datetime in the first column doesn't wrap to 2 lines

frankie567:
@copilot Please address the two comments of @pieterbeulque above. 
Besides, Since we'll put the available balance on front on the balance card, for consistency, we should also display that value on the home widgets and the mobile app widgets where we only shows the total balance currently.

emilwidlund:
> Some design feedback:
> 
> I'd flip the numbers as it feels more in line with bank apps and other dashboards.
> 
> **Balance** 
> $[Available balance] 
> Total balance: $[Total balance] `(i)`

I think we need to denote that the big number is the "Available Balance" somewhere, maybe in the title. Also - I'd like to try not have the (i) icon there, as it's always a bad escape hatch. Would love to find a way around that.

emilwidlund:
Nice work. I'm a little reluctant to call the status "Waiting". It sounds like it is waiting for something, and not that is a blocked resource. Not sure what would be a better fit. Maybe "On Hold" or "Locked".

frankie567:
Thank you for feedback, pals :) 

Here is the new version:

Capture d’écran 2026-05-11 à 08 57 49

---

* "Available Balance" title with the available balance amount in front
* Total balance below
* "Available soon" balance below, with the delay shown
    * 👉 Not 100% convinced, since part of this amount can be available before 7 days depending on the age of the transaction, but could be enough
* Changed "Waiting" status to "On hold", with a blue badge

clients/apps/web/src/components/Payouts/AccountBalance.tsx
```diff
@@ -115,7 +144,7 @@ const AccountBalance: React.FC<AccountBalanceProps> = ({
           <div className="text-4xl">
             {isLoading ? (
               <div className="animate-pulse rounded-lg bg-gray-200 text-gray-200">
-                &nbps;
```
joebon, line 118: 😺

joebon (approved):

clients/apps/web/src/components/Payouts/AccountBalance.tsx
frankie567: Credits to the LLM for catching that one :p

PR #11496, opened 2026-05-07
Support seat-based billing in AI create flow

pieterbeulque:
It works over MCP but the system prompt guards against it.


joebon (approved):
LGTM

PR #11500, opened 2026-05-07
Gate design system tokens based on usage

sebastianekstrom:
Currently you can do something like this:

```tsx
<Box backgroundColor="text-primary" />
<Box backgroundColor="border-secondary" />
<Text color="background-success" />
```

This PR fixes that by only allowing text/bg/border specific tokens to their intended props:

```tsx
// 🚨 Error
<Box backgroundColor="text-primary" />

// ✅ All good
<Box backgroundColor="background-primary" />
```

As well as making the autocomplete easier to parse since it only contains the relevant tokens:

Screenshot 2026-05-07 at 16 37 21

### DIFF
--- a/clients/apps/web/src/components/Onboarding/IntegrateStep.tsx
+++ b/clients/apps/web/src/components/Onboarding/IntegrateStep.tsx
@@ -400,7 +400,7 @@ const FrameworkCard = ({
         borderWidth={1}
         borderStyle="solid"
         borderColor={active ? 'border-secondary' : 'border-primary'}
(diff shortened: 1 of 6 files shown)


emilwidlund (approved):

pieterbeulque:
That's great, but then why do the token values still need the `background-` / `border-` / `text-` prefix?

sebastianekstrom:
@pieterbeulque that's a fair point, I opted for that to make it more explicit, but perhaps it's just redundant... I'll make a follow up to fix that!

PR #11510, opened 2026-05-08
feat(organization): add snoozed_until and snooze_type schema

psincraian:
## Summary

Some orgs get back to the review after 24 because they have lot of sales. I plan to add a way to snooze them for X days. 

This PR adds the migration:
- Adds `snoozed_until` (nullable `TIMESTAMP`) and `snooze_type` (nullable `VARCHAR`) columns on `organizations`.
- Adds the `SnoozeType` enum with `time_based` and `next_sale` modes.
- Migration backfills existing snoozed organizations with `next_sale` and `status_updated_at + 24h` so live behavior is unchanged.

The feature code that reads/writes these columns ships in a follow-up PR. Splitting per the repo's migration-isolation policy.

## Test plan

- [x] Backend lint + mypy clean
- [x] `alembic upgrade head` resolves to a single head (`9b1f3a4c2d8e` chained off `01607ddf3d7f`)

### DIFF
--- a/server/migrations/versions/2026-05-07-1000_add_organization_snooze_until_and_type.py
+++ b/server/migrations/versions/2026-05-07-1000_add_organization_snooze_until_and_type.py
@@ -1,0 +1,55 @@
+"""add organization snoozed_until + snooze_type columns
+
+Revision ID: 9b1f3a4c2d8e
+Revises: 01607ddf3d7f
+Create Date: 2026-05-07 10:00:00.000000
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# Polar Custom Imports
+
+# revision identifiers, used by Alembic.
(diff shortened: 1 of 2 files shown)


pieterbeulque:
>Some orgs get back to the review after 24 because they have lot of sales. I plan to add a way to snooze them for X days.

Sorry to push back on this, but isn't this a good thing? From a risk perspective, we'd want to be aware of that, no?

psincraian:
Uff, but the operations burden is there. For example, I have https://backoffice.polar.sh/organizations/d04b51a7-22dd-4944-b91b-1ec268a0ee3f that we will not review until next week and it appeared yesterday and will appear today again in the review.

Would be nice to have a capability that for some orgs, we push the review for X days (1 default and less than 7).

pieterbeulque (approved):

PR #11516, opened 2026-05-08
Check slug availability separately from organization creation

pieterbeulque:
Splitting into POST + PATCH allowed users to work around the onboarding AUP validation by abandoning the onboarding on step 3 and logging back in.

Also started tracking the AUP validation results in Posthog (for now) so that I can see a bit how we're performing, we shouldn't be flying blind there.

### DIFF
--- a/clients/apps/web/src/app/(main)/onboarding/validate-description/route.ts
+++ b/clients/apps/web/src/app/(main)/onboarding/validate-description/route.ts
@@ -1,10 +1,11 @@
-import { CONFIG } from '@/utils/config'
+import { getAuthenticatedUser } from '@/utils/user'
 import { createOpenAI } from '@ai-sdk/openai'
+import { withTracing } from '@posthog/ai'
 import { generateText, Output } from 'ai'
 import { readFileSync } from 'node:fs'
 import { join } from 'node:path'
-import { cookies } from 'next/headers'
 import { NextResponse } from 'next/server'
+import { PostHog } from 'posthog-node'
 import { z } from 'zod'
 import * as Sentry from '@sentry/nextjs'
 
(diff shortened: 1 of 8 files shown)


server/polar/organization/schemas.py
```diff
@@ -82,6 +82,31 @@ def validate_reserved_keywords(value: str) -> str:
...
+OrganizationSlugUnavailableReason = Literal[
+    "format", "reserved", "blocked", "taken"
+]
+
+
+class OrganizationSlugAvailability(Schema):
+    available: bool = Field(
+        description="Whether the slug is available for a new organization."
+    )
+    reason: OrganizationSlugUnavailableReason | None = Field(
```
Yopi: Do we want to leak this, or does it not matter?

clients/apps/web/src/app/(main)/onboarding/validate-description/route.ts
```diff
@@ -18,6 +19,12 @@ const openai = createOpenAI({
   baseURL: 'https://gateway-us.pydantic.dev/proxy/chat/',
 })
 
+const phClient = process.env.NEXT_PUBLIC_POSTHOG_TOKEN
```
Yopi, line 22: Should we track this in Posthog, or rely on Polar? Don't know if Posthog gives us something more

Yopi (approved):
Cool! Nice work

clients/apps/web/src/app/(main)/onboarding/validate-description/route.ts
pieterbeulque: It was out of laziness, ideally Polar, yes (but the organization doesn't exist at this point, so a bit hard to track properly)

server/polar/organization/schemas.py
pieterbeulque: Good point, we shouldn't.

clients/apps/web/src/app/(main)/onboarding/validate-description/route.ts
Yopi: We can ingest-- oh yeah. No that's true. Arguments for having a list of external_customer_ids linked to a single customer? 😄

pieterbeulque: I'd love that!

PR #11517, opened 2026-05-08
Fix PWYW amount validation

pieterbeulque:
We had a hardcoded maximum in client-side validation that wasn't currency-aware. This PR drops that.

We also have an unused field `maximum_amount` that still had a matching ceiling that was currency un-aware. That doesn't really matter, but I made it re-use the existing `validate_price_amount` just as a clean house kind of thing.

### DIFF
--- a/server/polar/product/schemas.py
+++ b/server/polar/product/schemas.py
@@ -194,7 +194,6 @@ class ProductPriceCustomCreate(ProductPriceCreateBase):
     )
     maximum_amount: PriceAmount | None = Field(
         default=None,
-        le=1_000_000,  # $10K
         description="The maximum amount the customer can pay.",
     )
     preset_amount: PriceAmount | None = Field(
@@ -208,7 +207,7 @@ class ProductPriceCustomCreate(ProductPriceCreateBase):
         ),
     )
 
-    @field_validator("minimum_amount", "preset_amount")
+    @field_validator("minimum_amount", "preset_amount", "maximum_amount")
     @classmethod
     def validate_amount_not_in_minimum_gap(
         cls, v: int | None, info: ValidationInfo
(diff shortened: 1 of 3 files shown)


server/polar/product/schemas.py
```diff
@@ -208,7 +207,7 @@ class ProductPriceCustomCreate(ProductPriceCreateBase):
         ),
     )
 
-    @field_validator("minimum_amount", "preset_amount")
+    @field_validator("minimum_amount", "preset_amount", "maximum_amount")
     @classmethod
     def validate_amount_not_in_minimum_gap(
```
joebon, line 212: From Claude:
Validator name is now misleading.
`validate_price_amount` checks both the currency minimum and the    
  global maximum. The name was already a bit narrow; with `maximum_amount` added it's actively wrong. Consider renaming to something like `validate_amount_within_currency_bounds.

server/polar/product/schemas.py
joebon, line 222: From Claude:
`allow_zero=True` doesn't make sense for `maximum_amount`. `PriceAmount` (`Annotated[int, Field(ge=1)]`) already rejects 0 at the type level, so `allow_zero` is dead for maximum_amount in practice — but it reads as if "max can be zero," which is nonsensical. It would be cleaner to call `validate_price_amount` with `allow_zero=False` for `maximum_amount` and `allow_zero=True` for `minimum_amount`/`preset_amount`. Either branch on the field name (`info.field_name`) or split the        
  validator. Minor, but worth tightening if you're touching this.

joebon (approved):
Nice! Added some optional nit picks from claude

pieterbeulque:
@joebon Those are valid points, but since maximum_amount isn't even used right now, I'm going to be pragmatic here 😁 thanks for the review

PR #11522, opened 2026-05-08
polar: wire up subscriptions

Yopi:
This PR wires up subscriptions with our internal API and our external API.

It adds a listing of (visible) products, subscriptions and creating checkouts.
It wires up the mock billing page with these APIs and enable checking out a subscription

### DIFF
--- a/server/polar/integrations/polar/service.py
+++ b/server/polar/integrations/polar/service.py
@@ -18,7 +18,7 @@ from polar.worker import enqueue_job
 from .client import get_client
 
 if TYPE_CHECKING:
-    from polar_sdk.models import BenefitGrant, Customer
+    from polar_sdk.models import BenefitGrant, Checkout, Customer, Product, Subscription
 
 
 BenefitGrantWebhookPayload = (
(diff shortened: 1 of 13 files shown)


emilwidlund:
Just made a slight adjustment to the Empty State. Otherwise all good!

emilwidlund (approved):

server/polar/integrations/polar/service.py
```diff
@@ -136,6 +156,71 @@ def enqueue_track_organization_review_usage(
             cost_usd=str(cost_decimal),
         )
 
+    async def list_plans(self) -> list["Product"]:
```
frankie567, line 159: Maybe premature optimization: could we cache it?

server/polar/integrations/polar/service.py
```diff
@@ -136,6 +156,71 @@ def enqueue_track_organization_review_usage(
...
+            raise PolarSelfNotConfigured()
+        await self._ensure_plan(product_id)
+        subscription = await get_client().get_active_subscription(
+            external_customer_id=str(organization_id)
+        )
+        if subscription is None:
+            raise PolarSelfNoActiveSubscription(organization_id)
+        return await get_client().update_subscription_product(
+            subscription_id=subscription.id,
+            product_id=product_id,
```
frankie567, line 221: We let the default proration behavior then?

frankie567 (approved):

server/polar/integrations/polar/service.py
Yopi: We can if you think that would be needed. I was thinking that a customer would probably do the same as this and just call the API. We could also add caching in a follow up PR

server/polar/integrations/polar/service.py
Yopi: Open for discussion I think, but yes. Not entirely sure what we would like to see here, so it's something we should think through

server/polar/integrations/polar/service.py
frankie567: Fine by me :)

PR #11530, opened 2026-05-08
Periodic cleanup of expired OAuth2 tokens

pieterbeulque:
(sorry, this shouldn't've created a PR, but wanted to ask if this is something we should/could/need to do?)

## Summary

The MCP "create product with AI" flow exchanges the user's session for a 1h OAuth2 access token via the custom `web` grant on every cache miss (cookie cached for `expires_in` seconds). The grant doesn't issue a refresh token, so each issuance is a brand-new row in `oauth2_tokens`, and there's no job that ever deletes them — `OAuth2TokenService.get_by_access_token` only checks `is_expired()` at lookup time, leaving expired rows behind forever.

This adds a daily cron actor that prunes safely-deletable expired tokens, mirroring the existing pattern from `customer_session.delete_expired` and `auth.delete_expired`.

### What gets deleted

A row is deleted only if **all** of the following hold:

1. The access token is past expiry: `issued_at + expires_in < now()` (`expires_at` is a Python property, not a column, so the SQL has to expand it).
2. There is no usable refresh path: `refresh_token IS NULL OR refresh_token_revoked_at != 0`. This preserves third-party OAuth apps that still have a valid refresh token they could use to mint a new access token.
3. The token's `client_id` is not the legacy Polar app client (`APP_CLIENT_ID`). The service intentionally accepts expired tokens from that client as a temporary workaround for an unfixed mobile auth flow — deleting those rows would break the workaround.

### Files

- `server/polar/oauth2/repository.py` (new) — `OAuth2TokenRepository.delete_expired(*, exclude_client_ids)`.
- `server/polar/oauth2/service/oauth2_token.py` — `OAuth2TokenService.delete_expired` wires the `APP_CLIENT_ID` exclusion.
- `server/polar/oauth2/tasks.py` (new) — `oauth2_token.delete_expired` cron actor (`hour=0, minute=0`, `TaskPriority.LOW`, `max_retries=0`).
- `server/polar/tasks.py` — registers the new tasks module.
- `server/tests/oauth2/service/test_oauth2_token.py` — `TestDeleteExpired` covering the four cases above.

## Test plan

- [ ] CI runs `TestDeleteExpired` (local execution blocked by a pre-existing Python 3.14rc2 / pydantic-ai import failure in this sandbox; ruff passes on all touched files).
- [ ] After deploy, confirm the cron actor appears in the scheduler logs at 00:00 UTC and that `oauth2_tokens` row count stops growing for client_id values without refresh tokens.

https://claude.ai/code/session_01LJttmtBWXZwktjxLRJnHpt

---
_Generated by [Claude Code](https://claude.ai/code/session_01LJttmtBWXZwktjxLRJnHpt)_

### DIFF
--- a/server/polar/oauth2/repository.py
+++ b/server/polar/oauth2/repository.py
@@ -1,0 +1,26 @@
+import time
+from collections.abc import Sequence
+
+from sqlalchemy import delete, or_
+
+from polar.kit.repository import RepositoryBase
+from polar.models import OAuth2Token
+
+
+class OAuth2TokenRepository(RepositoryBase[OAuth2Token]):
+    model = OAuth2Token
+
+    async def delete_expired(self, *, exclude_client_ids: Sequence[str] = ()) -> None:
+        now = int(time.time())
+        statement = delete(OAuth2Token).where(
+            OAuth2Token.issued_at + OAuth2Token.expires_in < now,
+            or_(
+                OAuth2Token.refresh_token.is_(None),
+                OAuth2Token.refresh_token_revoked_at != 0,
+            ),
+        )
+        if exclude_client_ids:
+            statement = statement.where(
+                OAuth2Token.client_id.notin_(exclude_client_ids)
+            )
+        await self.session.execute(statement)
--- a/server/polar/oauth2/service/oauth2_token.py
+++ b/server/polar/oauth2/service/oauth2_token.py
@@ -13,6 +13,7 @@ from polar.kit.crypto import get_token_hash
 from polar.kit.services import ResourceServiceReader
 from polar.logging import Logger
 from polar.models import OAuth2Token, User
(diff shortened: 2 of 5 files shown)


frankie567 (approved):

pieterbeulque:
@Yopi Do you have thoughts on the usefulness of this?

Yopi:
> @Yopi Do you have thoughts on the usefulness of this?

You mean out of performance / size reasons? It looks like we don't have a ton of entries here, but it also looks like we have no indexes for lookup so it can make sense to keep the table small

pieterbeulque:
Okay, I guess there's no harm 😊

frankie567:
Makes sense to clean this is up. No point in keeping stale tokens 🌍 💚

PR #11538, opened 2026-05-11
fix: recompute organization balance on refund and chargeback reversals

psincraian:
## Summary

Update organization.total_balance when a refund happens

## Test plan

- [x] New unit test: `tests/transaction/service/test_balance.py::TestCreateReversalBalance::test_recomputes_organization_total_balance`
- [x] Existing tests pass (`tests/transaction/service/`, `tests/refund/`, `tests/dispute/`)
- [x] `task lint` + `task lint_types` clean
- [x] E2E in local dev: subscription checkout → renewal → refund #1 → refund #2. Backoffice "Current Balance" matched DB at every step.

### DIFF
--- a/server/polar/organization/tasks.py
+++ b/server/polar/organization/tasks.py
@@ -77,6 +77,23 @@ async def organization_unsnooze_expired() -> None:
         await organization_service.unsnooze_expired_organizations(session)
 
 
+@actor(actor_name="organization.check_threshold", priority=TaskPriority.LOW)
+async def organization_check_threshold(account_id: uuid.UUID) -> None:
+    """Refresh the cached ``total_balance`` for the organization owning
+    ``account_id`` and re-evaluate the review threshold.
+
+    Enqueued by the transaction layer after balance / reversal-balance rows
+    are written, so the transaction service does not need to call into
+    organization service logic synchronously.
+    """
+    async with AsyncSessionMaker() as session:
+        repository = OrganizationRepository.from_session(session)
+        organization = await repository.get_by_account(account_id)
+        if organization is None:
+            return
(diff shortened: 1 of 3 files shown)


frankie567 (commented):
I'm a bit concerned that we tie the transactions service to more high-level logic like `check_review_threshold`. Could we find other ways to achieve this?

psincraian:
@frankie567 I thought about this. I think the clean fix would be to lean more into pub/sub, but that's quite an architectural change. So for now, I'd add a new organization.check_threshold task and enqueue it from the transaction layer instead of the sync call. That way the transaction service doesn't import or call into the org service anymore, which addresses the coupling you flagged.

on the long run I'd love to lean on SystemEvent as a real bus. We already emit order.paid, order.refunded, ... but there are no subscribers. The "check for review" logic would naturally fit as a listener on a few of those. Happy to have a chat about it if interesting.

psincraian:
@frankie567 thought about this a bit. The clean fix would be to lean more into pub/sub, but that's quite an architectural change. So for now, I added a new `organization.check_threshold` task and enqueue it from the transaction layer instead of the sync call. That way the transaction service doesn't import or call into the org service anymore, which addresses the coupling you flagged. It also matches the existing pattern (e.g. `organization.under_review` is already enqueued the same way).

Longer-term I'd love to lean on `SystemEvent` as a real bus. We already emit `order.paid`, `order.refunded`, `balance.refund`, etc., but nothing else fans out from them. The "check for review" logic would naturally fit as a listener on a few of those. Happy to have a chat about it if interesting.

Pushed as `0c576c484`.

frankie567 (approved):

PR #11555, opened 2026-05-11
Make duplicate product media more robust

sebastianekstrom:
We have gotten a few reports that duplicating a product that has images sometimes causes errors, making the duplication fail. This PR aims to make that a bit more robust by:

* Adding 3x retries for uploads
* Don't error out if a subset of the images fail to upload
* Add Sentry captions for errors


pieterbeulque (approved):
Such a frustrating issue, appreciate the attempts!

PR #11556, opened 2026-05-11
Fix discount calculation for seat subscription updates

pieterbeulque:
## Summary

**Related Issue**: 

Fix incorrect discount application in seat-based subscription updates. Previously, discounts were applied to the delta amount, which caused fixed discounts to be double-counted. Now discounts are applied to the effective amounts (base amount minus discount) before calculating the delta.

## What

Changed the discount calculation logic in `_generate_seats_subscription_update()` to:
1. Calculate discounts on the old and new effective amounts separately
2. Compute the delta from these effective amounts
3. Track only the change in discount amount (not the full discount) as the entry discount

Updated corresponding test expectations to reflect the corrected behavior.

## Why

The previous implementation applied discounts to the delta amount, which caused issues with fixed discounts:
- For a seat increase from 1 to 2 seats ($50 → $100) with a $10 fixed discount:
  - Old effective: $50 - $10 = $40
  - New effective: $100 - $10 = $90
  - Delta should be: $90 - $40 = $50 (full incremental charge)
  - Previous code incorrectly applied the $10 discount to the $50 delta, resulting in a $40 charge

Fixed discounts apply once to the subscription total, not to every seat change. The discount amount doesn't change in the delta calculation—only the base amount changes.

## How

1. Calculate `old_discount_amount` and `new_discount_amount` upfront based on the full base amounts
2. For reset mode: use the new effective amount (base - discount) as the charge
3. For proration mode: calculate delta from effective amounts and track only the discount change
4. Apply proration factor to the discount change, not the full discount amount

## Checklist

- [x] This PR addresses a single concern (discount calculation fix)
- [x] Tests updated to reflect corrected behavior

https://claude.ai/code/session_01U4NTDNFmfREunQASW2igk6

### DIFF
--- a/server/polar/subscription/update.py
+++ b/server/polar/subscription/update.py
@@ -207,13 +207,30 @@ def _generate_seats_subscription_update(
     old_base_amount = seat_price.calculate_amount(old_seats)
     new_base_amount = seat_price.calculate_amount(new_seats)
 
+    # Apply the discount to the old and new effective amounts, not to the delta.
+    # Computing the discount on the delta double-counts fixed discounts: a $10
+    # fixed discount applies once to the subscription total, not again to every
+    # seat change.
+    old_discount_amount = 0
+    new_discount_amount = 0
+    if subscription.discount and subscription.discount.is_applicable(
+        subscription.product, subscription.currency
+    ):
+        old_discount_amount = subscription.discount.get_discount_amount(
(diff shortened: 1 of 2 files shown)


frankie567 (approved):
Always bitten by fixed discounts :( Nice job!

PR #11558, opened 2026-05-11
fix(receipt): fall back to Customer.display_name to avoid None crash

pieterbeulque:
## Summary

- `Receipt.from_order` previously resolved `customer_name` via `billing_name → customer.name → customer.email` and then asserted non-`None`. Team customers can have all three values be `None`, which crashed receipt rendering with `AssertionError` and triggered the retry loop on the `invoices_and_receipts` worker.
- Reuse `Customer.display_name`, which already encodes the intended fallback (`name → email → "Team Customer"`), so receipt generation degrades gracefully. The redundant `assert` is dropped because `display_name` is `str` (never `None`).
- All other behavior is preserved, including the `customer_additional_info` logic that avoids duplicating the email when it's already used as `customer_name`.

## Test plan

- [x] Added unit tests in `tests/receipt/test_generator.py` covering the four fallback cases: billing_name preferred, customer.name fallback, email fallback (not duplicated in additional info), and the "Team Customer" fallback when all identifiers are `None`.
- [ ] CI passes.

Note: tests couldn't be executed locally in this worktree due to an unrelated Python 3.14.0rc2 + `pydantic_ai` incompatibility that fails at module import time, before any test runs. Syntax was verified via `ast.parse`.

https://claude.ai/code/session_013BBWCtvryoCJKYJm1cpAr4

---
_Generated by [Claude Code](https://claude.ai/code/session_013BBWCtvryoCJKYJm1cpAr4)_

### DIFF
--- a/server/polar/receipt/generator.py
+++ b/server/polar/receipt/generator.py
@@ -103,10 +103,7 @@ class Receipt(Invoice):
     ) -> Self:
         assert order.receipt_number is not None
 
-        customer_name = (
-            order.billing_name or order.customer.name or order.customer.email
-        )
-        assert customer_name is not None
+        customer_name = order.billing_name or order.customer.display_name
 
         sorted_payments = sorted(payments, key=lambda p: p.created_at)
         paid_at = sorted_payments[0].created_at if sorted_payments else None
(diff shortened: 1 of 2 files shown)


psincraian (approved):
It’s a shame I forgot about the nullable customer.email when I implemented it 🥲

frankie567 (approved):

PR #11564, opened 2026-05-11
checkout: Disable translation on checkout

Yopi:
To prevent the react rendering loop from throwing a fit because of google translate we can disable it for:
- The loading element for license keys
- The checkout confirmation page

### DIFF
--- a/clients/apps/web/src/app/(checkout)/layout.tsx
+++ b/clients/apps/web/src/app/(checkout)/layout.tsx
@@ -10,7 +10,7 @@ export default async function CheckoutLayout({
   children: React.ReactNode
 }) {
   return (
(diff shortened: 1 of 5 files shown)


pieterbeulque:
I'd add `translate="no"` on `<SelectContent>` for `CountryPicker` & `CountryStatePicker` too. You can see a `translate="no"` is already in there from a previous bugfix too.

pieterbeulque (approved):
@sebastianekstrom I wonder if we should look into removing the root layout completely and setting up a dedicated root layout in `(checkout)` and another one in `(main)`

sebastianekstrom:
@pieterbeulque yeah makes sense

PR #11568, opened 2026-05-11
subscription: Pass in product to enqueue_benefits_grants if we have it

Yopi:


frankie567:
Why this is needed? 😄

Yopi:
> Why this is needed? 😄

It might be an edge case, but we ran into an issue https://polar-olw4768.slack.com/archives/C0942KS0BD2/p1778512841870779

frankie567:
Ok, I see. Given the iterative nature of `update_product_benefits_grants`, I think a better longer bet could be to create a dedicated task that calls `enqueue_benefits_grants` for a given subscription and delay.

This way, we avoid issues with product that have thousands of subscriptions, no matter the complexity of what we do in `enqueue_benefits_grants`

Yopi:
> Ok, I see. Given the iterative nature of `update_product_benefits_grants`, I think a better longer bet could be to create a dedicated task that calls `enqueue_benefits_grants` for a given subscription and delay.
> 
> This way, we avoid issues with product that have thousands of subscriptions, no matter the complexity of what we do in `enqueue_benefits_grants`

Were you thinking something more like this?

frankie567 (approved):
Definitely :)

PR #11575, opened 2026-05-12
Small design tweaks on "Available balance"

pieterbeulque:


clients/apps/web/src/components/Payouts/AccountBalance.tsx
```diff
@@ -107,29 +97,25 @@ const AccountBalance: React.FC<AccountBalanceProps> = ({
...
-              {hasDelay && (
+        </WellContent>
+        <WellFooter>
+          <p className="dark:text-polar-500 text-gray-500">
+            {summary &&
+            summary.available_balance.amount !== summary.balance.amount ? (
+              <div className="dark:text-polar-500 space-y-1 text-gray-500">
                 <div>
-                  Available in {delayLabel}:{' '}
+                  Held balance:{' '}
```
frankie567: So we don't communicate the global delay applied to the account? Might be important since we may set account to a value different from the default

pieterbeulque: As you said in the other PR, I wasn't super convinced about the "Available in 7 days", as it could become available in as much as an hour… So relying only on the tooltip per transaction for now.

That said, I'm working up something else in development that will cover this too!

pieterbeulque: Do you consider this a blocker?

frankie567: Ah no, sorry :p

frankie567 (approved):

PR #11578, opened 2026-05-12
feat: sync Polar org subscription tier to Plain tenant

Yopi:
## Summary

Wires the polar-for-polar self-billing flow into Plain Tenants so each Polar organization corresponds to a Plain Tenant (`externalId = org.id`), members sync as tenant customers, and the support benefit grant drives the tenant's tier.

- `polar_self.create_customer` now also upserts a Plain Tenant for the organization.
- `polar_self.add_member` / `remove_member` add and remove the user as a tenant customer (keyed by user external_id).
- The support `benefit_grant.*` webhook handler reads `plain_tier_external_id` from benefit metadata and calls `plain.update_tenant_tier` with the matching tier external id; an absent key clears the tier.
- Metadata is strictly validated: non-string or empty values raise `SupportBenefitError` instead of silently clearing the tier.

Plain operations are idempotent (`upsert_tenant`; membership add/remove are no-ops if already in the desired state), and the existing Polar SDK calls already handle 409 conflicts as success, so task retries after a Plain failure are safe end-to-end.

## Test plan

- [x] `uv run task lint` passes
- [x] `uv run task lint_types` passes
- [x] `tests/integrations/polar/` passes (89 tests, including new coverage for tenant upsert, add/remove customer, tier update, and metadata validation)
- [ ] Set `plain_tier_external_id` on each of the four support benefits in Polar product config before enabling in production

### DIFF
--- a/server/polar/integrations/plain/service.py
+++ b/server/polar/integrations/plain/service.py
@@ -9,6 +9,7 @@ import pycountry
 import pycountry.db
 import structlog
 from plain_client import (
+    AddCustomerToTenantsInput,
     ComponentContainerContentInput,
     ComponentContainerInput,
     ComponentCopyButtonInput,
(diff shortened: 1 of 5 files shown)


frankie567 (approved):
Wow, that's an advanced integration 👏  Glad I invested in that Plain GraphQL client a while back :D

PR #11582, opened 2026-05-12
In-app support assistant

pieterbeulque:


clients/apps/web/src/app/(main)/feedback/question/mintlify.ts
```diff
@@ -0,0 +1,82 @@
...
+        method: 'POST',
+        headers: {
+          Authorization: `Bearer ${apiKey}`,
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify(body),
+      },
+    )
+    if (!response.ok) {
+      console.error(
```
psincraian: this should be logged to sentry or show a toast or similar

psincraian (commented):
I think this is the beginning of something big :D 

But, it would better to have the endpoints being on the backend. Mainly because we can have more data in the backend that don't need to get exposed, we can track the costs for polar for polar, maybe have that in the mobile

clients/apps/web/src/app/(main)/feedback/question/validate/route.ts
```diff
@@ -0,0 +1,64 @@
...
+
+const MAX_MESSAGE_LENGTH = 5000
+
+const requestSchema = z.object({
+  message: z.string().min(1).max(MAX_MESSAGE_LENGTH),
+  conversationId: z.string().min(1).max(64).optional(),
+  organizationId: z.string().min(1).max(64).optional(),
+})
+
+export async function POST(req: Request) {
```
psincraian, line 17: does this endpoints have rate limiting or similar? I would be worried if there is a bug or a malicius actor doing 100 calls seconds to sonnet

pieterbeulque: I was planning on setting up Vercel Firewall for this and our other Next.js LLM calling endpoints, yes.

psincraian (approved):
We agreed on shipping this as an experiment and iterate later on it later if successfull

PR #11584, opened 2026-05-12
chore: add script to reset unconfigured grandfathered orgs

psincraian:
## Summary

We agreed to clean up grandfathered orgs and move them back to created if:

- they don't have an active product, and
- they don't have a valid checkout link or api key

This means that they cannot fullfull and they cannot sell. If the org is on created, we remove the details_submitted_at so they need to fill the product description + aup

## Test plan

- [x] Lint + mypy clean
- [x] Tested end-to-end against local docker DB with three seed orgs (target ACTIVE, target CREATED, control)
- [x] Verified post-run state via `/v1/organizations/{id}/payment-status` and `/review` endpoints (merchant sees activation reset)
- [x] Idempotent re-run finds 0 targets

### DIFF
--- a/server/scripts/reset_unconfigured_grandfathered_orgs.py
+++ b/server/scripts/reset_unconfigured_grandfathered_orgs.py
@@ -9,8 +9,8 @@ grandfathered review, and resets the account's payout delay to the new
 
 Usage:
     cd server
-    uv run python -m scripts.revert_unconfigured_grandfathered_orgs            # dry-run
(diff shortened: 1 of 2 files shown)


pieterbeulque:
Can you also set their account payout delay back to the new 7 days default? cc @frankie567

frankie567 (commented):
Yep, good callout from @pieterbeulque: we should set `Account.payout_transaction_delay` to 7 days on those

frankie567 (approved):

PR #11587, opened 2026-05-12
feat: track docker dev instances in a per-machine registry

psincraian:
## Summary

- Adds a JSON registry at `~/.config/polar/docker-instances.json` that maps each worktree path to an instance number (lowest free 1..99), replacing the previous md5(path) hash.
- `dev docker list` shows every known worktree, its instance, status (`running` / `stopped` / `path missing`), and the current worktree marker.
- `dev docker prune` drops stale entries (path no longer exists) and tears down their app stack, postgres DB, redis data, and S3 buckets.
- `dev docker set-instance` now refuses to pin an instance already in use by another worktree.

## Test plan

- [x] `dev docker list` shows the current worktree as the only entry on a fresh machine.
- [x] `dev docker up` from a second worktree gets a different instance number; both appear in `list`.
- [x] After deleting a worktree, `dev docker prune` removes the entry and its DB / redis / buckets.
- [x] `dev docker set-instance N` refuses if another worktree already holds N.

### DIFF
--- a/dev/cli/commands/docker.py
+++ b/dev/cli/commands/docker.py
@@ -7,14 +7,17 @@ S3 bucket pair. Service-aware commands auto-route by service name to the
 right project. `dev docker up` starts shared (if needed) then this instance.
 """
 
-import hashlib
+import json
 import os
 import re
 import shutil
 import subprocess
+from datetime import UTC, datetime
+from pathlib import Path
 from typing import Annotated
 
 import typer
+from rich.table import Table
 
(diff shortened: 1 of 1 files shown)


Yopi (approved):
Very cool! Do we need to update the [conductor-run](https://github.com/polarsource/polar/blob/main/dev/conductor-run) / [conductor-archive](https://github.com/polarsource/polar/blob/main/dev/conductor-archive) as well? Or when does these get called (or only manually)?

PR #11588, opened 2026-05-12
skills: Add /render-env to seed an env variable into Terraform

Yopi:
Both varset and each individual state


Yopi:
From our discussion I thought this might be a worthwhile addition to make it easier and not have to explain all of the files the bot needs to touch

frankie567 (approved):
Ooooh yeah!

PR #11589, opened 2026-05-12
seed: Polar products should set visibility as the value not in metadata

Yopi:


frankie567 (approved):
Almost forgot about this feature 😂

PR #11591, opened 2026-05-12
polar: Remove inline imports

Yopi:
Since we're loading items from the Polar SDK anyway, there's no point in making the Polar client lazy loaded anymore.


frankie567 (approved):
Hopefully, Speakeasy will fix their lazy loading issue so we can avoid that :D

PR #11593, opened 2026-05-12
scripts: Backfill plain tenants

Yopi:
Backfills tenants into Plain from the Polar organizations and their users

### DIFF
--- a/server/scripts/backfill_plain_tenants.py
+++ b/server/scripts/backfill_plain_tenants.py
@@ -1,0 +1,189 @@
+import traceback
+import uuid
+from collections.abc import Sequence
+from dataclasses import dataclass, field
+
+import typer
+from rich.progress import Progress
+from sqlalchemy import select
+
+from polar.config import settings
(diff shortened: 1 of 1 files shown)


server/scripts/backfill_plain_tenants.py
```diff
@@ -0,0 +1,189 @@
...
+async def _load_organizations(
+    session: AsyncSession,
+    *,
+    self_org_id: uuid.UUID,
+    limit: int | None,
+) -> Sequence[Organization]:
+    statement = (
+        select(Organization)
+        .where(
+            Organization.can_authenticate,
```
pieterbeulque, line 40: Is it odd to check on `api_access` too?

pieterbeulque (approved):

server/scripts/backfill_plain_tenants.py
Yopi: Ah, I don't know. Will they ever get access again? 😄

PR #11554, opened 2026-05-11
feat(authz): RBAC Phase 3 — role-permission authorization cutover

joebon:
Closes #11400.

## Summary

Phase 3 of the RBAC rollout: switches policy-time authorization from `Account.admin_id` checks to a role-permission check against `UserOrganization.role`, ships the role-change endpoint, and adds the data-integrity invariants. The route-level token-scope check is unchanged and remains an independent first gate.

- **Policy cutover** (`d69c8063`): new `_require_permission` helper in `authz/policies/__init__.py`. `members.can_manage`, `organization.can_delete`, `organization.can_manage_payout_account`, `finance.can_read`, `finance.can_write` now consult `ROLE_PERMISSIONS[caller_role_in_org]`. Org-token subjects always pass; the old `_require_account_admin` helper is gone.
- **Member role exposure + removal invariants** (`d3d600ac`): `OrganizationMember.role` added; `is_admin` becomes a `@computed_field` derived from `role ∈ {owner, admin}`. `CannotRemoveOrganizationAdmin` renamed to `CannotRemoveOrganizationOwner` and now rejects when `user_org.role == owner`. New `OrganizationWouldHaveNoAdmins` invariant rejects removals that would drop admin-capable count to zero.
- **Role-change endpoint** (`99d099b8`): `PATCH /v1/organizations/{id}/members/{user_id}` route-gated by `members:write` scope, policy-gated by `members:set_role`. Body accepts `admin` / `member`; `owner` rejected at the schema layer (422). Ownership transfers continue to flow through `Account.admin_id` mutation (backoffice `change_admin`).

## Pre-merge verification (manual, gating the deploy)

- [ ] Every `Account.admin_id` user has the `owner` role on their org
- [ ] Exactly one owner per organization
- [ ] Every organization has admin capability (≥1 user in `role ∈ {owner, admin}`)

## Test plan

- [x] `uv run task test` — authz policies, organization endpoints, user_organization service
- [ ] Manual: promote a member to admin via `PATCH /members/{user_id}`, confirm new permission set takes effect
- [ ] Manual: attempt to demote the owner, confirm 403
- [ ] Manual: attempt to remove the last admin-capable user, confirm `OrganizationWouldHaveNoAdmins`

Issue #11402, opened 2026-05-05
Frontend adopts role


joebon:
**Scope.** Dashboard reads the new role and routes UI gating through the permission hook.

**Acceptance criteria**

- [ ] Per-user organizations endpoint augmented with `role`.
- [ ] Auth context populated with role per organization.
- [ ] `useRole(org)` and `hasPermission(perm, org)` hooks shipped.
- [ ] Permission tables generated from the backend source into a TypeScript artefact at build time.
- [ ] Settings → Members UI: role-change control gated by `members:set_role`; owner row's role-change and remove-member affordances disabled with tooltip pointing to support.
- [ ] Destructive controls disabled when the action would leave the org without admin capability.
- [ ] Finance pages migrated from implicit "is admin" check to the permission hook.
- [ ] Backend continues to return both `role` and `is_admin`.

**References.** [RFC §Phase 4](https://handbook.polar.sh/engineering/design-documents/rbac#phase-4--frontend-adopts-role), parent #6646. Depends on #11400.

PR #11609, opened 2026-05-12
authz: consolidate org/member permissions (1/5)

joebon:
## Summary

**Related Issue**: #11402

First of five PRs that incrementally enforce role-permissions across admin org-context endpoints. This one consolidates the role-permission vocabulary for the organization, members, OAT, and webhook surfaces, lands the foundation helpers that subsequent PRs in the stack will build on, and switches the corresponding admin endpoints onto the new permission gates. Closes a real auth gap on `submit_review` and restores a backoffice webhook defense-in-depth check along the way.

### Vocabulary changes

- `organizations:edit_settings` + `organizations:delete` + `organizations:manage_payout_account` → single `organization:manage`.
- `members:invite` + `members:remove` + `members:set_role` → single `members:manage` (alongside the existing `members:read`).
- New `finance:read` and `finance:manage` for the `GET /v1/organizations/{id}/account` finance guard.

### Foundation helpers (used in this PR; later PRs in the stack build on them)

- `assert_organization_permission(session, auth_subject, org_id, permission)` — service-layer mutation gate.
- `assert_resource_permission(session, auth_subject, resource, permission)` — same, reads `resource.organization_id`.
- `get_accessible_org_ids(session, auth_subject, *, permission=None)` — collection-read filter.
- `select_user_org_ids(user_id, *, permission=None)` — `Select` suitable for `Resource.organization_id.in_(...)` subqueries.
- `PERMISSION_DENIED_MESSAGE` registry — denial strings derived from the permission, no `denied_msg` argument at call sites.

### Guard refactor

- `OrgPolicyGuard(policy_fn=None, …)` — `policy_fn` is now optional; `AuthorizeOrgAccess` / `AuthorizeOrgAccessUser` go through the membership check only. The `_always_allow` stub is gone.
- `PayoutAccountPolicyGuard()` — `admin_id` ownership check inlined; the `policy_fn` parameter is gone.
- New `Authorize*` aliases: `AuthorizeOrgManage` (User|Org), `AuthorizeOrgManageUser` (User-only), `AuthorizeMembersRead`, `AuthorizeMembersManage`, `AuthorizeMembersSetRole`, `AuthorizeFinanceRead`.
- Removed: `AuthorizeOrgDelete`, `AuthorizeOrgManagePayoutAccount`, `AuthorizeFinanceWrite`, `AuthorizeOrgAccessWrite`.

### Endpoint cutover (the real behavior change)

`organization/endpoints.py`, `organization_access_token/{endpoints,service}.py`, `webhook/{endpoints,service}.py`:

- Org settings update, KYC/review state, OAT CRUD, webhook CRUD + redeliver are now gated by `organization:manage`. Members of an org without the admin/owner role can no longer perform these.
- `submit_review` previously used scope-only auth — now closed to `AuthorizeOrgManage`.
- `set_payout_account` and `DELETE /organizations/{id}` remain User-only (destructive / financial-plumbing actions); other org-context endpoints accept org-token callers as before.

### Other changes

- Policy denial messages no longer name a role: "Only an admin can …" → "You don't have permission to …". The role set can grow (custom roles, future ownership transfer) and naming a role in the response leaks an implementation detail.
- `polar/auth/routing.py` — `DocumentedAuthSubjectAPIRoute` now walks one level into a `Depends(...)` to find a nested `_Authenticator`, so endpoints wrapped by `OrgPolicyGuard` / `AccountPolicyGuard` / `PayoutAccountPolicyGuard` surface their required OAuth scopes in OpenAPI again (~30 endpoints regain their `**Scopes**: …` annotation).
- Backoffice webhook toggle re-validates `is_blocked_webhook_host` so a backoffice admin can't re-enable a webhook with a localhost / private-IP URL.

## Why

Two gaps were closing:

1. **API didn't enforce roles.** The `member` vs `admin`/`owner` distinction lived in the data model since RBAC Phase 3 (#11554) but wasn't enforced at the API layer. Any member of an org could call admin endpoints if their token had the right scope. This PR closes that gap for the org/members/OAT/webhook surface.
2. **Vocabulary churn.** Phase 3 left behind seven org/member permissions (`organizations:edit_settings`, `organizations:delete`, `organizations:manage_payout_account`, `members:invite`, `members:remove`, `members:set_role`, `members:read`) that mostly map 1:1 to single endpoints — fine for the dashboard's old per-action model, awkward for the section-shaped role model. Consolidate to one `*:manage` per section so new endpoints in the same section don't each need a new permission.

The next PRs in the stack will repeat the same shape for products, customers, sales/analytics/events, and finance.

## How

Two gates run at every protected site:

1. `required_scope ∈ token.scopes` — unchanged from today.
2. `required_permission ∈ permissions(role_in_org)` — new in this design.

Two call-site patterns are introduced:

- **Endpoint guard** (`{id}`-path endpoints): use one of the `Authorize*` aliases. `OrgPolicyGuard` resolves the org from the path, runs the membership check + optional policy, and returns an `AuthzContext`.
- **Service-layer assert** (post-fetch resources): call `assert_organization_permission(session, auth_subject, org_id, permission)` or `assert_resource_permission(session, auth_subject, resource, permission)` after `service.get(... auth_subject)` resolves the resource. Callers MUST gate the fetch first (404 on missing) — the assert raises 403, which would otherwise leak existence.

`owner` carries the same permission set as `admin`; the two are distinguished by invariants (singularity per org, member-removal exemption), not by additional permissions.

## Checklist


server/polar/auth/routing.py
```diff
@@ -8,6 +8,34 @@
 from polar.auth.dependencies import _Authenticator
 
 
+def _find_authenticator(dependency: typing.Any) -> _Authenticator | None:
```
frankie567, line 11: Ah, good catch 👍

server/polar/authz/repository.py
```diff
@@ -1,30 +1,68 @@
...
+    grants that permission.
+    """
+    stmt = (
+        select(UserOrganization.organization_id)
+        .join(Organization, UserOrganization.organization_id == Organization.id)
+        .where(
+            UserOrganization.user_id == user_id,
+            UserOrganization.is_deleted.is_(False),
+            Organization.can_authenticate,
+            Organization.status != OrganizationStatus.BLOCKED,
```
frankie567: I think it's redundant since `can_authenticate` cover that already. Could you confirm @psincraian? :)

server/polar/backoffice/webhooks/endpoints.py
```diff
@@ -266,9 +266,19 @@ async def toggle_enabled(
...
+            request,
+            "Cannot enable a webhook with a localhost or private IP URL.",
+            "error",
+        )
+        with tag.div(id="modal"):
+            pass
+        return
+
+    webhook = await repository.update(
+        webhook, update_dict={"enabled": not webhook.enabled}
```
frankie567, line 285: Looks unrelated?

frankie567 (approved):
Super solid, I like that we have a narrower set of roles :) A few comments, but otherwise good to go.

server/polar/backoffice/webhooks/endpoints.py
joebon: Yes, agree @frankie567 - sorry. A catch by Claude from running `/polar-code-review` on the whole changeset (i.e. before splitting up into 5 separate PRs). Made it into this first PR as some sort of "foundational findings/fixes".
Should I break this out into a standalone PR instead? Or just drop?

server/polar/authz/repository.py
joebon: Oh, right! Will fix

server/polar/backoffice/webhooks/endpoints.py
frankie567: Looks like quite an edge case to me 😄 I would say drop it.

joebon: Turns out this check is already present in the service, but since we've added permission checks in the service now we can't use that from the backoffice since the backoffice doesn't have a `auth_subject` to check against.
So instead backoffice is using the repository directly and then this check was re-implemented to open up this vulnerability. I'll add a comment clarifying that.

PR #11610, opened 2026-05-12
authz: consolidate product & custom_fields permissions (2/5)

joebon:
## Summary

**Related Issue**: #11402

Second in the stack. Enforces `products:read` / `products:manage` and `custom_fields:read` / `custom_fields:manage` across the catalog surface, using the foundation helpers from the previous PR.

**Stacked on top of #11609** — review/merge that one first; this PR's base auto-rebases onto `main` once it lands.

### Endpoints / services switched onto the new permissions

- `product/`: reads filtered by `products:read`; `create` / `update` / `update_benefits` / `get_validated_prices` gated by `products:manage`.
- `checkout_link/`: reads filtered by `products:read`; `create` / `update` / `delete` gated by `products:manage`.
- `discount/`: same shape — reads by `products:read`, mutations by `products:manage`.
- `benefit/`: same. `benefit/grant/service.list` reads filtered by `customers:read` (member-visible).
- `meter/`: same shape as products.
- `file/`: reads filtered by `products:read`; mutations gated by `products:manage` (covers `POST /files/{id}/uploaded`, `PATCH /files/{id}`, `DELETE /files/{id}`).
- `license_key/`: same shape. `PATCH /license-keys/{id}` mutation gated by `products:manage`. The customer-portal-side `/validate` and `/activate` endpoints are intentionally unchanged — they're authenticated by the license key itself, not the org role.
- `custom_field/`: reads filtered by `custom_fields:read`; `create` / `update` / `delete` gated by `custom_fields:manage`.

### Tests

Section test suites updated for the role-permission gating; the default `user_organization` fixture (owner role) continues to pass since admin/owner carry every `*:manage` permission.

### TS client

Regenerated.

## Why

Members of an org with the right token scope could previously call admin endpoints on the catalog surface — create/update/delete products, discounts, benefits, custom fields, etc. This PR closes that gap by requiring the admin/owner role for those mutations. Member-level reads (list/get) continue to work for all roles.

The `products:manage` permission covers everything product-adjacent (checkout_links, discounts, benefits, meters, files, license keys) intentionally — these are all surfaces a "product admin" reaches for, so folding them into one permission keeps the role model section-shaped and avoids needing a new permission per related entity.

## How

Two call-site shapes (introduced in the foundation PR):

- **Collection reads** — services call `get_accessible_org_ids(session, auth_subject, permission=products_read)` and pass the result to repository statements (e.g. `get_statement_by_org_ids(org_ids)`).
- **Resource mutations** — services / endpoints call `assert_resource_permission(session, auth_subject, resource, permission)` after fetching the resource via the service's own auth-aware `get(...)`. Callers always gate the fetch first (404 on missing) so the assert's 403 doesn't leak existence.

For repositories that need a permission-aware subquery (e.g. `discount/service._get_readable_discount_statement`, `custom_field/service._get_readable_custom_field_statement`), `select_user_org_ids(... permission=...)` returns the right `Select` for `Resource.organization_id.in_(...)`.

## Checklist

### DIFF
--- a/server/polar/meter/service.py
+++ b/server/polar/meter/service.py
@@ -22,7 +22,12 @@ from sqlalchemy import (
 from sqlalchemy.orm import joinedload
 
 from polar.auth.models import AuthSubject, Organization, User
-from polar.authz.service import get_accessible_org_ids
+from polar.auth.permission import OrganizationPermission
+from polar.authz.service import (
+    assert_organization_permission,
+    assert_resource_permission,
+    get_accessible_org_ids,
+)
 from polar.billing_entry.repository import BillingEntryRepository
 from polar.config import settings
 from polar.customer.repository import CustomerRepository
@@ -73,7 +78,9 @@ class MeterService:
         ],
     ) -> tuple[Sequence[Meter], int]:
         repository = MeterRepository.from_session(session)
-        org_ids = await get_accessible_org_ids(session, auth_subject)
+        org_ids = await get_accessible_org_ids(
+            session, auth_subject, permission=OrganizationPermission.products_read
+        )
         statement = repository.get_statement_by_org_ids(org_ids)
 
         if organization_id is not None:
@@ -111,7 +118,9 @@ class MeterService:
         id: uuid.UUID,
     ) -> Meter | None:
         repository = MeterRepository.from_session(session)
-        org_ids = await get_accessible_org_ids(session, auth_subject)
+        org_ids = await get_accessible_org_ids(
+            session, auth_subject, permission=OrganizationPermission.products_read
+        )
         statement = (
             repository.get_statement_by_org_ids(org_ids)
             .where(Meter.id == id)
@@ -128,6 +137,12 @@ class MeterService:
         repository = MeterRepository.from_session(session)
         organization = await get_payload_organization(
             session, auth_subject, meter_create
+        )
+        await assert_organization_permission(
+            session,
+            auth_subject,
+            organization.id,
+            OrganizationPermission.products_manage,
         )
 
         meter = await repository.create(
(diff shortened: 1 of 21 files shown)


server/polar/meter/service.py
```diff
@@ -73,7 +78,9 @@ async def list(
         ],
     ) -> tuple[Sequence[Meter], int]:
         repository = MeterRepository.from_session(session)
-        org_ids = await get_accessible_org_ids(session, auth_subject)
+        org_ids = await get_accessible_org_ids(
+            session, auth_subject, permission=OrganizationPermission.products_read
```
frankie567, line 82: Question for @pieterbeulque I guess: should we have a dedicated permission to read/manage meters rather than bucketing it in Products? Rationale might be that we could in the future nudge people to create meters for analytics purposes outside the context of a Product

frankie567 (approved):
Just a design question on the permissions themselves, but otherwise all good.

server/polar/meter/service.py
joebon: My reasoning for putting it under `products:read` is to make mental mapping to the dashboard sections and navigation easy: _Metrics_ are found/located under _Products_.

pieterbeulque: There's something to be said for both angles.

To François' point, I think it may be the more reasonable choice to map our permission namespaces closer to our database model than to how things are currently structured in the dashboard.

If/when we move _Meters_ out of _Products_ (an option we've explored to offer a lightweight product analytics feature), relying on `products:*` permissions is going to be weirder compared to right now having a `meters:*` permission that mirrors `products:*` for now.

joebon: Ok, so we should introduce `meters:read` here then?

pieterbeulque: Let's do it, and maybe do the same for discounts too?

joebon: To avoid answering question like "why `meters:read` and `discounts:read` but not for instance `benefits:read`" I'd say the above are valid points but can be addressed later on when the need actually surfaces. Unlike scopes, permissions don't end up in tokens that we need to support for the lifetime of the token. We can just add new permissions, switch what permission a given action/endpoint is protected by and connect the new permission to the relevant roles in one go.

So, with that reasoning and your permission., I'd say _let's cross that bridge when we get there_. Yes/no?

pieterbeulque: You have my blessing 🌞

PR #11611, opened 2026-05-12
authz: consolidate customers & seats permissions (3/5)

joebon:
## Summary

**Related Issue**: #11402

Third in the stack. Two changes for the customer surface:

1. Named `customers:read` / `customers:manage` gates on the admin customer endpoints and `subscription.create`. Per #11665 `customers:manage` is now a member-level permission, so the gates don't restrict members today — they're named hooks reserved for future custom roles. Behaviour for member-role callers is unchanged.
2. A real **cross-org IDOR fix** in `subscription.create`: previously the customer lookup ran with the caller's full set of accessible orgs, so an admin of OrgA who was also a member of OrgB could pair an OrgA product with an OrgB customer (creating a subscription owned by OrgA but referencing a customer in another org). The fix scopes the customer lookup to the product's org only and runs the permission assert before the lookup.

> Heads-up: the customer-seat work originally planned for this PR is gone — #11648 refactored seats independently on `main`, so I dropped those changes during the rebase and took main's version.

### Customer admin endpoints

`POST /v1/customers/`, `PATCH /v1/customers/{id}`, `PATCH /v1/customers/external/{external_id}`, `DELETE /v1/customers/{id}`, `DELETE /v1/customers/external/{external_id}` now go through the `customers:manage` named gate. List endpoints filter by `customers:read`. Member-role callers continue to pass the gate (per #11665); the gate documents intent and is the seam future custom roles will use.

### `subscription.create` cross-org pairing

Before: a multi-org caller (admin in OrgA, member in OrgB) could create a free subscription pairing an OrgA product with an OrgB customer. The flow looked up the customer with the caller's *full* set of accessible orgs, and asserted `customers:manage` only afterwards (on the product's org).

After: the permission assert runs first against the product's org. Then the customer lookup is restricted to that same org, so the customer must belong to the product's organization. Cross-org pairing now fails with "Customer does not exist."

### Test impact

`test_product_does_not_exist` no longer expects an accumulated "customer missing" error alongside the product-missing one — product validation now short-circuits before the customer lookup is attempted.

## Why

Two reasons:

1. **Named gates for future custom roles.** Even though `customers:manage` is currently member-level, the gate makes intent explicit and lets a future custom role drop the permission without changing every call site.
2. **Cross-org IDOR.** The `subscription.create` flow previously trusted the caller's whole-membership set for the customer lookup; the fix tightens that to the resource's own org.

## How

- **Endpoint-layer asserts** (`assert_resource_permission`) for the customer admin-API mutations, because `customer_service.update`/`delete` are shared with customer-portal and backoffice flows that legitimately should NOT do an org-role check. Pushing the asserts into the service would block those callers.
- **Service-layer assert** for `customer_service.create` (only called from the admin endpoint).
- **Service-layer assert + scoped lookup** for `subscription.create` — assert first, then look up the customer using `{product.organization_id}` as the org-id filter.

## Checklist

### DIFF
--- a/server/polar/customer/endpoints.py
+++ b/server/polar/customer/endpoints.py
@@ -5,7 +5,11 @@ from fastapi import Depends, Query, Response
 from fastapi.responses import StreamingResponse
 from pydantic import TypeAdapter
 
-from polar.authz.service import get_accessible_org_ids
+from polar.auth.permission import OrganizationPermission
+from polar.authz.service import (
+    assert_resource_permission,
+    get_accessible_org_ids,
+)
 from polar.exceptions import ResourceNotFound
 from polar.kit.csv import IterableCSVWriter
 from polar.kit.metadata import MetadataQuery, get_metadata_query_openapi_schema
@@ -134,7 +138,9 @@ async def export(
         )
 
         repository = CustomerRepository.from_session(session)
-        org_ids = await get_accessible_org_ids(session, auth_subject)
+        org_ids = await get_accessible_org_ids(
+            session, auth_subject, permission=OrganizationPermission.customers_read
+        )
         stream = repository.stream_by_organization(org_ids, organization_id)
 
         async for customer in stream:
@@ -303,6 +309,9 @@ async def update(
     if customer is None:
         raise ResourceNotFound()
 
+    await assert_resource_permission(
+        session, auth_subject, customer, OrganizationPermission.customers_manage
+    )
     return await customer_service.update(session, customer, customer_update)
 
 
@@ -327,6 +336,9 @@ async def update_external(
     if customer is None:
         raise ResourceNotFound()
 
+    await assert_resource_permission(
+        session, auth_subject, customer, OrganizationPermission.customers_manage
+    )
     return await customer_service.update(session, customer, customer_update)
 
 
@@ -375,6 +387,9 @@ async def delete(
     if customer is None:
         raise ResourceNotFound()
 
+    await assert_resource_permission(
+        session, auth_subject, customer, OrganizationPermission.customers_manage
+    )
     await customer_service.delete(session, customer, anonymize=anonymize)
 
 
@@ -410,4 +425,7 @@ async def delete_external(
     if customer is None:
         raise ResourceNotFound()
 
+    await assert_resource_permission(
+        session, auth_subject, customer, OrganizationPermission.customers_manage
+    )
     await customer_service.delete(session, customer, anonymize=anonymize)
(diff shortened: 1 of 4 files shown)


server/polar/customer_seat/endpoints.py
```diff
@@ -42,6 +46,53 @@
 )
 
 
+def _container_organization_id(container: Subscription | Order) -> UUID4:
+    """The org that owns a seat container (subscription or one-time order)."""
+    if isinstance(container, Subscription):
+        return container.product.organization_id
```
frankie567: TODO: denormalize organization_id on Subscription 😄

@stilla-bot Create an issue for that.

server/polar/customer_seat/endpoints.py
```diff
@@ -69,21 +120,28 @@ async def assign_seat(
     subscription: Subscription | None = None
     order: Order | None = None
 
+    typed_auth_subject: AuthSubjectType[User | Organization] | None = None
+    if not isinstance(auth_subject.subject, Anonymous):
+        typed_auth_subject = cast(AuthSubjectType[User | Organization], auth_subject)
```
frankie567: Question for @psincraian: why do we allow anonymous access on those endpoints? It feels weird since it's the Merchant API here. The anonymous claims should be part of the customer portal API from what I understand.

frankie567 (commented):
Not approving since I want clarifications on the customer seat API. It looks like we have a design issue here.

server/polar/customer_seat/endpoints.py
psincraian: Mmm 🤔 we are using the assign seat with the checkout id in checkout flow in polar. The user is not authenticated on that moment, that's why we rely on the checkout id.

I can move this to the customer-portal, but feels a bit stranger.

frankie567: Well, I think it's worth to discuss about it because there are a few things that feel strange:

* `SeatWriteOrAnonymous` is only useful for `assign_seat` with checkout ID. Other cases and endpoints all add a manual gate to block anonymous users.
* Why do we have this process of assigning anonymously with a checkout ID? Customers should automatically have access to a customer session token after a successful checkout.
* Invitation endpoints clearly targets customers, not merchants

psincraian: I agree with you, I can see that I can move to customer-portal endpoints to assign the seats. Raising a PR 👀

joebon: Sounds like a good find but something that'll be addressed elsewhere - correct?

frankie567: Yes, but could we maybe wait for Petru changes first and then rebase this one? Feels like it'll dramatically simplify the logic here.

joebon: Fo sho!

frankie567 (approved):

PR #11612, opened 2026-05-12
authz: consolidate sales, analytics & events permissions (4/5)

joebon:
## Summary

**Related Issue**: #11402

Fourth in the stack. Enforces `sales:read`, `analytics:read`/`analytics:manage`, and the new `events:ingest` permission across the corresponding surfaces.

**Stacked on top of #11611** — review/merge the previous PRs first; this PR's base auto-rebases onto `main` once they land.

### sales:read

Member-level read permission for the sales pipeline. Filters apply at the repository / service layer:

- `order/{repository,service}`: `get_readable_statement` filters the user branch by `sales:read`; the service-layer list uses `get_accessible_org_ids(... sales_read)` for filtering before delegating to the repository.
- `subscription/repository.get_readable_statement`: same.
- `dispute/service`, `payment/service`, `refund/service`, `checkout/service`: collection reads filtered by `sales:read`.

### analytics:read / analytics:manage

- `event_type/{endpoints,service}`: list / get filtered by `analytics:read`; `update` (label / label_property_selector) gated by `analytics:manage`.
- `metrics/{endpoints,service}`: metric reads + dashboard list / get filtered by `analytics:read`; dashboard create / update / delete gated by `analytics:manage`.

### events:ingest

- New permission, granted to all roles by default (apps and integrations commonly run as member-role users).
- `event/service.ingest` validates organization membership via `get_accessible_org_ids(... events_ingest)`. Previously the membership query was inline and had no role-permission filter, leaving ingest as scope-only auth even after the rest of the analytics READ side moved to `analytics:read`.

### TS client

Regenerated.

## Why

Sales/analytics/events were the broadest unfinished surface in the cutover. The asymmetry was visible: analytics reads were already admin-gated via `analytics:read` post-cutover (PR 1's foundation), but ingest — a *write* path that drives metering and billing — was still scope-only. Making `events:ingest` an explicit permission lets the role model evolve (e.g. a future custom role could withhold ingest) without changing the call site.

`sales:read` is granted to every role, but having it as a named permission makes the intent explicit in `get_readable_statement` and lets future custom roles drop it if desired.

`analytics:manage` is admin-only; member-role users can no longer create/update/delete metric dashboards or edit event-type labels.

## How

- **Collection reads** — `get_accessible_org_ids(session, auth_subject, permission=...)` at the service layer.
- **Repository-level subquery** — `select_user_org_ids(... permission=sales_read)` inside `Resource.organization_id.in_(...)` for `get_readable_statement` shapes.
- **Resource mutations** — `assert_resource_permission(session, auth_subject, resource, permission)` after fetch.

## Checklist

### DIFF
--- a/server/polar/customer/endpoints.py
+++ b/server/polar/customer/endpoints.py
@@ -5,7 +5,11 @@ from fastapi import Depends, Query, Response
 from fastapi.responses import StreamingResponse
 from pydantic import TypeAdapter
 
-from polar.authz.service import get_accessible_org_ids
+from polar.auth.permission import OrganizationPermission
+from polar.authz.service import (
+    assert_resource_permission,
+    get_accessible_org_ids,
+)
 from polar.exceptions import ResourceNotFound
 from polar.kit.csv import IterableCSVWriter
 from polar.kit.metadata import MetadataQuery, get_metadata_query_openapi_schema
@@ -134,7 +138,9 @@ async def export(
         )
 
         repository = CustomerRepository.from_session(session)
-        org_ids = await get_accessible_org_ids(session, auth_subject)
+        org_ids = await get_accessible_org_ids(
+            session, auth_subject, permission=OrganizationPermission.customers_read
+        )
         stream = repository.stream_by_organization(org_ids, organization_id)
 
         async for customer in stream:
@@ -303,6 +309,9 @@ async def update(
     if customer is None:
         raise ResourceNotFound()
 
+    await assert_resource_permission(
+        session, auth_subject, customer, OrganizationPermission.customers_manage
+    )
     return await customer_service.update(session, customer, customer_update)
 
 
@@ -327,6 +336,9 @@ async def update_external(
     if customer is None:
         raise ResourceNotFound()
 
+    await assert_resource_permission(
+        session, auth_subject, customer, OrganizationPermission.customers_manage
+    )
     return await customer_service.update(session, customer, customer_update)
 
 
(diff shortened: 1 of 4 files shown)


server/polar/refund/service.py
```diff
@@ -190,6 +193,10 @@ async def user_create(
                 ]
             )
 
+        await assert_resource_permission(
+            session, auth_subject, order, OrganizationPermission.finance_manage
```
frankie567: Gating refund creation under the finance permission is a bit problematic IMO. The Finance role is designed to get access to transactions and payouts. Refunds is a quite common day-to-day operation, so I think we need another permission — the same that would grant access to Subscription updates (cancellation, trials, etc.)

frankie567 (commented):
One concern regarding the permission required to perform refunds

server/polar/refund/service.py
joebon: Great! That's very important to capture. I opted to put subscription updates under `customers:manage` as it felt the most fitting unless carving out a dedicated permission namespace for it. Perhaps put refund actions into `customers:manage` also?

frankie567: Yep, that's probably make more sense. Would love @pieterbeulque input on this :)

pieterbeulque (commented):
> `analytics:manage` is admin-only; member-role users can no longer create/update/delete metric dashboards or edit event-type labels.

Can we maybe give everyone this permission to increase backwards compatibility?

server/polar/refund/service.py
pieterbeulque: Yes, agree. For example, a support team member should not be able to manage payouts but should be able to trigger refunds.

`customers:manage` is fine as a bucket permission if we don't want to opt for more granularity.

PR #11613, opened 2026-05-12
authz: consolidate finance permissions (5/5)

joebon:
## Summary

**Related Issue**: #11402

Final PR in the stack. Enforces `finance:read` and `finance:manage` across the payout and transaction surfaces.

**Stacked on top of #11612** — review/merge the previous PRs first; this PR's base auto-rebases onto `main` once they land.

### Payouts

- `payout_service.get` accepts a `permission` keyword (defaulting to `finance:read`); the collection lookup is filtered by `get_accessible_org_ids(... permission)`.
- `payout/endpoints.generate_invoice` passes `permission=finance_manage` to upgrade the gate — a side-effecting POST that triggers invoice generation now requires the admin/owner role's `finance:manage`.
- `payout/endpoints.get_estimate` asserts `finance:read` and `payout/endpoints.create` asserts `finance:manage` via `assert_organization_permission` against the resolved organization.

The `permission=` keyword on `payout_service.get` is the right shape for this resource (vs. the `assert_resource_permission(payout, ...)` pattern used elsewhere): `Payout` is keyed by `account_id`, and one account can back multiple organizations, so there's no single `payout.organization_id` to assert against. The permission filter at the lookup correctly resolves to "any org backing this payout's account that the caller has `finance:manage` on."

### Transactions

- `_get_readable_transactions_statement` joins on `UserOrganization` and `PaymentUserOrganization` now filter by `roles_with_permission(finance_read)` and `is_deleted = False` — so a user removed from an org (`deleted_at` set, role still `admin`) no longer leaks transaction visibility for that org.

## Why

Finance is the most sensitive surface in the cutover: payouts move real money out of the org's account, transaction lists expose revenue and fee detail. Pre-PR, both were gated by token scope only (`payouts:read`, `transactions:read`) — any member of an org with those scopes could view financial data. After this PR, only the admin/owner role's `finance:*` permissions unlock these endpoints.

The `is_deleted` filter on the transaction join closes a pre-existing latent bug surfaced during the cutover: removed members kept transaction visibility if their `UserOrganization.role` was still admin (the soft-delete state isn't reset on removal).

## How

Same patterns as previous PRs:

- **Lookup-time filter**: `get_accessible_org_ids(session, auth_subject, permission=finance_read)` filters payout lookups to orgs where the caller has the permission.
- **Endpoint-level assert**: `assert_organization_permission(session, auth_subject, organization.id, finance_manage)` for the payout endpoints that resolve an organization from a payload field rather than a `{id}` path.
- **Repository-level role filter**: `roles_with_permission(finance_read)` inside the existing `or_(...)` clause of `_get_readable_transactions_statement` — keeps the existing user/payment-user/org-admin tri-way readability shape and just tightens which org-admin roles count.

`Payout` doesn't get the `assert_resource_permission(payout, ...)` treatment because it lacks an `organization_id` field (account-keyed, multi-org). The `permission=` parameter on `payout_service.get` is the intentional alternative.

## Checklist

### DIFF
--- a/server/polar/customer/endpoints.py
+++ b/server/polar/customer/endpoints.py
@@ -5,7 +5,11 @@ from fastapi import Depends, Query, Response
 from fastapi.responses import StreamingResponse
 from pydantic import TypeAdapter
 
-from polar.authz.service import get_accessible_org_ids
+from polar.auth.permission import OrganizationPermission
+from polar.authz.service import (
+    assert_resource_permission,
+    get_accessible_org_ids,
+)
 from polar.exceptions import ResourceNotFound
 from polar.kit.csv import IterableCSVWriter
 from polar.kit.metadata import MetadataQuery, get_metadata_query_openapi_schema
@@ -134,7 +138,9 @@ async def export(
         )
 
         repository = CustomerRepository.from_session(session)
-        org_ids = await get_accessible_org_ids(session, auth_subject)
+        org_ids = await get_accessible_org_ids(
+            session, auth_subject, permission=OrganizationPermission.customers_read
+        )
         stream = repository.stream_by_organization(org_ids, organization_id)
 
         async for customer in stream:
@@ -303,6 +309,9 @@ async def update(
     if customer is None:
         raise ResourceNotFound()
 
+    await assert_resource_permission(
+        session, auth_subject, customer, OrganizationPermission.customers_manage
+    )
     return await customer_service.update(session, customer, customer_update)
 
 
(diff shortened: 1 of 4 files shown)


server/polar/payout/endpoints.py
```diff
@@ -84,6 +86,12 @@ async def get_estimate(
     if organization is None:
         raise ResourceNotFound()
 
+    await assert_organization_permission(
+        session,
+        auth_subject,
+        organization.id,
+        OrganizationPermission.finance_read,
```
frankie567: I think we should set `finance_manage` on this one: even if it's read-only, this endpoint only makes sense in the context of creating a payout, so there is no real interest to have a read access on that IMO.

joebon: Wonderful, will fix!

PR #11614, opened 2026-05-12
metrics: Fix FX-incorrect credit_order contribution in fast-path revenue

Yopi:
credit_order events don't carry their own exchange_rate, so the MV looked one up via a self-join to the customer's most recent balance.order. That join doesn't see prior insert batches when it runs inside an MV trigger, and balance.order has ingestion lag while credit_order is synchronous, so the two events basically never land in the same batch. FX silently fell back to 1.0 and non-USD amounts got stored as if they were USD.

Drop credit_order from the MV. Resolve it at query time in revenue_fast instead, with the same join revenue_direct already uses (it works fine outside an MV trigger).

Tests now ingest credit_order in a second batch so this class of bug actually reproduces. MV needs a repopulate after deploy.

### DIFF
--- a/server/tests/metrics/test_tinybird_metrics.py
+++ b/server/tests/metrics/test_tinybird_metrics.py
@@ -118,6 +118,7 @@ class OrderScenario:
     include_balance_net_amount: bool = True
     balance_exchange_rate: float | None = None
     balance_presentment_currency: str = "usd"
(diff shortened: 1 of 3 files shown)


pieterbeulque (approved):

pieterbeulque:
Blind approve 😇

frankie567 (approved):

PR #11637, opened 2026-05-13
server/order: reduce customer balance when voiding an order

frankie567:
Consider the following scenario:

- A subscription is cycled with an order of 100$. The payment is failing, and the payment is dunning.
- The customer downgrades to a 10$ plan, they get a credit of 90$ on their account.
- Later, the order is voided (either by Polar support or subscription cancellation). The customer still has a credit of 90$, even if the payment never succeeded.

This PR takes care of reducing the customer balance by the amount of the order (floored at 0) when voiding an order, so that in the scenario above, the customer would have a balance of 0$ after voiding the order.

We already had the same mechanism in place for refunds.

### DIFF
--- a/server/polar/order/service.py
+++ b/server/polar/order/service.py
@@ -754,7 +754,7 @@ class OrderService:
             )
             total_amount = net_amount + tax_amount
             customer_balance = await wallet_service.get_billing_wallet_balance(
-                session, customer, subscription.currency
+                session, customer, subscription.currency, for_update=True
             )
 
             # Calculate balance change and applied amount
@@ -1945,7 +1945,7 @@ class OrderService:
 
         # Reduce positive customer balance
         customer_balance = await wallet_service.get_billing_wallet_balance(
-            session, order.customer, order.currency
+            session, order.customer, order.currency, for_update=True
         )
         if customer_balance > 0:
             reduction_amount = min(customer_balance, order.due_amount)
--- a/server/polar/order/service.py
+++ b/server/polar/order/service.py
@@ -1943,6 +1943,20 @@ class OrderService:
             update_dict={"status": OrderStatus.void, "next_payment_attempt_at": None},
         )
 
+        # Reduce positive customer balance
+        customer_balance = await wallet_service.get_billing_wallet_balance(
+            session, order.customer, order.currency
+        )
+        if customer_balance > 0:
+            reduction_amount = min(customer_balance, order.due_amount)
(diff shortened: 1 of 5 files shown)


server/polar/order/service.py
```diff
@@ -1943,6 +1943,20 @@ async def void(self, session: AsyncSession, order: Order) -> Order:
             update_dict={"status": OrderStatus.void, "next_payment_attempt_at": None},
         )
 
+        # Reduce positive customer balance
+        customer_balance = await wallet_service.get_billing_wallet_balance(
```
Yopi, line 1947: Does this need to be `with_for_update` if we are voiding multiple orders at the same time?

Yopi (approved):
Good catch!

server/polar/order/service.py
```diff
@@ -1943,6 +1943,20 @@ async def void(self, session: AsyncSession, order: Order) -> Order:
...
+        )
+        if customer_balance > 0:
+            reduction_amount = min(customer_balance, order.due_amount)
+            await wallet_service.create_balance_transaction(
+                session,
+                order.customer,
+                -reduction_amount,
+                order.currency,
+                order=order,
+            )
```
joebon, line 1958: Red herring from Claude?
> Order creation already debits the wallet by `applied_balance_amount` at lines 750-815. The new void logic deducts again from current balance up to `order.due_amount`. Scenario: customer has $5 balance, order for $10 → at creation $5 is applied,            
  due_amount=$5. Customer tops balance back to $5 before void. Void then drains the new $5. Customer effectively loses $10 against a voided order they were never charged for. If the intent is to reverse the originally-applied balance, mirror             
  `refund/service.py:503-517` and credit `-order.applied_balance_amount` instead.

frankie567: I'm not sure I follow its reasoning. The point is to reduce customer balance for something that were never paid, so the customer doesn't lose money here.

joebon: I'm not sure I follow either, not even sure how the customer would "top balance back to $5" even. Just wanted to surface it in case it actually caught an edge case.

server/polar/order/service.py
frankie567: Good catch

frankie567: Added `for_update` when necessary :)

Yopi (approved):

PR #11641, opened 2026-05-13
fix: allow editing organization details after initial status

psincraian:
## Summary

Remove the status gate so `organization.details` can be updated via the org update endpoint regardless of the org's current status. We should be able to allow modifications of the details going forward, like asking orgs to complete all the details in here.

## Checklist

### DIFF
--- a/server/polar/organization/service.py
+++ b/server/polar/organization/service.py
@@ -494,7 +494,7 @@ class OrganizationService:
             },
         )
 
-        if update_schema.details and organization.status == OrganizationStatus.CREATED:
+        if update_schema.details:
             organization.details = cast(
                 OrganizationDetails, update_schema.details.model_dump()
             )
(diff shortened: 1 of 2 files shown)


pieterbeulque:
Does this expose us to bait-and-switch attacks?

frankie567 (commented):
Pieter Beulque concern seems valid. They could update the description after their review submission, isn't it?

psincraian:
Missed this. Mmm in one part it's a valid concern, but in another I think would be nice that in the future we trigger the agent review every time that the website, product description, identity, etc. is changed. 

We pass the product description to the agent review every single time, if that changes from "i'm a good citizen" to "i'm selling nsfw" the agent should detect and block it. Still, I don't think someone will put crazy stuff in there

frankie567 (approved):
Fine for me then :)

PR #11645, opened 2026-05-13
billing: Wire up billing address to the API

Yopi:
This wires up the billing address get / update in our own billing dashboard instead of going out to the customer portal.

### DIFF
--- a/server/polar/integrations/polar/service.py
+++ b/server/polar/integrations/polar/service.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING, Any
 
 import logfire
 from polar_sdk.models import (
+    CustomerPortalCustomerUpdate,
     WebhookBenefitGrantCreatedPayload,
     WebhookBenefitGrantRevokedPayload,
     WebhookBenefitGrantUpdatedPayload,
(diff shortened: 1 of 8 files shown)


server/polar/integrations/polar/service.py
```diff
@@ -258,6 +270,46 @@ async def list_orders(
...
+        update: OrganizationBillingDetailsUpdate,
+        external_member_id: str | None = None,
+    ) -> "CustomerPortalCustomer":
+        await self._ensure_polar_customer(organization_id)
+        # Only forward fields the client explicitly sent so we don't
+        # null out billing_address / tax_id on a partial update.
+        sdk_update = CustomerPortalCustomerUpdate.model_validate(
+            update.model_dump(exclude_unset=True, mode="json")
+        )
+        return await get_client().portal_update_customer(
```
frankie567, line 298: We should probably have proper 422 errors handling here. In particular with the Tax ID, it's highly probable people will put invalid garbage in there.

Yopi: Fair point!

frankie567 (approved):

PR #11648, opened 2026-05-13
refactor: drop anonymous branch from merchant seat-assignment API

psincraian:
## Summary

The post-checkout success page was the only caller of the anonymous branch of `POST /v1/customer-seats`, gated only by `checkout_client_secret`. The checkout confirm response already includes a fresh `customer_session_token`, so the frontend can authenticate against the customer portal seat endpoint instead. 

## What

- Backend: the merchant `assign_seat` now requires `SeatWrite`. Checked on logfire, we are the only ones using the merchant endpoint in the last 30 days.
- Frontend: `useAssignSeatFromCheckout` is a `useMutation` that POSTs to `/v1/customer-portal/seats` with the customer session token and `checkout_id`; backend resolves the container. `CheckoutSeatInvitations` is split into a guard and a panel so the panel only renders once the required props are present.

## Why

The anonymous merchant-API branch was an oversight, not a workaround for a missing capability: the customer session token has been issued at checkout confirm since long before the seats module existed. Removing it makes the merchant API uniformly authenticated and lets the frontend use the customer-portal path that the rest of the post-purchase surface already uses.

## How

- Switched the merchant endpoint dependency to `SeatWrite`, removed dead schema fields, and dropped the cast/null-narrowing boilerplate.
- Split `SeatAssign.validate_identifiers` into `validate_seat_source` and `validate_recipient`, with overridable `_seat_source_ids` / `_seat_source_error_message` helpers so the portal subclass adds `checkout_id` without touching the recipient identifier rules.
- Added `get_readable_by_id` / `get_readable_by_checkout_id` to `SubscriptionRepository` and `get_readable_by_id` / `get_earliest_readable_by_checkout_id` to `OrderRepository`, following the `get_readable_by_id` precedent in customer / meter / product / license_key repos. The portal service is pure orchestration — no inline `.where(...).options(...)` composition.
- Updated unauthorized tests from 403 → 401 (the dependency is now pure `SeatWrite`) and added explicit IDOR coverage for the `checkout_id` path on the customer-portal endpoint.

## Checklist

### DIFF
--- a/clients/apps/web/src/components/Checkout/CheckoutConfirmation.tsx
+++ b/clients/apps/web/src/components/Checkout/CheckoutConfirmation.tsx
@@ -228,7 +228,10 @@ export const CheckoutConfirmation = ({
           )}
           {status === 'succeeded' && (
             <>
-              <CheckoutSeatInvitations checkout={checkout} />
+              <CheckoutSeatInvitations
+                checkout={checkout}
+                customerSessionToken={customerSessionToken}
+              />
               {hasProductCheckout(checkout) &&
                 checkout.product_price.amount_type !== 'seat_based' && (
                   <CheckoutBenefits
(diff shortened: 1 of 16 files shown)


frankie567 (approved):
Thank you for this, Petru :) Feels much better!

PR #11653, opened 2026-05-13
server/auth: setup Reauth and add its models

frankie567:
- Add Reauth library to bring auth logic like TOTP and multi-factor authentication session
- Implement the corresponding database models
- I deliberately created a dedicated EmailOTP model even if very similar to LoginCode: I think we'll have a phase where we'll start to deploy the new auth flow behind feature flag or something; so didn't want to mess with the existing and working login.


server/pyproject.toml
```diff
@@ -174,6 +175,7 @@ warn_untyped_fields = true
 
 [tool.uv]
 exclude-newer = "1 week"
+exclude-newer-package = { reauth = false } # Reauth is brand new, but remember to remove this exclusion when stabilized
```
pieterbeulque, line 178: Do we trust this vendor? 🤡

joebon: who dis new phone

joebon (approved):

frankie567:
A bit too early, will re-open a more up-to-date PR to avoid piling up migrations.

PR #11654, opened 2026-05-13
scripts: Polar self owner drift

Yopi:
During the roll out of Polar-for-Polar owners/members, we have seen some drift. This scripts aims to fix it by reconciling each polar-for-polar customer's Member.role=owner with the organization's current Account.admin_id. For orgs where the admin user already has a Member row, ownership is transferred (the previous owner is demoted to billing_manager, matching the runtime change_admin flow).

### DIFF
--- a/server/scripts/backfill_polar_self_owner.py
+++ b/server/scripts/backfill_polar_self_owner.py
@@ -1,0 +1,223 @@
+import traceback
+import uuid
+from dataclasses import dataclass, field
+
+import dramatiq
+import structlog
+import typer
+from rich.progress import Progress
+from sqlalchemy import String, cast, select
+
+from polar import tasks  # noqa: F401
+from polar.config import settings
+from polar.kit.db.postgres import create_async_sessionmaker
+from polar.member.repository import MemberRepository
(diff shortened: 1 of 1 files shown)


server/scripts/backfill_polar_self_owner.py
```diff
@@ -0,0 +1,223 @@
...
+
+async def _load_drifted(
+    session: AsyncSession, self_org_id: uuid.UUID, limit: int | None
+) -> list[tuple[Organization, Customer, Member, User]]:
+    statement = (
+        select(Organization, Customer, Member, User)
+        .join(Account, Account.id == Organization.account_id)
+        .join(Customer, Customer.external_id == cast(Organization.id, String))
+        .join(Member, Member.customer_id == Customer.id)
+        .join(User, User.id == Account.admin_id)
```
joebon, line 44: Fyi, `Account.admin_id` will soon cease to exist, it's being replaced by `OrganizationRole.owner`. Perhaps better to join `UserOrganization` on `UserOrganization.role == OrganizationRole.owner` instead

frankie567 (approved):

PR #11657, opened 2026-05-13
fix gnarly stripe iframe color scheme bug

emilwidlund:
Implements a fix for the ugly dark mode flash on checkout which happens when Stripe injects its pre-auth iframe.

Learn more about the fix here — https://sergeyski.com/css-color-scheme-and-iframes-lessons-learned-from-disqus-background-bug/


sebastianekstrom (approved):
Great digg'arooski 🪏

PR #11659, opened 2026-05-13
ops: make sure to install UV dependencies with a frozen lockfile

frankie567:
- ops: make sure to install UV dependencies with a frozen lockfile


joebon (approved):
Nice catch

PR #11661, opened 2026-05-13
chore: add backfill script for failed checkouts

psincraian:
## Summary

The `failed` checkout status is deprecated and no longer assigned. Checkouts that hit a payment failure are now reset to `open` so the customer can retry. This PR adds a one-shot backfill script that rewrites any remaining `failed` rows to `expired`, so a follow-up PR can remove the enum value cleanly.

## What

- New script `server/scripts/checkout_failed_to_expired_backfill.py` that uses `run_batched_update` (batch size 5000, 0.1s sleep) to migrate `Checkout` rows from `status='failed'` to `status='expired'`.

## Why

`failed` is a dead enum value (the dashboard filter for it is unassigned), and the only thing keeping the value alive is a handful of legacy rows. Backfilling them lets us drop the enum entry and clean up the dashboard filter in a follow-up.

`checkouts` is one of the largest tables, so a single `UPDATE` in a migration would lock for too long. The batched helper avoids that.

## How

Mirrors the existing `checkout_require_3ds_backfill.py` pattern. Run via a Render One-Off Job on the worker service:

```
uv run python -m scripts.checkout_failed_to_expired_backfill
```

Once this has been deployed and the backfill has run to zero `failed` rows, the follow-up PR will:
- Remove `failed` from the `CheckoutStatus` enum
- Drop the dead branch in `checkout_service.handle_failure`
- Remove `failed` entries from the frontend status display maps (the dashboard filter is regenerated from the enum, so it disappears automatically)

## Checklist

### DIFF
--- a/server/scripts/checkout_failed_to_expired_backfill.py
+++ b/server/scripts/checkout_failed_to_expired_backfill.py
@@ -1,0 +1,53 @@
+import typer
+from sqlalchemy import select, update
+
+from polar.models import Checkout
+from polar.models.checkout import CheckoutStatus
+from scripts.helper import (
+    configure_script_logging,
+    limit_bindparam,
+    run_batched_update,
+    typer_async,
+)
+
+cli = typer.Typer()
+
+configure_script_logging()
+
+
+@cli.command()
+@typer_async
+async def checkout_failed_to_expired_backfill(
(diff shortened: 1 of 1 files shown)


frankie567 (approved):
Great! Thank you :)

PR #11673, opened 2026-05-15
authz: consolidate sales, analytics & events permissions (4/5)

joebon:
## Summary

**Related Issue**: #11402

Fourth in the stack. Three changes for the sales / analytics / events surface:

1. `sales:read` filter on order, subscription, dispute, payment, refund, and checkout reads — named hook for the sales pipeline.
2. `analytics:read` filter and `analytics:manage` gate on metric and event-type endpoints. Per #11665 `analytics:manage` is now member-level, so the gate documents intent and reserves the seam for future custom roles; member-role behaviour is unchanged.
3. A new `events:ingest` permission closing a real gap: analytics reads were already gated by `analytics:read` after the foundation PR landed, but `event/service.ingest` — a *write* path that drives metering and billing — was still scope-only. It now goes through the named permission.

### Sales reads

`GET /v1/orders`, `GET /v1/subscriptions`, `GET /v1/disputes`, `GET /v1/payments`, `GET /v1/refunds`, `GET /v1/checkouts` and their `{id}` siblings now filter the user-subject branch by `sales:read`. Granted to every role today; the hook is in place for a future role that drops it.

### Analytics reads & writes

`GET /v1/metrics`, `GET /v1/metrics/limits`, metric dashboard list / get, event-type list / get — filtered by `analytics:read`. Metric dashboard create / update / delete and event-type label edits — go through the `analytics:manage` named gate.

### `event/service.ingest`

Membership validation switches from an inline UserOrganization query (no role filter) to `get_accessible_org_ids(... events_ingest)`. `events:ingest` is granted to every role by default so apps and integrations running as member-role users keep working; the gate exists so a future custom role can withhold ingest.

## Why

Sales / analytics / events was the broadest unfinished surface from the earlier cutover. Naming the gates aligns intent with the rest of the role model. The `events:ingest` change closes a concrete asymmetry: the *read* side was role-aware, the *write* side wasn't.

## How

- **Collection reads** — `get_accessible_org_ids(session, auth_subject, permission=...)` at the service layer.
- **Repository-level subquery** — `select_user_org_ids(... permission=sales_read)` inside `Resource.organization_id.in_(...)` for `get_readable_statement` shapes.
- **Resource mutations** — `assert_resource_permission(session, auth_subject, resource, permission)` after the auth-aware fetch.

## Checklist

### DIFF
--- a/server/polar/event_type/service.py
+++ b/server/polar/event_type/service.py
@@ -2,7 +2,11 @@ from collections.abc import Sequence
 from uuid import UUID
 
 from polar.auth.models import AuthSubject
-from polar.authz.service import get_accessible_org_ids
+from polar.auth.permission import OrganizationPermission
+from polar.authz.service import (
+    assert_resource_permission,
+    get_accessible_org_ids,
+)
 from polar.customer.repository import CustomerRepository
 from polar.event.system import SYSTEM_EVENT_LABELS
 from polar.event.tinybird_repository import TinybirdEventRepository
@@ -27,7 +31,9 @@ class EventTypeService:
         id: UUID,
     ) -> EventType | None:
         repository = EventTypeRepository.from_session(session)
-        org_ids = await get_accessible_org_ids(session, auth_subject)
+        org_ids = await get_accessible_org_ids(
+            session, auth_subject, permission=OrganizationPermission.analytics_read
+        )
         statement = repository.get_statement_by_org_ids(org_ids).where(
             EventType.id == id
         )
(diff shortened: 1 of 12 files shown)


server/polar/event_type/service.py
```diff
@@ -27,7 +31,9 @@ async def get(
         id: UUID,
     ) -> EventType | None:
         repository = EventTypeRepository.from_session(session)
-        org_ids = await get_accessible_org_ids(session, auth_subject)
+        org_ids = await get_accessible_org_ids(
+            session, auth_subject, permission=OrganizationPermission.analytics_read
```
Yopi, line 35: It's probably unnecessary to introduce a new permission category, but I wonder if analytics_read and _manage makes sense for events and event types? Or if you should be able to read events without the metrics

Yopi (approved):

server/polar/event_type/service.py
joebon: That's a fair point, thanks @Yopi !
IMHO we can always make the permissions more fine grained down the line if/when needed. At this point my main target is to guard endpoints behind permissions that are fairly intuitive so merchants can easily analyze the consequences of assigning a role to a user (with the role's permissions listed).
And we won't have a role which will have access to `analyze:read/manage` but _not_ `metrics:read/manage` up front.
Makes sense?

Yopi: It makes sense! 😄

PR #11674, opened 2026-05-15
authz: consolidate finance permissions (5/5)

joebon:
## Summary

**Related Issue**: #11402

Final PR in the stack. Enforces `finance:read` and `finance:manage` across the payout and transaction surfaces. Unlike the previous two PRs in the stack, finance permissions *did* stay admin-only after #11665 — so this is a real behaviour tightening for member-role callers, not just a named hook.

### Payouts

`GET /v1/payouts/`, `GET /v1/payouts/estimate`, `GET /v1/payouts/{id}/csv`, `GET /v1/payouts/{id}/invoice` now require `finance:read` (admin/owner). `POST /v1/payouts/` and `POST /v1/payouts/{id}/invoice` require `finance:manage`. Member-role callers with `payouts:read` / `payouts:write` scopes no longer see payouts or trigger invoice generation.

### Transactions

`GET /v1/transactions/` now requires `finance:read`. Member-role callers with `transactions:read` no longer see transaction history.

### Removed members

A user removed from an org (`UserOrganization.is_deleted = true`) no longer sees that org's transactions — the soft-delete state wasn't being checked alongside the role, so removed admins kept visibility.

## Why

Payouts move real money out of the org's account and transaction lists expose revenue and fee detail. Pre-PR these were gated by token scope only — any org member with `payouts:read` / `transactions:read` could see financial data, and any with `payouts:write` could trigger invoice generation. After this PR the role check joins the scope check, locking these to admin/owner.

The removed-member bug was latent and surfaced naturally during the cutover: the existing transaction read path filtered by `UserOrganization.user_id` but never by `is_deleted`, so a removed admin's stale row still counted as access.

## How

Same patterns as previous PRs:

- **Lookup-time filter**: `get_accessible_org_ids(... permission=finance_read)` filters payout reads to orgs where the caller has the permission.
- **Endpoint-level assert**: `assert_organization_permission(... finance_manage)` for `payout.create` and `payout.generate_invoice` — both resolve their organization from a payload field rather than a `{id}` path.
- **Repository-level role filter**: `roles_with_permission(finance_read)` plus `is_deleted.is_(False)` inside the `or_(...)` clauses of `_get_readable_transactions_statement` — keeps the existing user / payment-user / org-admin tri-way readability shape and tightens which org-admin roles count.

`Payout` is keyed by `account_id` (one account can back multiple orgs), so the `assert_resource_permission(payout, ...)` pattern used elsewhere in the cutover doesn't apply — there's no single `payout.organization_id` to assert against. `payout_service.get` instead takes a `permission` keyword (defaulting to `finance:read`) and `generate_invoice` passes `finance_manage` to upgrade the lookup-time filter.

## Checklist

### DIFF
--- a/server/polar/payout/endpoints.py
+++ b/server/polar/payout/endpoints.py
@@ -3,6 +3,8 @@ from fastapi.responses import StreamingResponse
 from pydantic import UUID4
 from sqlalchemy.orm import joinedload
 
+from polar.auth.permission import OrganizationPermission
+from polar.authz.service import assert_organization_permission
 from polar.exceptions import ResourceNotFound
 from polar.kit.db.postgres import AsyncSessionMaker
 from polar.kit.pagination import ListResource, PaginationParamsQuery
@@ -84,6 +86,12 @@ async def get_estimate(
     if organization is None:
         raise ResourceNotFound()
 
+    await assert_organization_permission(
+        session,
+        auth_subject,
+        organization.id,
+        OrganizationPermission.finance_read,
+    )
     return await payout_service.estimate(session, organization)
 
 
@@ -106,6 +114,13 @@ async def create(
     )
     if organization is None:
         raise ResourceNotFound()
+
+    await assert_organization_permission(
+        session,
+        auth_subject,
+        organization.id,
+        OrganizationPermission.finance_manage,
+    )
     return await payout_service.create(session, locker, organization)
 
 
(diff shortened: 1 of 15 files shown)


server/polar/payout/endpoints.py
```diff
@@ -84,6 +86,12 @@ async def get_estimate(
     if organization is None:
         raise ResourceNotFound()
 
+    await assert_organization_permission(
+        session,
+        auth_subject,
+        organization.id,
+        OrganizationPermission.finance_read,
```
frankie567: I would put this one behind `finance_manage` as its only purpose is to give the estimated payout amount.

frankie567 (approved):

server/polar/payout/endpoints.py
joebon: Brilliant, thanks - will fix

Yopi (approved):

PR #11676, opened 2026-05-17
dx: remove GitHub Codespaces support

frankie567:
Wanted to do a bit of a cleanup of our dev environment, and I don't think anyone is actually using GitHub Codespaces.


pieterbeulque (approved):
Those are the nice diffs 😊

PR #11680, opened 2026-05-17
fix(order): prevent dunning retry for void orders

frankie567:
This fixes an inconsistent state where an Order could have `status=void` but also a non-null `next_payment_attempt_at` field.

## The Problem

When an order was voided, it correctly set `status: void` and `next_payment_attempt_at: None`. However, if a dunning task was already queued for this order, the following sequence would occur:

1. `process_dunning_order` task runs for the order
2. It loads the order with `status: void` and `next_payment_attempt_at: None`
3. Since `next_payment_attempt_at is None`, `handle_payment_failure` calls `_handle_first_dunning_attempt`
4. `_handle_first_dunning_attempt` **did not check if the order was void** and would set a new `next_payment_attempt_at` value

Result: Order has `status: void` and `next_payment_attempt_at: <some date>` — an inconsistent state.

## The Fix

- **`_handle_first_dunning_attempt`**: Added `is_void` check to return early with `next_payment_attempt_at` cleared
- **`process_dunning_order`**: Added `is_void` check to skip processing void orders and clear `next_payment_attempt_at`
- **`get_due_dunning_orders`**: Added `Order.is_void.is_(False)` filter to prevent void orders from being fetched for dunning

## Related Issues

This fix also helps address [SERVER-33A](https://polar-sh.sentry.io/issues/SERVER-33A) where `trigger_payment` was being called for non-pending (void) orders, resulting in `OrderNotPending` errors.

## Testing

Added two new test cases:
- `TestHandlePaymentFailure.test_void_order_with_null_next_payment_attempt`
- `TestProcessDunningOrder.test_process_dunning_order_void_order`

All existing tests continue to pass.

Generated by Mistral Vibe.
Co-Authored-By: Mistral Vibe <vibe@mistral.ai>

### DIFF
--- a/server/polar/order/repository.py
+++ b/server/polar/order/repository.py
@@ -149,6 +149,7 @@ class OrderRepository(
             .where(
                 Order.next_payment_attempt_at.is_not(None),
                 Order.next_payment_attempt_at <= utc_now(),
+                Order.is_void.is_(False),
                 Organization.is_deleted.is_(False),
                 Organization.can_renew_subscriptions.is_(True),
             )
--- a/server/polar/order/service.py
+++ b/server/polar/order/service.py
@@ -2167,6 +2167,17 @@ class OrderService:
         """Handle the first dunning attempt for an order, setting the next payment
         attempt date and marking the subscription as past due.
         """
+        # Don't schedule retries for void orders
+        if order.is_void:
+            log.info(
+                "Ignoring first dunning attempt for void order",
+                order_id=order.id,
+            )
+            repository = OrderRepository.from_session(session)
+            return await repository.update(
+                order, update_dict={"next_payment_attempt_at": None}
+            )
+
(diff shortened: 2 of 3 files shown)


pieterbeulque (approved):
Good catch! Do we need to fix any of this state in production? I guess the few lingering ones (if any) will just not trigger again with the changes to the repository method.

psincraian (approved):
Good catch!

frankie567:
> Good catch! Do we need to fix any of this state in production? I guess the few lingering ones (if any) will just not trigger again with the changes to the repository method.

I thought about having a migration, but since we make sure to reset the field here: 

https://github.com/polarsource/polar/blob/a4ecb7e8f59837323c043b7b505014953268c726/server/polar/order/service.py#L2394-L2402

They should naturally be fixed over time.

PR #11683, opened 2026-05-18
polar: Add subscription.revoke handler

Yopi:
If a subscription gets revoked, lets resubscribe the customer to the free plan


frankie567 (approved):

pieterbeulque:
Say this handling is delayed, what happens in the gap?

Yopi:
> Say this handling is delayed, what happens in the gap?

We still have the default values for the fees, which are the same as the free subscription. And we are only juggling the tier for the plain tenant in the benefit grant which will not remove them from Plain. So essentially nothing will happen I think

PR #11684, opened 2026-05-18
tailscale: Set render node to be ephemeral

Yopi:
Since it's running in Render, it might be recreated, so the node should be ephemeral rather than static


frankie567:
Just for my curiosity, what does it change? :)

Yopi:
> Just for my curiosity, what does it change? :)

Sorry, updated the description 😄. Basically Tailscale has two types of nodes: static and ephemeral. A static node might be your computer or something that will never change, whereas dockerised instances that might be recreated (during deploy or otherwise) count as ephemeral. The setting ensures that Tailscale removes the old nodes from being tracked since they will never come up again

PR #11253, opened 2026-04-28
Split AccountPage.tsx into multiple parts

sebastianekstrom:
As a preparation for https://github.com/polarsource/feedback/issues/113, this PR will refactor the `AccountPage.tsx` component, which is quite complex, and split it into multiple parts:

- `<AccountPageApproved />`
- `<AccountPageDetailsRequired />`
- `<AccountPageInReview />`

Where `page.tsx` is in charge of orchestrating what component should be rendered.

PR #11690, opened 2026-05-18
Add back payout account and verification sections for non-active orgs

sebastianekstrom:
Fix a regression from https://github.com/polarsource/polar/pull/11253 where we drop the `Payout account` and `Account verification` sections.

| Before | After  |
|--------|--------|
| Screenshot 2026-05-18 at 11 02 31 | Screenshot 2026-05-18 at 11 01 57 |


joebon (commented):
Hrm, I don't see how this solves the issue as I understand it:
> There is no UI path for going through identity verification.

AFAIK this is only adding back info cards encouraging users to go through review first.

What am I missing? 😬

pieterbeulque:
Yeah, @joebon is right. The original view had a difference between initial review (i.e. details required) and ongoing review organizations (in the latter people could still go through identity verification).

Let's tackle it differently, I'll open something up after lunch.

PR #11692, opened 2026-05-18
invoice: pick Polar's VAT number by the other party's address

pieterbeulque:
The VAT number shown on an invoice was hardcoded in `INVOICES_ADDITIONAL_INFO` (set via Terraform), so every customer saw the EU OSS VAT, including UK customers, who should see the UK VAT, and non-EU/non-UK customers who shouldn't see a VAT number at all.

Move the seller-side `additional_info` out of settings and into a helper that picks the right VAT registration from the other party's billing address. For now, that's EU OSS for EU members, GB VAT for the UK, none otherwise, but we can expand this easily even to a state-level if needed.

I couldn't think of another solution than moving this out from config into application code but I'd understand if you don't like that.


frankie567 (changes requested):
It's basically a map, right?

So I think we could have a configuration variable typed as `dict[str, str]` mapping a country code to a tax id. For EU, we could set the key as "eu" and reuse your logic to identify if it's a EU country.

Pydantic is fine with parsing environment variables in JSON format, so this should be fine.

We can also keep existing `invoices_additional_info` variable to keep the current top line.

pieterbeulque:
Yeah, that was going to be my initial approach until I thought of state-level. But maybe we should only fix state-level once state-level is a problem 😁

pieterbeulque:
@frankie567 Cleaned it up, decided to stay explicit on all EU OSS VAT member states for clarity.

frankie567 (approved):

pieterbeulque:
@nikitaphul - Can you confirm it's the correct approach to show EU VAT number for customers from all EU OSS member countries, UK VAT number for UK customers, and no VAT number for all other legislations? We can tweak this easily now per-country.

nikitaphul:
@pieterbeulque :

- for UK VAT we do need to show our UK VAT number on invoices issued to customers (GB458254961)

- For EU OSS member countries we should be showing our EU OSS reg no on invoices issued to customers (EU372061545). Exceptions here include Switzerland, Norway, Liechstentein and Iceland. 

- for US customers the US does not have a federal VAT system; instead it has sales tax administered at the state and local level, collected at the point of sale by the seller. There is no equivalent of a US VAT number to put on invoices (so do not include anything as a VAT no here as it causes confusion). Polar should appear on the invoice as the seller of record, along with the dollar amount of the sales tax collected, at the applicable state rate. 

This covers the main ones I think but the overall concept is that for each jurisdiction there is a specific VAT number that polar should apply (I am working on getting a list of these together so we are clear on what number should appear where).

pieterbeulque:
@nikitaphul - As you mentioned on Slack, I've now also removed our VAT number on the payout invoices as people are invoicing our US entity there.

Is this then fine to ship as-is or is this blocked until we've added other VAT numbers like Switzerland / Australia / etc. Not sure if we do more or less harm with a missing VAT number compared to our current situation of the potentially wrong VAT number (EU OSS) 😁

nikitaphul:
For the payout invoices it does more harm than good to have a random VAT number so i think we're good to ship as it is

terraform/production/render.tf
```diff
@@ -240,21 +240,71 @@ module "production" {
...
+    email_from_name            = "Polar"
+    email_from_domain          = "notifications.polar.sh"
+    frontend_base_url          = "https://polar.sh"
+    checkout_base_url          = "https://buy.polar.sh/{client_secret}"
+    jwks_path                  = "/etc/secrets/jwks.json"
+    log_level                  = "INFO"
+    testing                    = "0"
+    auth_cookie_domain         = "polar.sh"
+    invoices_additional_info   = "[support@polar.sh](mailto:support@polar.sh)"
+    invoices_vat_numbers = jsonencode({
```
Yopi, line 258: Nice!

And nice usage of jsonencode. We should probably use it for the others as well

Yopi (approved):
Nice work!

terraform/production/render.tf
pieterbeulque: Claude says thank you!

terraform/test/render.tf
```diff
@@ -202,9 +202,59 @@ module "test" {
     log_level                  = "INFO"
     testing                    = "0"
     auth_cookie_domain         = "test.polar.sh"
-    invoices_additional_info   = "[support@polar.sh](mailto:support@polar.sh)\nVAT: EU372061545"
-    tax_processors             = "[\"numeral\",\"stripe\"]"
-    tax_record_processor       = "numeral"
+    invoices_additional_info   = "[support@polar.sh](mailto:support@polar.sh)"
+    invoices_vat_numbers = jsonencode({
+      # EU One Stop Shop (OSS) registration
+      AT = "EU372061545"
```
pieterbeulque: @Yopi Do you know of a way to avoid this duplication for the test env?

Yopi: I was going to say that it is unavoidable, but there are a few ways you can do it.

One would be to:
1. Add a local variable in the global state
2. Add a variable for each env (prod, sandbox, test) in the respective variable set with a default value of the local variable
3. Set  invoices_vat_numbers = var.invoices_vat_numbers (or whatever name you chose)

Actually, you could put them in the global variable set instead, like here https://github.com/polarsource/polar/blob/main/terraform/global/main.tf#L214

Then you can avoid the local variable as well, and just a global variable with a default value

frankie567: Do we actually care about that value in test environment?

Yopi: No, not really 😄

pieterbeulque: Thanks for the explanation!

To François' point, I'll just make test match sandbox (i.e. no values) instead of production for these values.

PR #11697, opened 2026-05-18
billing: Disable upgrading before account is reviewed

Yopi:
Only allow upgrading to a paid plan until after the organization has been reviewed and approved

### DIFF
--- a/server/polar/integrations/polar/service.py
+++ b/server/polar/integrations/polar/service.py
@@ -15,7 +15,8 @@ from polar.account.repository import AccountRepository
 from polar.config import settings
 from polar.exceptions import PolarError
 from polar.integrations.plain.service import plain as plain_service
-from polar.postgres import AsyncSession
+from polar.organization.repository import OrganizationRepository
+from polar.postgres import AsyncReadSession, AsyncSession
 from polar.worker import enqueue_job
 
 from .client import get_client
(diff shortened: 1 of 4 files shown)


server/polar/integrations/polar/service.py
```diff
@@ -249,6 +268,20 @@ async def change_plan(
...
+        session: AsyncReadSession,
+        *,
+        organization_id: uuid.UUID,
+        target_product_id: str,
+    ) -> None:
+        if target_product_id == settings.POLAR_FREE_PRODUCT_ID:
+            return
+        review_repository = OrganizationReviewRepository.from_session(session)
+        review = await review_repository.get_by_organization(organization_id)
+        if review is None or not review.is_approved:
```
frankie567: Can you confirm @psincraian this is the way to go for this use-case?

Yopi: Ah, should I be looking at `Organization.is_active` instead?

psincraian: The onboarding process is changing right now.

Before, the AI review occurred when merchants submitted organisation details. This posed a problem because merchants without websites were denied or approved without sufficient information.

Now, the AI review will happen after details are submitted, a Stripe account is set up and identity is verified. 

Given this new flow, I recommend using Organisation.is_active. It also takes into account organization in offboarding, denied or blocked.

frankie567 (approved):

PR #11699, opened 2026-05-18
docs: revise Security Policy

frankie567:
- docs: revise Security Policy

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -624,7 +624,7 @@
       {
         "tab": "Support",
         "icon": "square-question",
-        "pages": ["support"]
+        "pages": ["support", "security"]
       }
     ]
   },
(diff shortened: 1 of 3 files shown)


docs/docs.json
```diff
@@ -642,6 +642,10 @@
       {
         "label": "Contact support",
         "href": "https://polar.sh/docs/support"
+      },
+      {
+        "label": "Security Policy",
+        "href": "https://github.com/polarsource/polar/blob/main/SECURITY.md"
```
frankie567: Might be better to add it as a proper doc page as well, but wasn't sure where to put it in the current hierarchy. Any advices, @pieterbeulque?

pieterbeulque: Maybe here: https://polar.sh/docs/support?

pieterbeulque (approved):

PR #11712, opened 2026-05-18
Disable Stripe assistant

pieterbeulque:


clients/apps/web/src/utils/stripe.ts
```diff
@@ -0,0 +1,6 @@
+import { loadStripe } from '@stripe/stripe-js'
+
+export const loadPolarStripe = () =>
```
pieterbeulque, line 3: I wonder if this should be `loadStripe` and we alias/wrap the underlying Stripe method.

sebastianekstrom (approved):

PR #11718, opened 2026-05-18
feat: add org member role management capabilities to dashboard

joebon:
## Summary

**Related Issue**: #11402

Frontend slice of the organization RBAC cutover. Wires the dashboard up to the new role/permission model and lets admins manage member roles from the members list, on top of the backend foundation shipped in #11554 and #11609.

## What

- **Backend**: expose `role` on the per-user organizations payload and add `GET /v1/organizations/{id}/roles` returning each role's permission set.
- **Frontend hooks**: `useHasPermission`, `useOrganizationRoles`, plus invite/remove/update-role member mutations.
- **Members page**: collapses the inline role `Select` + `Remove` button into a single MoreVert dropdown ("Change role" / "Remove from organization"). The change-role modal renders a Permission × Role matrix (Owner / Admin / Member) sourced from `/roles` so managers can compare grants before assigning.
- **Finance pages**: income/payouts gated on `finance:read`; access-restricted screen unified.
- **Members management UI**: gated behind `members:manage`.

#### Members page
image

#### Change role modal
image

#### Confirmation toast
image

## Why

Backend RBAC enforcement has landed (#11554, #11609), but the dashboard still gated UI on legacy role booleans and exposed a flat Select/Remove for member admin. This PR replaces those with permission-driven gates and an action surface that matches the new role model, so the UX reflects what the API will actually allow.

## How

- Backend adds the `/roles` endpoint backed by the existing `ROLE_PERMISSIONS` mapping, plus a `role` field on the per-user organizations join. No policy changes — read-only data.
- Frontend introduces a `useHasPermission(orgId, permission)` hook that resolves against the user's role for the org, and a `useOrganizationRoles` hook that caches the `/roles` response.
- Pages and conditional UI swap their old role checks for `useHasPermission`. Finance pages share a single `AccessRestricted` screen.
- Members list keeps the existing remove confirm + leave flows, but moves the per-row actions into a shadcn `DropdownMenu`. Role changes go through a dedicated modal that fetches `/roles` and renders the full permission matrix.

## Checklist


emilwidlund:
Neat @joebon! Great work on this. One small nitpick - the modal is a little bit too large. I'd be happy if we set an arbitrary max-height and let it `overflow-y-auto` within.

joebon:
> Neat @joebon! Great work on this. One small nitpick - the modal is a little bit too large. I'd be happy if we set an arbitrary max-height and let it `overflow-y-auto` within.

Thanks @emilwidlund, good point! I did it but immediately realized it caused the save and cancel buttons to go below the scroll view so I moved the overflow to the permissions table instead, like this:
image

WDYT?

clients/apps/web/src/hooks/permissions.ts
```diff
@@ -0,0 +1,29 @@
+import { useAuth } from '@/hooks/auth'
+import {
+  OrganizationPermission,
+  useOrganizationRoles,
+} from '@/hooks/queries/roles'
+import { schemas } from '@polar-sh/client'
+
+export const useOrganizationRole = (
```
pieterbeulque: Would it be better to put this under `useAuth` as `role` or `userRole` or something?

clients/apps/web/src/hooks/permissions.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+// Returns false while roles map is loading; callers can use this to disable
+// controls during fetch.
+export const useHasPermission = (
+  organizationId: string | undefined,
+  permission: OrganizationPermission,
+): boolean => {
+  const role = useOrganizationRole(organizationId)
+  const { data: roles } = useOrganizationRoles(organizationId)
```
pieterbeulque: Similarly here, I feel like we should be able to fetch the organization role once and keep this in context.

pieterbeulque: And I say that as someone who doesn't generally like to use context 😁

pieterbeulque: Another idea (just spitballing, might be absolutely shit), could the current user's permissions be part of the response of `/v1/users/me` and thus live on `const { currentUser } = useAuth()`?

We could then have a helper `const { hasPermissions } = useAuth()`.

Or is that weird because `/v1/users/me` is outside of the context of a single organization.

server/polar/organization/schemas.py
```diff
@@ -405,6 +407,28 @@ class Organization(OrganizationBase):
     )
 
 
+class OrganizationListItem(Organization):
```
pieterbeulque: Doesn't feel like the most descriptive name given the comment below

clients/apps/web/src/hooks/permissions.ts
joebon: Yeah, we can probably do a bit better with the hooks.

Good point re: `/v1/users/me`!
Unfortunately, just like you point out, `/v1/users/me` doesn't imply a specific organization.
We _could_ embed `organizations` to the response and then have each object describe the caller's role within that specific organization though. It's semantically better placed there (because `me`), but there's usually something _smelly_ around when you start embedding related objects in REST responses IMHO.

In this particular case though I think I vote for bringing in the orgs the user belongs to along with their role(s).

WDYT @pieterbeulque ?

server/polar/organization/schemas.py
joebon: Hah, good point! Perhaps just be silly clear and call it `OrganizationWithRole`?

server/polar/organization/endpoints.py
```diff
@@ -106,16 +111,17 @@
...
 )
-async def list(
+async def list_organizations(
     auth_subject: auth.OrganizationsRead,
     pagination: PaginationParamsQuery,
     sorting: sorting.ListSorting,
     slug: str | None = Query(None, description="Filter by slug."),
     session: AsyncReadSession = Depends(get_db_read_session),
-) -> ListResource[OrganizationSchema]:
+) -> ListResource[OrganizationListItem]:
```
frankie567: I'm not super comfortable with this change: having a specific schema only on the list endpoint is not really expected (this endpoint is public); esp. if it's just to output the role. 

It's somewhat kinda related to @pieterbeulque's concerns on `useHasPermission`. 

I'm wondering if we should introduce a `/v1/users/me/organizations/` for this purpose.

clients/apps/web/src/hooks/permissions.ts
pieterbeulque: I think you're right, `/v1/users/me` isn't the right place for it. We could still lift the second API call up into the auth context to keep everything together, maybe.

server/polar/organization/endpoints.py
joebon: Yeah, something like that. You're opposed to enriching `/v1/users/me` with an `organizations` key/list?

clients/apps/web/src/hooks/permissions.ts
joebon: Hah, god damn - I started to lean towards it _being_ the right place. That or @frankie567's suggestion to add `/v1/users/me/organizations` but I the org+role information seem tightly connected so I don't really see the point in having a separate endpoint for that tbh. In general I'd be happy either way.

pieterbeulque: Go for it! If nothing else then having everything bundled in a single network call is a good argument to ensure client state is in sync.

server/polar/organization/endpoints.py
frankie567: Not at all, that would work great as well :)

clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/members/MembersPage.tsx
```diff
@@ -198,16 +252,18 @@ export default function ClientPage({
...
+        ) : undefined
       }
     >
       <p className="dark:text-polar-500 text-gray-500">
-        Manage users who have access to this organization. All members are
-        entitled to view and manage organization settings, products,
-        subscriptions, etc.
+        Manage users who have access to this organization. Roles determine what
+        each member can do — admins can manage finance and members, while
+        members can manage day-to-day resources.
```
Comment (joebon) [inline clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/members/MembersPage.tsx]: Note to self: potentially reword

clients/apps/web/src/hooks/permissions.ts
joebon: We could, but it'd still need to be a method with an `orgId` argument - `useAuth` is not organization specific. What do you think about the ergonomics of this now that I moved the org roles to `/v1/users/me`?

clients/apps/web/src/hooks/permissions.ts
joebon: Fixed

server/polar/organization/endpoints.py
```diff
@@ -108,8 +110,9 @@
     summary="List Organizations",
     response_model=ListResource[OrganizationSchema],
     tags=[APITag.public],
+    operation_id="organizations:list",
 )
-async def list(
+async def list_organizations(
```
joebon, line 119: Not only cosmetic/noisy rename, apparently necessary to due to conflict with `roles` because of `response_model=list[…]` shadowing

emilwidlund (approved):

server/polar/user_organization/repository.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+
+    def __init__(self, session: AsyncReadSession) -> None:
+        self.session = session
+
+    @classmethod
+    def from_session(cls, session: AsyncReadSession) -> Self:
+        return cls(session)
+
+    async def get_roles_by_org_ids(
```
frankie567: This one is now unused it seems

frankie567 (approved):

server/polar/user_organization/repository.py
joebon: 🤦  thanks! removing...

PR #11719, opened 2026-05-18
Fix Transactions API: account_amount ignores platform fees during payout-delay window

frankie567:
Fix for #11705


joebon (approved):
Nice!

PR #11726, opened 2026-05-19
Avoid overriding customer name with cardholder

pieterbeulque:
I went back and forth quite a bit and I'm not even sure if this is the correct approach.

Basically, I noticed that when you have a company customer "ACME" that you create a checkout for, and you check out with _Cardholder name_ "John Doe", we'll update the customer's name to be "John Doe" after checkout.

That's very annoying in a B2B context.

### DIFF
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -2566,15 +2566,24 @@ class CheckoutService:
                 # Could be a malicious user trying to take over an existing customer's account by using their email
                 generate_customer_session = False
 
+        # Preserve any existing customer.name (e.g. a person's name) so it isn't
+        # overridden by the cardholder name on a wallet/payment method when an
+        # employee or third party checks out on the customer's behalf.
+        # customer.billing_name (the company name) is intentionally updatable.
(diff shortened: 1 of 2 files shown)


server/polar/checkout/service.py
```diff
@@ -2599,8 +2596,13 @@ async def _create_or_update_customer(
...
 
-        if checkout.customer_name is not None:
-            customer.name = checkout.customer_name
+        # Only populate customer.name when creating a new customer. For existing
+        # customers (linked via customer_id or matched by email),
+        # checkout.customer_name may be the cardholder name on a wallet/payment
+        # method — e.g. a CFO or office manager paying on behalf of a company
+        # customer — and must not overwrite the company's name.
+        if created and customer_name is not None:
+            customer.name = customer_name
```
pieterbeulque, line 2609: Basically, this is the change:

 - Only set/write name on `created`, avoiding overrides
 - Use billing name if available, then cardholder name

frankie567 (approved):
Makes sense!

PR #11728, opened 2026-05-19
polar: Add cleanup script to remove free subscriptions

Yopi:
Since we are rolling back our decision and going with no subscription = free tier, then we need remove the needlessly created subscriptions.

### DIFF
--- a/server/scripts/cleanup_polar_self_free_subscriptions.py
+++ b/server/scripts/cleanup_polar_self_free_subscriptions.py
@@ -2,6 +2,7 @@ import asyncio
 import uuid
 from typing import Any
 
+import structlog
 import typer
 from rich.progress import Progress
 from sqlalchemy import delete, or_, select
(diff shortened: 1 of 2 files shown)


server/scripts/cleanup_polar_self_free_subscriptions.py
```diff
@@ -0,0 +1,276 @@
...
+
+    organization_id = uuid.UUID(settings.POLAR_ORGANIZATION_ID)
+    free_product_id = uuid.UUID(settings.POLAR_FREE_PRODUCT_ID)
+
+    engine = create_async_engine("script")
+    sessionmaker = create_async_sessionmaker(engine)
+
+    try:
+        async with sessionmaker() as session:
+            subscription_ids = await _list_free_subscription_ids(
```
frankie567, line 202: Won't that blow up in memory? Not sure how many we have right now.

frankie567 (approved):
Good, but slightly worried we might blow up in memory by listing all subscriptions/orders/events ID.

Yopi:
> Good, but slightly worried we might blow up in memory by listing all subscriptions/orders/events ID.

Rightly so 😄. The footgun I love! I'll refactor it

server/scripts/cleanup_polar_self_free_subscriptions.py
Yopi: I think this one in particular is fine, it's only a few MB at most. However we were using them in an IN query later. I refactored this to batch it instead. Fairly sure it will work

PR #11730, opened 2026-05-19
Unify member and customer token in payment method SDK

sebastianekstrom:
* Unify the `customerSessionToken` and `memberSessionToken` into just `sessionToken`, the backend will figure out the type.
* Update README with docs on supported props/params

### DIFF
--- a/clients/.changeset/hot-rooms-cheat.md
+++ b/clients/.changeset/hot-rooms-cheat.md
@@ -1,0 +1,5 @@
+---
+"@polar-sh/checkout": patch
+---
+
+Simplify the payment method embed API with a single sessionToken prop/parameter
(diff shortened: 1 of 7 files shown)


Yopi (approved):
Very nice

pieterbeulque (commented):
Not against this, but curious what triggered this, and if this is a helpful pattern, should we maybe elevate this to also work on the customer portal itself?

sebastianekstrom:
@pieterbeulque didn't really see the reason for the consumer of the SDK to keep track of what _type_ of token it's passing since the backend will handle both either way, but happy to be challenged on that! I'm lacking some context on the both types of token so if it makes sense to be more explicit (`customerSessionToken` and `memberSessionToken`) I'm happy to go with that.

PR #11732, opened 2026-05-19
Send subscription create & cycle emails

pieterbeulque:


server/polar/integrations/polar/client.py
```diff
@@ -334,6 +335,27 @@ async def get_member_by_external_id(
             external_customer_id=external_customer_id,
         )
 
+    async def list_billing_contacts(self, *, customer_id: str) -> list[Member]:
```
pieterbeulque, line 369: I think this is super awkward.

server/polar/integrations/polar/service.py
```diff
@@ -464,6 +472,80 @@ async def handle_benefit_grant_event(
...
                     await self._apply_support(session, organization_id, active_grant)
 
+    async def handle_order_created_event(
+        self, payload: WebhookOrderCreatedPayload
+    ) -> None:
+        order = payload.data
+
+        if order.billing_reason not in (
+            OrderBillingReason.SUBSCRIPTION_CREATE,
+            OrderBillingReason.SUBSCRIPTION_UPDATE,
```
pieterbeulque, line 502:
```suggestion
```

Now that our free plan is no longer a product.

server/polar/integrations/polar/service.py
```diff
@@ -464,6 +472,80 @@ async def handle_benefit_grant_event(
...
+        recipients = sorted({contact.email for contact in contacts if contact.email})
+        if not recipients:
+            return
+
+        product_name = order.product.name if order.product is not None else "Polar"
+
+        if order.billing_reason in (
+            OrderBillingReason.SUBSCRIPTION_CREATE,
+            OrderBillingReason.SUBSCRIPTION_UPDATE,
+        ):
```
pieterbeulque:
```suggestion
        if order.billing_reason == OrderBillingReason.SUBSCRIPTION_CREATE:
```

server/polar/integrations/polar/service.py
```diff
@@ -464,6 +472,80 @@ async def handle_benefit_grant_event(
...
+            template_name = "polar_self_subscription_confirmation"
+            subject = f"Welcome to {product_name}"
+        else:
+            template_name = "polar_self_subscription_cycled"
+            subject = f"Your {product_name} subscription renewed"
+
+        attachments: list[Attachment] | None = None
+        if order.is_invoice_generated:
```
pieterbeulque: Not sure if this'll work or if this can be a runtime issue where the invoice isn't attached if it's not done yet. Should we wait for it / poll for it / …?

server/polar/integrations/polar/client.py
Yopi: Because of the indirection via Polar?

server/polar/integrations/polar/service.py
Yopi: Is it always generated? Do we need a separate webhook for it in that case?

Yopi (commented):
I think this makes sense!

server/polar/integrations/polar/client.py
pieterbeulque: No, more conceptually. It feels weird to require an extra API call altogether. It feels even more weird to do two.

I feel like most other billing platforms have a `billing_email_address` on a customer or something where you send this kind of notifications to.

frankie567 (commented):
Do we _really_ need custom emails for that?

pieterbeulque:
@frankie567 

Fair point. This is our built-in email.

image

I don't like that it links to the customer portal, which we don't use, and it leaks our custom benefits, which I don't like.

And "Merchant of Record services provided to Polar by Polar Software, Inc." in the footer is super silly.

So overall, I think it's a bit too meta to rely on our built-in emails, and things can be confusing if we keep them enabled.

server/polar/integrations/polar/client.py
Yopi: That's a fair point, and maybe it would make more sense to actually query this out from Polar rather than use the API for it.

Was discussing something similar with Joel today in terms of RBAC and the duplication between users and members. It might just be unnecessary noise after all?

server/polar/integrations/polar/service.py
```diff
@@ -492,6 +506,93 @@ async def handle_benefit_grant_event(
...
+        if order.billing_reason not in (
+            OrderBillingReason.SUBSCRIPTION_CREATE,
+            OrderBillingReason.SUBSCRIPTION_UPDATE,
+            OrderBillingReason.SUBSCRIPTION_CYCLE,
+        ):
+            return
+
+        # Every polar-self customer always has a subscription (free auto-created on
+        # org creation / after revocation). Free-priced orders shouldn't trigger
+        # the welcome or renewal email.
```
Yopi: No longer true

server/scripts/seeds_load.py
```diff
@@ -1421,6 +1423,19 @@ async def create_seed_data(session: AsyncSession, redis: Redis) -> None:
                 "seat_based_pricing_enabled": True,
                 "member_model_enabled": True,
             },
+            "customer_email_settings": {
```
Yopi, line 1410: 👍

Yopi (approved):

server/polar/integrations/polar/client.py
pieterbeulque: Ah, as in, our implementation uses its own logic to retrieve a billing address, we don't rely on Polar state for that.

I guess we could! It's a bit of a smell that we want to do that versus relying on Polar for that, but that's why we're doing this. 😊

server/polar/integrations/polar/service.py
```diff
@@ -492,6 +506,93 @@ async def handle_benefit_grant_event(
...
+        if order.net_amount == 0:
+            return
+
```
pieterbeulque:
```suggestion
```

PR #11733, opened 2026-05-19
fix: Allow org members to manage refunds

joebon:
## Summary

Org members couldn't issue refunds — refund creation was gated on the admin-only `finance_manage` permission. This PR introduces a `sales_manage` permission granted to all org roles and moves refund creation onto it. Payouts continue to gate on `finance_manage`.

## What

- Add `OrganizationPermission.sales_manage` (granted to `member`, `admin`, `owner`)
- Switch `refund_service.user_create` from `finance_manage` to `sales_manage`
- Add endpoint test that exercises refund creation as a `member`-role user

## Why

Refund listing already used the member-accessible `sales_read`, so members could see refunds but not create them. Refunds are operational customer-facing work, not finance-admin work like payouts.

## How

`finance_manage` stays admin-only and continues to gate payouts (money flowing out to the org's bank). Refunds get their own member-accessible permission under the `sales` namespace, mirroring the `*_read` / `*_manage` pattern used by customers, products, and analytics.

## Checklist

### DIFF
--- a/server/polar/auth/permission.py
+++ b/server/polar/auth/permission.py
@@ -42,6 +42,7 @@ class OrganizationPermission(StrEnum):
 
     # Sales — granted to all roles.
     sales_read = "sales:read"
+    sales_manage = "sales:manage"
 
     # Analytics.
     analytics_read = "analytics:read"
@@ -73,6 +74,7 @@ _MEMBER_PERMISSIONS: set[OrganizationPermission] = {
     OrganizationPermission.customers_read,
     OrganizationPermission.customers_manage,
     OrganizationPermission.sales_read,
+    OrganizationPermission.sales_manage,
     OrganizationPermission.analytics_read,
     OrganizationPermission.analytics_manage,
     OrganizationPermission.events_ingest,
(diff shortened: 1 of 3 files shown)


server/polar/auth/permission.py
```diff
@@ -42,6 +42,7 @@ class OrganizationPermission(StrEnum):
 
     # Sales — granted to all roles.
     sales_read = "sales:read"
+    sales_manage = "sales:manage"
```
joebon, line 45: Added this new permission for this. But could also reuse `customers:manage` if that's more fitting?

pieterbeulque (approved):
I like the explicitness

frankie567 (approved):
Let's go `sales_manage`! Didn't we talk about that refund permission before? I remember we had a similar debate.

joebon:
> Let's go `sales_manage`! Didn't we talk about that refund permission before? I remember we had a similar debate.

Yeah, I believe we've been through many of these before @frankie567 . Big surface area, easy to miss certain scenarios I guess

PR #11738, opened 2026-05-19
billing: cancel as downgrade

Yopi:
This removes free subscriptions as a subscription, and makes downgrade to free a cancelation.

If you are an early member, it shows up instead of the free plan.

Sets prorations according to if you are upgrading or downgrading.

### DIFF
--- a/server/polar/config.py
+++ b/server/polar/config.py
@@ -416,6 +416,8 @@ class Settings(BaseSettings):
     }
     PLATFORM_FEE_BASIS_POINTS: int = 400
     PLATFORM_FEE_FIXED: int = 40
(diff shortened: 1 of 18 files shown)


server/polar/config.py
```diff
@@ -416,6 +416,8 @@ def POLAR_SELF_ENABLED(self) -> bool:
     }
     PLATFORM_FEE_BASIS_POINTS: int = 400
     PLATFORM_FEE_FIXED: int = 40
+    PLATFORM_FEE_BASIS_POINTS_EARLY_ACCESS: int = 400
+    PLATFORM_FEE_FIXED_EARLY_ACCESS: int = 40
```
frankie567, line 420: Smart!

frankie567 (approved):

PR #11720, opened 2026-05-18
Add "Payment method" SDK

sebastianekstrom:
This PR will add a new embedded payment method feature.

The new embedded component borrows _a lot_ of code from our already existing [AddPaymentMethodModal.tsx](https://github.com/polarsource/polar/blob/ccfe617b46524e6e0ddc759b80b45ca6fd31bffe/clients/apps/web/src/components/CustomerPortal/AddPaymentMethodModal.tsx) component in our Customer Portal. The plan is to deprecate that component in favor of this new SDK.

It also mimics the `embed.ts` API with the same methods.

Three notable changes:

* `embed.ts` is now renamed to `checkout.ts` to better convey what it does
* `embed.ts` now exports the renamed `checkout.ts`  (old `embed.ts`)
* `embed.ts` now exports the new `payment-method.ts` (adds ~3KB to bundle size)

The other changes should not have any effect on the already existing embed checkout feature.

## Example usages

```ts
// Node
import { PolarEmbedPaymentMethod } from '@polar-sh/checkout/payment-method'

const embed = await PolarEmbedPaymentMethod.create({
  customerSessionToken, // or memberSessionToken
})

embed.addEventListener('success', (event) => {
  console.log({event})
})
```

```tsx
// React
import { PolarPaymentMethod } from '@polar-sh/checkout/react/payment-method'
return (
  <PolarPaymentMethod
    customerSessionToken={token} // or memberSessionToken
    onSuccess={(event) => console.log({event})}
  />
)
```

```html
// Code snippet, using the same embed.global.js path
<script
  defer
  data-auto-init
  src="https://cdn.jsdelivr.net/npm/@polar-sh/checkout@latest/dist/embed.global.js">
</script>
<button data-polar-payment-method="polar_cst_xxx">Add payment method</button>
```

## Screenshot

Screenshot 2026-05-18 at 16 36 05

PR #11739, opened 2026-05-19
Add the new "Payment method" SDK to /billing

sebastianekstrom:
This PR will add our new **Payment method** SDK, that was introduced in https://github.com/polarsource/polar/pull/11720 , to `/billing`. While dog fooding the new SDK I noticed a few issues that now is fixed:

* We now properly surface all (now typed) errors to the consumer
* 3DS errors will now allow you to close the modal

In addition to this I also added a new endpoint to retrieve the session token: `/v1/organizations/{id}/billing-customer-session ` which has the `organizations:write` scope. Meaning we can do the following to show the **Payment method** modal:

```tsx
import { PolarEmbedPaymentMethod } from '@polar-sh/checkout/payment-method'

const onAddPaymentMethod = async () => {
  const { data: session } = await api.POST(
    '/v1/organizations/{id}/billing-customer-session',
    { params: { path: { id: organization.id } } },
  )
 
  const embed = await PolarEmbedPaymentMethod.create({
    sessionToken: session.token,
  })

  embed.addEventListener('success', () => {
    console.log("success!")
  })
}
```

## Screenshots

| Before | After  |
|--------|--------|
| Screenshot 2026-05-19 at 18 54 31 | Screenshot 2026-05-19 at 18 48 31 |

| Before: 3DS error | After: 3DS error  |
|--------|--------|
| Screenshot 2026-05-19 at 17 20 49 | Screenshot 2026-05-19 at 18 49 17 |

## Redirected payment flow: Success
https://github.com/user-attachments/assets/6c822fcd-3bd9-4526-b0b5-e0db02988e89

## Redirected payment flow: Failure

https://github.com/user-attachments/assets/c12c7d0c-74f4-40a4-ab9f-e378d639b98f


pieterbeulque:
🤤 

One thing about this:

```ts
import { PolarEmbedPaymentMethod } from '@polar-sh/checkout/payment-method'

const onAddPaymentMethod = async () => {
  const { data: session } = await api.POST(
    '/v1/organizations/{id}/billing-customer-session',
    { params: { path: { id: organization.id } } },
  )
 
  const embed = await PolarEmbedPaymentMethod.create({
    sessionToken: session.token,
  })

  embed.addEventListener('success', () => {
    console.log("success!")
  })
}
```

That introduces a network call on click adding a bit of delay/friction. Would it make more sense to prefetch a token on mount?

sebastianekstrom:
@pieterbeulque absolutely! The code in the description was just a simplified example (sorry I should've made that more clear)

server/polar/organization/endpoints.py
```diff
@@ -935,6 +936,27 @@ async def list_payment_methods(
     )
 
 
+@router.post(
+    "/{id}/billing-customer-session",
```
Yopi: I guess this could just as well be just /customer-session given that's what it is

Yopi (approved):
Nice work! We can kill the other endpoints that I added, but we can do it in a follow up PR

server/polar/organization/endpoints.py
sebastianekstrom: Good idea, fixed in the last commit

frankie567 (approved):

Yopi (approved):
Nice work

PR #11741, opened 2026-05-19
fix: ensure `dev up` with tinybird works on orbstack

joebon:
The `--profile` option only works on the top level of `docker compose` on orbstack.
So switch `docker compose up -d --profile tinybird` to `docker compose --profile up -d` and then everyone's happy


Yopi (approved):
😺

frankie567:
Computers are _fun_, they said 😆

Issue #11404, opened 2026-05-05
Drop Account.admin_id


joebon:
**Scope.** Re-home every remaining read off `Account.admin_id`, install role-side enforcement, and drop the column. Ships after Phase 5 has soaked in production long enough to trust the role system as the live authorization signal.

**Acceptance criteria**

- [ ] `repository.get_admin_user` queries `UserOrganization WHERE role = 'owner'`.
- [ ] Backoffice `change_admin` flow mutates role directly (KYC/identity-verification check inline) instead of `Account.admin_id`.
- [ ] Notification routing and Polar-for-Polar owner-side lookup re-anchored on the user holding `owner`.
- [ ] Any helper that closed over `Account.admin_id` migrated.
- [ ] Partial unique index on `UserOrganization (organization_id) WHERE role = 'owner'` added.
- [ ] `IdentityVerificationStatus.verified` gate moved from `account_service.change_admin` to role transitions into `owner`.
- [ ] Pre-deploy check: grep gate + production audit logs show no remaining `Account.admin_id` reads over the bake period.
- [ ] Destructive migration: `Account.admin_id` dropped; dual-write code removed.

**Bake-time gate.** Concrete signals before opening this issue: zero policy-denial reports tied to role lookups; audit logs show no `Account.admin_id` reads in production for N days. Pick N before starting.

**References.** [RFC §Phase 6](https://handbook.polar.sh/engineering/design-documents/rbac#phase-6--drop-accountadmin_id), parent #6646. Depends on #11403.

Issue #11403, opened 2026-05-05
Frontend cleanup


joebon:
**Scope.** Retire the legacy `is_admin` API field. Pick and execute a stale-tab strategy.

**Acceptance criteria**

- [ ] Phase 5 strategy decided (default per RFC: accept the brief break).
- [ ] `is_admin` removed from API responses.
- [ ] No frontend code references `is_admin`.
- [ ] `Account.admin_id` ↔ `owner` dual-write **stays in place** through this phase (retired in Phase 6).

**Stale-tab strategy options**

1. **Time-based** — wait long enough after Phase 4 that weekly-active sessions have refreshed at least once.
2. **Bundle-mismatch prompt** — ship a "new version available — please refresh" prompt as part of Phase 4.
3. **Accept the break** — ship Phase 5 immediately after Phase 4 and accept that any tab still on the previous bundle will see errors until the user reloads.

RFC default is (3).

**References.** [RFC §Phase 5](https://handbook.polar.sh/engineering/design-documents/rbac#phase-5--frontend-cleanup), parent #6646. Depends on #11402.

PR #11747, opened 2026-05-20
chore: drop OrganizationMember.is_admin from API responses

joebon:
## Summary

**Related Issue**: #11403

Phase 5 of the RBAC rollout — retires the transitional `is_admin` field on `OrganizationMember` from API responses.

## What

- Drop the `is_admin` computed field from `OrganizationMember` (`server/polar/user_organization/schemas.py`)
- Update the two `TestSetMemberRole` assertions that read `is_admin` from the role-update response
- Regenerate `clients/packages/client/src/v1.ts` so the SDK no longer types the field

`Account.admin_id` ↔ `owner` dual-write **stays in place** per the issue AC; that retires in Phase 6 (#11404).

## Why

Phase 4 (#11402, merged in 528932643) migrated the dashboard from the boolean `is_admin` to the typed `role` and the `hasPermission` hook. The field was kept on the API as a transitional alias while the frontend caught up. With Phase 4 shipped and no remaining frontend callsites referencing it, the alias is dead weight on the public schema — leaving it in place risks new code grabbing the legacy boolean instead of the role/permission system that's meant to replace it.

## How

**Stale-tab strategy: accept the brief break** (RFC default). Any tab still on the pre-Phase-4 bundle hitting `/v1/organizations/{id}/members` will see `is_admin === undefined` on each member. Practical impact is minimal — the surviving callsites would only fire inside a refresh-pending tab, and Phase 4 has been live in `main` for long enough that weekly-active sessions have largely rotated.

Removing the `computed_field` is sufficient — there's no Pydantic deprecation step needed because the alias has no database backing; it was always derived from `role` at serialization time.

## Checklist

### DIFF
--- a/clients/packages/client/src/v1.ts
+++ b/clients/packages/client/src/v1.ts
@@ -24015,11 +24015,6 @@ export interface components {
       avatar_url: string | null
       /** @description The user's role on the organization. */
       role: components['schemas']['OrganizationRole']
-      /**
-       * Is Admin
-       * @description Whether the user has admin capability on the organization. Derived from `role`: true for `owner` or `admin`. Kept as a transitional alias; prefer reading `role` directly.
-       */
-      readonly is_admin: boolean
     }
     /** OrganizationMemberInvite */
     OrganizationMemberInvite: {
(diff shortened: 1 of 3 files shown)


frankie567 (approved):
🧹

PR #11749, opened 2026-05-20
Handle subscription fee in Polar for Polar

frankie567:
- server/integrations/polar: add subscription fee handling in platform fee benefit
- server/account: introduce dynamic subscription fee property


server/polar/integrations/polar/service.py
frankie567, line 203: Wasn't sure if we wanted to explicitly communicate the subscription fee? Thoughts?

server/migrations/versions/2026-05-20-0914_add_account_platform_subscription_fee_.py
```diff
@@ -0,0 +1,33 @@
...
+# Polar Custom Imports
+
+# revision identifiers, used by Alembic.
+revision = "69a2e3ae542c"
+down_revision = "b13c8489dbea"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
```
Yopi, line 21: I think the subscription fee is making it's final exit with this change, so it might be unnecessary to introduce the configurability. But I guess it also doesn't hurt to do it

server/polar/config.py
```diff
@@ -415,8 +416,10 @@ def POLAR_SELF_ENABLED(self) -> bool:
     }
     PLATFORM_FEE_BASIS_POINTS: int = 400
     PLATFORM_FEE_FIXED: int = 40
+    PLATFORM_SUBSCRIPTION_FEE_BASIS_POINTS: int = 50
```
pieterbeulque, line 418:
```suggestion
    PLATFORM_SUBSCRIPTION_FEE_BASIS_POINTS: int = 0
```

Or is this only for when we flip the switch?

server/polar/integrations/polar/service.py
Yopi: No I think not 😄

server/polar/integrations/polar/service.py
```diff
@@ -565,10 +567,12 @@ async def _apply_transaction_fee(
             return
 
         if grant is None:
-            fee_percent, fee_fixed = None, None
+            fee_percent, fee_fixed, subscription_fee_percent = None, None, None
         else:
-            fee_percent, fee_fixed = self._extract_transaction_fee(
-                grant.benefit.metadata or {}, grant.benefit_id
+            fee_percent, fee_fixed, subscription_fee_percent = (
```
Yopi, line 576: Reminder to add it to the metadata

server/polar/config.py
Yopi: All three should probably be updated, but felt safer to wait until we flip the switch

Yopi (approved):

server/migrations/versions/2026-05-20-0914_add_account_platform_subscription_fee_.py
frankie567: But how do you solve early access fee then?

server/tests/integrations/polar/test_service.py
```diff
@@ -617,6 +625,7 @@ async def test_revoked_with_no_remaining_grant_resets_fees(
         assert set_platform_fee_mock.await_args.kwargs == {
             "fee_percent": None,
             "fee_fixed": None,
+            "subscription_fee_percent": None,
```
frankie567: Wondering if we should explicitly write the configuration defaults instead of `None`? I noticed you did this when first creating an account @Yopi

server/polar/integrations/polar/service.py
frankie567: If I run the `server/scripts/seed_polar_for_polar.py`, it'll automatically update it, right?

server/tests/integrations/polar/test_service.py
Yopi: I started doing it now since we wanted to differentiate between the grandfathered and the non grandfathered subscriptions. So we probably need to do it here as well

server/migrations/versions/2026-05-20-0914_add_account_platform_subscription_fee_.py
Yopi: I was thinking we could check the plan they are on, but you're right that it becomes a bit more involved to do that just for this

Yopi: So then as per the other comment:
Write all existing accounts to the current value, and update it on subscription change

server/tests/integrations/polar/test_service.py
frankie567: Great, agree!

server/polar/integrations/polar/service.py
Yopi: Yes, but I think it's out of sync with reality now given the removal of the free subscription

server/tests/integrations/polar/test_service.py
frankie567: Changed to always write the defaults back to the record

server/migrations/versions/2026-05-20-0914_add_account_platform_subscription_fee_.py
frankie567: Added an UPDATE statement to write this fee on all records

server/polar/integrations/polar/service.py
frankie567: What would you recommend then? Do it manually from dashboard?

Yopi: Since we're using the seed script for local seed as well, I think it makes sense to update the script

frankie567: Ok, the changes are fine, I guess?

But what should we do to rollout that in sandbox/prod?

Yopi: The changes look good! I can take care of that and run the update

frankie567: Ok, I'll merge then :)

PR #11750, opened 2026-05-20
server/auth: add EmailOTP and TOTP factors

frankie567:
Add new routes for email OTP and TOTP factors.

Deliberately not reusing the existing login routes to not break existing flows, until we're not completely done here.

### DIFF
--- a/server/polar/config.py
+++ b/server/polar/config.py
@@ -121,6 +121,11 @@ class Settings(BaseSettings):
         "http://127.0.0.1:8000/v1/checkout-links/{client_secret}/redirect"
     )
 
+    # Authentication session
+    AUTHENTICATION_SESSION_TTL: timedelta = timedelta(minutes=15)
+    AUTHENTICATION_SESSION_COOKIE_KEY: str = "polar_auth_session"
+    AUTHENTICATION_SESSION_COOKIE_DOMAIN: str = "127.0.0.1"
+
     # User session
     USER_SESSION_TTL: timedelta = timedelta(days=31)
     USER_SESSION_COOKIE_KEY: str = "polar_session"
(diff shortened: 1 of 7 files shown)


server/polar/config.py
```diff
@@ -121,6 +121,11 @@ class Settings(BaseSettings):
         "http://127.0.0.1:8000/v1/checkout-links/{client_secret}/redirect"
     )
 
+    # Authentication session
+    AUTHENTICATION_SESSION_TTL: timedelta = timedelta(minutes=15)
+    AUTHENTICATION_SESSION_COOKIE_KEY: str = "polar_auth_session"
```
pieterbeulque, line 126: Probably unnecessary / unrelated but I noticed that we sometimes have cookies leak between sandbox & production as they all use `polar.sh` as the domain.

pieterbeulque (approved):

server/polar/config.py
frankie567: Oh right, good heads-up! Given the architecture of our HTTP services, we have to use the same domain.

We solved that properly for the actual user session by having a different cookie name:

Capture d’écran 2026-05-21 à 10 11 03

We should apply this strategy for all cookies.

joebon: Can't we set same site setting to `strict` instead of `lax`?
Lax allows for other subdomains on the same domain, strict doesn't. That'd stop the leaking between sandbox and prod.

frankie567: Unfortunately, no, as we need the Next.js app to access it from `polar.sh`/`sandbox.polar.sh`: we have a server-side middleware that pre-checks the validity of the session.

joebon: Got it - was worth a try! 😅

PR #11752, opened 2026-05-20
polar_self: Retry until invoice is ready

pieterbeulque:
Defer ``polar_self.webhook.order.created`` processing by one minute via Dramatiq's built-in delay so payment settlement has time to complete.

 - Refactor the handler to refetch the order over the API (the webhook payload's ``is_invoice_generated`` will be stale upon retries)
 - Skip the email entirely for free orders (``net_amount == 0``)
 - Retry rather than silently sending the email without an invoice attached when the order isn't paid yet (422 ``NotPaidOrder``) or the PDF hasn't been written.

### DIFF
--- a/server/polar/integrations/polar/service.py
+++ b/server/polar/integrations/polar/service.py
@@ -26,7 +26,7 @@ from polar.email.sender import Attachment, enqueue_email_template
 from polar.integrations.plain.service import plain as plain_service
 from polar.organization.repository import OrganizationRepository
 from polar.postgres import AsyncReadSession, AsyncSession
-from polar.worker import enqueue_job, get_retries
+from polar.worker import enqueue_job
 
 from .client import get_client
 from .exceptions import (
@@ -35,6 +35,7 @@ from .exceptions import (
     PolarSelfNoActiveSubscription,
     PolarSelfNotApproved,
     PolarSelfNotConfigured,
+    PolarSelfNotPaidOrder,
     PolarSelfOrderNotFound,
     PolarSelfPlanNotFound,
     PolarSelfWebhookError,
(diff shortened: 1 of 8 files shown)


Yopi (approved):
Looks good!

server/polar/integrations/polar/service.py
```diff
@@ -520,17 +529,18 @@ async def handle_order_created_event(
...
-            invoice_expected = await client.trigger_order_invoice_generation(
-                order_id=order.id
-            )
-
         if invoice_expected:
+            if not order.is_invoice_generated:
+                # Kick off PDF generation if the API hasn't done so yet, then
+                # retry — generation runs asynchronously on Polar's side.
+                try:
+                    await client.trigger_order_invoice_generation(order_id=order.id)
```
pieterbeulque, line 539: @Yopi If the retry then triggers before the invoice is generated, we'll trigger a new invoice generation, but that second invoice generation should be the same invoice as the first?

I just wonder if we can potentially add fuel to fire if we build up some backlog on the invoice generation and these retries keep triggering new ones.

frankie567:
Wouldn't it be simpler to actually change the main logic to always generate the invoice even if we don't send the mail? Typically, in `_on_order_created`?

server/polar/integrations/polar/service.py
Yopi: Yes it should, but I don't think it will be a huge problem

Yopi:
> Wouldn't it be simpler to actually change the main logic to always generate the invoice even if we don't send the mail? Typically, in `_on_order_created`?

Don't know if there is any historical reasons why we werent? Or is just that it felt unnecessary?

pieterbeulque:
Happy to reconsider when/how we generate invoices and simplify this logic! It's not the easiest to work with, and to say that I've recommended this workaround before to merchants 🫣

frankie567:
At the beginning, we were not automatically generating invoices, but later decided to include it directly in the email; so we did that synchronously when sending the email. Effectively, for most orgs, it means we generate the invoice automatically.

So I think it's safe/better to just always generate it after order creation

pieterbeulque:
@frankie567 Would you then do it inline before triggering the `order.created` webhook, so that it's always generated when that webhook lands? Otherwise it doesn't really solve the core problem here.

frankie567:
Yep, that would be the intent

PR #11783, opened 2026-05-20
billing: Include return_url on checkout

Yopi:


frankie567 (approved):

server/polar/integrations/polar/schemas.py
```diff
@@ -216,6 +216,7 @@ def from_sdk(cls, subscription: "Subscription") -> "OrganizationSubscription":
 class OrganizationCheckoutRequest(Schema):
     product_id: str = Field(description="Polar product ID to subscribe to.")
     success_url: str | None = None
+    return_url: str | None = None
```
frankie567: We should probably add validation on both of those using `get_safe_return_url`, otherwise we're open to open redirection attacks.

frankie567 (changes requested):

server/polar/integrations/polar/schemas.py
Yopi: Absolutely

frankie567 (approved):

PR #11802, opened 2026-05-20
Expose metadata types in product create

sebastianekstrom:
Expose the metadata type in product creation/editing. The backend supports `string | bool | number` but we only expose strings in the UI. This can make it tricky to know exactly what a value is.

* String: Text input
* Number: Number input
* Boolean: True/False dropdown

Also sprinkled some of that sweet sweet monospace font in there to better convey what does what.

### Before
Screenshot 2026-05-20 at 20 57 38

### After
Screenshot 2026-05-20 at 21 10 59

### DIFF
--- a/clients/apps/web/src/components/Products/ProductMetadataForm.tsx
+++ b/clients/apps/web/src/components/Products/ProductMetadataForm.tsx
@@ -1,8 +1,16 @@
-import { FormField, FormLabel } from '@polar-sh/ui/components/ui/form'
+import { FormField } from '@polar-sh/ui/components/ui/form'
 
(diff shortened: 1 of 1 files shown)


emilwidlund (approved):

pieterbeulque (commented):
Can you also do the back-end changes to cast the submitted data to the correct data type again before storing?

On its own, it's pretty unexpected to actively declare something is a boolean only to then get `"true"` back.

frankie567:
@pieterbeulque Agree, but I think the casting should be done on the UI's side when posting to the API. The backend has no data type concept, so this would require a complex schema change.

pieterbeulque:
Oh, yes, definitely, that's even cleaner/easier!

PR #11804, opened 2026-05-21
docs: add team management section

joebon:


docs/features/team-management.mdx
```diff
@@ -0,0 +1,113 @@
+---
+title: "Team Management"
+sidebarTitle: "Team Management"
+description: "Invite teammates to your organization, manage their roles, and control what they can access in the Polar dashboard."
+---
+
```
joebon, line 6: Should we perhaps have a screenshot here? Like in *Customer management*

docs/features/team-management.mdx
```diff
@@ -0,0 +1,113 @@
...
+| Role | Description |
+| --- | --- |
+| **Owner** | There is always exactly one owner per organization. Has the same permissions as an admin. |
+| **Admin** | Full access to the organization — including managing members, finances, and organization settings. |
+| **Member** | Operational access to day-to-day work (products, customers, orders, analytics) but cannot manage other members, finances, or organization settings. |
+
+### Permissions
+
+The exact permissions granted by each role:
```
joebon, line 23: Not convinced this should be here because of _risk of drift_, but it's helpful for sure.

docs/features/team-management.mdx
pieterbeulque: Screenshots tend to go stale really quickly, so I've mostly been removing them as I'm updating the documentation. It forces you to write better too, in a way.

docs/features/team-management.mdx
pieterbeulque: Detail.dev should be able to keep these in check!

docs/features/team-management.mdx
```diff
@@ -0,0 +1,113 @@
...
+| View & manage products | ✓ | ✓ | ✓ |
+| View & manage customers | ✓ | ✓ | ✓ |
+| View & manage custom fields | ✓ | ✓ | ✓ |
+| View & manage sales (orders, subscriptions, refunds) | ✓ | ✓ | ✓ |
+| View & manage analytics | ✓ | ✓ | ✓ |
+| Ingest events | ✓ | ✓ | ✓ |
+| View other members | ✓ | ✓ | ✓ |
+| Invite, remove & change member roles | | ✓ | ✓ |
+| View & manage finances (balance, payouts) | | ✓ | ✓ |
+| Manage organization settings | | ✓ | ✓ |
```
pieterbeulque: I would explicitly call out managing webhooks & API keys too, right now that's buried under organization settings but people will look for these explicitly (as I noticed in support)

docs/features/team-management.mdx
joebon: Fair point - though I do remember a discussion regarding automating screenshots. But until then!

docs/features/team-management.mdx
```diff
@@ -0,0 +1,113 @@
...
+The exact permissions granted by each role:
+
+| Permission | Member | Admin | Owner |
+| --- | :---: | :---: | :---: |
+| View & manage products | ✓ | ✓ | ✓ |
+| View & manage customers | ✓ | ✓ | ✓ |
+| View & manage custom fields | ✓ | ✓ | ✓ |
+| View & manage sales (orders, subscriptions, refunds) | ✓ | ✓ | ✓ |
+| View & manage analytics | ✓ | ✓ | ✓ |
+| Ingest events | ✓ | ✓ | ✓ |
```
pieterbeulque: This isn't something you can do from the dashboard, so this feels a bit out of place.

docs/features/team-management.mdx
joebon: Good catch, thanks!

docs/features/team-management.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+<AccordionGroup>
+  <Accordion title="Do invited users need an existing Polar account?">
+    No. If the invited email doesn't have a Polar account yet, one is created automatically. The user signs in to claim it when they accept the invitation.
+  </Accordion>
+  <Accordion title="Can a user belong to more than one organization?">
+    Yes. A user can be a member of any number of organizations and may have different roles in each one. You can switch between them from the organization picker in the dashboard.
+  </Accordion>
+  <Accordion title="How do I transfer ownership of an organization?">
+    Ownership can only be transferred to another member of the organization who has completed identity verification. Every organization always has exactly one owner — the previous owner becomes an admin once the transfer completes.
```
pieterbeulque, line 107: Is this self-serve now (i.e. you can make anyone the owner who has gone through identity verification?) or do they also need to go through support? If the latter, then we need to iterate on this a bit.

docs/features/team-management.mdx
```diff
@@ -0,0 +1,113 @@
...
+  </Accordion>
+  <Accordion title="What happens to a member's API tokens if they're removed?">
+    Personal access tokens belong to the user, not the organization. Once a user is removed, their tokens lose access to the organization's resources. For long-lived, organization-scoped automation, use an [Organization Access Token](/integrate/authentication) instead.
+  </Accordion>
```
pieterbeulque: PAT are deprecated and can't be created anymore so I think we can drop this.

docs/features/team-management.mdx
joebon: Still through support. I added a callout and link to support in another info box above.

joebon: Added the same here just now

pieterbeulque (approved):

docs/features/team-management.mdx
joebon: Good point - removed!

docs/features/team-management.mdx
joebon: Updated

PR #11806, opened 2026-05-21
feat: Add identity verification section to user settings

joebon:
Make it possible for a user to verify their identity outside of the org onboarding flow.
Verified identity is a prerequisite for ownership transfer and without this that's not possible.


pieterbeulque (approved):
Smart man

PR #11810, opened 2026-05-21
chore: Account admin_id cleanup and ownership transfer flow update

joebon:
## Summary

**Related Issue**: #11404

Phase 6 of the RBAC rollout: retires `Account.admin_id` and routes ownership transfer entirely through `UserOrganization.role`.

## What

- New `UserOrganizationService.transfer_ownership` does the atomic demote-current-owner + promote-new-owner swap, with the `IdentityVerificationStatus.verified` gate fired inline on the role transition.
- `OrganizationRepository.get_admin_user` rewritten as `get_owner_user`, querying `UserOrganization WHERE role = 'owner'`. All 8 callers follow.
- `account_service.change_admin` → `change_owner`; backoffice `make_admin` route → `make_owner`. Stops writing `Account.admin_id`.
- Partial unique index `(organization_id) WHERE role = 'owner' AND deleted_at IS NULL` added (CONCURRENTLY) to enforce single-owner-per-org at the DB level.
- Destructive migration drops `accounts.admin_id`; model relationship and dual-write helper removed.
- Backoffice team section: requirements card refreshed to match what's actually enforced; "Make Owner" dropdown gated on the member being identity-verified.
- Manual-payout-account creation now picks the org owner instead of an arbitrary `users[0]`.

## Why

Phase 4 migrated the frontend to read `role` and Phase 5 retired the legacy `is_admin` API field. `Account.admin_id` is now the only authorization signal still living outside the role model — keeping it around invites drift and gives future code two sources of truth for "who owns this org".

## How

Three semantic commits plus follow-ups: (1) partial unique index, (2) re-home authorization reads off `Account.admin_id` (introduces `transfer_ownership`, refactors `change_admin`), (3) destructive column drop + remove dual-write. Then a rename pass (admin→owner across identifiers, routes, UI text, test classes), review-feedback cleanups, and backoffice UI polish.

**Pre-deploy check**: no org should have more than one live owner row.
```sql
SELECT organization_id, COUNT(*) FROM user_organizations
WHERE role = 'owner' AND deleted_at IS NULL
GROUP BY organization_id HAVING COUNT(*) > 1;
```

## Checklist

### DIFF
--- a/server/polar/account/service.py
+++ b/server/polar/account/service.py
@@ -26,9 +26,9 @@ class AccountServiceError(PolarError):
     pass
 
 
-class CannotChangeAdminError(AccountServiceError):
+class CannotChangeOwnerError(AccountServiceError):
     def __init__(self, reason: str) -> None:
-        super().__init__(f"Cannot change account admin: {reason}")
+        super().__init__(f"Cannot change account owner: {reason}")
 
 
 class AccountService:
@@ -101,7 +101,7 @@ class AccountService:
         session: AsyncSession,
         *,
         organization_id: uuid.UUID,
-        new_admin_user: User,
+        new_owner_user: User,
     ) -> None:
         if not settings.POLAR_SELF_ENABLED:
             return
@@ -115,17 +115,17 @@ class AccountService:
             str(organization_id), polar_organization_id
         )
         if customer is None:
-            raise CannotChangeAdminError(
+            raise CannotChangeOwnerError(
                 f"Polar self customer not found for organization {organization_id}"
             )
 
         member_repository = MemberRepository.from_session(session)
         target_member = await member_repository.get_by_customer_id_and_external_id(
-            customer.id, str(new_admin_user.id)
+            customer.id, str(new_owner_user.id)
         )
         if target_member is None:
-            raise CannotChangeAdminError(
-                f"Polar self member not found for user {new_admin_user.id}"
+            raise CannotChangeOwnerError(
+                f"Polar self member not found for user {new_owner_user.id}"
             )
 
         if target_member.role != MemberRole.owner:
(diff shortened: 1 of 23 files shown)


server/polar/user_organization/service.py
```diff
@@ -194,6 +207,65 @@ async def set_role(
...
+            )
+
+        await session.execute(
+            sql.update(UserOrganization)
+            .where(
+                UserOrganization.organization_id == organization_id,
+                UserOrganization.role == OrganizationRole.owner,
+            )
+            .values(role=OrganizationRole.admin)
+        )
```
psincraian: we have those sql operations inside a repository

psincraian (approved):
This looks good! 

One small comment but approved :)

server/polar/user_organization/service.py
joebon: Right! Fixing...

joebon: Fixed in 59bee1a8f — moved the demote+promote updates into `UserOrganizationRepository.demote_current_owner` / `promote_to_owner`. The `IntegrityError → AlreadyOwner` translation stays in the service since it's a service-level concern.

server/polar/account/service.py
```diff
@@ -153,87 +136,23 @@ async def _sync_polar_self_customer_owner(
                 allow_ownership_transfer=True,
             )
 
-    async def change_admin(
+    async def change_owner(
```
frankie567, line 139: Semantically, this probably should move to organization service, but it can be done in a follow-up PR since this one is already quite big :)

frankie567 (approved):
Awesome, it really feels good to get rid of that one 👏

PR #11816, opened 2026-05-21
billing: Remove internal payment method endpoints

Yopi:
Since we are now embedding our embeddable payment method editor, we don't need to have these as a side option


frankie567 (approved):
The shortest code lifetime ever in this code base 😂

PR #11831, opened 2026-05-22
fix: preserve scheduled product changes when seats change (#11129)

### LINKED ISSUE #11129: Seat changes override scheduled subscription changes

psincraian:
## Summary

Closes #11129. Allow to have product and seat updates on the SubscriptionUpdate.

## How

- `SubscriptionUpdateRepository.upsert` merge fields conditionally instead of overwriting
- `SubscriptionService.update_seats` during immediate-proration, if a pending product change exists, keep it. Clear only the stale `seats` column on the pending row.
- `SubscriptionService.update_seats` when new seats and old seats match, drop only the seats instead of soft-deleting the whole row.

## Checklist

- [x] Lint and type checks pass locally (`uv run task lint`, `uv run task lint_types`).
- [x] Tests added for the new behaviour.
- [x] PR is focused and small.

### DIFF
--- a/server/polar/subscription/service.py
+++ b/server/polar/subscription/service.py
@@ -1476,13 +1476,16 @@ class SubscriptionService:
         )
 
         if old_seats == seats:
-            # No-op for the seat count itself, but a stale next-period
-            # SubscriptionUpdate would otherwise survive — clear it so the
-            # subscription doesn't change at the next billing cycle.
+            # Re-asserting the current seat count cancels a pending seat
+            # change. Drop the row if nothing else is scheduled on it.
             pending = subscription.pending_update
             if pending is not None and pending.seats is not None:
-                await subscription_update_repository.soft_delete(pending)
-                subscription.pending_update = None
+                if pending.product_id is None:
+                    await subscription_update_repository.soft_delete(pending)
+                    subscription.pending_update = None
+                else:
+                    pending.seats = None
+                    await subscription_update_repository.update(pending)
             return subscription
 
         event = await event_service.create_event(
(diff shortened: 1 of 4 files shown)


server/polar/subscription/service.py
```diff
@@ -1476,13 +1476,16 @@ async def update_seats(
...
-            # No-op for the seat count itself, but a stale next-period
-            # SubscriptionUpdate would otherwise survive — clear it so the
-            # subscription doesn't change at the next billing cycle.
+            # Re-asserting the current seat count cancels a pending seat
+            # change. Drop the row if nothing else is scheduled on it.
             pending = subscription.pending_update
             if pending is not None and pending.seats is not None:
-                await subscription_update_repository.soft_delete(pending)
-                subscription.pending_update = None
+                pending.seats = None
```
joebon: Nit pick: I think this would read even easier if this assignment is only made inside the else block below

joebon (approved):

server/polar/subscription/service.py
psincraian: firxed :D

joebon: 💅 💄

frankie567 (approved):

PR #11834, opened 2026-05-22
fix: Ensure tasks with optional fields can be enqueued from backoffice

### LINKED ISSUE #11709: [Detail Bug] Backoffice task enqueue form marks optional TypedDict (total=False) fields as required
# Detail Bug Report

https://app.detail.dev/org_ecfbc7cf-6d21-4ce1-8c61-cda1d650ccf7/bugs/bug_3babcb8c-29d2-41b4-b995-f8ad340e6663

Introduced in [#11672](https://github.com/polarsource/polar/pull/11672) by @birkjernstrom on May 15, 2026

# Summary
- **Context**: The backoffice task enqueue form dynamically builds Pydantic models from Dramatiq task function signatures to render forms for admins to manually trigger tasks.
- **Bug**: When a task function uses `Unpack[TypedDict]` with `total=False`, all TypedDict fields are incorrectly treated as required form fields instead of optional.
- **Actual vs. expected**: TypedDict fields marked as optional (via `total=False` or `NotRequired[]`) should produce optional form fields, but they produce required fields instead.
- **Impact**: Admins cannot enqueue tasks like `loops.update_contact` or `loops.send_event` without filling in 12+ fields that should be optional, making the form effectively unusable.

# Code with Bug
`server/polar/backoffice/tasks/forms.py`

```python
if is_typeddict(type_hints_args[0]):
    for k, v in get_type_hints(type_hints_args[0]).items():
        yield k, v, inspect.Parameter.empty  # <-- BUG 🔴 ignores TypedDict optionality; makes all fields required
    return
```

# Explanation
The form builder uses `_get_function_arguments` output to decide whether a generated Pydantic field has a default. For TypedDict-backed `Unpack[...]` params, this code yields `inspect.Parameter.empty` as the default for every key. Downstream, this is interpreted as “no default” (Pydantic `...`/undefined), so every TypedDict field becomes required.

This contradicts TypedDict semantics: for `total=False`, `Properties.__required_keys__` is empty and all fields appear in `__optional_keys__`, so these fields should be optional in the enqueue form. Runtime evidence shows validation fails unless many optional properties (e.g., `firstName`, `lastName`, `createdAt`, etc.) are provided.

# Recommended Fix
When unpacking a TypedDict, consult `__optional_keys__` (and by implication `__required_keys__`) and yield `None` (or another non-empty default) for optional keys so they become optional fields:

```python
if is_typeddict(type_hints_args[0]):
    typeddict_class = type_hints_args[0]
    optional_keys = typeddict_class.__optional_keys__
    for k, v in get_type_hints(typeddict_class).items():
        if k in optional_keys:
            yield k, v, None  # Optional field
        else:
            yield k, v, inspect.Parameter.empty  # Required field
    return
```

# History
This bug was introduced in commit c3cd825a0. The commit "fix(backoffice): Respect requirement of enqueued task params" correctly added logic to respect parameter defaults for regular function arguments by yielding `parameter.default`, but for TypedDict fields it unconditionally yielded `inspect.Parameter.empty` without checking the `__optional_keys__` attribute. The original implementation (commit 46cebde48d) treated all fields as required, and this fix commit should have distinguished optional TypedDict fields but missed the `__optional_keys__` check.

joebon:
## Summary

**Related Issue**: #11709

Ensure tasks with optional fields can be enqueued from backoffice

### DIFF
--- a/server/tests/backoffice/tasks/test_forms.py
+++ b/server/tests/backoffice/tasks/test_forms.py
@@ -1,0 +1,51 @@
+import inspect
+from typing import NotRequired, Required, TypedDict, Unpack
+
+from polar.backoffice.tasks.forms import _get_function_arguments
+
+
+class _AllOptional(TypedDict, total=False):
+    first_name: str
+    last_name: str
+
+
+class _AllRequired(TypedDict):
+    user_id: str
+    email: str
+
+
+class _Mixed(TypedDict):
+    user_id: Required[str]
+    first_name: NotRequired[str]
+    last_name: NotRequired[str]
+
+
+def test_typeddict_total_false_yields_none_default() -> None:
+    def task(**kwargs: Unpack[_AllOptional]) -> None: ...
+
+    args = dict(((k, default) for k, _, default in _get_function_arguments(task)))
+
+    assert args == {"first_name": None, "last_name": None}
+
+
+def test_typeddict_total_true_yields_empty_default() -> None:
+    def task(**kwargs: Unpack[_AllRequired]) -> None: ...
+
+    args = {k: default for k, _, default in _get_function_arguments(task)}
+
+    assert args == {
+        "user_id": inspect.Parameter.empty,
+        "email": inspect.Parameter.empty,
+    }
+
+
+def test_typeddict_mixed_required_and_not_required() -> None:
+    def task(**kwargs: Unpack[_Mixed]) -> None: ...
+
+    args = {k: default for k, _, default in _get_function_arguments(task)}
+
+    assert args == {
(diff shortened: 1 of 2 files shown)


server/tests/backoffice/tasks/test_forms.py
joebon, line 1: Perhaps overkill but felt so meta to just ship the fix - not super easy to verify that it does what it says on the tin from just looking at the code

frankie567: No, that's a good thing! This thing is quite hairy, so that's great :)

frankie567 (approved):

PR #11836, opened 2026-05-22
Fix: show team members' actual role(s) in backoffice

joebon:
Fixes so we don't show either `Owner` or `Member` (if not owner) but show the actual role instead


pieterbeulque (approved):
🧼

PR #11847, opened 2026-05-22
fix(backoffice): show each team member's own verified country

pieterbeulque:
One could also wonder if this is useful information to show there?


frankie567 (approved):
Yeah, don't really see the point, but otherwise, good to go.

pieterbeulque:
@psincraian - I saw that you added it previously, do you remember why that's in there and if we can maybe remove it instead?

psincraian:
I remember Rishi asked for that

PR #11853, opened 2026-05-25
fix(benefit): stop resetting meters when refreshing benefit grants

### LINKED ISSUE #11427: Editing benefits on a product wipes customer meter's balance
When a merchant edits a product's benefit list, the `update_product_benefits_grants` flow dispatches a per-subscription benefit-revoke pipeline. The pipeline's completion callback (`benefit.reset_meters_and_enqueue_grants`) unconditionally calls `subscription_service.reset_meters`, which emits a `meter_reset` event for every meter on the subscription. Because `meter_reset` is scoped to `(customer, meter)` globally, the customer's balance for that meter is wiped — including credits granted by entirely separate active subscriptions and credits the customer has already paid for.

## Reproduction

 - Org with meter M.
 - Product ABC: benefits [X, Y (say, a custom benefit), Z (meter_credit, 50 units of M)] + metered price on M.
 - Product DEF: benefits [X, Z (meter_credit, 3000 units of M)] + metered price on M.

1. Customer subscribes to ABC, then upgrades to DEF.
2. As the merchant, edit product ABC and remove benefit Y.
3. Observe: customer balance for M is now 0.

## Expected behavior

Removing benefit Y from a product should revoke Y for affected subscriptions only and leave unrelated meters untouched. Meter resets should fire on subscription cycle/renewal, not on product-benefit edits.

joebon:
## Summary

**Related Issue**: #11427

Editing a product's benefits no longer wipes customer meter balances. The benefit grant pipeline no longer emits a global `meter_reset` event; meter resets are now scoped to the order layer (cycle/cancel orders) and per-grant cycle events, where they belong.

## What

- `server/polar/benefit/tasks.py` — removed the `subscription_service.reset_meters` call from the `benefit.reset_meters_and_enqueue_grants` task body. Dropped two now-unused imports. Actor name retained for in-flight Dramatiq message compatibility (documented inline).
- `server/polar/benefit/grant/service.py` — refreshed the pipeline comment (no behavior change).
- `server/tests/benefit/test_tasks.py` — collapsed three reset-related tests into one regression test (`test_does_not_reset_meters`) that asserts `subscription_service.reset_meters` is never called from the pipeline.
- `server/tests/benefit/strategies/test_meter_credit.py` — added `test_upgrade_balance_without_meter_reset` asserting the revoke→grant balance trajectory (100 → 0 → 1000) on an upgrade, locking in that the math is symmetric without a reset.

## Why

The previous implementation called `subscription_service.reset_meters` whenever the pipeline ran with a `subscription_id` scope. Because `meter_reset` is scoped to `(customer, meter)` globally — not `(customer, meter, subscription)` — *any* meter reset wiped the customer's balance across every subscription, including credits the customer had paid for via separate subscriptions. The pipeline fires from many triggers (product benefit edits, dunning, refunds, customer seat updates, subscription product changes, …), so the impact was wide-ranging.

The repro from the issue:
- Customer has subscriptions on products ABC (50 credits on meter M via benefit Z) and DEF (3000 credits on meter M via benefit Z'). Balance: 3050.
- Merchant edits ABC and removes benefit Y. The pipeline ran for ABC's subscription and called `reset_meters`, emitting a `meter_reset` for (customer, M). Balance: 0.

This also aligns the benefit pipeline with the explicit "`subscription_update` is intentionally excluded" design already documented in `server/polar/order/service.py:820-823`.

## How

Removed the `reset_meters` block from `benefit_enqueue_grants` entirely. The math still settles correctly because `meter_credit.revoke()` emits a matching negative-units event for the previously credited amount, so revoke→grant on a product change nets to exactly the new product's credit total. The new test exercises this end-to-end.

Meter resets continue to fire on:
- `subscription_cycle`, `subscription_cycle_after_trial`, `subscription_cancel` orders (`order.service.create_subscription_order`)
- per-grant renewal via `meter_credit.cycle()` (with dedup against the order-layer reset)

The Dramatiq actor name `benefit.reset_meters_and_enqueue_grants` is now stale but kept to avoid orphaning in-flight messages during deploy. A rename can land in a follow-up PR after this is fully rolled out.

## Checklist

### DIFF
--- a/server/polar/benefit/tasks.py
+++ b/server/polar/benefit/tasks.py
@@ -11,8 +11,6 @@ from polar.exceptions import PolarTaskError
 from polar.logging import Logger
 from polar.models.benefit_grant import BenefitGrantScopeArgs
 from polar.product.repository import ProductRepository
-from polar.subscription.repository import SubscriptionRepository
-from polar.subscription.service import subscription as subscription_service
 from polar.worker import (
     AsyncSessionMaker,
     RedisMiddleware,
@@ -100,6 +98,11 @@ async def enqueue_benefits_grants(
         )
 
 
+# Actor name kept for backwards compatibility with in-flight Dramatiq messages;
+# this task no longer resets meters. Meter resets happen on
+# subscription_cycle / subscription_cycle_after_trial / subscription_cancel
+# orders (see order.service.create_subscription_order) and via
+# meter_credit.cycle() per benefit grant.
 @actor(
     actor_name="benefit.reset_meters_and_enqueue_grants", priority=TaskPriority.MEDIUM
 )
@@ -109,15 +112,6 @@ async def benefit_enqueue_grants(
     member_id: uuid.UUID | None = None,
     **scope: Unpack[BenefitGrantScopeArgs],
 ) -> None:
-    if subscription_id := scope.get("subscription_id"):
-        async with AsyncSessionMaker() as session:
-            repository = SubscriptionRepository.from_session(session)
-            subscription = await repository.get_by_id(
-                subscription_id, options=repository.get_eager_options()
-            )
-            if subscription:
-                await subscription_service.reset_meters(session, subscription)
-
     for benefit_id in grant_benefit_ids:
         enqueue_job(
             "benefit.grant",
--- a/server/polar/benefit/tasks.py
+++ b/server/polar/benefit/tasks.py
@@ -98,14 +98,7 @@ async def enqueue_benefits_grants(
         )
 
 
-# Actor name kept for backwards compatibility with in-flight Dramatiq messages;
-# this task no longer resets meters. Meter resets happen on
-# subscription_cycle / subscription_cycle_after_trial / subscription_cancel
-# orders (see order.service.create_subscription_order) and via
-# meter_credit.cycle() per benefit grant.
-@actor(
-    actor_name="benefit.reset_meters_and_enqueue_grants", priority=TaskPriority.MEDIUM
-)
+@actor(actor_name="benefit.enqueue_grants", priority=TaskPriority.MEDIUM)
 async def benefit_enqueue_grants(
     customer_id: uuid.UUID,
     grant_benefit_ids: list[uuid.UUID],
@@ -120,6 +113,25 @@ async def benefit_enqueue_grants(
             member_id=member_id,
             **scope,
         )
+
+
+# Deprecated alias. Kept registered to drain in-flight Dramatiq messages
+# enqueued before the rename. Remove once the queue is drained after deploy.
+# This task previously also reset meters; resets are now the responsibility
+# of order.service.create_subscription_order (cycle/cancel orders) and
+# meter_credit.cycle() per benefit grant.
+@actor(
+    actor_name="benefit.reset_meters_and_enqueue_grants", priority=TaskPriority.MEDIUM
+)
+async def benefit_reset_meters_and_enqueue_grants(
+    customer_id: uuid.UUID,
+    grant_benefit_ids: list[uuid.UUID],
+    member_id: uuid.UUID | None = None,
+    **scope: Unpack[BenefitGrantScopeArgs],
+) -> None:
+    await benefit_enqueue_grants(
+        customer_id, grant_benefit_ids, member_id=member_id, **scope
+    )
 
 
 @actor(actor_name="benefit.grant", priority=TaskPriority.MEDIUM)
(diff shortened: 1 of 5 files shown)


pieterbeulque:
Given the explicitness of the initial implementation (e.g. the actor name and config), I'm starting to wonder if this is is a bugfix or a behavior change. Apart from that, I think the behavior as described in the issue is unexpected and feels "broken".

Or put differently I wonder if fixing this can break something else for someone else?

frankie567:
How does it behave with those changes when a new subscription is created? I think the purpose of this was to start with fresh meters whenever a customer starts a paid subscription, so they don't have to pay for usage prior their subscription.

joebon:
Both fair concerns — pulling on the threads:

**@frankie567** on new subscriptions: the checkout path is unaffected. `create_or_update_from_checkout` does an explicit `reset_meters` at `subscription/service.py:659` before enqueuing benefit grants, so fresh meters on every new paid subscription still hold. The programmatic `subscription_service.create` endpoint is free-only by design (`endpoints.py:270`), so the "don't pay for prior usage" intent doesn't apply there. The path that *did* change is `subscription_service.update` (API-driven product change on an existing subscription) — the pipeline reset that commit 311ade8ec added is gone in this PR.

**@pieterbeulque** on bugfix-vs-behavior-change: you're right, in its current form this is both. The pure bugfix is for `update_product_benefits_grants` (product-benefit edits). The behavior change is the removal of meter reset on `subscription_service.update`. The two got bundled because the same actor was doing both.

Suggested narrower path: gate the reset behind a `reset_meters: bool = False` parameter threaded through `subscription_service.enqueue_benefits_grants` → `benefit.enqueue_benefits_grants` task → `benefit_grant_service.enqueue_benefits_grants` → `benefit.reset_meters_and_enqueue_grants` task. Default `False` (fixes #11427 — product-benefit edits don't reset). Pass `True` only at the `subscription_service.update` callsite, preserving 311ade8ec's behavior on product change.

That leaves one residual issue: even on legitimate subscription update, the `meter_reset` event is scoped to `(customer, meter)` globally, so a customer with two active subscriptions on the same meter still loses the unrelated one's credits when they upgrade either side. Fixing that needs scoped reset events (`(customer, meter, subscription)`) and is a bigger architectural change — worth a separate issue rather than bundling here.

Want me to switch this PR to the gated approach?

joebon:
Claude got a little too eager and decided to respond here. Sorry. Consider the above as pending my _actual_ analysis and recommendation.

joebon:
This seem quite intricate. From what I can gather this is the history of this:
-  On Jan 19th we added [6e1d2513e - fix: prevent meter reset on subscription_update orders](https://github.com/polarsource/polar/commit/6e1d2513e)
- On Feb 6th we added [311ada8ec - benefits: Revoke benefits immediately on product change](https://github.com/polarsource/polar/commit/311ade8ec)

They appear to be contradicting. The first one could be described with the following scenario:
> - Customer on plan A: 100 credits granted, 80 used, balance 20.
> - Upgrade to plan B (1000 credits) via API.
> - **No reset** (`6e1d2513e`'s intent):
>     - `credited_units` = running sum [+100, -100, +1000] clamped at each step → 1000
>     - `consumed_units` = 80 (persists)
>     - **balance = 920**

This is reasonably correct IMHO.

After `311ada8ec` the balance was instead reset to *1000* because it inadvertently missed/erased the consumed credits while solving the ordering issue.

I've added a complementary test to ensure we catch this properly.

WDYT @pieterbeulque @frankie567 ?

frankie567:
Yes, that all makes sense! Thank you @joebon for the archeological research :)

server/polar/benefit/tasks.py
```diff
@@ -100,6 +98,11 @@ async def enqueue_benefits_grants(
         )
 
 
+# Actor name kept for backwards compatibility with in-flight Dramatiq messages;
+# this task no longer resets meters. Meter resets happen on
+# subscription_cycle / subscription_cycle_after_trial / subscription_cancel
+# orders (see order.service.create_subscription_order) and via
+# meter_credit.cycle() per benefit grant.
```
frankie567: Can we already clone the actor with the "right" name and start enqueueing it? I would like to be sure we don't forget to change that name in the future :)

frankie567 (changes requested):

pieterbeulque (approved):

server/polar/benefit/tasks.py
joebon: Ah, yes - good point! Fixing

joebon: Done

frankie567 (approved):
Awesome!

PR #11878, opened 2026-05-25
Use /to/ links in our docs

sebastianekstrom:
With our new `/to/` links we can make our docs better by instead of asking the user to navigate to a specific page, simply create a link to that page.

| Before | After  |
|--------|--------|
| Monosnap Team Management - Polar 2026-05-25 16-17-04 | Monosnap Team Management - Polar 2026-05-25 16-17-28 |

### DIFF
--- a/docs/features/benefits/feature-flags.mdx
+++ b/docs/features/benefits/feature-flags.mdx
@@ -15,7 +15,7 @@ The Feature Flag benefit is a lightweight way to grant feature access to custome
 
 ## Create Feature Flag Benefit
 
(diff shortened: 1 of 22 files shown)


pieterbeulque (approved):
You're a genius.

My only concern is that people may go from sandbox → docs → production and not notice, but I don't think that should hold this back, it's definitely an improvement.

sebastianekstrom:
@pieterbeulque good point! Wonder if we could solve that on the `/to/` route itself, detecting if they have been in sandbox mode and then to the redirect in that case

PR #11884, opened 2026-05-25
Handle sandbox on /to/ redirects

sebastianekstrom:
We have an issue with our `/to/` redirect links (e.g https://polar.sh/to/dashboard/products) which is that if the user is in our sandbox environment, they will be redirected to production which can be confusing. This PR solves that by setting a `polar_env` cookie which our proxy then handles.

For a sandbox user clicking https://polar.sh/to/dashboard/products from the docs:

1. Browser requests https://polar.sh/to/dashboard/products
2. Our proxy reads `polar_env=sandbox` cookie and sees the current host is `production` meaning they mismatch
3. We redirect to https://sandbox.polar.sh/to/dashboard/products
4. Our proxy passes the env check (`sandbox === sandbox`)
5. `/to/[...path]/page.tsx` reads the org cookie and redirects to https://sandbox.polar.sh/dashboard/acme/products

### DIFF
--- a/clients/apps/web/src/utils/cookies.ts
+++ b/clients/apps/web/src/utils/cookies.ts
@@ -2,6 +2,9 @@ import { schemas } from '@polar-sh/client'
 import { ReadonlyRequestCookies } from 'next/dist/server/web/spec-extension/adapters/request-cookies'
 
 const lastVisitedOrg = 'last_visited_org'
+export const POLAR_ENV_COOKIE = 'polar_env'
+
+export type PolarEnv = 'production' | 'sandbox'
 
 export const setLastVisitedOrg = (
   organization: string,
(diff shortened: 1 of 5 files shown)


clients/apps/web/src/utils/cookies.ts
```diff
@@ -20,3 +23,13 @@ export const getLastVisitedOrg = (
   }
   return organizations.find((org) => org.slug === lastVisitedOrgSlug)
 }
+
+export const setLastVisitedEnv = (
+  env: PolarEnv,
+  maxAge: number = 30 * 86400, // Expires in 30 days
```
pieterbeulque: I was thinking this should be a short TTL, say 5 minutes, and be cleared by the next `/to` redirect.

I don't like the idea of it lingering, say you went from sandbox → docs, then don't click a link. Google something four days later, find the docs, click the link. You're still taken to sandbox because of this action four days earlier.

sebastianekstrom: I think 5 min might be a bit too aggressive, changing it to 30min if you don't have strong opinions!

pieterbeulque: Works for me!

pieterbeulque (approved):
Would it make sense to try and set and automatically clear the `POLAR_ENV_COOKIE` based on user intent (i.e. only set when they navigate to the docs, and clear when `/to/` redirects)?

PR #11886, opened 2026-05-26
amount v2: Migrate billing_entry amounts to BigInt

Yopi:
The first step in our journey to support large amounts

### DIFF
--- a/server/polar/models/billing_entry.py
+++ b/server/polar/models/billing_entry.py
@@ -3,7 +3,10 @@ from enum import StrEnum
 from typing import TYPE_CHECKING, Self
 from uuid import UUID
 
-from sqlalchemy import TIMESTAMP, ForeignKey, Index, String, Uuid
+from alembic_utils.pg_function import PGFunction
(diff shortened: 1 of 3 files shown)


server/polar/models/billing_entry.py
```diff
@@ -139,3 +148,48 @@ def from_metered_event(
...
+                NEW.discount_amount_v2 := NEW.discount_amount;
+            END IF;
+        END IF;
+        RETURN NEW;
+    END
+    $$ LANGUAGE plpgsql;
+    """,
+)
+
+billing_entry_sync_v2_amounts_trigger = PGTrigger(
```
frankie567, line 180: Temporary trigger during the migration I guess?

frankie567 (approved):

server/polar/models/billing_entry.py
Yopi: Yeah, removed in https://github.com/polarsource/polar/pull/11888

PR #11888, opened 2026-05-26
amount v2: Drop old billing entry amount columns

Yopi:


server/polar/models/billing_entry.py
```diff
@@ -154,48 +140,3 @@ def from_metered_event(
             subscription=subscription_product_price.subscription,
             event=event,
         )
-
-
-billing_entry_sync_v2_amounts_function = PGFunction(
```
frankie567, line 159: Good, I have the answer to my question 😆

frankie567 (approved):

PR #11889, opened 2026-05-26
feat(organization): limit slug length to 64 characters

pieterbeulque:
Enforce a 64-character cap on organization slugs in the Pydantic SlugInput, mirror the limit in the slug availability check and in the backoffice forms, and update the onboarding/sandbox forms so longer values can't be entered or auto-generated client-side.

The rationale behind 64 characters is that according to [RFC 2821](https://www.ietf.org/rfc/rfc2821.txt) the local-part of an email address can be maximum 64 characters, and we currently use `organization-slug@notifications.polar.sh` as the sender of merchant notifications.

I already cleaned up our database to fix the organization names that are currently longer than 64 characters.

### DIFF
--- a/server/polar/organization/schemas.py
+++ b/server/polar/organization/schemas.py
@@ -75,9 +75,12 @@ def validate_reserved_keywords(value: str) -> str:
     return value
 
 
+SLUG_MAX_LENGTH = 64
+
+
 SlugInput = Annotated[
     str,
-    StringConstraints(to_lower=True, min_length=3),
+    StringConstraints(to_lower=True, min_length=3, max_length=SLUG_MAX_LENGTH),
     SlugValidator,
     AfterValidator(validate_reserved_keywords),
     AfterValidator(validate_blocked_words),
@@ -85,7 +88,10 @@ SlugInput = Annotated[
 
 
 class OrganizationSlugCheck(Schema):
-    slug: str = Field(description="The slug to check availability for.")
+    slug: str = Field(
+        description="The slug to check availability for.",
+        max_length=SLUG_MAX_LENGTH,
+    )
 
 
 class OrganizationSlugAvailability(Schema):
--- a/server/polar/organization/schemas.py
+++ b/server/polar/organization/schemas.py
@@ -88,7 +88,7 @@ SlugInput = Annotated[
 
 
 class OrganizationSlugCheck(Schema):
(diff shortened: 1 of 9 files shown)


server/polar/organization/schemas.py
```diff
@@ -75,17 +75,23 @@ def validate_reserved_keywords(value: str) -> str:
...
+    StringConstraints(to_lower=True, min_length=3, max_length=SLUG_MAX_LENGTH),
     SlugValidator,
     AfterValidator(validate_reserved_keywords),
     AfterValidator(validate_blocked_words),
 ]
 
 
 class OrganizationSlugCheck(Schema):
-    slug: str = Field(description="The slug to check availability for.")
+    slug: str = Field(
```
frankie567: Why this one doesn't use `SlugInput`? It looks like it can create validation inconsistencies between the two flows.

server/polar/organization/service.py
```diff
@@ -277,7 +278,11 @@ async def check_slug_availability(
         """
         normalized = slug.lower()
 
-        if len(normalized) < 3 or slugify(normalized) != normalized:
+        if (
+            len(normalized) < 3
+            or len(normalized) > SLUG_MAX_LENGTH
+            or slugify(normalized) != normalized
```
frankie567: Ah, we do it manually! Let's move to `SlugInput` in the schema and remove this manual condition.

frankie567 (changes requested):

server/polar/organization/service.py
```diff
@@ -272,18 +272,13 @@ async def check_slug_availability(
...
-        if len(normalized) < 3 or slugify(normalized) != normalized:
-            return OrganizationSlugAvailability(available=False)
-
         try:
-            validate_reserved_keywords(normalized)
-            validate_blocked_words(normalized)
-        except ValueError:
+            normalized = _slug_input_adapter.validate_python(slug)
+        except PydanticValidationError:
             return OrganizationSlugAvailability(available=False)
```
frankie567, line 282: We prefer to not distinguish invalid from unavailable slugs then?

frankie567 (approved):

server/polar/organization/service.py
pieterbeulque: Yes, first version had something that returned the reason, but then decided not to do that, security by obscurity 🤡

PR #11890, opened 2026-05-26
alembic utils: Explicitly opt-in to the entity types we track

Yopi:
Otherwise we start detecting drift for other things, such as PGGrant for the polar_read role, which we are not tracking in alembic migrations.

This will make the autogenerate work (again/better) by not filling in all of those grants in the upgrade/downgrade.

If we add new entity types that we are tracking, we simply add them to the list of entities we track here.

### DIFF
--- a/server/polar/kit/db/models/base.py
+++ b/server/polar/kit/db/models/base.py
@@ -2,6 +2,8 @@ from datetime import datetime
 from uuid import UUID
 
 from alembic_utils.pg_extension import PGExtension
+from alembic_utils.pg_function import PGFunction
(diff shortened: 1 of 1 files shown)


frankie567 (approved):
This isn't happening on my machine; but I guess it's because I don't have the `polar_read` role configured.

Yopi:
> This isn't happening on my machine; but I guess it's because I don't have the `polar_read` role configured.

Yeah I think so 😄. Maybe the solution is for me to not have the polar_read role, but I think this is change is fine

PR #11895, opened 2026-05-26
Implement OAuth2 factors

frankie567:


server/polar/auth/exception_handlers.py
```diff
@@ -0,0 +1,19 @@
...
+
+from polar.config import settings
+
+from .exceptions import PolarAuthRedirectionError
+
+
+async def auth_redirection_error_exception_handler(
+    request: Request, exc: PolarAuthRedirectionError
+) -> Response:
+    error_url_params = urlencode({"error": exc.message})
```
Yopi: I wonder if it would be nicer if we would do
```
async def auth_redirection_error_exception_handler(
    request: Request, exc: HTTPException
) -> Response:
    assert isinstance(exc, PolarAuthRedirectionError)
    error_url_params = urlencode({"error": exc.message})
```
to avoid the type ignore in `app.py`?

server/polar/auth/oauth2/factor.py
```diff
@@ -0,0 +1,97 @@
...
+    async def update(
+        self: OAuth2FactorProtocol, enrollment: OAuth2EnrollmentDataclass
+    ) -> None:
+        profile = await self.get_profile(enrollment.access_token)
+        email = await typing.cast(OAuth2FactorMixin, self).get_email(enrollment)
+        statement = (
+            update(OAuthAccount)
+            .where(
+                OAuthAccount.user_id == enrollment.identity_id,
+                OAuthAccount.platform == OAuthPlatform.google,
```
Yopi: Is it only google that does an update, or why don't we do `OAuthAccount.platform == provider,` here?

And if that is the case, should we add an assertion  that the enrollment is for google above this?

server/polar/auth/oauth2/state.py
```diff
@@ -0,0 +1,61 @@
...
+from sqlalchemy import delete, select
+
+from polar.config import settings
+from polar.models import OAuth2State
+from polar.postgres import AsyncSession, get_db_session
+
+TOKEN_PREFIX = "polar_oauth2_"
+
+
+class OAuth2StateService(OAuth2StateServiceBase):
```
Yopi, line 19: Doesn't this feel more like a repository? Is it a service since we're extending `OAuth2StateServiceBase`?

Yopi (approved):
I'm not 100% familiar with how this flow works, but overall looks good to me! A few small questions

server/polar/auth/exception_handlers.py
frankie567: Nice idea, but doesn't work neither :( 

```
polar/app.py:229: error: Argument 2 to "add_exception_handler" of "Starlette" has incompatible type "Callable[[Request[State], HTTPException], Coroutine[Any, Any, Response]]"; expected "Callable[[Request[State], Exception], Response | Awaitable[Response]] | Callable[[WebSocket, Exception], Awaitable[None]]"  [arg-type]
```

I don't really understand what's wrong with mypy here. The type of exception handler is the following:

```py
HTTPExceptionHandler = Callable[["Request", Exception], "Response | Awaitable[Response]"]
WebSocketExceptionHandler = Callable[["WebSocket", Exception], Awaitable[None]]
ExceptionHandler = HTTPExceptionHandler | WebSocketExceptionHandler
```

I'm wondering if `Exception` subclasses are _slightly_ different than other type of subclasses, which might make the type checker refusing a subclass.

server/polar/auth/oauth2/factor.py
frankie567: That's a bug, sir :D Good catch!

server/polar/auth/oauth2/state.py
frankie567: Yes, it's somewhat a repository, but since the upstream library requires a specific protocol, I just subclassed it and implemented methods directly.

We could implement a proper repository for OAuth2State and then wrap it to conform with this protocol, but felt a bit too much since we'll only use it here.

server/polar/auth/exception_handlers.py
frankie567: Ah no, this is a contra-variance issue 🙃 

https://chat.mistral.ai/chat/8759880c-dcf2-496c-b2c5-352cfbf38f09

server/polar/auth/oauth2/state.py
Yopi: Definitely!

server/polar/auth/exception_handlers.py
Yopi: Oh, but then if you were to do `exc: Exception` wouldn't it work?

frankie567: It would, yes :)

frankie567: Fixed this one and all its friends in 925e334d3cbb4ee6fb589a873fa1617ebe4f009a

PR #11898, opened 2026-05-26
Allow changing plans during trial

pieterbeulque:
## Summary

**Related Issue**: polarsource/feedback#132

I approached it incrementally, so if you go through the commits one by one, it might be a bit easier to follow. First, allow changing to plans with equal or longer trials, then, compare against trial elapsed instead of total trial length, then allow any product changes.

Includes client-side notices in the situations where the customer will be charged immediately.

Allowed from both merchant dashboard & customer portal.

### DIFF
--- a/server/polar/subscription/service.py
+++ b/server/polar/subscription/service.py
@@ -1129,10 +1129,10 @@ class SubscriptionService:
             new_trial_end = product.trial_interval.get_end(
                 subscription.trial_start, product.trial_interval_count
             )
-            if new_trial_end < subscription.trial_end:
+            if new_trial_end <= utc_now():
                 raise TrialingSubscription(
                     subscription,
-                    "Can't change to a product with a shorter trial period during a trial.",
+                    "Can't change to a product whose trial period would already be over given the current trial usage.",
                 )
 
         # Add event for the subscription plan change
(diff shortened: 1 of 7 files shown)


server/polar/subscription/service.py
```diff
@@ -1169,19 +1174,34 @@ async def update_product(
...
+                    subscription.current_period_end = new_trial_end
             session.add(subscription)
             await session.flush()
 
-            # Invoice and attempt to pay immediately
-            if (
-                proration_behavior.is_immediate() or interval_changed
-            ) and billing_entries:
+            if was_trialing and ends_trial:
+                subscription = await self.cycle(session, subscription)
```
pieterbeulque, line 1197: This is the thing that I'm least sure of.

frankie567: Technically, I _think_ it would work. 

The other approach could simply be to call `update_trial`, which will set the periods correctly and create related events. Then, we let it naturally cycle because their new current period is past. That's the approach we take when we end the trial immediately. It might be safer IMO.

pieterbeulque: I looked into it and I think it's not worth it.

`update_trial` synchronously cycles just like here, it doesn't defer to the scheduler to cycle.

Because `update_trial` triggers its own `subscription.updated` webhook and other side-effects, we'd then have to make sure we don't trigger duplicate side-effects from the trial-ending branch in `update_product`.

frankie567: Ah yes, you're right. Let's go then.

frankie567 (approved):

frankie567 (approved):

PR #11899, opened 2026-05-26
amount v2: Migrate checkouts amounts to BigInt

Yopi:
Since the checkout amounts are not nullable, we will remove the nullability on the new columns and add it to the old columns in the following PR. The new columns are added as nullable so that we can run the backfill.

### DIFF
--- a/server/migrations/versions/2026-05-27-0929_add_checkouts_v2_amount_columns.py
+++ b/server/migrations/versions/2026-05-27-0929_add_checkouts_v2_amount_columns.py
@@ -1,0 +1,138 @@
+"""Add checkouts v2 amount columns (int → bigint widening)
+
+Revision ID: 1635867d733d
+Revises: c89b877bd235
+Create Date: 2026-05-26 13:06:00.000000
+
+Phase 1 of widening checkouts.amount, checkouts.net_amount and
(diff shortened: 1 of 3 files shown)


frankie567 (approved):
Great!

Heads-up: backfill on checkouts are quite long :)

Yopi:
> Great!
> 
> Heads-up: backfill on checkouts are quite long :)

Haha, yeah 🥲

server/migrations/versions/2026-05-27-0929_add_checkouts_v2_amount_columns.py
```diff
@@ -0,0 +1,138 @@
...
+down_revision = "8493a1870f36"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+CHECKOUTS_SYNC_V2_AMOUNTS = PGFunction(
+    schema="public",
+    signature="checkouts_sync_v2_amounts()",
+    definition="""RETURNS trigger AS $$
+    BEGIN
```
Yopi, line 41: @frankie567 I made this bidirectional in this PR instead. I am hoping that it will avoid the false reads that we got in the billing entries.

PR #11900, opened 2026-05-26
Skip invoice regeneration when inputs are unchanged

joebon:
## Summary

`POST /v1/orders/{id}/invoice` always enqueues a regeneration task today, even when nothing affecting the rendered invoice has changed. This adds an `orders.invoice_checksum` column so the endpoint can short-circuit those redundant regenerations.

## What

- New `orders.invoice_checksum` column (nullable string, autogenerated migration)
- `InvoiceService.compute_order_checksum(order)` — SHA-256 over a JSON-encoded payload of `INVOICE_TEMPLATE_VERSION`, `billing_name`, `billing_address`, `tax_id`, and `customer.locale`
- `OrderService.trigger_invoice_generation` short-circuits when the stored checksum matches the recomputed one and an invoice already exists; endpoint still returns 202
- `OrderService.generate_invoice` persists the freshly computed checksum alongside `invoice_path`

## Why

Each `POST /invoice` call enqueues a worker task that renders a PDF and uploads to S3. When nothing has changed (the common case for retries, customer-portal refreshes, etc.) this is pure waste. The checksum lets us recognize that case cheaply and skip the work without changing the API contract.

## How

The checksum captures every practically-mutable input to the rendered invoice:
- `billing_name` and `billing_address` — exposed in `OrderUpdate`
- `customer.locale` — mutable via customer profile
- `tax_id` — currently set once at order creation, but included for safety
- `INVOICE_TEMPLATE_VERSION` — manually bumped when the template or seller-side settings (`INVOICES_NAME`, `INVOICES_ADDRESS`) change

Payload is serialized with `json.dumps(..., sort_keys=True)` to avoid delimiter ambiguity between user-controlled fields.

Legacy rows (no checksum yet) compare `None == <hash>` → falsy, so they fall through to regeneration as before. After the first regeneration each row captures its checksum, so no explicit backfill is needed.

The `InvoiceService.compute_order_checksum` method asserts that `order.customer` is loaded, since `Order.customer` uses `lazy="raise"` — both current call sites eager-load via `OrderRepository.get_eager_options()`.

## Checklist


pieterbeulque:
Would you backfill the checksums or just allow one more potentially wasteful regeneration?

The thing is that we changed our Merchant of Record information so quite a few regenerations would actually render a different PDF than what is currently stored.

joebon:
> Would you backfill the checksums or just allow one more potentially wasteful regeneration?
> 
> The thing is that we changed our Merchant of Record information so quite a few regenerations would actually render a different PDF than what is currently stored.

Good point @pieterbeulque !  I realized the same after a while and shifted to _not_ backfill. So the job populates the stored checksum once completed and then we reasonably would know they're up to date.

server/polar/invoice/service.py
```diff
@@ -31,6 +38,34 @@ def __init__(self, account: Account) -> None:
...
 class InvoiceService:
+    def compute_order_checksum(self, order: Order) -> str:
+        """Checksum over the mutable inputs to invoice generation, used to
+        short-circuit regeneration when nothing affecting the rendered invoice
+        has changed since the last generation.
+
+        Caller must ensure ``order.customer`` is loaded (e.g. via
+        ``OrderRepository.get_eager_options()``).
+        """
+        assert "customer" in order.__dict__, (
```
frankie567: That assertion is a bit weird IMO. If `order.customer` is not eagerly loaded, we'll get a "lazy raise" exception.

server/tests/invoice/test_service.py
```diff
@@ -27,3 +28,69 @@ async def test_create_order_invoice(
     )
 
     invoice_path = await invoice_service.create_order_invoice(order)
+
+
+@pytest.mark.asyncio
+class TestComputeOrderChecksum:
+    async def _order(
```
frankie567: A pytest fixture would work :)

frankie567 (approved):

psincraian (approved):

server/polar/invoice/service.py
joebon: Yeah, it's odd I agree. Was hesitant to include it in the first place, happy to remove

server/tests/invoice/test_service.py
joebon: Haha, yes of course 🤦‍♂️

PR #11905, opened 2026-05-26
fix(order): correct off-by-one in dunning retry exhaustion

pieterbeulque:
The Stripe webhook handler upserts the failed Payment row before
delegating to the order's dunning state machine, so
count_failed_payments_for_order returns a count that already includes
the current failure (and the initial subscription_cycle failure).

The previous logic used `failed_attempts >= len(DUNNING_RETRY_INTERVALS)`
and `intervals[failed_attempts]`, which silently skipped intervals[1]
(the 5d wait) and revoked subscriptions after only 3 retries instead of
4, finalizing dunning ~5 days before the configured past_due_deadline.

Use `>` for the exhaustion check and `failed_attempts - 1` for the
interval index. Update the affected tests to seed the database with the
extra failure row that production always records first.

https://claude.ai/code/session_015sepPrBrdWi2vCKiyce8Ts

### DIFF
--- a/server/polar/order/service.py
+++ b/server/polar/order/service.py
@@ -2233,8 +2233,11 @@ class OrderService:
         now = utc_now()
         subscription = order.subscription
 
+        # failed_attempts includes the current failure (upserted by the Stripe
+        # webhook handler before we get here) and the initial cycle failure
+        # which _handle_first_dunning_attempt already paired with intervals[0].
         if (
-            failed_attempts >= len(settings.DUNNING_RETRY_INTERVALS)
+            failed_attempts > len(settings.DUNNING_RETRY_INTERVALS)
             or (
                 subscription is not None
                 and subscription.past_due_deadline
@@ -2253,7 +2256,7 @@ class OrderService:
             return order
 
         # Schedule next retry using the appropriate interval
-        next_interval = settings.DUNNING_RETRY_INTERVALS[failed_attempts]
+        next_interval = settings.DUNNING_RETRY_INTERVALS[failed_attempts - 1]
         next_retry_date = now + next_interval
 
         order = await repository.update(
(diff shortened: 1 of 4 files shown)


server/polar/order/service.py
```diff
@@ -2233,8 +2233,11 @@ async def _handle_consecutive_dunning_attempts(
         now = utc_now()
         subscription = order.subscription
 
+        # failed_attempts includes the current failure (upserted by the Stripe
+        # webhook handler before we get here) and the initial cycle failure
+        # which _handle_first_dunning_attempt already paired with intervals[0].
         if (
-            failed_attempts >= len(settings.DUNNING_RETRY_INTERVALS)
+            failed_attempts > len(settings.DUNNING_RETRY_INTERVALS)
```
pieterbeulque, line 2240: Should I add an explicit + 1 instead for clarity?

server/tests/order/test_service.py
```diff
@@ -2816,6 +2825,8 @@ async def test_final_attempt_canceled_subscription(
         mocker: MockerFixture,
     ) -> None:
         """Test that order service cancels subscription after final retry attempt"""
+        from polar.config import settings
```
frankie567: Can we put it as top level import?

server/tests/order/test_tasks.py
```diff
@@ -403,8 +410,10 @@ async def test_payment_retry_failure_final_attempt_marks_canceled(
         order.next_payment_attempt_at = very_old_time
         await save_fixture(order)
 
-        # Create 4 failed payments to exhaust all retry attempts
-        for _ in range(4):
+        # Initial cycle failure + all DUNNING_RETRY_INTERVALS retries failed
+        from polar.config import settings
```
frankie567: Another one :)

server/polar/order/service.py
frankie567: I think it's fine like this :)

frankie567 (changes requested):

frankie567 (approved):

psincraian (approved):
Upss 🙈

PR #11910, opened 2026-05-26
scripts: Bring back billing entry amount backfill

Yopi:
We might need it for local dev for a while longer


frankie567 (commented):
We could also tweak the delete column migration to actually make the backfill? In prod, since the backfill has passed, it'll be basically a no-op. WDYT?

Yopi:
> We could also tweak the delete column migration to actually make the backfill? In prod, since the backfill has passed, it'll be basically a no-op. WDYT?

Oh. Yes, that's a lot smarter. Good idea!

PR #11928, opened 2026-05-27
Re-sync seat-based benefits on subscription product change

### LINKED ISSUE #11808: Benefits grants aren't updated for members on product change
When subscription_service.update_product() is called to change a seat-based subscription from Product A → Product B mid-cycle, the subscription is updated and an invoice is generated, but the benefit grants of users on already-claimed seats are not updated. They keep Product A's benefits indefinitely (or until the seat is manually re-claimed), even when proration_behavior=invoice.

1. Create two seat-based recurring products, A and B, with different benefits attached.
2. Subscribe a customer to A with N ≥ 1 seats.
3. Have at least one seat claimed by another user. Confirm that user receives A's benefits.
4. Call `PATCH /v1/subscriptions/{id}` (or the equivalent service call) with `product_id=B.id `and `proration_behavior=invoice`.

Observe:
 - ✅ `subscription.product_id` is now B
 - ✅ Prorated order/invoice created and paid
 - ❌ Existing seat-holder benefit grants still reference Product A
 - ❌ No new grants for Product B's benefits are created for the already-claimed seat holders

Re-claiming the seat (revoke + reclaim) does grant Product B's benefits correctly, which confirms the gap is specifically in the plan-change path.

For seat-based products, `enqueue_benefits_grants()` short-circuits when the subscription is active. This invariant — "for active seat-based subscriptions, benefits are granted only at seat-claim time" — holds for the initial subscription lifecycle but breaks under a product change. The comment "we don't need to do anything here" is no longer true once the underlying product (and its benefit set) has changed out from under the existing grants.

A helper that does exactly the right re-sync already exists: `customer_seat_service.update_product_benefits_grants()`. It is currently only called from the "merchant edits a product's benefits" flow, not from `update_product()`.

joebon:
## Summary

**Related Issue**: #11808

When a seat-based subscription was moved from product A → product B mid-cycle, existing claimed seats kept A's benefit grants indefinitely. This wires the seat re-sync into the product change path.

## What

- New `seat_service.update_subscription_benefits_grants(subscription)` helper that enqueues `benefit.enqueue_benefits_grants` (task=`grant`) for each claimed seat of a single subscription, scoped to the subscription's current product.
- `subscription_service.update_product()` calls the new helper after `enqueue_benefits_grants` when the new product is seat-based and the subscription is active.
- Test in `TestUpdateProduct` confirming the re-sync fires for a claimed seat (and ignores pending/revoked seats) with the new product's id.

## Why

`subscription_service.enqueue_benefits_grants` short-circuits for active seat-based subscriptions on the assumption that benefits are granted at seat-claim time. That invariant breaks under a product change: the underlying product (and its benefit set) swapped out, but no one re-synced the existing seat holders' grants. Re-claiming a seat (revoke + reclaim) worked, confirming the gap is specific to the plan-change path.

## How

A per-subscription seat helper mirrors the existing `update_product_benefits_grants(product)` used by the merchant-edits-product-benefits flow, but scoped to a single subscription instead of sweeping every subscription on the product. Each enqueued grant job flows through `benefit_grant_service.enqueue_benefits_grants`, which:

1. Computes the new product's benefits to grant.
2. Calls `list_outdated_grants(new_product, subscription=...)` to find still-granted benefits no longer attached to the new product.
3. Runs revoke-then-grant as a Dramatiq pipeline.

This delegates grant/revoke logic to the same pipeline the merchant-benefit-edit flow already uses, so behavior matches the existing well-tested code path.

## Checklist

### DIFF
--- a/server/polar/customer_seat/service.py
+++ b/server/polar/customer_seat/service.py
@@ -1107,5 +1107,34 @@ class SeatService:
             )
             index += 1
 
+    async def update_subscription_benefits_grants(
+        self, session: AsyncReadSession, subscription: Subscription
+    ) -> None:
+        """Re-sync benefit grants for claimed seats of a single subscription.
+
+        Used when a subscription's product changes mid-cycle: existing
+        seat holders must pick up the new product's benefits (and lose any
+        outdated grants from the previous product).
+        """
+        repository = CustomerSeatRepository.from_session(session)
+        seats = await repository.list_by_subscription_id(subscription.id)
+        claimed_seats = [
+            seat
+            for seat in seats
+            if seat.status == SeatStatus.claimed and seat.customer_id is not None
+        ]
+
+        calculate_delay = make_bulk_job_delay_calculator(len(claimed_seats))
+        for index, seat in enumerate(claimed_seats):
+            enqueue_job(
+                "benefit.enqueue_benefits_grants",
+                task="grant",
+                customer_id=seat.customer_id,
+                product_id=subscription.product_id,
+                member_id=seat.member_id,
+                subscription_id=seat.subscription_id,
+                delay=calculate_delay(index),
+            )
+
 
 seat_service = SeatService()
--- a/server/polar/subscription/service.py
+++ b/server/polar/subscription/service.py
@@ -1205,6 +1205,14 @@ class SubscriptionService:
 
             await self.enqueue_benefits_grants(session, subscription)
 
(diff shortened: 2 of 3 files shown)


frankie567 (approved):
Great catch!

joebon:
> Great catch!

Thanks - but credit goes to @pieterbeulque !

psincraian (approved):
Looks good!

PR #11939, opened 2026-05-27
Allow upgrading from non-seat to seat

pieterbeulque:
Fixes #133

## Summary by CodeRabbit

* **New Features**
  * Plan upgrades from standard to seat-based subscriptions are now fully supported with automatic customer and seat assignment.

* **Bug Fixes**
  * Enhanced validation to clearly reject incompatible subscription plan transitions.

* **Documentation**
  * Clarified plan change requirements including currency matching and seat-based transition rules.

### DIFF
--- a/server/polar/subscription/service.py
+++ b/server/polar/subscription/service.py
@@ -1098,13 +1098,15 @@ class SubscriptionService:
 
         old_has_seat_prices = any(is_seat_price(p) for p in previous_prices)
         new_has_seat_prices = any(is_seat_price(p) for p in currency_prices)
-        if old_has_seat_prices != new_has_seat_prices:
+
+        # Seat → non-seat plan changes are not yet supported.
+        if old_has_seat_prices and not new_has_seat_prices:
             raise PolarRequestValidationError(
                 [
                     {
                         "type": "value_error",
                         "loc": ("body", "product_id"),
-                        "msg": "Can't switch between seat-based and non-seat-based products.",
+                        "msg": "Can't switch from a seat-based to a non-seat-based product.",
                         "input": product_id,
                     }
                 ]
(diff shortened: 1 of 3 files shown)


server/polar/subscription/service.py
```diff
@@ -1203,6 +1228,20 @@ async def update_product(
...
+                    session, subscription.customer
+                )
+                await seat_service.assign_seat(
+                    session,
+                    subscription,
+                    customer_id=subscription.customer_id,
+                    immediate_claim=True,
+                )
+
             await self.enqueue_benefits_grants(session, subscription)
```
psincraian, line 1245: nitpick: assign_seat immediately already enqueues the benefit grants

server/polar/subscription/service.py
```diff
@@ -490,13 +490,7 @@ async def create(
...
         # Auto-upgrade customer to 'team' type when subscribing to a seat-based product
         if product.has_seat_based_price:
-            customer_type = customer.type or CustomerType.individual
-            if customer_type == CustomerType.individual:
-                await customer_repository.update(
-                    customer,
-                    update_dict={"type": CustomerType.team},
-                    flush=True,
-                )
+            await self._maybe_upgrade_customer_to_team(session, customer)
```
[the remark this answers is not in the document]
psincraian: MMmm I know that we can update the product prices, but I don't think we are able to add or remove seat prices from a product

psincraian (approved):

server/polar/subscription/service.py
pieterbeulque: Should we then skip it for the path where we set the owner to avoid duplicate runs?

frankie567 (approved):

server/polar/subscription/service.py
psincraian: I think that makes sense. But, checking again I think it will be a no-op as we check for existing grants already. So we can ignore this and trigger twice the message

PR #11941, opened 2026-05-27
Wait for fulfillment before emitting embed success event

sebastianekstrom:
Our embedded checkout iframe was posting `event: 'success'` to the parent the moment `confirm()` returned (`status === 'confirmed'`), not when the payment actually settled (`status === 'succeeded'`). Merchants listening for that event were fulfilling customers whose charges later got blocked by Stripe risk rules or declined.

👇 Notice how the `success` event is fired immediately after `confirmed` before the fix

| Before the fix | After the fix  |
|--------|--------|
| without-fix | with-fix |

## Summary by CodeRabbit

* **Bug Fixes**
  * Checkout confirmation now waits for actual fulfillment more broadly (including embed/internal and external flows) before showing success.
  * If fulfillment fails or times out, users are redirected to the confirmation page to prevent premature success state.

[Review Change Stack](https://app.coderabbit.ai/change-stack/polarsource/polar/pull/11941?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/clients/apps/web/src/hooks/checkout.ts
+++ b/clients/apps/web/src/hooks/checkout.ts
@@ -44,15 +44,21 @@ export const useCheckoutConfirmedRedirect = (
         )
       }
 
-      // For external success URL, make sure the checkout is processed before redirecting
-      // It ensures the user will have an up-to-date status when they are redirected,
-      // especially if the external URL doesn't implement proper webhook handling
-      if (!isInternalSuccessURL && listenFulfillment) {
+      // Wait for actual payment fulfillment before declaring success when
+      // either the redirect destination can't observe in flight state
+      // (external success URL) or a merchant is listening on the parent
+      // page (embed_origin). confirm() returning only means the payment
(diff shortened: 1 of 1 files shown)


clients/apps/web/src/hooks/checkout.ts
```diff
@@ -44,15 +44,20 @@ export const useCheckoutConfirmedRedirect = (
         )
       }
 
-      // For external success URL, make sure the checkout is processed before redirecting
-      // It ensures the user will have an up-to-date status when they are redirected,
-      // especially if the external URL doesn't implement proper webhook handling
-      if (!isInternalSuccessURL && listenFulfillment) {
```
pieterbeulque, line 50: I don't think this is the fix! For `isInternalSuccessURL` I believe we immediately redirect to the confirmation view, so that should trigger `confirm`, but then the bug is that it triggers `success` too soon.

frankie567 (approved):

PR #11943, opened 2026-05-27
fix: collapse duplicate org review runs and make confirm atomic

psincraian:
## Summary

Two compounding bugs caused duplicate AI-agent review runs on orgs that crossed their next-review threshold.

## What

- Added `debounce_key=…` to the `organization.check_threshold` Dramatiq actor (mirrors the `customer_meter.update_customer` pattern, default 15s/15min thresholds).
- Rewrote `confirm_organization_reviewed` to issue an atomic `UPDATE … WHERE status IN (REVIEW, SNOOZED) RETURNING` with server-side doubling via `func.greatest(next_review_threshold * 2, _MIN)`. Return type is now `Organization | None` — `None` means another worker won the race.
- `handle_ongoing_review_verdict` reports `True/False` based on whether *this* worker won, so only the winner records the agent decision.
- Two new race-safety tests: race-loss returning `None`, and doubling reading from the DB row rather than a stale in-memory snapshot.

## Why

The root cause is commit `0c576c484` (*refactor: enqueue organization.check_threshold instead of calling org service from transaction layer*). That refactor turned an inline, in-session `for` loop into per-balance-row Dramatiq enqueues.

The pre-refactor version was correct by accident: a single order writes several balance rows (charge + platform fee + reversal fees), and the inline loop called `check_review_threshold` sequentially against the same in-memory `Organization`. Once the first call mutated `status` from `ACTIVE` to `REVIEW`, subsequent calls in the same loop observed the new state and bailed at the `if status != ACTIVE` guard. One business event → one `under_review` enqueue → one agent run. The implicit serialization came from sharing one Python object across the calls.

After the refactor, N concurrent Dramatiq workers each open their own session, read the org fresh from the database, never see the others' in-memory mutation, and all flip `ACTIVE → REVIEW` + enqueue `under_review` + spawn an agent run.

The post-verdict doubling in `confirm_organization_reviewed` then read the in-memory `next_review_threshold`, so N parallel verdicts after a long LLM call all doubled from the same starting value — collapsing into a single effective doubling and leaving the org's threshold stuck.

The refactor's commit message acknowledged one trade-off (`organization.total_balance` becoming eventually consistent) but missed the load-bearing one: replacing an inline call with an enqueue removed the sequential-Python serialization that the read-modify-write on `status` was relying on.

## How

- The debounce restores the "one business event, one threshold check" semantic by collapsing per-balance-row enqueues back into a single execution per window. Redis dedup replaces the implicit in-memory state-sharing the refactor accidentally relied on.
- The atomic UPDATE uses the existing `cast(CursorResult[…], …).rowcount` idiom from `order/repository.py`.
- The repository method `confirm_review_atomic` takes the doubling input as `int | None`; `None` triggers the SQL-side `GREATEST` doubling, an explicit int writes it verbatim (preserves the backoffice override path).
- All callers (auto-approve worker + backoffice approve flows) compile against the new `Organization | None` return type without changes — they all discard the return value today. Worth a follow-up to surface a toast in the backoffice when the race is lost; tracked separately.

## Operational follow-ups (not in this PR)

- Backfill `next_review_threshold` for orgs whose value is stuck below `default × 2^approvals` due to collapsed doublings.
- Audit cases where parallel agent runs reached opposite verdicts within seconds — the verdict that "stuck" was determined by the race winner, not by deliberation.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrency handling for organization review confirmations via atomic updates to prevent race conditions; auto-approval now respects race outcomes and may not apply if another worker wins.
  * Added debouncing to threshold monitoring tasks to prevent duplicate simultaneous executions in multi-worker environments.
* **Tests**
  * Expanded concurrency-focused tests to verify race behavior and threshold update correctness.

[Review Change Stack](https://app.coderabbit.ai/change-stack/polarsource/polar/pull/11943?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/polar/organization/repository.py
+++ b/server/polar/organization/repository.py
@@ -1,9 +1,8 @@
 from collections.abc import Sequence
 from datetime import datetime
-from typing import cast
 from uuid import UUID
 
-from sqlalchemy import CursorResult, Select, func, select, update
+from sqlalchemy import Select, func, select, update
 from sqlalchemy.orm import joinedload
 
 from polar.authz.types import AccessibleOrganizationID
@@ -289,12 +288,13 @@ class OrganizationRepository(
         min_threshold: int,
         active_capabilities: OrganizationCapabilities,
         now: datetime,
-    ) -> bool:
+    ) -> Organization | None:
         """Atomically transition a REVIEW or SNOOZED organization to ACTIVE.
 
-        Returns ``True`` if the row was updated, ``False`` if it was no
-        longer in a confirmable state (e.g. another worker already won the
-        race and set the org back to ACTIVE).
+        Returns the updated ``Organization`` (also merged into the session's
+        identity map via ``populate_existing``), or ``None`` if the row was
+        no longer in a confirmable state — typically because another worker
+        already won the race and flipped the org back to ACTIVE.
 
         When ``next_review_threshold`` is ``None``, the threshold is doubled
         server-side from the current row (floored at ``min_threshold``) so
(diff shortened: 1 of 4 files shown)


frankie567 (approved):
Great stuff! TIL that our business rule is now to double the threshold at each review 😆

PR #11953, opened 2026-05-27
Fix multiline names in /sales

sebastianekstrom:
| Before | After  |
|--------|--------|
| Screenshot 2026-05-27 at 16 08 44 | Screenshot 2026-05-27 at 16 08 49 |

## Summary by CodeRabbit

* **Style**
  * Updated the customer column styling in the sales table to prevent text wrapping and improve readability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/polarsource/polar/pull/11953?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #11944, opened 2026-05-27
fix: partial unique index on organization_reviews

psincraian:
## Summary

The unique index on `organization_reviews(organization_id)` did not filter on `deleted_at`, so a soft-deleted row blocked any subsequent INSERT for the same organization. The SUBMISSION review worker then crashed with `UniqueViolationError`, retried once, and DLQed; affected dashboards polled `/review-status` indefinitely showing "Reviewing organization…" with no verdict and no recovery path.

## What

- Drop the total unique index `ix_organization_reviews_organization_id`.
- Recreate it as a partial unique index with `WHERE deleted_at IS NULL`.
- Update the `OrganizationReview` model's `__table_args__` to match.

Migration uses `CREATE/DROP INDEX CONCURRENTLY` inside `autocommit_block` so no table-level write lock is held. Each step is idempotent (IF EXISTS guards + temp `_new` name), so a failed deploy retries cleanly.

## Why

This is the established pattern in Polar for soft-deletable models with per-parent uniqueness (`members`, `subscription_update`, `user_organization`, `discount`, `organization_review_feedback`, and others).

## Checklist

- [x] This PR addresses a single concern (one bug fix, one feature, one refactor)
- [x] The diff is reasonably sized and easy to review
- [ ] New functionality is covered by tests
- [x] Linting and type checking pass (\`uv run task lint && uv run task lint_types\`)
- [x] No unrelated changes or drive-by fixes are included

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database constraint handling for organization reviews so soft-deleted entries no longer block new reviews, allowing multiple stored reviews per organization with only one active at a time.
* **Chores**
  * Added a database migration to apply the constraint change safely without downtime.

[Review Change Stack](https://app.coderabbit.ai/change-stack/polarsource/polar/pull/11944?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #11961, opened 2026-05-28
ci(deploy): diff against last-deployed SHA via git tags

psincraian:
## Problem

The current deploy workflow can silently drop commits when GitHub Actions concurrency cancels a pending run.

**The mechanics:**
- `concurrency: group: build-deploy` allows only one in-progress + one pending run per group. When a third push lands, the pending one is **cancelled**.
- `dorny/paths-filter@v4` on `push` events diffs against `github.event.before` (the immediate parent of the new push).
- Combined: if commit B's run is cancelled, the next commit C's filter only diffs `C..B`. Any backend/frontend changes that were in B (and earlier dropped commits) are never detected as "changed" and never deployed.

**Real incident** (run [26517009590](https://github.com/polarsource/polar/actions/runs/26517009590)):

| Time | Commit | PR | Type | Outcome |
|------|--------|----|------|--------|
| 14:02:29 | `8a6c421f6` | #11952 | backend | ✅ deployed (bundled #11944's migration with it) |
| 14:02:59 | `8d58080a0` | #11943 | backend-only | ❌ run cancelled, **changes never deployed** |
| 14:07:09 | `bde397456` | #11951 | frontend | ❌ run cancelled |
| 14:14:55 | `cdead53d8` | #11953 | frontend | ✅ frontend deployed, backend skipped (paths-filter saw only `cdead53d8..bde397456` = frontend) |

PR #11943's backend changes sat in `main` but never reached Render.

## Fix

Replace the implicit "diff against parent" diff base with an explicit "diff against the last successfully deployed SHA for this component + environment." State lives in four lightweight git tags:

```
deployed/backend-sandbox
deployed/backend-production
deployed/frontend-sandbox
deployed/frontend-production
```

Each tag advances only when its component deploys successfully to its environment. The `changes` job in [deploy.yml](.github/workflows/deploy.yml) reads each tag and passes it as `base:` to `dorny/paths-filter@v4` — once per environment. Cancelled intermediate runs are now harmless: the next run still sees the dropped commits because the tag hasn't moved.

### Walk-through of the same incident under this fix

After the `8a6c421f6` deploy: `backend-prod = 8a6c421f6`, `frontend-prod = 8a6c421f6`.

The runs for `8d58080a0` and `bde397456` get cancelled by concurrency. `cdead53d8` runs:
- Backend filter: `git diff 8a6c421f6..cdead53d8 -- server/` → **server/ files changed** (8d58080a0's commit) → backend = true ✅
- Frontend filter: `git diff 8a6c421f6..cdead53d8 -- clients/` → clients/ files changed → frontend = true

Backend image builds with 8d58080a0's fix baked in, both deploy, both tags advance to `cdead53d8`. No commit lost.

### Why per-environment

Sandbox can be ahead of production (failed prod deploy leaves sandbox tag advanced, prod tag stays). Each environment needs its own diff base to compute its own pending changes correctly. The `changes` job runs `paths-filter` twice — once per env — and emits per-env outputs (`backend-sandbox`, `backend-production`, etc.) consumed by the respective deploy jobs.

### Why per-component tag-update jobs

The tag-update logic lives in `deploy-environment.yml` as two separate jobs:
- `update-backend-tag` — `needs: deploy-backend`, gated on `deploy-backend.result == 'success' && skip-backend != true`
- `update-frontend-tag` — `needs: promote-frontend`, gated on `promote-frontend.result == 'success' && skip-frontend != true`

Splitting them ensures a frontend failure doesn't pin the backend tag at a stale SHA. If they shared a single job gated on overall success, a Vercel hiccup would leave the backend tag at the old SHA, causing the next push to redundantly redeploy backend to Render.

### Bonus: free observability win

Any engineer can run this locally to see exactly what's pending deploy:

```bash
git fetch --tags
git log deployed/backend-production..main -- server/
git log deployed/frontend-production..main -- clients/
```

## Bootstrap (REQUIRED before merge)

The four tags don't exist yet. Workflow falls back to `github.event.before` when a tag is missing, so a missing-tag is degraded behavior (today's behavior), not broken. But for the first deploy after merge to behave correctly, create the tags pointing at the current `main` HEAD just before merging:

```bash
SHA=$(git rev-parse origin/main)
for tag in deployed/backend-sandbox deployed/backend-production \
           deployed/frontend-sandbox deployed/frontend-production; do
  git tag "$tag" "$SHA"
  git push origin "refs/tags/$tag"
done
```

The first push after merge will then see no diff against the tags (correctly: nothing has been deployed since the tag was created) and skip both components. The deploy will not happen — that's expected, since nothing has changed. The next real change advances the tag and the system is in steady state.

## Rollback story

If something goes wrong after merge:
1. Revert this PR. Workflow goes back to old behavior immediately on next push.
2. The `deployed/*` tags can stay or be deleted — they're orphaned but harmless.

If a tag gets force-pushed or deleted accidentally during normal operation:
```bash
git tag -f deployed/backend-production <known-good-sha>
git push -f origin refs/tags/deployed/backend-production
```

## Permissions

The new tag-update jobs in `deploy-environment.yml` declare `permissions: contents: write` so `GITHUB_TOKEN` can push tags. No org-level changes needed. The `changes` job in `deploy.yml` doesn't push anything and keeps the default permissions.

## Testing

Tested via:
- YAML parses cleanly (`yaml.safe_load`)
- Walked through the bug scenario above by hand against the new diff logic
- Walked through frontend-only, backend-only, and mixed PR scenarios

## Checklist

- [x] Workflow YAML parses
- [x] No regression to manual `workflow_dispatch` (still forces all deploys via existing `|| github.event_name == 'workflow_dispatch'` guards)
- [x] No regression to migration ordering (migrations filter is per-env and still drives the API-first behavior in [deploy_server.sh](.github/workflows/deploy_server.sh))
- [x] No regression to sandbox-before-production gate
- [x] Per-component tag updates are independent (frontend failure won't pin backend tag)
- [ ] Bootstrap tags created (do this just before merging)

## Summary by CodeRabbit

* **Infrastructure & Deployment**
  * Automatic lightweight git tags now mark deployed frontend and backend commits; tag updates run with retry attempts, are gated to main branch for promoted runs, and are best-effort.
  * Error-tracking release publishing is best-effort and won’t fail deployments on auth/rate issues.
  * Change detection is environment-aware (sandbox vs production) for more precise deploy decisions.
  * Workflow permissions updated to allow necessary write operations for deployments.

### DIFF
--- a/.github/workflows/deploy-environment.yml
+++ b/.github/workflows/deploy-environment.yml
@@ -153,3 +153,82 @@ jobs:
         env:
           VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
           VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+
+  # Tag-advance jobs: after a successful deploy of a component, push a
+  # lightweight git tag deployed/{component}-{environment} pointing at the
+  # SHA we just deployed. The next workflow run's `changes` job uses this
+  # tag as the diff base, so cancelled intermediate runs never cause a
+  # commit to be skipped silently — the next run still sees the pending
+  # changes because the tag hasn't advanced.
+  #
+  # Each component has its own job, gated only on its own deploy's success.
+  # This way a frontend failure doesn't pin the backend tag at a stale SHA
+  # (which would cause Render to be re-deployed redundantly on the next
+  # push), and vice versa.
+
+  update-backend-tag:
+    name: "Update deployed/backend-${{ inputs.environment }} tag"
+    needs: deploy-backend
+    if: always() && needs.deploy-backend.result == 'success' && inputs.skip-backend != true
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 1
+
+      - name: Push deploy tag
+        env:
+          TAG: deployed/backend-${{ inputs.environment }}
+          DEPLOYED_SHA: ${{ github.sha }}
+        run: |
+          set -euo pipefail
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          # Retry a few times in case of transient push failures. A failed
+          # tag push means the next deploy will redundantly redeploy this
+          # SHA (safe, idempotent, but noisy) — worth retrying.
+          for attempt in 1 2 3; do
+            git tag -f "$TAG" "$DEPLOYED_SHA"
+            if git push -f origin "refs/tags/$TAG"; then
+              echo "Advanced $TAG -> $DEPLOYED_SHA"
+              exit 0
+            fi
+            echo "Push attempt $attempt failed; retrying..."
+            sleep 2
+          done
+          echo "Failed to push $TAG after 3 attempts"
+          exit 1
+
+  update-frontend-tag:
+    name: "Update deployed/frontend-${{ inputs.environment }} tag"
+    needs: promote-frontend
+    if: always() && needs.promote-frontend.result == 'success' && inputs.skip-frontend != true
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 1
+
+      - name: Push deploy tag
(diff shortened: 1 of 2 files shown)


frankie567 (approved):
That looks quite complex, but makes sense from the high level :)

PR #11962, opened 2026-05-28
Extract visibility enum to shared kit module

pieterbeulque:
I want to roll out visibility to benefits & meters as well, so I thought it's best to reuse what we already have from products.

## How

Refactors the `ProductVisibility` enum into a reusable `Visibility` enum in the kit module, with a `VisibilityMixin` for database column mapping. This enables visibility to be used across multiple models while maintaining backward compatibility with the `ProductVisibility` enum for API schemas.

This change enables visibility to be a shared concept across multiple models (not just products) while maintaining the existing API contract. The `ProductVisibility` enum remains for backwards compatibility in the SDK, but the actual database column and business logic now use the shared `Visibility` enum.

## What

- Created `polar/kit/visibility.py` with:
  - `Visibility` enum (draft, private, public)
  - `VisibilityMixin` for SQLAlchemy column mapping
- Updated `Product` model to use `VisibilityMixin` instead of defining its own visibility column
- Kept `ProductVisibility` enum as a distinct type for OpenAPI/SDK schema naming
- Updated all service and test files to import and use `Visibility` instead of `ProductVisibility`

https://claude.ai/code/session_01TPxMM4dmcLwC64LLMExcRG

## Summary by CodeRabbit

* **Refactor**
  * Improved internal visibility state management system across product-related services and models for better code organization and reusability.
  * Updated test fixtures and seed data to align with refactored visibility system.

[Review Change Stack](https://app.coderabbit.ai/change-stack/polarsource/polar/pull/11962?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/clients/packages/client/src/v1.ts
+++ b/clients/packages/client/src/v1.ts
@@ -27367,7 +27367,7 @@ export interface components {
         | null
     }
     /**
-     * ProductVisibility
+     * Visibility
      * @enum {string}
      */
     ProductVisibility: 'draft' | 'private' | 'public'
@@ -30942,6 +30942,11 @@ export interface components {
         [key: string]: string
       }
     }
+    /**
+     * Visibility
+     * @enum {string}
+     */
+    Visibility: 'draft' | 'private' | 'public'
     /**
      * Wallet
      * @description A wallet represents a customer's balance in your organization.
@@ -57713,6 +57718,9 @@ export const userUpdateCountryAnyOf0Values: ReadonlyArray<
   'ZM',
   'ZW',
 ]
+export const visibilityValues: ReadonlyArray<
+  FlattenedDeepRequired<components>['schemas']['Visibility']
+> = ['draft', 'private', 'public']
 export const walletSortPropertyValues: ReadonlyArray<
   FlattenedDeepRequired<components>['schemas']['WalletSortProperty']
 > = ['created_at', '-created_at', 'balance', '-balance']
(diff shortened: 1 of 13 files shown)


server/polar/models/product.py
```diff
@@ -48,13 +49,16 @@ class ProductBillingType(StrEnum):
     recurring = "recurring"
 
 
+# Kept as a distinct enum (rather than reusing `Visibility`) so the public
+# OpenAPI/SDK component stays named `ProductVisibility`. The product schemas
+# reference this; the DB column comes from `VisibilityMixin` (`Visibility`).
 class ProductVisibility(StrEnum):
```
frankie567: Worth to double check, but I _believe_ we can do a black magic like this:

```py
from polar.kit.schemas import SetSchemaReference

ProductVisibility = Annotated[
    Visibility, SetSchemaReference("ProductVisibility")
]
```

frankie567 (approved):
I like that!

clients/packages/client/src/v1.ts
```diff
@@ -27367,7 +27367,7 @@ export interface components {
         | null
     }
     /**
-     * ProductVisibility
+     * Visibility
```
pieterbeulque, line 27370: @frankie567 Looks like it worked, but this comment isn't matching, not sure if that has any downstream issues (I don't think so?)

frankie567: I think that'll be fine

PR #11964, opened 2026-05-28
feat: off-session charges / draft orders API

joebon:
## Summary

**Related Issue**: polarsource/feedback#127

Lets merchants charge a customer's saved payment method for an existing product over the API, without sending the customer through checkout — the stop-gap for merchant-rolled credit/wallet flows until managed wallets ship. API-only, gated behind the per-org `off_session_charges_enabled` feature flag.

## What

- **`POST /v1/orders/`** — creates an order with the new `draft` status (no charge, no invoice number). Takes `organization_id`, `customer_id`, `product_id`, optional `amount` (custom/PWYW prices, bounds-checked) / `seats` / `metadata`. One-time products only.
- **`POST /v1/orders/{id}/finalize`** — synchronously charges the saved card (explicit `payment_method_id` or the customer's default). On success → `paid`, invoice number assigned, benefit grants fire before the response. On failure → reverts to `draft` with a typed 4xx (402 decline/SCA, 412 not-draft, 403 flag off). The order ID is the idempotency key.

**Auth**: create resolves the org from the body and asserts `sales:manage` on it (product + customer scoped to that org); update + finalize move onto an `OrderPolicyGuard` dependency.

## Notes

- New `OrderStatus.draft`; `orders.invoice_number` made nullable (migration). `off_session_charges_enabled` added to org feature settings.
- Finalize claims the draft atomically (`draft → pending`) and settles the charge inline; the `charge.succeeded` webhook remains an idempotent safety net.
- 3DS/`requires_action` is a failure for v1 (merchant re-auths via the portal); no dunning/retries.

### Out of scope (per the issue)
Next-invoice line items, arbitrary amounts without a product, managed wallets, dashboard UI, dunning, recurring-product drafts, updating drafts beyond billing.

## Checklist

- [x] Single concern (the off-session charge feature)
- [x] Reasonably sized / easy to review
- [x] Covered by tests
- [x] Lint + type checking pass (server and clients)
- [x] No unrelated changes (the checkout touch is the shared-validator consolidation the order path depends on)

## Summary by CodeRabbit

* **New Features**
  * Draft order creation and finalize flow for off‑session payments (with organization feature flag)
  * New draft order status with UI indicators (gray status pill and list/detail support)

* **Bug Fixes**
  * Invoice numbers are optional until an order is finalized

* **Chores**
  * Centralized/custom-price validation logic consolidated and reused

[Review Change Stack](https://app.coderabbit.ai/change-stack/polarsource/polar/pull/11964?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


server/polar/order/service.py
```diff
@@ -244,6 +248,47 @@ def __init__(self, reason: PaymentFailedReason) -> None:
...
+    """
+
+    def __init__(self, order: Order, payment_intent_id: str) -> None:
+        self.order = order
+        self.payment_intent_id = payment_intent_id
+        message = (
+            f"Order {order.id} payment requires customer authentication "
+            "(e.g. 3DS challenge); off-session charge cannot complete."
+        )
+        super().__init__(message, 402)
```
joebon, line 268: Does `402 Payment Required` make sense? I'm a bit torn. `412 Precondition Failed` could work aswell

server/polar/order/schemas.py
```diff
@@ -272,6 +279,42 @@ class Order(CustomFieldDataOutputMixin, MetadataOutputMixin, OrderBase):
...
+            "The ID of the organization the order belongs to. "
+            "**Required unless you use an organization token.** "
+            "The customer and product must belong to this organization."
+        ),
+    )
+    customer_id: UUID4 = Field(
+        description="The ID of the customer the order is for. "
+        "Must belong to the order's organization."
+    )
+    product_id: UUID4 = Field(
```
frankie567, line 298: Actually, I think the use-case is to let merchants create fully arbitrary orders. So I would expect an `items` list consisting of label/amount/currency to charge for. 

WDYT @pieterbeulque?

server/polar/order/service.py
frankie567: 402 feels great to me :)

server/polar/order/endpoints.py
```diff
@@ -190,29 +202,79 @@ async def get(
...
-    await assert_resource_permission(
-        session, auth_subject, order, OrganizationPermission.sales_manage
-    )
 
-    return await order_service.update(session, order, order_update)
+@router.post(
+    "/{id}/finalize",
+    summary="Finalize Order",
+    response_model=OrderSchema,
+    responses={404: OrderNotFound},
```
frankie567: Could be nice to document the `PaymentActionRequired`, `OrderNotDraft` and `PaymentFailed` here

server/polar/order/service.py
```diff
@@ -2309,14 +2815,19 @@ async def _calculate_subscription_order_tax(
         customer: Customer,
         product: Product,
         tax_exempted: bool,
+        billing_address_override: Address | None = None,
```
frankie567: We could make `billing_address` a mandatory argument so we explicitly pass `customer.billing_address` in other cases.

frankie567 (commented):
Awesome, I think we're mostly there.

The only thing we need to align on is if we impose it to be under a given product, or if we allow fully arbitrary line items.

server/polar/order/endpoints.py
joebon: Good point! Fixed

server/polar/order/schemas.py
joebon: Hrm, that's not how I read the specs but it's certainly an alternative way forward.
But how would you in that case for instance grant benefits with that approach?

server/polar/order/service.py
pieterbeulque: "Stripe uses this code for failed payments where parameters were correct, for example blocked fraudulent payments"

Seems to match that!

server/polar/order/schemas.py
pieterbeulque: I would also link it to a product, I think it'll be good to have that product act as a "bucket" and link these off-session charges back into known Polar entities (e.g. filtering analytics by your "Credits" product to see all these charges bundled for your 'credit revenue').

I think just allowing arbitrary charges not linked to any domain object will backfire with follow-up work: analytics, grant benefits upon charge, …

We'll probably have to figure out how to support arbitrary amounts, yes, but I was thinking to rely on our ad-hoc pricing for that and match that API?

joebon: How do we reach a decision?

frankie567: If that solves the need, then let it be `product_id` :)

joebon: @pieterbeulque do we need to support discounts applied to these purchases?

pieterbeulque: I'd say yes, I feel that the closer it matches an actual user-interacted checkout, the least follow-up questions we're going to get. Feel free to push back if it introduces a lot of complexity.

joebon: @pieterbeulque I agree it makes sense but also realize it's reasonably a scope worthy its own follow-up PR

server/polar/order/schemas.py
```diff
@@ -272,6 +279,42 @@ class Order(CustomFieldDataOutputMixin, MetadataOutputMixin, OrderBase):
...
+    amount: int | None = Field(
+        None,
+        ge=0,
+        description=(
+            "Amount in the smallest currency unit. Required for "
+            "pay-what-you-want / custom-priced products; ignored otherwise. "
+            "Must respect the price's configured minimum and maximum."
+        ),
+    )
+    seats: int | None = Field(
```
pieterbeulque: I wonder if we should block seat-based products from this as well to reduce potential blast radius?

server/polar/order/schemas.py
```diff
@@ -272,6 +279,42 @@ class Order(CustomFieldDataOutputMixin, MetadataOutputMixin, OrderBase):
...
+        description="The ID of the one-time product to charge for. "
+        "Must belong to the order's organization. "
+        "Subscription products are not supported."
+    )
+    amount: int | None = Field(
+        None,
+        ge=0,
+        description=(
+            "Amount in the smallest currency unit. Required for "
+            "pay-what-you-want / custom-priced products; ignored otherwise. "
```
pieterbeulque: Implicitly, this means we're recommending people to set up a private "Credits" product with PWYW pricing if they want to do custom charges.

Just spelling it out to see how it feels 😁

joebon:
This got to big. Closing as I split it into two new PRs:
- #12089 
- #12090

PR #11969, opened 2026-05-28
Make product visibility non-nullable

pieterbeulque:
## Summary by CodeRabbit

* **Chores**
  * Product visibility is now required and defaults to "public", improving data consistency.
* **Bug Fixes**
  * Eliminated cases where products lacked visibility to reduce unexpected behavior.
* **Tests**
  * Updated test fixtures to supply explicit visibility values to reflect the new required behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/polarsource/polar/pull/11969?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


server/migrations/versions/2026-05-28-1206_make_product_visibility_non_nullable.py
```diff
@@ -0,0 +1,42 @@
...
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column(
+        "products",
+        "visibility",
+        existing_type=sa.VARCHAR(),
+        nullable=False,
```
Yopi, line 52: So I know you said that the products table is small, so I believe this is totally fine.

For large tables (or if we always want to do so) you can add
```
ALTER TABLE products
ADD CONSTRAINT products_visibility_not_null
CHECK (visibility IS NOT NULL) NOT VALID
```
```
ALTER TABLE products VALIDATE CONSTRAINT products_visibility_not_null
```

Before altering the table, and then removing this constraint after the alter has been completed. It ensures that you don't do a full table scan when adding the not-null constraint in the alter.

Yopi (approved):
Nice

server/migrations/versions/2026-05-28-1206_make_product_visibility_non_nullable.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # Step 3: set NOT NULL — because a validated check constraint already proves
+    #         no NULLs exist, Postgres 12+ skips the full table scan and holds
+    #         ACCESS EXCLUSIVE for only a very short time.
+    op.alter_column(
+        "products",
+        "visibility",
+        existing_type=sa.VARCHAR(),
+        nullable=False,
+        existing_server_default=sa.text("'public'::character varying"),
```
frankie567, line 53: Do we really have that server default set? I don't think it'll impact anything here, but this surprised me.

frankie567 (approved):

server/migrations/versions/2026-05-28-1206_make_product_visibility_non_nullable.py
pieterbeulque: image

Yes. Why did it surprise you?

frankie567: Because it's not specified in `mapped_column` argument 😅 In general, I'm not super comfortable having server defaults.

PR #11987, opened 2026-05-29
Add composite `(customer_id, status)` index on subscriptions

pieterbeulque:
To fix #11983 

## Summary by CodeRabbit

* **Chores**
  * Optimized database performance for improved query execution speed.

[Review Change Stack](https://app.coderabbit.ai/change-stack/polarsource/polar/pull/11987?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/migrations/versions/2026-05-29-0918_add_subscriptions_customer_id_status_index.py
+++ b/server/migrations/versions/2026-05-29-0918_add_subscriptions_customer_id_status_index.py
@@ -1,0 +1,47 @@
+"""add (customer_id, status) index on subscriptions
+
+Revision ID: f1a3c7429eba
+Revises: 0c12d2aaab31
+Create Date: 2026-05-29 09:18:38.849185
+
+Adds the composite index alongside the existing single-column
+ix_subscriptions_customer_id index. The old index is dropped in a follow-up
+migration once this one is deployed.
+
+If a previous CIC was killed it leaves an INVALID index behind. Re-running
+fails on the name collision. To recover, run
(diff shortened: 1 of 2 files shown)


server/migrations/versions/2026-05-29-0918_add_subscriptions_customer_id_status_index.py
```diff
@@ -0,0 +1,47 @@
...
+# revision identifiers, used by Alembic.
+revision = "f1a3c7429eba"
+down_revision = "0c12d2aaab31"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
+    with op.get_context().autocommit_block():
+        op.execute("SET lock_timeout = '5s'")
```
Yopi, line 31: This feels like a new thing that Claude always wants to thrown in

Yopi: Since it's concurrently I don't think it's needed, but also doesn't really matter

Yopi (approved):
Nice

frankie567 (approved):
That's great but why not bundling this with #11983 directly? :D

pieterbeulque:
@frankie567 I was thinking since we create the index concurrently, there could have been a gap where there is no index on `customer_id`. Probably overthinking it.

pieterbeulque:
Ah, with #11983? We shouldn't have migrations in a feature PR, right?

frankie567:
Yeah, right, but since it's an index, I would have 👍

PR #11999, opened 2026-05-29
amount v2: Migrate payments amounts to BigInt

Yopi:
Adds the new amount columns as _v2 and a backfill script. Reads will move over in the second PR and the legacy columns will be dropped in the third.

## Summary by CodeRabbit

* **New Features**
  * Expanded payment amount capacity from 32-bit to 64-bit integers, enabling support for larger transaction values.
  * Automatic synchronization between legacy and new amount fields for seamless data handling.

* **Chores**
  * Added database migration and backfill utility to populate payment amounts in the new field.

### DIFF
--- a/server/scripts/backfill_amount_v2_payments.py
+++ b/server/scripts/backfill_amount_v2_payments.py
@@ -1,0 +1,54 @@
+import asyncio
+from functools import wraps
+
+import typer
+from sqlalchemy import select, update
+
+from polar.models import Payment
+from scripts.helper import (
+    configure_script_logging,
+    limit_bindparam,
+    run_batched_update,
+)
+
+cli = typer.Typer()
+
+configure_script_logging()
+
+
+def typer_async(f):  # type: ignore
+    @wraps(f)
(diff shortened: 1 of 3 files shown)


server/scripts/backfill_amount_v2_payments.py
```diff
@@ -0,0 +1,58 @@
...
+        (
+            update(Payment)
+            .values(
+                amount_v2=Payment.amount,
+            )
+            .where(
+                Payment.id.in_(
+                    select(Payment.id)
+                    .where(
+                        or_(
```
joebon: If I'm reading this right then this `or_()` is pointless (as it only has one argument) and confusing. Let's drop

joebon (approved):
Nice! Left a non-blocking comment, but I'd recommend fixing for clarity

frankie567 (approved):

server/scripts/backfill_amount_v2_payments.py
Yopi: Fair yes, it's because all of the other ones had more amounts 😄

PR #12035, opened 2026-05-29
 meter: treat LIKE/NOT LIKE values literally in SQL clauses

### LINKED ISSUE #12024: [Detail Bug] Metering: LIKE/NOT LIKE filters treat '_' and '%' inconsistently between SQL and ingestion matching, dropping billable events
# Detail Bug Report

https://app.detail.dev/org_ecfbc7cf-6d21-4ce1-8c61-cda1d650ccf7/bugs/bug_e858be05-a75a-49e9-9413-a7c526872853

Introduced in [#8793](https://github.com/polarsource/polar/pull/8793) by @Yopi on Jan 7, 2026

# Summary
- **Context**: The `FilterClause` class in `server/polar/meter/filter.py` implements event filtering with both SQL generation (`get_sql_clause()`) and in-memory matching (`matches()`) for meter aggregation.
- **Bug**: The `like` and `not_like` operators treat `%` and `_` characters as SQL wildcards in database queries but as literal characters in Python matching, causing inconsistent filter results.
- **Actual vs. expected**: SQL path interprets `%` as "match any sequence" and `_` as "match any single character", while Python `matches()` treats them as literal characters. Both should perform substring containment (literal match) as documented.
- **Impact**: Revenue leakage - events are displayed in usage counts but not billed.

# Code with Bug
```python
# server/polar/meter/filter.py
elif self.operator == FilterOperator.like:
    return attr.like(f"%{value}%")  # <-- BUG 🔴 does not escape SQL LIKE wildcards (% and _)
elif self.operator == FilterOperator.not_like:
    return attr.notlike(f"%{value}%")
```

```python
# server/polar/meter/filter.py
elif self.operator == FilterOperator.like:
    return str(expected) in str(actual)  # <-- BUG 🔴 treats % and _ as literals; disagrees with SQL
elif self.operator == FilterOperator.not_like:
    return str(expected) not in str(actual)
```

# Explanation
`FilterClause` is evaluated in two places for the same ingested events:
- During ingestion, `_create_meter_events(...)` uses `FilterClause.matches()` (Python substring check).
- In the background task that activates meters, SQL is generated via `get_sql_clause()`.

Because SQL `LIKE` treats `_`/`%` as wildcards but Python treats them as literal characters, an event can be counted by SQL-based usage queries but excluded from the `MeterEvent` table during ingestion. Billing later joins through `MeterEvent`, so excluded events are not billed.

Minimal repro from the report:
- Filter value: `api_test`
- Event name: `apiXtest`
- `matches()` returns `False` (literal underscore mismatch)
- SQL `LIKE '%api_test%'` matches `apiXtest` (`_` matches `X`)

## Codebase Inconsistency
Another area previously fixed the same pattern by enabling escaping (using `autoescape=True`), indicating the intended semantics are literal “contains”, not SQL-pattern matching.

# Recommended Fix
Use literal contains semantics in SQL (or otherwise escape `%` and `_`) so SQL and Python agree, e.g.:
```python
elif self.operator == FilterOperator.like:
    return attr.contains(value, autoescape=True)
elif self.operator == FilterOperator.not_like:
    return attr.notlike(f"%{value}%", escape="\\")
```

# History
This bug was introduced in commit 1f5a74ee when `matches()` was added for ingestion-time filtering, implementing `like`/`not_like` as literal substring checks while the pre-existing SQL implementation continued to use `LIKE '%value%'` with wildcard behavior, creating inconsistent results between ingestion (billing) and SQL queries (usage).

Yopi:
Fix #12024

The `like`/`not_like` operators generated SQL `LIKE '%value%'`, so `%` and `_` in the filter value were interpreted as SQL wildcards. The ingestion-time `FilterClause.matches()` check, however, treats them as literal characters

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed LIKE and NOT LIKE string pattern matching to treat underscore and percent characters as literal values instead of wildcards.

* **Tests**
  * Added test cases validating wildcard character handling in string filters across SQL and in-memory filtering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/polarsource/polar/pull/12035?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/polar/meter/filter.py
+++ b/server/polar/meter/filter.py
@@ -116,9 +116,9 @@ class FilterClause(BaseModel):
         elif self.operator == FilterOperator.lte:
             return attr <= value
         elif self.operator == FilterOperator.like:
-            return attr.like(f"%{value}%")
+            return attr.contains(value, autoescape=True)
         elif self.operator == FilterOperator.not_like:
-            return attr.notlike(f"%{value}%")
+            return ~attr.contains(value, autoescape=True)
         raise ValueError(f"Unsupported operator: {self.operator}")
 
     def _get_str_value(self) -> str:
--- a/server/tests/meter/test_filter.py
+++ b/server/tests/meter/test_filter.py
@@ -332,6 +332,87 @@ class TestFilter:
         assert len(matching_events) == 1
         assert matching_events[0].id == events[1].id
 
+    async def test_like_treats_wildcards_as_literal(
+        self,
+        save_fixture: SaveFixture,
+        session: AsyncSession,
+        organization: Organization,
+    ) -> None:
+        """`_` must be matched literally, not as a SQL LIKE wildcard, so the
+        SQL path agrees with `FilterClause.matches()`."""
+        literal_match = await create_event(
+            save_fixture, organization=organization, name="api_test"
+        )
+        wildcard_only = await create_event(
+            save_fixture, organization=organization, name="apiXtest"
+        )
+
+        clause = FilterClause(
+            property="name", operator=FilterOperator.like, value="api_test"
+        )
+        filter = Filter(conjunction=FilterConjunction.and_, clauses=[clause])
+
+        repository = EventRepository.from_session(session)
+        statement = repository.get_base_statement().where(filter.get_sql_clause(Event))
+        matching_events = await repository.get_all(statement)
+
+        assert {e.id for e in matching_events} == {literal_match.id}
+        assert clause.matches(literal_match) is True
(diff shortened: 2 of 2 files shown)


frankie567 (approved):
If I understand correctly, the problem happened only if the meter clause value contained a "%" or "_", is that it?

Yopi:
> If I understand correctly, the problem happened only if the meter clause value contained a "%" or "_", is that it?

Correct then it would be counted as a wild card. As far as I could see there were no impacted events so this is just about covering our bases

PR #12047, opened 2026-06-01
server/auth: handle hard-coded app review login in email OTP flow

frankie567:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue in the email OTP flow so OTPs are now reliably persisted and verified during the app-review path.

* **Improvements**
  * More reliable email OTP handling to improve authentication stability and user identification.

### DIFF
--- a/server/polar/auth/factors.py
+++ b/server/polar/auth/factors.py
@@ -5,6 +5,7 @@ from math import ceil
 import structlog
 from fastapi import Depends
 from reauth.authentication_session import AuthenticationSession
+from reauth.crypto import get_token_hash
 from reauth.factors import FactorBase
 from reauth.factors.backup_codes import (
     BackupCodesEnrollment as BackupCodesEnrollmentDataclass,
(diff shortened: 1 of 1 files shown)


server/polar/auth/factors.py
```diff
@@ -104,6 +120,18 @@ async def request(
             authentication_session_id=authentication_session.id,
         )
 
+        if (
+            email_otp.email == settings.APP_REVIEW_EMAIL
```
psincraian, line 124: what's this APP_REVIEW_EMAIL?

frankie567: So, for Apple Store and Google Play reviewers, they require a username/password combo (yes, they don't want OAuth or Passkeys). So our solution was to add a hardcoded email/otp code pair we give them to review the app. Shitty, but that's the only workaround right now.

psincraian (approved):
LGTM

PR #12051, opened 2026-06-01
Send questions straight to Plain

pieterbeulque:
## Summary by CodeRabbit

* **New Features**
  * Question-type feedback submissions now enqueue background processing to automatically create Plain support threads.
  * Thread subjects are generated via an AI-driven helper with robust fallback handling.
  * When present, transcripts are posted as customer messages and preserved; otherwise original messages become internal notes.
  * Feedback records are updated to persist support thread URLs and mark triaged status after thread creation.

### DIFF
--- a/server/polar/integrations/plain/service.py
+++ b/server/polar/integrations/plain/service.py
@@ -1,7 +1,6 @@
 # pyright: reportCallIssue=false
 import asyncio
 import contextlib
-import json
 import uuid
 from collections.abc import AsyncIterator, Awaitable, Callable
 
@@ -49,10 +48,7 @@ from plain_client import (
     UpsertCustomerUpsertCustomer,
     UpsertTenantInput,
 )
-from plain_client.exceptions import (
-    GraphQLClientGraphQLMultiError,
-    GraphQLClientHttpError,
-)
+from plain_client.exceptions import GraphQLClientGraphQLMultiError
 from pydantic_ai import Agent
 from sqlalchemy import func, or_, select
 from sqlalchemy.orm import contains_eager, joinedload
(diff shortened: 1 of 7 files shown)


server/polar/integrations/plain/service.py
```diff
@@ -84,6 +92,132 @@ def plain_thread_url(thread_id: str) -> str:
     return f"https://app.plain.com/workspace/{PLAIN_WORKSPACE_ID}/thread/{thread_id}"
 
 
+# Markers used by the question chat flow when it builds the feedback message
```
pieterbeulque, line 91: This is arguably terrible, but I'm fine with it… I can look into another format if it's too smelly for you.

server/polar/integrations/plain/service.py
```diff
@@ -84,6 +88,100 @@ def plain_thread_url(thread_id: str) -> str:
...
+                {} if model_name.startswith("gpt-5.5") else {"temperature": 0}
+            ),
+        )
+        result = await asyncio.wait_for(agent.run(content[:4000]), timeout=10)
+        subject = result.output.strip().strip('"').strip()
+        return subject or fallback
+    except Exception:
+        return fallback
+
+
```
frankie567, line 184: So... All that stuff _just_ to generate the subject line? 😅

pieterbeulque: 😂 no, no. The front-end does this

```
Optional user provided escalation message

---

## Transcript

**User**: Initial message

**Assistant**: Initial reply

[...]
```

And the whole thing above makes it so that we create a Plain thread with an initial message impersonating the customer (so, faking an inbound email) with a generated subject line, but more importantly, using the user question as the email body (so either "Optional user provided escalation message" or "Initial message"), and then adds the full chat transcript as an internal note in Plain to help the CS team by showing what was already replied by the AI.

frankie567: Ok, but then is `_generate_thread_subject` really necessary? :D Spin up an agent in an event loop, calling GPT5.5, that's quite heavy.

pieterbeulque: It's a very nice QOL improvement but I can see why it's not great. The thing is it's hard to organize Plain if everything is titled "Re: your recent question"

frankie567: Ok, let's go then :D

frankie567 (approved):

PR #12053, opened 2026-06-01
Block OAT management screen for members

pieterbeulque:
## Summary by CodeRabbit

* **New Features**
  * Implemented permission-based access control for developer settings. Users without organization management permissions now see a restricted access message instead of access token configuration options.

* **Style**
  * Refined the visual appearance of the access restriction message with adjusted spacing and typography for better layout consistency.


clients/apps/web/src/components/Finance/AccessRestricted.tsx
joebon, line 1: Perhaps take the opportunity to move this component outside of *Finance*? Since it's not only used there and nor should it be

joebon (approved):

PR #12069, opened 2026-06-01
amount v2: Increase max amount for specific currencies

Yopi:
Now that all of the amounts have been migrated to bigint, we can increase the max amount for specific currencies.

This commit sets it to k for the ones that are specified. To be updated.

## Summary by CodeRabbit

* **New Features**
  * Per-currency maximum price configuration with a sensible default fallback and case-insensitive currency handling.
  * Minimum price now uses a configurable default fallback (case-insensitive) instead of a single hardcoded value.

* **Documentation**
  * Product pricing descriptions updated to show per-currency maximums and clarify maximum/minimum amount behavior.

### DIFF
--- a/clients/packages/client/src/v1.ts
+++ b/clients/packages/client/src/v1.ts
@@ -26769,12 +26769,29 @@ export interface components {
        *     - YER: 200
        *     - ZAR: 9
        *     - ZMW: 10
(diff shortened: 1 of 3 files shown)


pieterbeulque (approved):
The rabbit has a point about the docstring

PR #12085, opened 2026-06-02
server/migrations: add lock_timeout by default

frankie567:
To prevent the app from going down when running a migration, set a lock timeout by default. This will cause the migration to fail if it cannot acquire a lock within the specified time, rather than hanging indefinitely.

## Summary by CodeRabbit

* **Chores**
  * Enhanced database migration configuration with automatic lock timeout settings to improve deployment stability.

### DIFF
--- a/server/migrations/script.py.mako
+++ b/server/migrations/script.py.mako
@@ -19,8 +19,12 @@ depends_on: tuple[str] | None = ${repr(depends_on)}
 
 
 def upgrade() -> None:
+    # Ensures we don't break app by applying a deadlock-inducing migration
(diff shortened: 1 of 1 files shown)


joebon (approved):
Nice! AFAICT this can also be overridden inside the actual migrations if/when needed

frankie567:
> Nice! AFAICT this can also be overridden inside the actual migrations if/when needed

Yes, actually this is just a tweak to the migration template so we nudge engineers to have that timeout; but yes, they can always tweak or remove it

Yopi (approved):

psincraian (approved):
I didn't know that we were able to have this nice templates 👏

PR #12093, opened 2026-06-02
Add optimistic updates for customer/account notifications

sebastianekstrom:
Our notification settings in the dashboard can be toggled very quickly, which means that we can get into the following state:

1. I **toggle on** notification A, then B, then C
2. We fire a request for notification A
3. Request comes back and sets A as on, however this request also sets B and C as off
4. UI gets in a bad state

This PR solves that with an optimistic update approach instead. Meaning the following now happens:

1. I **toggle on** A, then B, then C. Each switch flips instantly in the UI
2. A request is sent for the first change (toggle A on). While we wait for the response, toggling B and C does not fire new requests it just updates the local state
3. When we get the response back, another request is sent with the latest state (A+B+C on).
4. Because only one request is sent at a time, responses can't arrive out of order and overwrite each other
5. The UI is in a good state

If a request fails, the switches roll back to the last successfully saved state and an error toast is shown.

### Before

https://github.com/user-attachments/assets/618acb2c-0e03-479a-a8a3-9cfd9566f5b3

### After

https://github.com/user-attachments/assets/29b40fdf-dcf3-463a-a8b1-6783649b0871

## Summary by CodeRabbit

* **New Features**
  * Notification and customer email settings now update instantly with optimistic saves, providing immediate visual feedback.

* **Refactor**
  * Streamlined settings interfaces by removing unnecessary form validation steps.

* **Tests**
  * Added comprehensive test coverage for optimistic update behavior, including concurrent updates and error recovery scenarios.

### DIFF
--- a/clients/apps/web/src/components/Settings/OrganizationCustomerEmailSettings.tsx
+++ b/clients/apps/web/src/components/Settings/OrganizationCustomerEmailSettings.tsx
@@ -1,17 +1,9 @@
 import { useUpdateOrganization } from '@/hooks/queries'
-import { useAutoSave } from '@/hooks/useAutoSave'
-import { extractApiErrorMessage, setValidationErrors } from '@/utils/api/errors'
-import { isValidationError, schemas } from '@polar-sh/client'
+import { useOptimisticSave } from '@/hooks/useOptimisticSave'
+import { extractApiErrorMessage } from '@/utils/api/errors'
+import { schemas } from '@polar-sh/client'
 import Switch from '@polar-sh/ui/components/atoms/Switch'
-import {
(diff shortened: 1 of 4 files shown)


pieterbeulque (approved):
Better! I think the `useOptimisticSave` would benefit from a debounce parameter to optimize this even more, and avoid the situation where users move away and updates were only partially saved.

sebastianekstrom:
@pieterbeulque a debounce here would delay the network call which I think would make the UX worse. Right now, as soon as the initial request comes back for toggle A, it will immediately fire a single request for toggle B, C... X. Adding a debounce would delay that second request.

PR #12098, opened 2026-06-02
worker: Set up rails to run workers in lambda

Yopi:
The basic theory is that we'll have:
1 SQS queue and 1 lambda per task. We'll move over a task at a time based on the list of tasks to run in AWS.

This will give us an individual knob to tune worker concurrency per task, and they will stop bottlenecking each other.

Follow up work for this would be:
- Set up the infra in AWS
- Set up the env variables in TF
- Pick a task to start testing out with

## Summary by CodeRabbit

* **New Features**
  * AWS Lambda entrypoint and SQS-based worker dispatch for running tasks in Lambda
  * LocalStack service and healthcheck for local SQS/Lambda development
  * Configurable SQS routing with actor allowlisting and per-queue naming/prefix
  * New CLI commands to run and poll SQS worker tasks locally
  * Batched SQS send with error reporting and compact message envelopes

* **Tests**
  * Added tests verifying SQS routing behavior for enqueued jobs

[Review Change Stack](https://app.coderabbit.ai/change-stack/polarsource/polar/pull/12098?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


server/polar/worker/aws_lambda.py
```diff
@@ -0,0 +1,51 @@
...
+                    receive_count=receive_count,
+                    source_correlation_id=correlation_id,
+                )
+            )
+        except Exception:
+            log.error(
+                "polar.worker.sqs_task_failed", message_id=message_id, exc_info=True
+            )
+            batch_item_failures.append({"itemIdentifier": message_id})
+    return {"batchItemFailures": batch_item_failures}
```
frankie567, line 51: Is it a builtin feature of Lambda/SQS to trigger a retry?

frankie567 (commented):
So, if I read that correctly, we have a home-made runner that polls the SQS queue and executes the task in an event loop. Is that it?

server/polar/worker/aws_lambda.py
Yopi: Yes https://docs.aws.amazon.com/lambda/latest/dg/services-sqs-errorhandling.html

Yopi:
> So, if I read that correctly, we have a home-made runner that polls the SQS queue and executes the task in an event loop. Is that it?

The home-made runner is only for local testing. I'm pretty sure we can hook the lambda up to the handler function directly (https://github.com/polarsource/polar/pull/12098/changes#diff-21f135a0e1f178db21e4f30b40991b862c9c03e9a773ead4a89bebd8ff3cdaebR28-R38) without the need to run it as an ad-hoc command. And if we do then it will come in via `run` for a single task (https://github.com/polarsource/polar/pull/12098/changes#diff-aceb0c8dd123c4920b407a368f180e89d7a05ed568d3384dff17c6ab7457e7f6R26-R31).

The polling is automatic between SQS and lambda

frankie567 (approved):
Looking forward to see a demo of that 🙂

I guess that when we're confident with this approach, we can get rid of Dramatiq and avoid the mimicking of CurrentMessage.

Main concern will be regarding the debounce feature, I think it's important to support it as well before winding down our current worker.

Yopi:
> Looking forward to see a demo of that 🙂
> 
> I guess that when we're confident with this approach, we can get rid of Dramatiq and avoid the mimicking of CurrentMessage.
> 
> Main concern will be regarding the debounce feature, I think it's important to support it as well before winding down our current worker.

Yeah I think we can roll this out end-to-end and then try it a bit for a single task in sandbox to start with, and play around with it. Maybe even with a dummy task

PR #12102, opened 2026-06-02
tf: Circumvent bootstrap issue for render

Yopi:
We've had an issue bootstrapping an environment due to us reading the data source. Since the deploys happen outside of terraform anyway, this attempts to disregard it fully and ignore changes to the image in terraform

## Summary by CodeRabbit

* **Chores**
  * Updated service deployment configuration to use latest container images instead of pinned versions for API and worker services across all environments.
  * Simplified image management by removing manual digest tracking from configuration files.
  * Added safeguards to prevent Terraform from reverting image updates made outside of Terraform workflows.

### DIFF
--- a/terraform/modules/render_service/main.tf
+++ b/terraform/modules/render_service/main.tf
@@ -253,12 +253,19 @@ resource "render_web_service" "api" {
   health_check_path  = "/healthz"
   pre_deploy_command = "uv run task pre_deploy"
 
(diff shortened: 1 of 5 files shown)


joebon (approved):
Nice! This felt awkward so glad if it can be dropped. LGTM!

frankie567 (approved):

PR #12113, opened 2026-06-03
feat(organisation): gate UI settings behind canManage flag

maximevast:
## What

Gate organisation settings behind `canManageOrganization` flag. They are already gated server side (HTTP 403)

## Why

This will make clearer what a `member` can or cannot do since we are moving notification settings to the user level

### CanManage

https://github.com/user-attachments/assets/b36342c4-79c2-4eb4-b769-76ec9a6cb982

### !CanManage

https://github.com/user-attachments/assets/9b6badc8-4b08-4fed-b012-32f5f789e4e4


maximevast:
@coderabbitai generate pr description

pieterbeulque (approved):
I _think_ it may be cleaner to name the prop on all settings `readOnly` instead of `canManageOrganization` so that it's less implementation-specific and more self-explanatory, then set `readOnly={!canManageOrganization}` on the caller.

Issue #12094, opened 2026-06-02
Allow to pass `discount_id` or `trial_end` in `SubscriptionUpdateProduct`


frankie567:
## Problem

In the current state, when updating an existing subscription, you can either:

* Change the Product
* Add or remove a discount
* Update the trial period
* Update the period end

Those four operations are *exclusive*, meaning you can't both change the product and apply a discount at the same time.

For some use-cases, it might be challenging. Typically, you may want to change the product with a discount set so the first prorated invoice is already discounted.

## Proposal

Enrich `SubscriptionUpdateProduct` to support both `discount_id` and `trial_end`, so we can apply those changes atomically.

PR #12155, opened 2026-06-04
server/subscription: add an update context to unify side-effects

frankie567:
Preparatory work for #12094

## Problem

In the current implementation, we have several ways to update a subscription: update product, update trial, cancel, revoke, etc. Each one is responsible to trigger the side effects, like the updated webhook or cycle the subscription.

We also have a global `update` method that simply routes to the specific update method, but it doesn't trigger any side effect.

However, in #12094, we'll need to "bundle" several updates into one (like product + trial + discount). In that case, we don't want to trigger the side effects for each update, but only once at the end of the process.

The challenge is then the following: we want to trigger the side effects once when calling `update`, but we also want to trigger them when calling the specific update methods directly.

Here is a schema of all the update methods and how they're called:

```mermaid
flowchart LR
    subgraph Update
        Product
        Discount
        Trial
        Seats
        BillingPeriod
        Cancel
        Uncancel
        Revoke
        ClearPendingUpdate
    end
    API
    CPAPI[Customer Portal API]
    Dispute
    OD[Order Dunning]
    Backoffice
    API --> Update
    API --> Revoke
    CPAPI --> Uncancel
    CPAPI --> ClearPendingUpdate
    CPAPI --> Seats
    CPAPI --> Revoke
    Backoffice --> Uncancel
    Backoffice --> BillingPeriod
    Dispute --> Revoke
    OD --> Revoke
```

## Proposed solution

In this PR, we implement a context manager that's responsible to keep track of the updates happening inside, and trigger the side effects when exiting:

1. Track the billing side effect: immediate invoice or cycle. Cycle will always take precedence over immediate invoice.
2. Track the updated metadata to build the subscription updated system event.
3. Trigger the `_after_subscription_updated` callback at the end of the context.

With this approach, we can be totally flexible in the number of individual updates we want to perform, and make sure we trigger the right side effects at the end of the process.

> [!NOTE]
> This PR doesn't change any behavior, but simply refactors the code to use the new context manager. The next PR will be the one that actually implements the product + trial + discount update, and will show the real benefits of this new approach.

---
## Summary by cubic
Adds an async SubscriptionUpdateContext to batch subscription update/cycle side effects and run them once at the end. No behavior changes; prepares bundled updates for #12094.

- **Refactors**
  - Introduced SubscriptionUpdateContext to collect billing actions (cycle > invoice-now), aggregate SubscriptionUpdatedMetadataFields, and trigger _after_subscription_updated once.
  - Updated service methods to accept a ctx and register effects: update, update_product, update_trial, update_seats, cancel, revoke, cycle.
  - Wrapped entry points with the context: public API, Customer Portal, Backoffice, dispute and dunning revocations, batch cancel script, and the subscription_cycle task; updated tests.
  - Simplified subscription.updated event metadata to a single shape and regenerated the OpenAPI client in `clients/packages/client`.

- **Migration**
  - When calling update/cancel/revoke/cycle directly, create an async SubscriptionUpdateContext and pass ctx; side effects run once on context exit.

<sup>Written for commit f94620b7f416fb8c43c638c63828260e09aff101. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12155?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/packages/client/src/v1.ts
+++ b/clients/packages/client/src/v1.ts
@@ -29876,20 +29876,6 @@ export interface components {
        */
       trial_end: string | 'now'
     }
-    /** SubscriptionUpdatedBillingPeriodMetadata */
-    SubscriptionUpdatedBillingPeriodMetadata: {
-      /** Subscription Id */
-      subscription_id: string
-      /** Billing Period End */
-      billing_period_end: string
-    }
-    /** SubscriptionUpdatedDiscountMetadata */
-    SubscriptionUpdatedDiscountMetadata: {
-      /** Subscription Id */
-      subscription_id: string
-      /** Discount Id */
-      discount_id: string | null
-    }
     /**
      * SubscriptionUpdatedEvent
      * @description An event created by Polar when a subscription is updated.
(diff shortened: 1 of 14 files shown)


Yopi (approved):
Oh, this is super neat!

PR #12162, opened 2026-06-04
Make subscriptions.organization_id non-nullable

pieterbeulque:
## Summary

**Related Issue**: 

Denormalizes `organization_id` directly onto the `Subscription` model as a non-nullable column, eliminating the need to traverse through the product relationship to determine subscription ownership.

## What

- Added database migration to make `subscriptions.organization_id` non-nullable using PostgreSQL's NOT VALID/VALIDATE pattern to minimize locking
- Updated `Subscription` model to make `organization_id` a required UUID column with a direct relationship to `Organization`
- Refactored all subscription queries to filter by `Subscription.organization_id` instead of `Product.organization_id`
- Updated subscription creation and update logic to set `organization` relationship instead of `organization_id`
- Added eager loading of `Subscription.organization` throughout the codebase to prevent N+1 queries

## Why

This change improves query efficiency and data consistency:
- Eliminates unnecessary joins to the `Product` table when filtering subscriptions by organization
- Provides direct access to organization information without traversing relationships
- Simplifies authorization checks by making organization ownership explicit on the subscription
- Reduces query complexity and improves performance for subscription list operations

## How

1. **Database Migration**: Uses PostgreSQL's constraint validation pattern to safely add NOT NULL constraint without full table locks:
   - Backfills any remaining NULL values from product organization
   - Adds NOT VALID check constraint (SHARE UPDATE EXCLUSIVE lock)
   - Validates constraint (SHARE UPDATE EXCLUSIVE lock)
   - Sets column NOT NULL (brief ACCESS EXCLUSIVE lock)
   - Drops redundant check constraint

2. **Model Changes**: 
   - Changed `organization_id` from nullable to non-nullable
   - Replaced `AssociationProxy` with direct `relationship` to `Organization`

3. **Query Refactoring**:
   - Updated `SubscriptionRepository` methods to filter by `Subscription.organization_id`
   - Removed unnecessary `Product` joins in authorization checks
   - Added `joinedload(Subscription.organization)` to eager loading options

4. **Data Updates**:
   - Updated seed data and test fixtures to provide `organization_id` or `organization` relationship
   - Updated subscription creation endpoints to set organization relationship

## Checklist

- [x] This PR addresses a single concern (denormalizing organization_id for query efficiency)
- [x] The diff is reasonably sized and focused on the denormalization
- [x] Existing tests pass with the schema changes
- [x] Migration uses safe PostgreSQL patterns to avoid locking issues

https://claude.ai/code/session_01KbVZWRgkFKk6HF7pdKH8gt

---
## Summary by cubic
Make `subscriptions.organization_id` required and add a direct `Subscription.organization` relationship. This removes `Product` joins from auth and listing paths, trims redundant eager-loads in read-only lists, and speeds up subscription queries.

- **Migration**
  - Backfills NULLs from `products.organization_id` as a safety net.
  - Uses NOT VALID → VALIDATE → SET NOT NULL to minimize locks; then drops the check.
  - Sets `lock_timeout` to 5s to avoid long waits.

- **Refactors**
  - Model: `organization_id` is non-nullable; replace association proxy with a direct `organization` relationship (`lazy="raise"`).
  - Repositories: filter by `Subscription.organization_id`; remove `Product` joins; eager-load `Subscription.organization`.
  - Services/endpoints: list/get filter on subscription columns; explicitly join `Subscription.product` where needed; backoffice cancel/uncancel and customer portal get-by-id eagerly load `Subscription.organization`; drop redundant eager-loads in read-only customer portal and claimed-seat lists.
  - Creation paths: assign the `organization` relationship on create; in checkout flows set from `Checkout.organization` (not `Product.organization`) to avoid `lazy="raise"` errors.
  - Seeds/scripts/tests: seeding sets `organization_id`; dev script and test fixture assign the `organization` relationship.

<sup>Written for commit 4b8e811e3597871ca6218713d55380dbb3183400. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12162?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/migrations/versions/2026-06-04-1000_make_subscriptions_organization_id_non_nullable.py
+++ b/server/migrations/versions/2026-06-04-1000_make_subscriptions_organization_id_non_nullable.py
@@ -1,0 +1,84 @@
+"""Make subscriptions.organization_id non nullable
+
+Revision ID: c9e9440f7d03
+Revises: 5263a32cf4dd
+Create Date: 2026-06-04 10:00:00.000000
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# Polar Custom Imports
+
+# revision identifiers, used by Alembic.
+revision = "c9e9440f7d03"
+down_revision = "5263a32cf4dd"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
+    # Ensures we don't break app by applying a deadlock-inducing migration
+    op.execute("SET LOCAL lock_timeout = '5s'")
+
+    # Production is expected to have been backfilled via
+    # scripts/backfill_subscription_organization_id.py before this migration
+    # runs; this UPDATE is a safety net for local environments where the script
+    # may not have been executed.
+    op.execute(
+        """
+        UPDATE subscriptions
+           SET organization_id = products.organization_id
+          FROM products
+         WHERE subscriptions.product_id = products.id
+           AND subscriptions.organization_id IS NULL
+        """
+    )
+
+    # Use the NOT VALID / VALIDATE pattern to avoid a full-table ACCESS EXCLUSIVE
+    # lock when setting NOT NULL.
+    #
+    # Step 1: add the check constraint as NOT VALID — takes only SHARE UPDATE
+    #         EXCLUSIVE, so concurrent reads and writes are unblocked.
+    op.execute(
+        """
+        ALTER TABLE subscriptions
+            ADD CONSTRAINT subscriptions_organization_id_not_null
+            CHECK (organization_id IS NOT NULL) NOT VALID
+        """
+    )
+
+    # Step 2: validate the constraint — also SHARE UPDATE EXCLUSIVE.  Postgres
+    #         scans the table for violations here, but concurrent DML is still
+    #         allowed (it must simply not introduce new NULLs).
+    op.execute(
+        """
+        ALTER TABLE subscriptions
+            VALIDATE CONSTRAINT subscriptions_organization_id_not_null
+        """
+    )
+
+    # Step 3: set NOT NULL — because a validated check constraint already proves
+    #         no NULLs exist, Postgres 12+ skips the full table scan and holds
+    #         ACCESS EXCLUSIVE for only a very short time.
+    op.alter_column(
+        "subscriptions", "organization_id", existing_type=sa.UUID(), nullable=False
+    )
+
+    # Step 4: the check constraint is now redundant — the NOT NULL column
+    #         constraint takes over.
+    op.execute(
+        """
+        ALTER TABLE subscriptions
+            DROP CONSTRAINT subscriptions_organization_id_not_null
+        """
+    )
+
+
+def downgrade() -> None:
+    # Ensures we don't break app by applying a deadlock-inducing migration
+    op.execute("SET LOCAL lock_timeout = '5s'")
+    op.alter_column(
+        "subscriptions", "organization_id", existing_type=sa.UUID(), nullable=True
+    )
--- a/server/polar/customer_portal/endpoints/customer_seat.py
+++ b/server/polar/customer_portal/endpoints/customer_seat.py
@@ -205,6 +205,7 @@ async def list_claimed_subscriptions(
         auth_subject
     ).options(
         joinedload(Subscription.customer),
+        joinedload(Subscription.organization),
         joinedload(Subscription.product).options(
             selectinload(Product.product_medias),
             joinedload(Product.organization),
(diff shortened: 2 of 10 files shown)


server/migrations/versions/2026-06-04-1000_make_subscriptions_organization_id_non_nullable.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def upgrade() -> None:
+    # Ensures we don't break app by applying a deadlock-inducing migration
+    op.execute("SET LOCAL lock_timeout = '5s'")
+
+    # Production is expected to have been backfilled via
+    # scripts/backfill_subscription_organization_id.py before this migration
+    # runs; this UPDATE is a safety net for local environments where the script
+    # may not have been executed.
```
frankie567, line 28: @cubic-dev-ai This is a key thing to remember: when a PR is a follow up of a data migration / backfill, it's important to also apply the equivalent backfill in the migration: this way, local environments are updated automatically, and it's still safe to run in prod because the backfill has run, making it a no op.

server/polar/customer_portal/endpoints/customer_seat.py
```diff
@@ -205,6 +205,7 @@ async def list_claimed_subscriptions(
         auth_subject
     ).options(
         joinedload(Subscription.customer),
+        joinedload(Subscription.organization),
         joinedload(Subscription.product).options(
             selectinload(Product.product_medias),
             joinedload(Product.organization),
```
frankie567, line 210: I think we can remove that joinedload then.

server/polar/customer_portal/service/subscription.py
```diff
@@ -126,6 +127,7 @@ async def get_by_id(
             .where(Subscription.id == id)
             .options(
                 joinedload(Subscription.customer),
+                joinedload(Subscription.organization),
                 joinedload(Subscription.product).options(
                     selectinload(Product.product_medias),
                     joinedload(Product.organization),
```
frankie567, line 132: Here too

server/polar/customer_portal/service/subscription.py
frankie567, line 85: We can also probably optimize this one

frankie567 (changes requested):

server/polar/customer_portal/endpoints/customer_seat.py
pieterbeulque: The nested `CustomerSubscriptionProduct` schema also serializes `organization` (sourced from `product.organization`), so `joinedload(Product.organization)` is still needed to render that nested field. That's why I added `joinedload(Subscription.organization)` for the new top-level `organization` field — the old `association_proxy("product", "organization")` that reused the product's org is gone now.

To load the org only once, I'd keep `joinedload(Product.organization)` and instead source the top-level `CustomerSubscription.organization` from the already-loaded `product.organization` (via an alias), dropping `joinedload(Subscription.organization)`. Want me to take that route here and in the two `customer_portal/service/subscription.py` spots? Or did you have a different shape in mind?

---

frankie567: Claude, can you check if we really need to serialize `organization` in `CustomerSubscriptionProduct` then?

pieterbeulque: Checked — I think we should keep it. `organization: CustomerOrganization` is serialized consistently on **all three** customer-portal product schemas: `CustomerOrderProduct` (`schemas/order.py`), `CustomerProduct` (`schemas/organization.py`), and `CustomerSubscriptionProduct`. It's how the customer portal surfaces the merchant org for a product, and it's already in the public/generated client — dropping it from just the subscription one would be inconsistent and a breaking API change.

So `joinedload(Product.organization)` is needed. The genuinely redundant one was my `joinedload(Subscription.organization)`, which the read-only lists don't use (they serialize the org via `product.organization`). I removed it from the subscription list and this claimed-seats list in 4b8e811.

I kept it in `customer_portal/service/subscription.py::get_by_id`, since that loader also feeds the cancel/update/uncancel mutations, which read the direct `subscription.organization` when building system events — dropping it there would hit `lazy="raise"`. Happy to revisit if you'd prefer those mutations re-load via `get_eager_options` instead.

---

server/migrations/versions/2026-06-04-1000_make_subscriptions_organization_id_non_nullable.py
frankie567: @cubic-dev-ai Yes, please record it.

frankie567 (approved):

PR #12172, opened 2026-06-04
Prepare discount calculation for multiple static prices

pieterbeulque:
Allow a discount to be consumed across multiple static prices through a single new `Discount.allocate_discount_amounts(amounts, currency)` primitive. Fixed amount discounts are allocated as a waterfall: the base fee absorbs its share first, the remainder flows to the seat charge instead of being capped independently per price. This would've over-discounted when a fixed discount spanned more than one static price.

`DiscountFixed` waterfalls, fixed fee first.
`DiscountPercentage` distributes independently (unchanged)

To ensure consistency on the waterfall, `PriceSet.get_static_prices()` now returns a canonical order.

This is a behavior-preserving refactor for every configuration reachable today (a single static price per currency returns exactly what the old per-price code did, verified by the unchanged single-static regression tests).

This is in preparation of fixed + seat pricing composition.

---
## Summary by cubic
Distributes discounts across multiple static prices with one allocation step. Fixed discounts now split proportionally across proratable amounts; percentage discounts still apply per amount.

- New Features
  - Added `Discount.allocate_discount_amounts(amounts, currency)`; prorations (credit/debit) and seat updates now use it (debit allocates only if the discount applies to the new product).
  - Allocation is order-independent: fixed discounts split proportionally with rounding leftovers given to the largest remainders so parts sum to the capped discount; `PriceSet.get_static_prices()` returns natural order again.

- Bug Fixes
  - Prevented over-discounting when a fixed discount spanned multiple prices during product changes and prorations.
  - Seat-change prorations no longer double-count fixed discounts; the seat uses its slice from the `[fixed, seat]` allocation.

<sup>Written for commit bad4ed3ba236012ce0012d1610155d9eb0558ec4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12172?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/models/discount.py
+++ b/server/polar/models/discount.py
@@ -103,11 +103,6 @@ class Discount(MetadataMixin, RecordModel):
         raise NotImplementedError()
 
     def allocate_discount_amounts(self, amounts: list[int], currency: str) -> list[int]:
-        """Allocate this discount across an ordered list of amounts.
-
-        Returns the discount applied to each amount, in the same order. Each
-        concrete discount type defines how the discount is distributed.
-        """
         raise NotImplementedError()
 
     def is_applicable(self, product: "Product", currency: str) -> bool:
--- a/server/polar/models/discount.py
+++ b/server/polar/models/discount.py
@@ -175,15 +175,21 @@ class DiscountFixed(Discount):
         return min(self.amounts[currency], amount)
 
     def allocate_discount_amounts(self, amounts: list[int], currency: str) -> list[int]:
-        # Waterfall: the first amount absorbs as much of the discount as it can,
-        # the remainder flows to the next, and so on. This guarantees the parts
-        # sum to min(discount, total)
-        remaining = self.amounts[currency]
-        allocations: list[int] = []
-        for amount in amounts:
-            allocated = min(remaining, amount)
-            allocations.append(allocated)
-            remaining -= allocated
+        # Distribute the discount across the amounts proportionally, capped at
+        # the combined total. The units lost when flooring each share are handed
+        # to the largest remainders, so the parts always sum to the discount on
+        # the combined total (i.e. `get_discount_amount` of the sum).
+        total = sum(amounts)
+        if total == 0:
+            return [0] * len(amounts)
+
+        discount = min(self.amounts[currency], total)
+        weighted = [discount * amount for amount in amounts]
+        allocations = [weight // total for weight in weighted]
+        for index in sorted(
+            range(len(amounts)), key=lambda i: weighted[i] % total, reverse=True
+        )[: discount - sum(allocations)]:
+            allocations[index] += 1
         return allocations
 
     @property
--- a/server/polar/models/discount.py
+++ b/server/polar/models/discount.py
@@ -102,6 +102,14 @@ class Discount(MetadataMixin, RecordModel):
     def get_discount_amount(self, amount: int, currency: str) -> int:
         raise NotImplementedError()
 
+    def allocate_discount_amounts(self, amounts: list[int], currency: str) -> list[int]:
+        """Allocate this discount across an ordered list of amounts.
+
+        Returns the discount applied to each amount, in the same order. Each
+        concrete discount type defines how the discount is distributed.
+        """
+        raise NotImplementedError()
+
     def is_applicable(self, product: "Product", currency: str) -> bool:
         raise NotImplementedError()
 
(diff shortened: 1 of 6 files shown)


server/polar/models/discount.py
```diff
@@ -205,6 +225,13 @@ def get_discount_amount(self, amount: int, currency: str) -> int:
         discount_amount_float = amount * (self.basis_points / 10_000)
         return polar_round(discount_amount_float)
 
+    def allocate_discount_amounts(self, amounts: list[int], currency: str) -> list[int]:
+        # Each amount is discounted independently. Because each share is rounded
+        # on its own, the parts can sum to ±(n-1) minor units off the discount on
+        # the combined total; this is accepted because prorations are not
+        # re-pooled at the order level.
```
pieterbeulque, line 233: A 10% discount on [995, 995] would round up to a [100, 100] = $2 discount, while you'd only have a $1.99 discount on the total of $19.90.

Since it's limited to ±1 minor unit I decided to not go much deeper into it…

server/tests/subscription/test_service_prorations.py
```diff
@@ -1413,3 +1365,338 @@ async def test_e2e_immediate_seats_with_pending_product_change(
...
+
+# --- Composed static price (fixed + seat) discount allocation -----------------
+#
+# These exercise the not-yet-reachable "fixed base fee + seat price" composition
+# (PR 2 enables it at the product layer). The factories build the prices and
+# subscription rows directly, bypassing product-price validation. Cases use
+# values where the fixed-amount waterfall *diverges* from the old per-price
+# `min(discount, amount)` behaviour, so they fail on the previous code and pass
+# now. Period 2025-06-01 → 2025-07-01 with an update at 2025-06-16 gives a clean
+# proration factor of 0.5.
```
pieterbeulque: These tests are premature but I figured let's already test that this is fixing what it's intending to fix 😊

frankie567:
Hmm... I'm wondering why we need to do it this way. From my perspective, the discount is applied on the "grand total". So the logic would rather be `SUM(static_prices) - discount`; so we don't need to ventilate across the prices. Or am I missing something?

pieterbeulque:
@frankie567 - I thought so too at first, but Claude was very convincing…

For full billing cycles it's functionally equivalent. For flat discounts, `SUM(static_prices) − discount` and the per-price split always produce the same total. I believe it already works like this for subscription cycles too.

However, I believe this will be needed for prorations, where discounts can't be applied on the grand total.

For a fixed + seat product, switching plans would produce four separate billing entries:

```
  credit: old fixed price   (its own row, product_price_id = old_fixed)
  credit: old seat price    (its own row, product_price_id = old_seat)
  debit:  new fixed price   (its own row, product_price_id = new_fixed)
  debit:  new seat price     (its own row, product_price_id = new_seat)
```

There are no totals here, and the discount has to be baked into these net amounts.  So to fill in each row's amount, you must answer "how much of the discount belongs to this line" (`allocate_discount_amounts`) computes. Discounts are thus applied at the billing-entry level, and we then emit one static line item per price, each of which becomes its own (discounted) order item. There's no single "grand total" object at this layer to subtract from.

For seat changes, the `SUM - discount` would be possible too (as only a single billing entry is created), but it felt better to keep it consistent then.

So to do `SUM − discount` for prorations, we'd have to lift the discounting up to the order-aggregation layer, which is a bigger refactor of how prorations work?

server/polar/models/discount.py
```diff
@@ -171,6 +174,18 @@ def is_applicable(self, product: "Product", currency: str) -> bool:
     def get_discount_amount(self, amount: int, currency: str) -> int:
         return min(self.amounts[currency], amount)
 
+    def allocate_discount_amounts(self, amounts: list[int], currency: str) -> list[int]:
+        # Waterfall: the first amount absorbs as much of the discount as it can,
```
joebon: Why do we do it this way, i.e. the waterfall way? Why don't we distribute the discount across the line items proportionally?
I.e. like this, with a $100 discount example:
```
Fixed monthly fee    $200     -$87           $113
3 seats à $10        $30      -$13           $17
Total                $230     -$100          $130
```

pieterbeulque: It's an arbitrary choice, but it's driven by the fixed + seats case (the only static + static combination we'll allow). Applying the discount as a waterfall reads better than spreading it proportionally.

Take your example: the fixed monthly fee is $200 and wil likely include some free seats, and there's a $200 discount code. With the waterfall the whole $200 lands on the base fee, so the included seats effectively come for free, which feels logical. Proportional distribution would smear the discount across both line items and force more mental gymnastics to reason about, for no real upside (?)

frankie567:
@pieterbeulque So the problem arises only for prorations then. Ideally, I would say it's better to keep those calculations for this context, and have this approach only when needed in prorations. I'm really worried this makes billing really complicated to understand; for us, for merchants and for customers.

server/polar/models/discount.py
joebon: > Applying the discount as a waterfall reads better than spreading it proportionally.
> […]
> With the waterfall the whole $200 lands on the base fee, so the included seats effectively come for free, which feels logical.

Hrm. Either I don't understand your point or I disagree with it.
But I think I'm coming at it from a more general invoice structure PoV, in which I found find the waterfall approach very unexpected (I've never seen it applied anywhere else).
The reason I bring it up is that if we want to do discounting this thorough (i.e. applied on line items instead of just a simple subtraction on the invoice total) then I think we should follow industry established patterns and best-practices and this feels off imho.

frankie567 (approved):
Looks better like this 👏

joebon (approved):

PR #12173, opened 2026-06-04
slack: Add slack/benefit model

Yopi:
The idea is that this table will hold the integration details between a Slack app on an organization, and the benefit it links to.

The benefit_id is nullable to allow the slack connection to be created before the benefit is saved.

---
## Summary by cubic
Add `SlackApp` to store Slack app installation details per organization, with a new `slack_apps` table and ORM export for use across the app.

- **New Features**
  - Added `SlackApp` model and `slack_apps` table; exported in `polar.models`.
  - Stores app/team/bot/user IDs, tokens/secrets, scopes, and install/revoke timestamps.
  - Unique on `slack_app_id`; cascade delete via `organization_id`; indexes on `organization_id`, `team_id`, `created_at`, and `deleted_at`.

<sup>Written for commit 6916397c83f57348df6e979bd17c6442e3e8c22c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12173?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/migrations/versions/2026-06-04-1650_add_slack_apps.py
+++ b/server/migrations/versions/2026-06-04-1650_add_slack_apps.py
@@ -1,0 +1,102 @@
+"""add slack apps
+
+Revision ID: bc2540d6d2c0
+Revises: c9e9440f7d03
+Create Date: 2026-06-04 16:50:37.766909
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# Polar Custom Imports
+
+# revision identifiers, used by Alembic.
+revision = "bc2540d6d2c0"
+down_revision = "c9e9440f7d03"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
+    # Ensures we don't break app by applying a deadlock-inducing migration
+    op.execute("SET LOCAL lock_timeout = '5s'")
+    op.create_table(
+        "slack_apps",
+        sa.Column("id", sa.Uuid(), nullable=False),
+        sa.Column("created_at", sa.TIMESTAMP(timezone=True), nullable=False),
+        sa.Column("modified_at", sa.TIMESTAMP(timezone=True), nullable=True),
+        sa.Column("deleted_at", sa.TIMESTAMP(timezone=True), nullable=True),
+        sa.Column("organization_id", sa.Uuid(), nullable=False),
+        sa.Column("display_name", sa.String(length=35), nullable=False),
+        sa.Column("slack_app_id", sa.String(length=32), nullable=True),
+        sa.Column("client_id", sa.String(length=255), nullable=True),
+        sa.Column("client_secret", sa.String(length=255), nullable=True),
+        sa.Column("signing_secret", sa.String(length=255), nullable=True),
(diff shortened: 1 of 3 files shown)


server/polar/models/benefit_slack_integration.py
```diff
@@ -0,0 +1,76 @@
...
+from polar.kit.db.models.base import RecordModel
+
+from .benefit import Benefit
+from .organization import Organization
+
+
+class BenefitSlackIntegration(RecordModel):
+    __tablename__ = "benefit_slack_integrations"
+
+    benefit_id: Mapped[UUID | None] = mapped_column(
```
frankie567: Now that I see it, I'm wondering if we should have the foreign key to `benefit_id`. Maybe we can reuse that for other means?

server/polar/models/benefit_slack_integration.py
```diff
@@ -0,0 +1,76 @@
...
+from sqlalchemy.orm import Mapped, declared_attr, mapped_column, relationship
+
+from polar.kit.db.models.base import RecordModel
+
+from .benefit import Benefit
+from .organization import Organization
+
+
+class BenefitSlackIntegration(RecordModel):
+    __tablename__ = "benefit_slack_integrations"
```
frankie567: Therefore, maybe call that `SlackApp`/`slack_apps`.

frankie567 (commented):
Going a bit backward to what I said previously, but since we have this full setup on its own, might be beneficial to extract it from the "benefits" world

server/polar/models/benefit_slack_integration.py
Yopi: Hmm, skip this foreign key all together you mean?

server/polar/models/benefit_slack_integration.py
Yopi: One idea was to be even more generic and call it `OrganizationApp` or something, but it might be nice to start in the specific area.

server/polar/models/benefit_slack_integration.py
frankie567: Yes, so this model is not necessarily tied to a benefit.

server/polar/models/benefit_slack_integration.py
frankie567: So you foresee other cases than Slack?

Yopi: I have tried to foresee others, but I haven't come up with any yet. So I think it makes sense to make this specific for now, and then when we implement a second one we can generalize it and rename this one. I think that's probably the best course of action

server/polar/models/benefit_slack_integration.py
Yopi: Makes sense, and I agree with you. I'll circle back to this

Yopi: Updated

frankie567 (approved):

PR #12209, opened 2026-06-05
Support combined product/discount/trial update

### LINKED ISSUE #12094: Allow to pass `discount_id` or `trial_end` in `SubscriptionUpdateProduct`

frankie567:
Fix #12094

> [!NOTE]
> I recommend to review it commit by commit

## Commit 1

Tweak the API to allow to update the product, discount and trial at the same time. Basically, we merged the three schemas into one.

## Commit 2

This is a bit more complex. The problem when we allow to update a product _and_ discount at the same time is that it impacts **prorations** :bufo-oh-no:

The solution in this case is to compute the credit proration with the previous discount and the debit proration with the new discount. Basically, we "refund" the previous discount and we "apply" the new discount.

This required a bit of refactor because discount was then required to be part of SubscriptionUpdate. This is why it's now recorded in the dedicated table.

---
## Summary by cubic
Enable combined product, discount, and trial updates in a single API call with correct prorations. Discounts are validated against the target product and saved on `SubscriptionUpdate` for charge previews and next-cycle processing (POL-12094).

- **New Features**
  - Unified `SubscriptionUpdateBase` to update `product_id`, `discount_id` (set to `null` to remove), `trial_end`, and `proration_behavior` together; updated API client and web UIs.
  - Persist pending discount via `discount_id`/`discount_unset` for `next_period`; applied in previews and end-of-cycle.

- **Bug Fixes**
  - Correct proration for product+discount changes: credit uses the previous discount; debit uses the new discount.
  - Validate discount against the new product when switching; web UI fixes for optional selects and “now” trial end; expanded tests for combined updates and proration cases.

<sup>Written for commit ca4409b55dff0f79a3782957035f026154f47ba2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12209?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/polar/subscription/schemas.py
```diff
@@ -269,8 +269,9 @@ class SubscriptionCreateExternalCustomer(SubscriptionCreateBase):
 SubscriptionCreate = SubscriptionCreateCustomer | SubscriptionCreateExternalCustomer
 
 
-class SubscriptionUpdateProduct(Schema):
-    product_id: UUID4 = Field(
+class SubscriptionUpdateBase(Schema):
```
frankie567, line 272: Not too happy with that name. Open to suggestions

server/polar/subscription/schemas.py
```diff
@@ -281,26 +282,38 @@ class SubscriptionUpdateProduct(Schema):
...
+    @property
+    def has_trial_end(self) -> bool:
+        return "trial_end" in self.model_fields_set
+
+    @property
+    def discount(self) -> UUID4 | Literal["unset"] | None:
+        if self.discount_id is not None:
+            return self.discount_id
+        if "discount_id" in self.model_fields_set:
+            return "unset"
```
frankie567, line 314: We have an internal helper value "unset" to carry on the information the user wanted to remove the discount.

server/polar/subscription/schemas.py
pieterbeulque: Just `SubscriptionUpdate`? `SubscriptionUpdateShared`?

server/polar/subscription/service.py
```diff
@@ -774,6 +775,11 @@ async def cycle(
                     )
                 else:
                     pending_update.product = None
+                if pending_update.discount_id is not None:
```
pieterbeulque, line 778: I'm probably misunderstanding but should you check for `unset` here?

frankie567: No, actually, this is a quite shitty way of loading the relationship (`pending_update.discount`). I think I'll revisit that in a follow-up PR.

server/polar/subscription/schemas.py
frankie567: Problem is `SubscriptionUpdate` is already the union of all sub-schemas 🤔 

@stilla-bot What would you suggest?

frankie567: @stilla-bot No but `SubscriptionUpdate` is the following:

```py
SubscriptionUpdate = Annotated[
    SubscriptionUpdateBase
    | SubscriptionUpdateSeats
    | SubscriptionUpdateBillingPeriod
    | SubscriptionCancel
    | SubscriptionRevoke
    | SubscriptionUpdateClear,
    SetSchemaReference("SubscriptionUpdate"),
]
```

That's why we need a specific name like `SubscriptionUpdateBase`. Suggest better alternatives.

pieterbeulque (approved):

PR #12211, opened 2026-06-05
fix creating checkout link or benefit showing stale empty state

strandhvilliam:
## Summary

https://github.com/user-attachments/assets/d491452a-57d4-40f9-bfc6-ea37d55e0e4d

https://github.com/user-attachments/assets/2d7d10e1-7398-4fdb-86c6-000fa279815a

Small fix for post-create navigation on checkout links and benefits so newly created items open automatically, without a full page reload or failed client-side routing.

Before when starting from an empty state and creating a new checkout link, the router didn’t navigate correctly and the page stayed on the empty state until the item was clicked in the sidebar. I’ve seen similar issues with nuqs racing with other navigations before.

One possible fix would have been to await setProductIds([]) (as it returns a promise)ƒ, but I instead included the query string directly in router.push. That fixed the issue and now the new item opens automatically. It also fixes creating additional items, which before left the view on the previously visible item.

This wasn’t an issue in the Benefits component because it used window.location.href, but that caused a full page flash. I added the same client-side navigation behavior there as well. (can be seen in _before_ video)

Please let me know if I missed anything or if this does not look good to you. See videos for before and after.

---
## Summary by cubic
Fixes post-create navigation for checkout links and benefits so the newly created item opens immediately via client-side routing, avoiding stale empty states and full page reloads.

- **Bug Fixes**
  - Checkout links: use `usePushRouteWithoutCache` on modal close to navigate to the new link, preserving query params while removing `productId`. Prevents `nuqs`/routing race and stale sidebar selection.
  - Benefits: replace `window.location.href` with `usePushRouteWithoutCache` and remove `query`/`create_benefit` before navigating to the new benefit. Eliminates page flash and ensures smooth client-side routing.

<sup>Written for commit 8745b202f26788995ad860f7e5a4d8908223c64a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12211?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/apps/web/src/components/Benefit/BenefitListSidebar.tsx
+++ b/clients/apps/web/src/components/Benefit/BenefitListSidebar.tsx
@@ -10,6 +10,7 @@ import { useModal } from '@/components/Modal/useModal'
 import Spinner from '@/components/Shared/Spinner'
 import { useInfiniteBenefits } from '@/hooks/queries'
 import { useInViewport } from '@/hooks/utils'
+import { usePushRouteWithoutCache } from '@/utils/router'
 import AddOutlined from '@mui/icons-material/AddOutlined'
 import ArrowDownward from '@mui/icons-material/ArrowDownward'
 import ArrowUpward from '@mui/icons-material/ArrowUpward'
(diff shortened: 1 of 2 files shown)


pieterbeulque (approved):
fastest approve ever

sebastianekstrom (approved):
Amazing 🙌

PR #12212, opened 2026-06-05
Change OTA policy to appVersion instead of fingerprint

sebastianekstrom:
The `fingerprint` policy for OTA updates is fragile and prone to false positives, this PR changes the policy to `appVersion` instead.

It also adds a guard rail to prevent us from shipping _native_ code/dependencies as an OTA. It's quite hacky, and hard to test properly since everything in Expo uses `fingerprint` as a policy right now. But I'll test it properly after we pushed a new build (with `appVersion` as policy)

---
## Summary by cubic
Switches OTA runtime policy to `appVersion` and adds a preflight guard to stop native changes from shipping as OTAs. Improves safety and makes it clear when a new native build is required.

- **New Features**
  - Set `runtimeVersion.policy` to `appVersion` in `app.config.js`.
  - Added `pnpm ota` script with a preflight that checks the latest finished build per platform on the target channel, compares native fingerprints (ignores pnpm virtual-store noise), blocks when no matching runtime build exists or when native drift is detected, supports `--check-only`, then forwards to `eas update`.
  - Updated README with the new workflow and added ESLint settings for `tooling/**` CommonJS scripts.

- **Migration**
  - Publish with `pnpm ota --channel <name> --message "<desc>"` instead of raw `eas update`.
  - If blocked, bump `version` in `app.config.js`, build and submit a new binary, then OTA JS-only fixes.
  - Use `pnpm ota --check-only` to validate without publishing.

<sup>Written for commit e13ce80212fcaf8dd02649c35336d36d6d55f730. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12212?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/apps/app/README.md
+++ b/clients/apps/app/README.md
@@ -74,17 +74,27 @@ This is how `expo-updates` works by default: it downloads updates in the backgro
 
 ### Publishing an OTA update
 
+Publish through `pnpm ota`, **not** raw `eas update`. It runs a preflight guard
+(`tooling/ota-preflight/`) that refuses to publish if the update can't safely
+reach devices then forwards to `eas update` for you:
+
 ```bash
-# Test on preview first
-eas update --channel preview --message "Description of changes"
+# Test on preview first if it's a larger change
+pnpm ota --channel preview --message "Description of changes"
 
 # Then push to production
-eas update --channel production --message "Description of changes"
+pnpm ota --channel production --message "Description of changes"
 ```
 
-### Runtime version
+The guard verifies, per platform, against the latest finished build on the channel:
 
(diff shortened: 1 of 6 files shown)


strandhvilliam (approved):
Think this is very smart, especially handling the pnpm peer deps hash stuff!

Only thing I was thinking about is if someone forgets about `pnpm ota` , and uses `eas update` and putting it all in a bad state.

Maybe some way to block this? A shim script or so? Might be overkill and not a real problem idk.

PR #12223, opened 2026-06-08
Create fixed seat count checkout links

pieterbeulque:
Add support to pre-configure # of seats on a checkout link. The use case is to send a customer-specific checkout link for a pre-agreed upon number of seats. 

I decided to implement `seats` only and make it a fixed / locked number of seats. It's forwards-compatible with eventually adding min seats & max seats if we eventually want to do that, but I don't really see the need for it right now.

---
## Summary by cubic
Add fixed seat count support to checkout links so you can send customer-specific links with a locked number of seats. The lock applies only when all products are seat-based and the count fits each product’s tiers; if tiers drift by the time the link is opened, the lock is ignored and checkout falls back to the tier minimum.

- **New Features**
  - API: optional `seats` on checkout links (create/update/read) with strict validation across all products; product-only changes that invalidate a lock auto-clear it.
  - Checkout: when valid, sessions set seats with min=max=seats; on tier drift, fall back to the tier minimum and keep unlocked.
  - Web: “Lock number of seats” toggle and input shown only when all selected products are seat-based; validates against intersected tier bounds and auto-clears when incompatible or when products aren’t all seat-based.
  - SDK: update `@polar-sh/client` types to include `seats`.

- **Migration**
  - Adds nullable `seats` column to `checkout_links`. Run DB migrations.

<sup>Written for commit 88fdc96b61ad06b3419fe3a83b7fdc39b53cf812. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12223?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/models/checkout_link.py
+++ b/server/polar/models/checkout_link.py
@@ -40,6 +40,8 @@ class CheckoutLink(TrialConfigurationMixin, MetadataMixin, RecordModel):
         Boolean, nullable=False, default=False
     )
 
+    # Preconfigured seat count for seat-based pricing. When set, checkout sessions
+    # created from this link are locked to this number of seats (min == max == seats).
     seats: Mapped[int | None] = mapped_column(Integer, nullable=True, default=None)
 
     discount_id: Mapped[UUID | None] = mapped_column(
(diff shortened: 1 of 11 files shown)


server/polar/models/checkout_link.py
```diff
@@ -40,6 +40,10 @@ class CheckoutLink(TrialConfigurationMixin, MetadataMixin, RecordModel):
         Boolean, nullable=False, default=False
     )
 
+    # Preconfigured seat count for seat-based pricing. When set, checkout sessions
+    # created from this link are locked to this number of seats (min == max == seats).
```
frankie567, line 44: Verbosy LLM comment

frankie567 (approved):

PR #12224, opened 2026-06-08
handbook: API versioning RFC

frankie567:
Proposal to define our API versioning strategy.

> [!IMPORTANT]
> This is a quite aggressive proposal, on purpose. I would love for us to set a bar when we can provide some stability guarantees to our users, without compromising our ability to iterate and improve our API quickly. Let's discuss and find the right balance.

## [📖 Read Preview](https://polar-handbook-rfc-api-versioning.mintlify.app/engineering/design-documents/api-versioning)

---
## Summary by cubic
Adds an API versioning RFC proposing quarterly, date-based versions (`YYYY-MM`), maintaining three versions (~9 months) with a `Polar-Version` header defaulting to Current. Includes a backend POC for header-based routing and per‑version OpenAPI with schema‑lock tests, and revamps the SDK to versioned submodules (e.g., `v2026_07`) with SemVer and auto‑set headers, plus a concrete rollout plan.

<sup>Written for commit cf7c61ca1cdc0c2229b38f9739a670f09a74c563. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12224?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/handbook/engineering/design-documents/api-versioning.mdx
+++ b/handbook/engineering/design-documents/api-versioning.mdx
@@ -164,11 +164,44 @@ When the `Polar-Version` header is omitted, requests are routed to the **Current
 
 ## SDK Strategy
 
-The SDK will follow the API versioning with patch versions for bug fixes:
+The SDK will support the three supported versions at any time, through submodules. For example:
 
-- **Format**: `2026-01.0`, `2026-01.1`, `2026-04.0`, etc.
-- **Support**: Exactly 3 SDK versions maintained, matching the API policy
-- **Versioning**: Patch versions (`2026-01.1`) include only bug fixes; major versions (`2026-04`) include contract changes
+```
+sdk/
+├── v2026_04/
+│   ├── __init__.py
+│   ├── client.py
+│   ├── models.py
+│   └── services/
+│       └── customers.py
+├── v2026_07/
+│   ├── __init__.py
+│   ├── client.py
+│   ├── models.py
+│   └── services/
+│       └── customers.py
+├── v2026_10/
+│   ├── __init__.py
+│   ├── client.py
+│   ├── models.py
+│   └── services/
+│       └── customers.py
+└── README.md
+```
+
+The user would use the client by importing the version they want:
+
+```py
+from polar.v2026_07 import Polar
+```
+
+Under the hood, this version would automatically use the schemas and services corresponding to that version and automatically set the version header when issuing requests. The good side effect is that it forces users to explicitly set a version.
+
+### Versioning strategy
+
+- Bug fixes, performance improvements to the generated code: **Patch** version
+- Added features, updates to the **Next** API version (models or endpoints): **Minor** version
+- API version cycle: **Major** version. Since we're removing a version, it's a breaking change — users will need to change their import.
 
 ## Implementation
 
@@ -414,6 +447,24 @@ We should have tests that lock the schema for N-1 and N versions to ensure no co
 ### Monitoring
 
 We should track adoption of each version and monitor for errors. After one or two release cycles, we can evaluate if the cadence and support window are working as intended or if adjustments are needed.
+
+### Rollout Plan
+
+#### ASAP
+
+- Start to tag API version in OpenAPI file.
+- Continue to work as we do today, i.e. "best-effort" backward compatibility
+
+#### Before September 2026
+
+- Implement the API versioning logic in the server
+- Have the SDK working as described in [SDK Strategy](#sdk-strategy)
+
+#### October 2026
+
+- **Freeze** the API as it is under the version `2026-10`
+- Start the cycles by setting up the version `2027-01` and start to work on this version only
+- Release the SDK with those two versions
 
 ## Appendix: Version Timeline Examples
 
--- a/handbook/engineering/design-documents/api-versioning.mdx
+++ b/handbook/engineering/design-documents/api-versioning.mdx
@@ -46,39 +46,34 @@ Releases occur on a **quarterly cadence** during the **first week of each quarte
 
 The release is **calendar-driven**: it happens on the scheduled date regardless of feature completion. Incomplete work is deferred to the Next version.
 
-<Note>
-  The release cadence (quarterly vs. semi-annual) is open for discussion. See
-  [Alternatives Considered](#alternatives-considered).
-</Note>
-
 ### Version Lifecycle
 
-Exactly **3 versions** are maintained at any time:
+Exactly **4 versions** are maintained at any time:
 
 ```
-[Deprecated] → [Current: default] → [Next: in development]
+[N-1: deprecated] → [N: default] → [Next: in development]
 ```
 
 #### State Transitions
 
 On each release:
 
-1. The **oldest version (Deprecated)** is **removed**
-2. The **Current version** becomes **Deprecated**
-3. The **Next version** becomes **Current** (and default)
+1. **N-1** is **removed**
+2. **N** becomes **N-1** and is **deprecated**
+3. The **Next version** becomes **N** and is the new default. It is **frozen** for contract changes.
 4. A **new Next version** is created for ongoing development
 
 #### Timeline Example
 
 ```
 Before Q2 2026 release:
-  2025-10  (Deprecated, to be removed)
-  2026-01  (Current, default)
+  2025-10  (N-1, to be removed)
+  2026-01  (N, default)
   2026-04  (Next, in development)
 
 After Q2 2026 release:
-  2026-01  (Deprecated, to be removed on next release)
-  2026-04  (Current, default) ← was Next
+  2026-01  (N-1, to be removed on next release)
+  2026-04  (N, default) ← was Next
   2026-07  (Next, in development)
   (2025-10 is now removed)
 ```
@@ -91,8 +86,9 @@ Each version follows this lifecycle:
 - **Total**: ~9 months of support
 
 <Note>
-  The support window duration (9 months vs. 12 months) is open for discussion.
-  See [Alternatives Considered](#alternatives-considered).
+  If this is seen as too aggressive, we can consider a longer support window
+  (e.g., 12 months) by maintaining 4 versions instead of 3. See Appendix for
+  timeline examples.
 </Note>
 
 ### Change Policy
@@ -120,6 +116,8 @@ Each version follows this lifecycle:
 3. New contract changes target the new Next version
 4. Bug fixes can be applied to any maintained version
 
+Polar dashboard and mobile app will always use the **Next** version, so it can benefit from latest features immediately. Breaking changes are less disruptive for internal clients since we control both sides of the contract.
+
 ### Version Increment
 
 The version **always increments on schedule**, even if no contract changes have been made in that quarter.
@@ -129,11 +127,6 @@ The version **always increments on schedule**, even if no contract changes have 
 ### Removal Policy
 
 Deprecated versions are **removed on the next release date** (hard cutoff).
-
-<Note>
-  Adding a grace period (e.g., 30 days after deprecation window) is open for
-  discussion.
-</Note>
 
 ### Version Naming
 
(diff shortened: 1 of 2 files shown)


handbook/engineering/design-documents/api-versioning.mdx
```diff
@@ -0,0 +1,308 @@
...
+**Comments**: All aspects of this proposal are open for debate, particularly release cadence and support window duration.
+
+</Info>
+
+## Abstract
+
+This RFC proposes a date-based API versioning strategy for Polar with the following characteristics:
+
+- **Version lifecycle**: 3 versions maintained with ~9-month support window — open for discussion
+- **Release cadence**: Quarterly (January, April, July, October) — open for discussion
```
Yopi, line 20: I think it would make sense out of a promise perspective that we can communicate that this is the ambition, but we can do it more often (or more seldom) depending on our needs.

pieterbeulque:
I like it! Two random thoughts:
 
- First week of January feels like a weird week to launch a new API version with the holidays before them. Detail.
- I don't understand how we can deprecate an API field. Can future deprecations be published to current as patch updates, so that they can be dropped in the next release, or do you need one release cycle to update the field to mark as deprecated, then another release cycle to actually remove the field?

handbook/engineering/design-documents/api-versioning.mdx
```diff
@@ -0,0 +1,308 @@
...
+- **Change policy**: Strict freeze on Current and Deprecated versions; all contract changes through Next version
+- **Version naming**: `YYYY-MM` format (e.g., `2026-01`, `2026-04`)
+- **Versioning mechanism**: `Polar-Version` header, defaulting to Current version
+
```
Yopi, line 24: I've thought a lot about this in the past, and I also think like this is a good versioning scheme, the only thing is that we don't have the ability to do multiple releases a month if we would ever need it.

frankie567:
> * First week of January feels like a weird week to launch a new API version with the holidays before them. Detail.

Yeah, I guess we can either tweak the anchor or give us some room to allow for week 2.

> * I don't understand how we can deprecate an API field. Can future deprecations be published to current as patch updates, so that they can be dropped in the next release, or do you need one release cycle to update the field to mark as deprecated, then another release cycle to actually remove the field?

The way I see it, the deprecation is at the version level. If you want to drop a field, you do it in `Next`, without prior notice. When we cycle versions, the field is removed in `Current` — but still available in `Deprecated`.

`Current` and `Deprecated` should really be seen as frozen and shouldn't accept changes.

handbook/engineering/design-documents/api-versioning.mdx
```diff
@@ -0,0 +1,308 @@
...
+The goal is to establish a predictable, disciplined API evolution process that balances innovation with stability.
+
+## Motivation
+
+### Problems with Current Approach
+
+Currently, Polar lacks a formal API versioning strategy. Our current approach is "best-effort", where we _try_ to limit breaking changes and maintain backward compatibility, but we have no guarantees or clear policies. This leads to several issues:
```
Yopi, line 31: I think it would make sense to add a definition of what a breaking change is so that we all know, and our customers also know

handbook/engineering/design-documents/api-versioning.mdx
```diff
@@ -0,0 +1,308 @@
...
+
+- No strong guarantees for consumers about stability and support duration
```
joebon, line 33: Have we had feedback on this "lack of stability" from our merchants?

handbook/engineering/design-documents/api-versioning.mdx
```diff
@@ -0,0 +1,308 @@
...
+[Deprecated] → [Current: default] → [Next: in development]
+```
+
+#### State Transitions
+
+On each release:
+
+1. The **oldest version (Deprecated)** is **removed**
+2. The **Current version** becomes **Deprecated**
+3. The **Next version** becomes **Current** (and default)
```
Yopi: I think it would be better to remove the oldest version either on each release -if more than 9 months have passed- or multiple versions are deprecated.

handbook/engineering/design-documents/api-versioning.mdx
```diff
@@ -0,0 +1,308 @@
...
+
+- **Format**: `2026-01.0`, `2026-01.1`, `2026-04.0`, etc.
+- **Support**: Exactly 3 SDK versions maintained, matching the API policy
+- **Versioning**: Patch versions (`2026-01.1`) include only bug fixes; major versions (`2026-04`) include contract changes
+
+## Implementation
+
+### Backend: Header-based routing POC
+
+A way to implement this is to consider each version as a separate FastAPI app. The main app routes requests to the appropriate version based on the `Polar-Version` header. In each business logic module (e.g. `customers`), we'll have a submodule for each version, containing the schemas and endpoints for that version. Common logic, like `service.py`, can be shared across versions.
```
Yopi: I think we should consider a way to do this without duplicating the entire app. I.e. if we have made no changes to a route in an API version, which might happen, it feels unnecessary to duplicate it into a new folder.

Instead if we can have a "default" route or similar for routes that haven't changed, if we can do it in a way that feels neat.

Yopi (commented):
I really like it, good suggestions François. Added a few comments on different areas I think we should clarify

handbook/engineering/design-documents/api-versioning.mdx
```diff
@@ -0,0 +1,308 @@
...
+The version **always increments on schedule**, even if no contract changes have been made in that quarter.
+
+**Rationale**: Maintaining a regular cadence with predictable version numbers simplifies planning and avoids gaps in the version sequence.
+
+### Removal Policy
+
+Deprecated versions are **removed on the next release date** (hard cutoff).
+
+<Note>
+  Adding a grace period (e.g., 30 days after deprecation window) is open for
```
joebon: Hrm, what would be the point of that? It's deprecated for 9 months and then _really_ deprecated for an additional 30 days?

handbook/engineering/design-documents/api-versioning.mdx
joebon: Nice, this feels like a reasonable containment and helps us guard rail our architecture in a nice way IMHO.

handbook/engineering/design-documents/api-versioning.mdx
```diff
@@ -0,0 +1,308 @@
...
+
+1. **Version selector**: Allow users to view docs for each maintained version
+2. **Version indicators**: Clearly mark deprecated versions in documentation
+3. **Changelog**: Maintain per-version changelog
+4. **Migration guides**: Provide guides for migrating between versions
+
+### Monitoring
+
+1. Track usage of each version
+2. Alert when deprecated version usage exceeds thresholds
```
joebon: What thresholds could that be?

handbook/engineering/design-documents/api-versioning.mdx
```diff
@@ -0,0 +1,308 @@
...
+
+**Proposal**: Maintain 4 versions with 12-month support window.
+
+```
+Lifecycle: Next (3 mo) → Current (6 mo) → Deprecated (3 mo) → Removed
+Total: 12 months support
+Versions maintained: 4
+```
+
+**Tradeoff**: Longer support window improves enterprise compatibility but increases maintenance burden.
```
joebon: IMHO we don't have that enterprise merchant base at the moment, I'd suggest to not take on such burden until expressively necessary.

handbook/engineering/design-documents/api-versioning.mdx
frankie567: That's the thing I guess: we can do all our changes live in the `Next` version. Even breaking ones.

handbook/engineering/design-documents/api-versioning.mdx
```diff
@@ -0,0 +1,308 @@
...
+```
+
+#### State Transitions
+
+On each release:
+
+1. The **oldest version (Deprecated)** is **removed**
+2. The **Current version** becomes **Deprecated**
+3. The **Next version** becomes **Current** (and default)
+4. A **new Next version** is created for ongoing development
```
joebon, line 64: We can start consuming the `Next` version ourselves in the Dashboard, customer portal etc, correct? So we won't ourselves need to wait up to 3 months to start consuming new stuff.

handbook/engineering/design-documents/api-versioning.mdx
Yopi: That's fair, yes 😄. I like it!

joebon (commented):
Some initial thoughts and questions

handbook/engineering/design-documents/api-versioning.mdx
frankie567: In this approach, any change to a schema or endpoint is breaking: no change are allowed in `Deprecated`/`Current`. `Next` can do anything.

handbook/engineering/design-documents/api-versioning.mdx
frankie567: Well, I think this is already quite tight: basically, an endpoint or field can disappear in only 6 months.

handbook/engineering/design-documents/api-versioning.mdx
frankie567: Yes, that'll need some digging

handbook/engineering/design-documents/api-versioning.mdx
frankie567: Yes, in particular with the SDK. Marc Lou on Twitter has been particularly vocal: https://x.com/marclou/status/2026235846567874782

handbook/engineering/design-documents/api-versioning.mdx
frankie567: Yeah, might not make sense indeed. Maybe better to support a full cycle if we want to bring more stability

handbook/engineering/design-documents/api-versioning.mdx
frankie567: Absolutely no idea :p I guess we'll first need to collect metrics when we start having several versions in parallel to see how long a version sticks in reality.

handbook/engineering/design-documents/api-versioning.mdx
frankie567: Exactly :) This is somehow how we work today: we introduce new API/schemas and start using them right away in our product.

handbook/engineering/design-documents/api-versioning.mdx
pieterbeulque: Does this mean that we build our dashboard against _Next_?

frankie567: Yes, I believe so. This is somewhat what we already do: we spin up endpoints and schemas when we need them

handbook/engineering/design-documents/api-versioning.mdx
joebon: Good! That was my biggest worry but then I understood it correctly and that makes a lot of sense + doesn't slow us down

handbook/engineering/design-documents/api-versioning.mdx
joebon: Yeah, if we can solve for that without introducing too much indirection that makes sense!

handbook/engineering/design-documents/api-versioning.mdx
Yopi: The comment associated a bit with the other ones, in regards of releasing more versions than next, if we were to have more versions than deprecated, current and next. In which case we'd probably want to have multiple deprecated versions living at the same time

handbook/engineering/design-documents/api-versioning.mdx
frankie567: I've revamped my copy here and propose a different way using `@version` decorator. Tell me what you think!

handbook/engineering/design-documents/api-versioning.mdx
```diff
@@ -0,0 +1,448 @@
...
+- **Support**: Exactly 3 SDK versions maintained, matching the API policy
+- **Versioning**: Patch versions (`2026-01.1`) include only bug fixes; major versions (`2026-04`) include contract changes
+
+## Implementation
+
+### Backend: Header-based routing POC
+
+We introduce a new `version` decorator to put on top of routes. By default, without the decorator, the route is available on all versions. If the decorator is used, the route is only available on the specified version(s). A more specific version takes precedence over a more general one.
+
+Custom routing classes will then take care to create sub-FastAPI apps that expose the right endpoints and schemas for each version. The main app will route requests to the appropriate sub-app based on the `Polar-Version` header.
```
Yopi, line 212: I think this is a very neat solution. Saves us having to copy over everything initially.

Yopi (approved):
I like it!

handbook/engineering/design-documents/api-versioning.mdx
```diff
@@ -0,0 +1,499 @@
...
+
+### Release Schedule
+
+Releases occur on a **quarterly cadence** during the **first week of each quarter** (January, April, July, October).
+
+The release is **calendar-driven**: it happens on the scheduled date regardless of feature completion. Incomplete work is deferred to the Next version.
+
+### Version Lifecycle
+
+Exactly **4 versions** are maintained at any time:
```
psincraian: I have the same question as cubic. And, can the users use the next verision? Can we make breaking changes on the next if it's not released?

handbook/engineering/design-documents/api-versioning.mdx
```diff
@@ -0,0 +1,499 @@
...
+
+#### Strict Freeze Rule
+
+**Deprecated and Current versions are completely frozen for contract changes.**
+
+| **Change Type**                          | **Deprecated** | **Current** | **Next** |
+| ---------------------------------------- | -------------- | ----------- | -------- |
+| Add endpoint                             | ❌             | ❌          | ✅       |
+| Remove endpoint                          | ❌             | ❌          | ✅       |
+| Modify schema (add/remove/change fields) | ❌             | ❌          | ✅       |
```
psincraian, line 104: even non breaking changes are only on the next version? 🤔

handbook/engineering/design-documents/api-versioning.mdx
psincraian: this looks really smart 👏

psincraian (approved):
That looks really neat!

handbook/engineering/design-documents/api-versioning.mdx
frankie567: Yeah, sorry, it's 3 versions :D 

Users can use the next version if they specify it in `Polar-Version`, but with no guarantee that it won't break.

handbook/engineering/design-documents/api-versioning.mdx
frankie567: Yes. IMO, trying to say which change is breaking or not can be quite tricky and even subjective. Adding an optional field is apparently non-breaking; but it still requires the client schema to be updated.

Having this freezing approach is easier, I think: no change allowed at all.

handbook/engineering/design-documents/api-versioning.mdx
psincraian: understood. But then in reality there is at most 6 months of support, which I think it's more than fine.

Issue #12205, opened 2026-06-05
Replace `free` `amount_type` with `fixed` + `amount: 0`


pieterbeulque:

pieterbeulque:
@frankie567 I assume this has your blessing?

frankie567:
Amen. A good one for @maximevast I think ;)

PR #12225, opened 2026-06-08
Allow 0 for fixed price products and treat them as a free price

joebon:
**Related issue**: https://github.com/polarsource/polar/issues/12205

## Migration plan

1. Make a `fixed` price of `0` behave identically to a `free` price everywhere    <-- _this step_
2. Stop producing `free`: Change the frontend “Free” option and the API to create `fixed`-`0` prices instead of `free` ones (while still _accepting_ `free` for back-compat), freezing the `free` population so no new rows are added.
3. Backfill: Convert all existing `free` (and `legacy_free`) rows to `fixed` with `price_amount = 0`.
4. Remove `ProductPriceFree` & `LegacyRecurringProductPriceFree`

---
## Summary by cubic
Treats a fixed price of 0 as a free price across the app, so zero-amount fixed prices behave like the old free type. This is the first step toward removing the dedicated free price type.

- **New Features**
  - `ProductPriceFixed.is_free` returns true when `price_amount == 0`.
  - Checkout `is_free_product_price` now checks `price.is_free`, so zero fixed prices skip the payment form.
  - Backoffice displays a zero fixed price as “Free” (excluding seat prices).
  - `ProductPriceFixedCreate` accepts `price_amount: 0` and documents zero as the free option.
  - Regenerated client to describe fixed `price_amount` as “Set to `0` for a free price”.

- **Migration**
  - Create free prices by setting fixed `price_amount` to `0`; no other changes needed.

<sup>Written for commit 6fb400867a157b1c876f14943adf4cc151526f61. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12225?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/models/checkout.py
+++ b/server/polar/models/checkout.py
@@ -33,7 +33,6 @@ from polar.kit.trial import TrialConfigurationMixin, TrialInterval
 from polar.kit.utils import utc_now
 from polar.product.guard import (
     is_discount_applicable,
-    is_free_price,
     is_metered_price,
 )
 from polar.tax.calculation import TaxBreakdownItem
@@ -322,7 +321,7 @@ class Checkout(
     def is_free_product_price(self) -> bool:
         if self.product_prices is None:
             return False
-        return all(is_free_price(price) for price in self.product_prices)
+        return all(price.is_free for price in self.product_prices)
 
     @property
     def has_metered_prices(self) -> bool:
--- a/clients/packages/client/src/v1.ts
+++ b/clients/packages/client/src/v1.ts
@@ -27123,7 +27123,7 @@ export interface components {
       tax_behavior?: components['schemas']['TaxBehaviorOption'] | null
       /**
        * Price Amount
-       * @description The price in cents.
+       * @description The price in cents. Set to `0` for a free price.
        *     Minimum amounts per currency:
        *     - USD: 0.5
        *     - AED: 2
--- a/server/polar/backoffice/products/endpoints.py
+++ b/server/polar/backoffice/products/endpoints.py
@@ -14,7 +14,6 @@ from polar.product import sorting
 from polar.product.guard import (
     is_custom_price,
     is_fixed_price,
-    is_free_price,
     is_metered_price,
     is_seat_price,
 )
@@ -30,7 +29,7 @@ router = APIRouter()
 
 def _format_price_display(price: ProductPrice) -> str:
     """Format a price for display based on its amount type."""
-    if is_free_price(price):
+    if price.is_free and not is_seat_price(price):
         return "Free"
     elif is_custom_price(price):
         parts = []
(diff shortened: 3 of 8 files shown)


server/polar/models/checkout.py
```diff
@@ -322,7 +321,7 @@ def is_discount_applicable(self) -> bool:
     def is_free_product_price(self) -> bool:
         if self.product_prices is None:
             return False
-        return all(is_free_price(price) for price in self.product_prices)
+        return all(price.is_free for price in self.product_prices)
```
frankie567, line 324: We need to keep `is_free_price` check as well if we want existing free prices to continue working

frankie567 (changes requested):

server/polar/models/checkout.py
joebon: Hrm, are you sure?
`price.is_free` should be `True` both for `ProductPriceFree`, `LegacyRecurringProductPriceFree` (because of [_ProductPriceFree.is_free](https://github.com/polarsource/polar/blob/main/server/polar/models/product_price.py#L306) and `ProductPriceFixed` (because of the addition on `_ProductPriceFixed.is_free` in this PR).
Am I missing something?

pieterbeulque:
Will we have a gap with the dashboard UI? In between two deploys, I mean.

server/polar/models/checkout.py
frankie567: Oh good catch, didn't remember it had the method as well!

joebon:
> Will we have a gap with the dashboard UI? In between two deploys, I mean.

@pieterbeulque the 4 step rollout/migration is _meant_ to avoid that and other issues. The idea:
1. Modify/relax fixed prices so zero prices are allowed => same behavior as the _free_ price products
2. When creating "free" prices in the Dashboard UI (and API), _actually_ create fixed prices with price set to `0` instead of the dedicated `free` type
3. Backfill/convert all existing `free` prices to `fixed` prices with price `0`. This should then be safe because behavior alignment in step 1 and ensuring only the latter variant is produced going forward in step 2
4. Drop models `ProductPriceFree` & `LegacyRecurringProductPriceFree` and related constructs from the code base

WDYT?

pieterbeulque (approved):
Sorry, I misread, maybe because of the branch name. Sounds good!

It assume this will still reject values in the gap between $0 & $0.50.

frankie567 (approved):

PR #12249, opened 2026-06-08
slack: Add shared channel benefit

Yopi:
Extracted out from #12188

---
## Summary by cubic
Adds a Slack Shared Channel benefit that provisions customer-specific channels with Slack-safe templates and a channel-name preview API. Creation now requires selecting a Slack integration on the benefit (integration ID is required), gated by an org flag, with safer grant/update flows and smarter channel lifecycle.

- **New Features**
  - New benefit type: `slack_shared_channel` (schemas, grant properties, service, customer portal; `slack_integration_id` is required `UUID4`).
  - API: POST `/v1/benefits/slack/preview-channel-name` (scope `benefits:write`) to render a channel name; returns 422 on invalid templates.
  - Templates support `{customer_name}`, `{customer_email_local}`, and `{metadata.<key>}` with Slack-safe formatting (max 80 chars); dots in `<key>` aren’t traversed.
  - Require a Slack integration when creating the benefit; removed `/v1/integrations/slack/link`. Regenerated `@polar-sh/client` and updated server routes (`/benefits/slack`).
  - UI/i18n: hide this type unless `organization.feature_settings.slack_benefit_enabled`; added `benefitTypes.slack_shared_channel` with translations and grant-view mapping. Server also blocks creation if the flag is disabled.

- **Bug Fixes**
  - Preserve `grant.properties` when `BenefitActionRequiredError` occurs on grant/update (supports passing properties via the exception).
  - Customer portal: allow changing `invited_email` and re-send the invite.
  - Prevent benefit updates from unlinking Slack apps; handle transient vs non-recoverable errors safely.
  - Provisioning improvements: use a Polar lock, reuse a shared Slack client, try channel creation before search, reuse sibling grants, and don’t resend the welcome message when attaching to existing channels.
  - Only archive a channel on revoke if no other grant references it.

<sup>Written for commit dcd745153a8cd2b2bdcaa00d56b97be16ea26efe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12249?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/benefit/strategies/slack_shared_channel/schemas.py
+++ b/server/polar/benefit/strategies/slack_shared_channel/schemas.py
@@ -1,0 +1,84 @@
+from typing import Annotated, Literal
+
+from pydantic import Field, StringConstraints
+from pydantic.json_schema import SkipJsonSchema
+
+from polar.kit.schemas import Schema
+from polar.models.benefit import BenefitType
+
+from ..base.schemas import (
+    BenefitBase,
+    BenefitCreateBase,
+    BenefitSubscriberBase,
+    BenefitUpdateBase,
+)
+
+ChannelNameTemplate = Annotated[str, StringConstraints(min_length=1, max_length=80)]
+WelcomeMessage = Annotated[str | None, StringConstraints(max_length=4000)]
+
+
+class BenefitSlackSharedChannelProperties(Schema):
+    slack_integration_id: str | None = Field(
+        default=None,
+        description="Polar Slack integration ID linked to this benefit, if any.",
+    )
+    channel_name_template: ChannelNameTemplate = Field(
+        description=(
+            "Template for the channel name. Supports placeholders: "
+            "{customer_name}, {customer_email_local}, and {metadata.<key>} "
+            "for any value stored in customer user metadata."
+        ),
+    )
+    private: bool = Field(
+        default=True,
+        description="Create the channel as private (recommended).",
+    )
+    welcome_message: WelcomeMessage = Field(
+        default=None,
+        description="Optional message posted to the channel right after creation.",
+    )
+    archive_on_revoke: bool = Field(
+        default=True,
+        description="Archive the channel when the benefit is revoked.",
+    )
+    team_invitees: list[str] = Field(
+        default_factory=list,
+        description=(
+            "Slack user IDs from the merchant workspace to invite to every "
+            "channel created for this benefit."
+        ),
+    )
+
+
+class BenefitSlackSharedChannelCreateProperties(Schema):
+    slack_integration_id: str | None = None
+    channel_name_template: ChannelNameTemplate
+    private: bool = True
+    welcome_message: WelcomeMessage = None
+    archive_on_revoke: bool = True
+    team_invitees: list[str] = Field(default_factory=list)
+
+
+class BenefitSlackSharedChannelCreate(BenefitCreateBase):
+    type: Literal[BenefitType.slack_shared_channel]
+    properties: BenefitSlackSharedChannelCreateProperties
+
+
+class BenefitSlackSharedChannelUpdate(BenefitUpdateBase):
+    type: Literal[BenefitType.slack_shared_channel]
+    properties: BenefitSlackSharedChannelCreateProperties | None = None
+
+
+class BenefitSlackSharedChannel(BenefitBase):
+    type: Literal[BenefitType.slack_shared_channel]
+    properties: BenefitSlackSharedChannelProperties
+    is_tax_applicable: SkipJsonSchema[bool] = Field(deprecated=True)
+
+
+class BenefitSlackSharedChannelSubscriberProperties(Schema):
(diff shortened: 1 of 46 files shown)


server/polar/benefit/strategies/slack_shared_channel/template.py
```diff
@@ -0,0 +1,143 @@
...
+    customer_name: str
+    customer_email_local: str
+    metadata: dict[str, Any] = field(default_factory=dict)
+
+
+class InvalidTemplateError(ValueError):
+    pass
+
+
+def validate_template(template: str) -> None:
```
frankie567, line 27: Given that it's pure logical, sans-IO validation, I think we could apply it directly at schema level on `BenefitSlackSharedChannelCreate` and `BenefitSlackSharedChannelUpdate`. This would avoid the need to introduce `validate_properties_update` on the protocol.

server/polar/benefit/strategies/slack_shared_channel/template.py
```diff
@@ -0,0 +1,143 @@
...
+
+def _malformed_placeholder(placeholder: str) -> InvalidTemplateError:
+    return InvalidTemplateError(
+        f"Malformed placeholder: {placeholder}. "
+        f"Allowed: {{customer_name}}, {{customer_email_local}}, "
+        f"or {{metadata.<key>}} for customer metadata."
+    )
+
+
+def _slugify(value: str) -> str:
```
frankie567: We have the slugify library at hand to offload that logic

server/polar/integrations/slack/service.py
frankie567: I don't really understand when this flow is called. It looks to me that we just check the existence of the Slack integration and if it's free before wiring it to the Benefit.

From my (partial) understanding, I would say it's actually a great candidate to leverage the `validate_properties` of the benefit service:

* You make `integration_id` a required property in `BenefitSlackSharedChannelCreateProperties`.
* You validate it's existing/available in `validate_properties`
* -> You get a Benefit with `integration_id` clean and ready to use in one call

frankie567 (commented):
Huge work 👏 

My main concern/interrogation is how we validate the benefit properties, in particular the `integration_id`. It looks to me we could streamline this with the structure we have already.

server/polar/benefit/strategies/slack_shared_channel/template.py
Yopi: Good idea!

server/polar/integrations/slack/service.py
Yopi: I think it's doable, but we need to re-arrange the setups and flow in that case. Right now we only allow one benefit to be attached to one slack app, and inside of `validate_properties` we don't know if we are updating the same benefit or updating a different benefit and trying to attach it to the same app

frankie567: Is there a reason to limit the number of benefits attached to a Slack app?

Yopi: There is no external reason per se. I was thinking that there might be a risk that we end up in a weird state or that you unnecessarily break one benefit by updating the slack app connected to another. If we enforce 1:1 then there is no risk in any of it.

So I think that we can loosen that requirement

frankie567: Yep, if it allows to do a more streamlined benefit creation flow, I would prefer to drop that constraint.

Yopi: Followed your suggestion here https://github.com/polarsource/polar/pull/12249/changes/4279c4bbf23f4c258a5b903a46b18d34f763f821

server/polar/benefit/strategies/slack_shared_channel/schemas.py
```diff
@@ -0,0 +1,96 @@
...
+
+ChannelNameTemplate = Annotated[str, StringConstraints(min_length=1, max_length=80)]
+ValidatedChannelNameTemplate = Annotated[
+    ChannelNameTemplate, AfterValidator(_validate_channel_name_template)
+]
+WelcomeMessage = Annotated[str | None, StringConstraints(max_length=4000)]
+
+
+class BenefitSlackSharedChannelProperties(Schema):
+    slack_integration_id: UUID4 | None = Field(
```
frankie567: I guess it's non nullable now, then? :)

frankie567 (approved):
Great!

server/polar/benefit/strategies/slack_shared_channel/schemas.py
Yopi: Haha, yes, thanks!

PR #12266, opened 2026-06-09
tf: Add module for the AWS task worker

Yopi:
This sets up the infra that will connect AWS queues to lambdas.

Not used today

---
## Summary by cubic
Adds a reusable Terraform module for an AWS task worker: an SQS queue with DLQ wired to a container-image Lambda via event source mapping. Not used yet; no behavior change.

- **New Features**
  - SQS task queue + DLQ with visibility timeout >= 180s and timeout+60 headroom; redrive capped at 5.
  - Lambda (Image) with IAM role, log group, env vars (incl. secrets) and optional VPC; sets POLAR_DATABASE_POOL_SIZE=1.
  - SQS→Lambda event source mapping with batch size, enabled toggle, and batch item failure reporting.
  - Tuning: timeout, memory, reserved concurrency, batch size, retries, log retention, tags.
  - Outputs for CI: queue URL/ARN, DLQ ARN, function name/ARN, Lambda role ARN.

<sup>Written for commit 032591b9729ef2e357c099d6aab649b92993ae91. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12266?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/terraform/modules/aws_task_worker/main.tf
+++ b/terraform/modules/aws_task_worker/main.tf
@@ -1,0 +1,130 @@
+locals {
+  in_vpc = length(var.subnet_ids) > 0
+}
+
+resource "aws_sqs_queue" "dlq" {
+  name                      = "${var.queue_name}-dlq"
+  message_retention_seconds = 1209600
+
+  tags = var.tags
+}
+
+resource "aws_sqs_queue" "task" {
+  name = var.queue_name
+
+  # Visibility must exceed the function timeout so a slow task is not redelivered while still running.
+  visibility_timeout_seconds = max(180, var.timeout_seconds + 60)
+
+  redrive_policy = jsonencode({
+    deadLetterTargetArn = aws_sqs_queue.dlq.arn
+    maxReceiveCount     = min(var.max_retries + 1, 5)
+  })
+
+  tags = var.tags
+}
+
+data "aws_iam_policy_document" "lambda_assume" {
+  statement {
+    actions = ["sts:AssumeRole"]
+    principals {
+      type        = "Service"
+      identifiers = ["lambda.amazonaws.com"]
+    }
+  }
+}
+
+resource "aws_iam_role" "lambda" {
+  name               = var.function_name
+  assume_role_policy = data.aws_iam_policy_document.lambda_assume.json
+  tags               = var.tags
+}
+
+data "aws_iam_policy_document" "lambda" {
+  statement {
+    sid = "Logs"
+    actions = [
+      "logs:CreateLogGroup",
+      "logs:CreateLogStream",
+      "logs:PutLogEvents",
+    ]
+    resources = ["${aws_cloudwatch_log_group.task.arn}:*"]
+  }
+
+  statement {
+    sid = "TaskQueue"
+    actions = [
+      "sqs:ReceiveMessage",
+      "sqs:DeleteMessage",
+      "sqs:GetQueueAttributes",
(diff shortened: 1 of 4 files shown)


terraform/modules/aws_task_worker/main.tf
```diff
@@ -0,0 +1,130 @@
...
+
+resource "aws_sqs_queue" "task" {
+  name = var.queue_name
+
+  # Visibility must exceed the function timeout so a slow task is not redelivered while still running.
+  visibility_timeout_seconds = max(180, var.timeout_seconds + 60)
+
+  redrive_policy = jsonencode({
+    deadLetterTargetArn = aws_sqs_queue.dlq.arn
+    maxReceiveCount     = min(var.max_retries + 1, 5)
```
joebon, line 20: Can of course always be changed later, but shouldn't the fallback be 20 here since that's what we have currently?

terraform/modules/aws_task_worker/variables.tf
```diff
@@ -0,0 +1,87 @@
...
+variable "memory_size" {
+  description = "Lambda memory in MB."
+  type        = number
+  default     = 512
+}
+
+variable "reserved_concurrency" {
+  description = "null leaves the function unreserved (-1). Bounds warm containers, and therefore DB connections."
+  type        = number
+  default     = null
```
joebon: Let's default this to a non-null value as a dead man's switch. 50?

terraform/modules/aws_task_worker/variables.tf
```diff
@@ -0,0 +1,87 @@
...
+}
+
+variable "max_retries" {
+  description = "DLQ redrive maxReceiveCount is min(max_retries + 1, 5)."
+  type        = number
+  default     = 4
```
joebon, line 37: Again, I think our default retry limit currently is 20

joebon (approved):
Nice! Added some suggestions but since this won't pick up anything immediately they're non-blocking

terraform/modules/aws_task_worker/main.tf
Yopi: SQS is a bit different, as we will need to solve for something similar to our exponential backoff with this. Today, as the solution is built, it has a fixed backoff and retry. So basically it will retry every 180s if it fails, for the specified amount of times. 

So it's not 1:1 to how it works today, and I don't think 20 will be equivalent in time, but possibly not `min(var.max_retries + 1, 5)` either

terraform/modules/aws_task_worker/variables.tf
Yopi: Makes sense, yes!

terraform/modules/aws_task_worker/main.tf
joebon: Hah, yes - sorry! And the elapsed time between the first and last attempt is 20 days which we should _not_ mimic in this fixed-delay setup. Let's stick with 5 attempts; if it doesn't work after 15 minutes it likely needs manual intervention anyways. 👍 sorry for the noise

PR #12304, opened 2026-06-10
chore(docs): add learning from previous session to claude.md

maximevast:
---
## Summary by cubic
Add backend conventions and examples to `server/CLAUDE.md` to align new code with review standards and avoid common pitfalls. Covers schema patterns (`IDSchema`, enum typing), service singleton naming, creating ORM objects via relationships (not IDs), `lazy="raise"` loading with explicit eager-loads, using status-coded `PolarError`s (not 422) with OpenAPI responses, and returning ORM models from endpoints with `response_model`.

<sup>Written for commit 65e6611b87c5307f5e345dd23ab24072da46d7bd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12304?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/CLAUDE.md
```diff
@@ -285,6 +286,53 @@ class TestListResources:
...
+### Creating ORM objects — pass objects, not ids
+Set the related ORM object, not its foreign-key id: it's type-safe and populates the relationship in-session.
+```python
+SupportCaseMessage(case=case, author_user=user)   # ✅
+SupportCaseMessage(case_id=case.id, author_user_id=user.id)   # ❌
+```
+Add the `relationship()` to the model if it doesn't exist yet.
+
+### Relationships are `lazy="raise"`
+All relationships use `lazy="raise"`: accessing an unloaded one raises instead of silently emitting a query (no async lazy-load `MissingGreenlet`, no N+1). Eager-load what the request needs:
```
frankie567: Unless there is a legitimate reason to always eager load: typically, many-to-many relationship tables.

server/CLAUDE.md
```diff
@@ -285,6 +286,53 @@ class TestListResources:
...
+select(X).options(joinedload(X.rel))   # or selectinload / contains_eager
+```
+(Populating a backref into an unloaded collection — e.g. `case=case` — is passive and does **not** trigger the raise.)
+
+### Errors → status-coded `PolarError`, not validation errors
+Logical/conflict errors are `PolarError` subclasses carrying their own `status_code`; the global exception handler renders them. Don't catch and re-raise as `PolarRequestValidationError` — that's only for request-*payload* validation (→ 422). Declare them on the endpoint's `responses=` so the OpenAPI client gets the schema:
+```python
+class CaseClosedError(PolarError):
+    def __init__(self) -> None:
+        super().__init__("This case is closed.", 409)
```
frankie567: Mention that it's important to set a proper status code on the errors that should be handled as normal business operations, like here. The default, 500, will still be rendered correctly by the handler _but_ Sentry will report it as a hard error.

frankie567 (commented):
A few more details :) 

BTW, if one day you have time to merge `AGENTS.md` and `CLAUDE.md`, would be super nice. Codex & Mistral are a thing too ;)

maximevast:
@claude did we loose any documentation by normalizing claude.md with agents.md ?

frankie567 (approved):

PR #12306, opened 2026-06-10
Expose and persist benefit visibility in the benefit API

strandhvilliam:
Part 2 for the Benefit Visibility feature in: https://github.com/polarsource/feedback/issues/159.

## Summary
  - Adds `visibility` to benefit create/update/response schemas
  - Persists visibility on create via `BenefitType.resolve_visibility()`
  - Rejects visibility updates for non-configurable benefit types (Discord, GitHub, downloadables)
  - Regenerates the API client (`v1.ts`) with `BenefitVisibility` types
 
Depends on PR 1 (migration + model). After this merges, visibility can be set via API. Customer-facing surfaces still show all benefits until PR 4 where it will be enforced (after pr 3 which will include UI controls for visibility).

Null values will be assumed to be public until visibility column is backfilled and set to not null.

---
## Summary by cubic
Adds benefit visibility (`draft`, `private`, `public`) to the API so orgs can control what subscribers see. Visibility is persisted with defaults, validated on update, and filterable in the list endpoint.

- New Features
  - Added `visibility` to create/update and org-facing responses; exposed `visibility_configurable` and coerce `null` to `public`. Subscriber responses now omit `metadata` and do not expose `visibility`.
  - Persisted via `BenefitType.resolve_visibility()` with type defaults (feature flags temporarily default to `public`); updates are rejected for non-configurable types (Discord, GitHub repository, downloadables).
  - Regenerated TS client (`v1.ts`) with `BenefitVisibility` enum/fields and `benefitVisibilityValues`; list endpoint supports `visibility` filtering.

<sup>Written for commit 1ab9560b11a89b5c4d78391e78b4111270dcbeae. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12306?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/polar/benefit/strategies/base/schemas.py
```diff
@@ -60,6 +69,16 @@ class BenefitPublicBase(TimestampedSchema, IDSchema):
...
     )
+    visibility: BenefitVisibility = Field(
+        description="The visibility of the benefit in the customer portal."
+    )
+
+    @field_validator("visibility", mode="before")
+    @classmethod
+    def default_null_visibility(cls, value: Visibility | None) -> Visibility:
+        if value is None:
+            return Visibility.public
```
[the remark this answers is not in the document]
strandhvilliam: Will backfill the null visibility column to 'public' in later PR while setting the column to NOT_NULL. Thus this coercion is temporary while we allow null values.

server/polar/benefit/service.py
```diff
@@ -188,6 +190,21 @@ async def update(
                 ]
             )
 
+        if (
+            benefit_update.visibility is not None
```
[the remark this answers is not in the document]
strandhvilliam: should be fixed in [ba51ea1](https://github.com/polarsource/polar/pull/12306/commits/ba51ea1b5e966c7aefd7dcb5b63b666e916da78d)

server/polar/benefit/service.py
```diff
@@ -69,6 +71,9 @@ async def list(
         if query is not None:
             statement = statement.where(Benefit.description.ilike(f"%{query}%"))
 
+        if visibility is not None:
+            statement = statement.where(Benefit.visibility.in_(visibility))
```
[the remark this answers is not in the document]
strandhvilliam: We’ll backfill all visibility values to 'public' before exposing using the visibility filter, so we don’t need the NULL OR public

server/polar/benefit/strategies/base/schemas.py
```diff
@@ -60,6 +69,21 @@ class BenefitPublicBase(TimestampedSchema, IDSchema):
     organization_id: UUID4 = Field(
         ..., description="The ID of the organization owning the benefit."
     )
+    visibility: BenefitVisibility = Field(
+        description="The visibility of the benefit in the customer portal."
+    )
```
frankie567, line 77: I wouldn't put it in `BenefitPublicBase`: this schema will be used by Customer Portal API (i.e. targeted at customers), so we'll likely hide the invisible benefits from there. So we don't want to "leak" that info there.

I would put it under `BenefitBase` instead.

frankie567 (changes requested):

server/polar/benefit/strategies/base/schemas.py
strandhvilliam: Aah makes sense. Will fix! Thanks! 🙏

strandhvilliam: Fixed it in newer commit.

Follow up question: If I understand correctly the BenefitSubscriberBase is used on some public endpoints (Like List Public Benefit Grants) but that one inherits from BenefitBase. Wouldn't that leak it too? Or should BenefitSubscriberBase inherit from the BenefitPublicBase? I might just not understand the whole picture though.

frankie567: Hmm, good catch 🤔 I think indeed `BenefitSubscriberBase` should inherit from `BenefitPublicBase`, it's a mistake (without consequence since there wasn't extra property anymore in `BenefitBase`). But yeah, good to fix that :) 

For the record `BenefitPublic` is also used directly via `CustomerOrganizationData`, which is a public endpoint to build the customer portal.

strandhvilliam: Ok, appreciate the context! Made the `BenefitSubscriberBase` inherit from `BenefitPublicBase` now 🫡

frankie567 (approved):

PR #12335, opened 2026-06-11
Add backfill script for benefit visibility

strandhvilliam:
Part 3 of benefit visibility rollout.

Backfill script to set all current NULL values to `public` in the benefit visibility column.

After previous PR all newly created benefits should be set to public automatically, and after this backfill no NULL visibility values should ever exist.

The next step is a migration script to set the column to NOT NULL, then finally I'll push the final rollout PR

Please let me know if you have any advice or feedback.

---
## Summary by cubic
Backfills `Benefit.visibility` by setting all NULL values to `public` so no existing benefits are hidden in the customer portal during the visibility rollout.

- **New Features**
  - Adds `server/scripts/backfill_benefit_visibility.py` to update NULLs to `Visibility.public` in batched chunks.
  - Idempotent and safe to re-run; only rows with NULL visibility are updated.
  - Tests in `server/tests/scripts/test_backfill_benefit_visibility.py` verify update behavior and idempotency.

<sup>Written for commit cfe62d5b45162caccd125b4bcf021b6d9a103401. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12335?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/scripts/backfill_benefit_visibility.py
+++ b/server/scripts/backfill_benefit_visibility.py
@@ -1,0 +1,55 @@
+"""Backfill the visibility column for existing benefits.
+
+All existing benefits are set to public so nothing is unexpectedly hidden in
+the customer portal. New benefits created after this backfill get their
+visibility from `BenefitType.resolve_visibility()` in `BenefitService.user_create`
+(e.g. feature flags defaulting to private once that rollout is enabled).
+
+The update is idempotent: once a row has visibility set it no longer matches
+the filter, so the script can be safely re-run or resumed.
+"""
+
+import typer
+from sqlalchemy import select, update
+
+from polar.kit.visibility import Visibility
+from polar.models import Benefit
+from scripts.helper import (
+    configure_script_logging,
+    limit_bindparam,
+    run_batched_update,
+    typer_async,
+)
+
+cli = typer.Typer()
+
+configure_script_logging()
+
+subquery = (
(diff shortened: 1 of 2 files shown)


pieterbeulque (approved):

frankie567 (approved):
Perfect!

Open question though: shouldn't we backfill feature flags directly as private? cc @pieterbeulque

strandhvilliam:
I was thinking that this might cause confusion for current customers that expects all benefits to be viewable since when we haven't had an option to hide up to now. At least in next pr when UI is added.

Like if a merchant from nowhere gets customers saying that they can't see their feature flag who always has been visible. When UI is implemented, then it's expected.

But @pieterbeulque can correct me if I'm thinking about it wrong 🙈

pieterbeulque:
I went back and forth because feature flags should be private by default, but eventually I agree that a backfill should be backwards compatible and don't change any implied default, so I agree with Villiam's call to make everything public, just for consistency's sake.

PR #12336, opened 2026-06-11
chore: Remove the free product price type

joebon:
**Related issue**: https://github.com/polarsource/polar/issues/12205
**Follow-up to**: #12272

Completes the migration that replaces the `free` `amount_type` with a `fixed` price of amount `0`. All existing `free` records were backfilled in #12272, so this removes the type entirely.

## Manual checks

```
SELECT COUNT(*) FROM product_prices WHERE amount_type='free';
```

yielded `0` both in `production` and `sandbox`

## Backend
- Remove the `ProductPriceFree` / `LegacyRecurringProductPriceFree` ORM models and output schemas.
- Remove the `FreePrice` type alias and `is_free_price()` guard; callers use `is_fixed_price` / the `is_free` property instead.
- Remove `ProductPriceFreeCreate` and the `free` value from `ProductPriceAmountType` — the API no longer accepts `amount_type: "free"` (returns `422`).
- Preserve the previous free-price behavior now that free == fixed-0: zero-amount prices are skipped in proration (`_collect_proratable_amounts`) and are not discountable (`is_discount_applicable`).
- Drop the spent `free_price_to_fixed_backfill` script + test and update the seed scripts.

## Frontend
- Regenerate the API SDK (no more `ProductPriceFree*` types).
- Keep the dashboard **"Free"** pricing option as a UI-only price type (`FreeProductPriceCreate`) that is converted to a fixed price of `0` before submitting — the API only ever receives `fixed`.
- Detect free plans via amount (`fixed` + `price_amount === 0`) instead of `amount_type === 'free'`.

> [!NOTE]
> Despite dropping the public schemas, we will still accept `amount_type: "free"` in the API and coerce it to a `fixed` price with `amount: 0` since some merchant clients seem to use it.

> [!NOTE]
> `docs/openapi.yaml` is intentionally left untouched — it's regenerated out-of-band via the dedicated "docs: update OpenAPI schema" PRs.

---
## Summary by cubic
Remove the `free` product price type. Free pricing is now represented as a `fixed` price with amount `0` across backend and frontend; the API still accepts `amount_type: "free"` on input for backward compatibility and coerces it to `fixed` with `price_amount: 0`.

- **Refactors**
  - Removed `ProductPriceFree` / legacy variants, schemas, and the `is_free_price` guard; dropped `free` from `ProductPriceAmountType`.
  - Kept zero-amount prices non-discountable and excluded from proration; off-session now only allows fixed prices.
  - Accept legacy `amount_type: "free"` on product price input via a coercing validator (rewritten to `fixed` with `price_amount: 0`); excluded from OpenAPI and the `@polar-sh/client` SDK (no `Free*` types).
  - Updated sorting to remove `free` from ordering; added `isFreePrice` helpers; updated seeds, tests, and UI to treat fixed `0` as free; removed the backfill script.

- **Migration**
  - Stop sending `amount_type: "free"`. Use `amount_type: "fixed"` with `price_amount: 0`. The API will still accept `free` and convert it to fixed `0`.
  - Frontend: detect free plans via `fixed` + `price_amount === 0`. The dashboard “Free” option remains UI-only and is converted to a fixed `0` on submit.

<sup>Written for commit e9c0a807aba917a3114554dbbdb8ac94090bdbb3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12336?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pieterbeulque (approved):

server/polar/product/repository.py
```diff
@@ -108,10 +108,6 @@ def get_sorting_clause(self, property: ProductSortProperty) -> SortingClause:
                 return Product.name
             case ProductSortProperty.price_amount_type:
                 return case(
-                    (
```
pieterbeulque, line 111: This will have a slight behavior change but I think it's fine, not sure how it'll surface too

joebon: Yeah, good point but effectively I think this changes _very_ little. The Dashboard UI currently doesn't offer to sort by price amount type, and sorting by price amount (which is possible) would only move "free" products from "top-top" (`-2`)  to "top" (`0`)

server/polar/product/repository.py
```diff
@@ -108,10 +108,6 @@ def get_sorting_clause(self, property: ProductSortProperty) -> SortingClause:
...
-                        ProductPrice.amount_type == ProductPriceAmountType.free,
-                        1,
-                    ),
                     (
                         ProductPrice.amount_type == ProductPriceAmountType.custom,
                         2,
```
frankie567: Nitpick, but we could shift the indices :D

server/polar/product/service.py
```diff
@@ -578,17 +576,6 @@ async def get_validated_prices(
             else:
                 model_class = price_schema.get_model_class()
                 price_kwargs = price_schema.model_dump()
-                # We no longer create `free` prices: we're dropping the free price
-                # type in favor of a fixed price with an amount of 0. Incoming `free`
-                # prices are still accepted for backward compatibility, but persisted
-                # as a fixed price with an amount of 0.
-                if isinstance(price_schema, ProductPriceFreeCreate):
```
frankie567, line 585: I think we should keep the compat layer for a while: I quickly checked in [Logfire](https://logfire-us.pydantic.dev/polar/polar?q=attributes-%3E%3E%27http.target%27+%3D+%27%2Fv1%2Fproducts%2F%27+AND+attributes-%3E%3E%27http.method%27+%3D+%27POST%27+AND+attributes-%3E%27fastapi.arguments.values%27-%3E%27product_create%27-%3E%27prices%27-%3E0-%3E%3E%27amount_type%27+%3D+%27free%27&last=%222d%22) and it seems some users are programmatically creating this kind of prices.

server/polar/product/schemas.py
```diff
@@ -389,7 +375,6 @@ def get_model_class(self) -> builtins.type[ProductPriceMeteredUnitModel]:
 ProductPriceCreate = Annotated[
     ProductPriceFixedCreate
     | ProductPriceCustomCreate
-    | ProductPriceFreeCreate
```
frankie567, line 392: We could wrap it with `SkipJsonSchema` so the runtime still accepts it, but hide from the OpenAPI schema, effectively removing it from future SDK versions

frankie567 (changes requested):
I think we should keep the backward compat layer for now.

server/polar/product/service.py
pieterbeulque: Oh, good call

server/polar/product/repository.py
joebon: 🧹👍

server/polar/product/service.py
joebon: Makes sense, will fix

server/polar/product/schemas.py
joebon: Good idea, thanks!

joebon: `SkipJsonSchema` still leaves the component around, despite not used everywhere, so I went with a `BeforeValidator` to coerce/cast it to a fixed price instead. LMK if that's too hurtful for the eyes and I'll switch to `SkipJSonSchema`

server/polar/product/service.py
joebon: Addressed at the schema level instead so we don't need to carry this legacy around in the service layer (or elsewhere)

server/polar/product/repository.py
joebon: Fixed

server/polar/product/schemas.py
frankie567: Yeah, actually that's even better since it allows us to get rid of the logic in the service. Great 👏

frankie567 (approved):

psincraian (approved):

sebastianekstrom (approved):

PR #12340, opened 2026-06-11
feat(support-case): emails for new message in appeal thread

maximevast:
## Summary
Emails an organization's members when Polar staff post a new message on their support case — a reply or a decision — so merchants don't have to refresh the case to find out. v1 covers the human-review appeal case.

Screenshot 2026-06-11 at 15 40 15

## What

- New `support_case_organization_new_message` email + React template (subject/body driven by a per-type `case_label`, so the copy reads "your appeal" now and generalizes to future case types).
- `support_case.notify_organization_of_new_message` background task: resolves the case's organization and emails all its members.
- Triggered from `appeal_case.add_reply` (visible replies only, not internal notes) and `appeal_case.record_decision` (approve/deny).

## Why

Case activity has to reach merchants by email, at least until the notification system is refactored. We deliberately **bypass the legacy notification system and enqueue an email directly**. A decision is treated as just another message on the case, so it flows through the same path rather than being a separate notification.

## How

- Reuses the existing `enqueue_email_template` pipeline;
- Fired from a background task so it only sends after the request commits (a rolled-back reply/decision sends nothing) and resolution happens off the request path.
- Naming is recipient-scoped (`…Organization…`), reserving a `…Customer…` variant for when customer-facing case types arrive.
- Tests cover the enqueue triggers (reply, decision, not internal notes) and the task's fan-out + non-staff skip.
## Checklist

---
## Summary by cubic
Emails organization members when Polar staff post a new message (reply or decision) on their appeal support case. Sends a short, no-reply email with a dashboard link so merchants stay updated without refreshing.

- **New Features**
  - Added `support_case_organization_new_message` email and React template with a generic `case_label` ("appeal"); no message excerpt.
  - Background actor `support_case.notify_organization_of_new_message` runs after commit, emails all org members; staff messages only; internal notes aren’t enqueued.
  - Triggered on `appeal_case.add_reply` (visible replies) and `appeal_case.record_decision`.
  - Direct send via `enqueue_email_template` (bypasses legacy notifications); from `noreply@` with no reply-to; includes a dashboard link. Tests cover triggers, fan-out, and skip behavior.

<sup>Written for commit d0b7f35a309a6daf28773a788b1479b5074325dc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12340?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/polar/support_case/tasks.py
```diff
@@ -0,0 +1,90 @@
...
+    never emails.
+    """
+    async with AsyncSessionMaker() as session:
+        message = await SupportCaseMessageRepository.from_session(session).get_by_id(
+            message_id
+        )
+        if message is None:
+            return
+        # Only staff messages notify; merchants/customers see their own.
+        if message.author_kind != SupportCaseMessageAuthorKind.platform:
```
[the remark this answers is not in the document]
maximevast: It's already gated in `add_reply`

server/emails/src/emails/support_case_organization_new_message.tsx
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <WrapperPolar preview={`Update on your ${organization_name} ${case_label}`}>
+      <Intro>
+        There's an update on your {case_label} for{' '}
+        <Text as="span" weight="bold">
+          {organization_name}
+        </Text>
+        .
+      </Intro>
+      <Text>{message_excerpt}</Text>
```
pieterbeulque: I'm unsure if we should include an excerpt as it may trick people into thinking they can reply directly to the email, which would go to Plain instead and beat the purpose.

@sebastianekstrom - If we keep it, can you help by putting this excerpt in some sort of card to visually distinguish between the two, not sure if we have an email component ready for that?

maximevast: You're right, let's remove it. By the way, we should probably have a "noreply" email don't you think?

frankie567 (approved):

server/emails/src/emails/support_case_organization_new_message.tsx
pieterbeulque: In general, I think companies with a noreply email do something wrong, but until we've set up proper inbound reply parsing, we can consider doing it.

The coolest thing would of course be if replies to these emails just go directly to the support case 😏

frankie567: https://resend.com/docs/dashboard/receiving/introduction 🥸

maximevast: I'm not doing inbound for the moment :)

maximevast: Screenshot 2026-06-12 at 10 26 46

I'm using noreply and dropped the reply to field so nothing ends up in plain. 
I also updated the copy. let me know what you think @pieterbeulque @frankie567

pieterbeulque (approved):

pieterbeulque:
(force merged to override Petru's codeowner on organization review)

PR #12356, opened 2026-06-12
feat(benefit): benefit visibility

strandhvilliam:
## Summary

Part 5 Rollout of Benefit visibility.

Introduces configurable **benefit visibility** in dashboard, letting organizations mark a benefit as hidden so it is still granted after purchase but not shown on customer-facing surfaces.

| Area | Changes |
|---|---|
| **Benefit API** | Adds a configurable `visibility` field to benefits, resolves it on create/update based on the benefit type, and surfaces it through the public API schema. Includes the regenerated API client. |
| **Customer-facing surfaces** | Filters non-public benefits out of the customer portal, product benefit lists, order/subscription emails, and benefit grant listings. |
| **Dashboard controls** | Adds a visibility field in the benefit create/update forms, a row-level visibility toggle, and visibility indicators on the benefit detail page. |

## Screenshots

| Product Form Benefits | Benefit visibility field |
|---|---|
| Screenshot 2026-06-14 at 22 26 52 | Screenshot 2026-06-12 at 11 08 05 |
## Notes

- Hidden benefits are still **granted** on purchase — they are only filtered from customer-facing views.
- `feature_flag` benefits default to `private`; other configurable types default to `public`. Non-configurable types (Discord, GitHub, Downloadables, Slack) remain always visible.

---
## Summary by cubic
Adds configurable benefit visibility so orgs can hide a benefit from customer-facing views while still granting it after purchase. Enforces allowed visibility per type and filters non‑public benefits across checkout, customer portal, product schemas, and emails.

- **New Features**
  - API: Add `visibility` to benefits and resolve it on create/update by type; reject non‑public visibility for non‑configurable types with 422. `feature_flag` defaults to private; `custom`, `license_keys`, `meter_credit` default to public; Discord, GitHub repository, Downloadables, Slack shared channel are always public. Regenerated `@polar-sh/client`.
  - Customer surfaces: Filter non‑public benefits from checkout/checkout‑links, customer portal (products, subscriptions, orders, benefit grants), and order/subscription emails.
  - Dashboard: Add a visibility radio for configurable types with sensible defaults and show visibility indicators on the benefit detail page and list rows.

- **Migration**
  - `visibility_configurable` remains as a computed field for compatibility with `@polar-sh/client`.
  - Customer-facing APIs now return only public benefits; non‑configurable types reject non‑public visibility on create/update.

<sup>Written for commit 8fb374be0d61462ca306945ae17d28dbd6b9613a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12356?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


sebastianekstrom (approved):
Client LGTM!

clients/apps/web/src/components/Benefit/BenefitForm.tsx
```diff
@@ -115,6 +121,12 @@ const BenefitForm = ({
       />
 
       {!update ? <BenefitTypeSelect organization={organization} /> : null}
+      {type !== 'usage' && isBenefitVisibilityConfigurable(type) ? (
```
pieterbeulque: What does `usage` mean and why is it not in the `isBenefitVisibilityConfigurable` check? 🤔

clients/apps/web/src/components/Products/Benefits/components/BenefitRow.tsx
```diff
@@ -115,12 +120,24 @@ export const BenefitRow = ({
...
-            <span className="dark:text-polar-500 text-xs text-gray-500">
-              {benefitsDisplayNames[benefit.type]}
-            </span>
+            <div className="flex items-center gap-2">
+              <span className="dark:text-polar-500 text-xs text-gray-500">
+                {benefitsDisplayNames[benefit.type]}
+              </span>
+              {selected && visibilityConfigurable && !isPublic ? (
+                <span className="dark:bg-polar-700 dark:text-polar-500 inline-flex items-center rounded-full bg-gray-100 px-2 text-xs text-gray-500">
+                  Hidden from customers
```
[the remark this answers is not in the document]
pieterbeulque: We don't use i18n in the merchant dashboard.

clients/apps/web/src/components/Benefit/BenefitForm.tsx
strandhvilliam: I'm not 100% sure _why_ `usage` is in this form. The `type !-- 'usage' check is added since TS was complaining about isBenefitVisibiltyConfigurable only checking the backend schema from `schemas['BenefitType']` were `usage` isn't included.

I assume it's sent as extra type option as prop (img) to the component so that BenefitForm will work in some other schenario.

Screenshot 2026-06-12 at 13 57 35

I'll add `usage` as an option on the type sent to `isBenefitVisibilityConfigurable` just to be safe, but if I understand correctly it is not a valid benefit type in backend and should not be an valid type option, but I may have missed something. 🤷‍♂️

clients/apps/web/src/components/Benefit/updateBenefitPayload.ts
```diff
@@ -9,7 +10,10 @@ export const prepareBenefitUpdatePayload = (
 ): BenefitUpdate => {
   const payload = { ...benefitUpdate }
 
-  if (!benefit.visibility_configurable && 'visibility' in payload) {
+  if (
+    !isBenefitVisibilityConfigurable(benefit.type) &&
+    'visibility' in payload
+  ) {
     delete payload.visibility
   }
```
pieterbeulque: I believe this is no longer necessary as the visibility will be ignored by the API. It might be better to not duplicate that logic then.

clients/apps/web/src/components/Products/Benefits/components/BenefitRow.tsx
```diff
@@ -130,9 +134,16 @@ export const BenefitRow = ({
...
               {benefit.description}
             </span>
-            <span className="dark:text-polar-500 text-xs text-gray-500">
-              {benefitsDisplayNames[benefit.type]}
-            </span>
+            <div className="flex items-center gap-2">
+              <span className="dark:text-polar-500 text-xs text-gray-500">
+                {benefitsDisplayNames[benefit.type]}
+              </span>
+              {selected && visibilityConfigurable && !isPublic ? (
```
pieterbeulque: I don't think this needs the extra `visibilityConfigurable` check, can be confusing if we eventually add a `private` non-configurable benefit type (who knows!)

clients/apps/web/src/components/Benefit/BenefitVisibilityField.tsx
```diff
@@ -0,0 +1,117 @@
...
+              value={field.value}
+              onValueChange={field.onChange}
+              className="grid grid-cols-1 gap-3"
+            >
+              <BenefitVisibilityOption
+                id="benefit-visibility-public"
+                value="public"
+                selected={field.value === 'public'}
+                label="Visible"
+                description="Customers can see this benefit in customer-facing product and benefit lists."
```
pieterbeulque: Suggestion:

"Customers can see this benefit after purchase, including in order confirmations and the Customer Portal."

clients/apps/web/src/components/Benefit/BenefitVisibilityField.tsx
```diff
@@ -0,0 +1,117 @@
...
+              />
+              <BenefitVisibilityOption
+                id="benefit-visibility-private"
+                value="private"
+                selected={field.value === 'private'}
+                label="Hidden"
+                description="Still granted after purchase, but not visible in customer-facing product or benefit lists."
```
pieterbeulque: Suggestion:

"Granted after purchase, but is never shown to customers."

or

"Customers can't see this benefit, but it is still granted after purchase."

pieterbeulque (approved):
I'm not sure about the `visibility_configurable` on a benefit but that wasn't added in this PR, was it?

strandhvilliam:
I dont' like the `visibility_configurable` either and instead prefer the specific function to check for configurability.

It was added in another earlier PR for benefit visibility, however when I tried to remove it, it resulted in merge conflicts and for some reason tests failing in CI (but not in my local env) so I kept them in. Think it became dependent on from this pr: https://github.com/polarsource/polar/pull/12345.

Can try to remove it in separate PR.

clients/apps/web/src/components/Benefit/BenefitVisibilityField.tsx
strandhvilliam: Copy skills on point! Will fix ☀️

clients/apps/web/src/components/Benefit/BenefitVisibilityField.tsx
strandhvilliam: Nice, less technical. Like it!

clients/apps/web/src/components/Products/Benefits/components/BenefitRow.tsx
strandhvilliam: Smart! Didn't think of that. Will remove.

clients/apps/web/src/components/Benefit/updateBenefitPayload.ts
strandhvilliam: Will remove 🫡

server/polar/benefit/service.py
```diff
@@ -162,12 +162,13 @@ async def user_create(
             organization,
             create_schema.properties.model_dump(mode="json", by_alias=True),
         )
+        visibility = create_schema.type.resolve_visibility(create_schema.visibility)
```
frankie567, line 184: If I understand correctly, `resolve_visibility` will implicitly convert to `public` benefits that are not configurable, right?

I'm wondering if it would be more explicit to actually throw an error if the user tries to do so. Something like:

```py
if not create_schema.type.is_visibility_configurable() and create_schema.visibility != 'public':
    raise PolarRequestValidationError(...)
```

strandhvilliam: Yeah exactly that was my thinking with the `resolve_visibility` to implicitly convert to `public`

But I can see what you're saying with it being more explicit. To throw when a non-configurable is attempted to be set as non public value, but if nothing is sent in it resolves to the default.

Will update the resolve_visibility method with this 🫡

frankie567: Awesome, thanks 🙏

frankie567 (approved):

PR #12364, opened 2026-06-12
server/transaction: fix a TOCTOU bug when handling charge.updated events

### LINKED ISSUE #12362: TOCTOU race in create_payment(): duplicate Transaction rows when charge.updated fires twice concurrently
## Bug

`PaymentTransactionService.create_payment()` uses a check-then-insert pattern that is not safe under concurrent execution:

```python
# polar/transaction/service/payment.py L49-52
existing_transaction = await self.get_by_charge_id(session, charge.id)
if existing_transaction is not None:
    return existing_transaction
# ... proceeds to INSERT a new Transaction row
```

When Stripe fires `charge.updated` twice within \~100ms (a documented Stripe behaviour), two workers can both execute this check before either has committed, pass the `is not None` guard, and both insert a `Transaction(type=payment, charge_id=X)` — resulting in two rows with the same `charge_id`.

There is no DB-level `UniqueConstraint` on `(type, charge_id)` — only an index — so nothing prevents the duplicate insert.

The duplicate then causes `order.balance` to crash with `sqlalchemy.exc.MultipleResultsFound` in `balance_transaction_service.get_by(charge_id=...)` (which calls `.one_or_none()`), failing indefinitely on retries.

## Potential fixes

**Option A — DB unique constraint + catch IntegrityError (recommended)**

```python
# models/transaction.py
UniqueConstraint("charge_id", postgresql_where="type = 'payment'")

# payment.py create_payment()
try:
    session.add(transaction)
    await session.flush()
except IntegrityError:
    await session.rollback()
    return await self.get_by_charge_id(session, charge.id)
```

**Option B — Postgres upsert (**`INSERT ... ON CONFLICT DO NOTHING`**)**
Replace the manual check-then-insert with a single atomic upsert statement that returns the existing row if a conflict is detected.

**Option C — Advisory lock on** `charge_id`
Acquire a `pg_advisory_xact_lock` keyed on the `charge_id` hash before the read, serialising concurrent calls for the same charge.

**Option D — Idempotent query in** `create_order_balance`
Change `get_by()` (which uses `.one_or_none()`) to use `.first()` for the payment transaction lookup — a short-term mitigation that avoids the crash while the root cause is fixed.

## Related

* Stripe docs note that the same event can be delivered more than once
* A similar issue was previously fixed for payment intents via upsert: #7171
* `order.balance` task: `polar/order/tasks.py:165`

<sub>Sent by @frankie567 from [Investigate](https://app.stilla.ai/m/memo_01ktxasp7nf2yajs7315f8m9g4).</sub>

frankie567:
Fix #12362

- server/transaction: fix a TOCTOU bug when handling charge.updated events
- dx/solo: workaround a bug of Solo with working_dir

---
## Summary by cubic
Fixes a race in `stripe` charge.updated handling that could create duplicate payment transactions by enforcing DB uniqueness and returning the existing record on conflicts. Also updates Solo dev commands to run from the correct directories. Addresses pol-12362.

- **Bug Fixes**
  - Add partial unique index on `transactions` for (`type`, `charge_id`) where `type='payment'`; set a 5s `lock_timeout` and mirror the index in the model.
  - On `IntegrityError` during insert, fetch and return the existing transaction; skip fee job enqueue.
  - Add a test that simulates the race; update tests to use distinct `charge_id`s.
  - Solo: prefix `Containers`, `API`, `Worker`, and `Web` commands with `cd` to work around the `working_dir` bug.

<sup>Written for commit 15444d38d549f243f2cce80fd046b338b72cb22a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12364?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/polar/transaction/service/payment.py
```diff
@@ -103,7 +104,14 @@ async def create_payment(
         )
 
         session.add(transaction)
-        await session.flush()
+        try:
+            await session.flush()
+        except IntegrityError:
+            # This means another worker has created the transaction after we checked for it
+            existing_transaction = await self.get_by_charge_id(session, charge.id)
+            if existing_transaction is not None:
```
joebon, line 116: Perhaps add some logging here so it's not _completely_ silent?

joebon (approved):
LGTM! Non-blocking suggestion to log if a duplicate record is identified

Yopi (approved):

pieterbeulque (approved):

PR #12393, opened 2026-06-15
fix(customer-portal): show downloadables to members via benefit grants

pieterbeulque:
I believe this was never finished, and I think it makes more sense to _remove_ `member_id` from downloadables instead of duplicating a downloadables row per member, whereas for e.g. license keys it was required to generate a license key _per member_.

## Generated description

Members in the customer portal could not see downloadables. The repository
filtered by `Downloadable.member_id == member.id`, but a downloadable row is
unique per (customer, file, benefit) and shared across a customer's members,
so its member_id can only ever reference one of them. On seat-based products
the same file/benefit is granted to several members, they collide on that one
row, and only the first-granted member gets attributed — everyone else sees
nothing.

A downloadables benefit defines its files at the benefit level, so every
member holding an active grant for that benefit is entitled to the same files.
Gate member visibility on the member's benefit grants (the source of truth for
per-member entitlement, which already works) instead of on the shared
Downloadable.member_id column. This also fixes legacy rows that never had
member_id backfilled. No migration required.

---
## Summary by cubic
Fixes member visibility of downloadables in the customer portal by checking active benefit grants instead of Downloadable.member_id. Restores access for seat-based products and legacy rows without member_id.

- Bug Fixes
  - Filter member results by customer_id and active, non-deleted benefit grants.
  - Supports shared rows across members (seat-based products) and legacy rows without member_id.
  - Added API tests for member with grant, without grant, and shared-row collision; no migration needed.

<sup>Written for commit d2f10570945f98a59a1a5b9366dbcde396794b30. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12393?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/customer_portal/repository/downloadable.py
+++ b/server/polar/customer_portal/repository/downloadable.py
@@ -13,7 +13,7 @@ class DownloadableRepository(RepositoryBase[Downloadable]):
     def get_customer_statement(
         self, auth_subject: AuthSubject[Customer | Member]
     ) -> Select[tuple[Downloadable]]:
-        from polar.models import Benefit
+        from polar.models import Benefit, BenefitGrant
 
         statement = (
             sql.select(Downloadable)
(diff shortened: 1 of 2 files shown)


psincraian (changes requested):
I would love to discuss this approach. I thought I removed the uniqueness of customer_id in every benefit type, but seems I was wrong. 

My main idea is that benefits are member aware mainly, as we grant benefits for each member. For me the `downloadable` should follow the same pattern as the others, so we should replace the uniqueness to be member and benefit_id aware only.

pieterbeulque:
I'm fine with that too. What are the steps involved in that approach then? I imagine it'll need a backfill?

psincraian:
that will become more a burden, yes 🥲 as it's a shitty thing that I forget, I can fix it.

The idea would be to remove the customer uniqueness and add a member uniqueness. Then do a backfill of missing downloads (looping through customer members)

psincraian (approved):
Ok, we will fix this for existing customers and next week we are gonna ship and improved version that allows the grants to flow to all members

pieterbeulque:
Will be fixed in #12556 instead

PR #12411, opened 2026-06-16
Send chargeback prevention notices

pieterbeulque:
image

---
## Summary by cubic
Send chargeback-prevention notices to organization owners and admins when a dispute-prevention refund is issued. Adds a new email template and a background task that fans out to the right recipients.

- **New Features**
  - Added `chargeback_prevention_refund` email template with order number, customer name, amount, currency, and refund date; includes computed `formatted_amount`.
  - Enqueues `refund.send_chargeback_prevention_notice` when a `dispute_prevention` refund is created from a dispute, and when a pending Stripe refund transitions to `succeeded` (ignored for other reasons).
  - Task loads the order, builds subject “Chargeback prevented for order {order_number}”, and emails org `owner` and `admin` members only; falls back to order ID if no invoice number.
  - Registered the task in `server/polar/tasks.py` and exposed the template in `emails/src/emails/index.ts`; added OpenAPI and Python schemas for template props.
  - Tests cover enqueue timing (create vs. transition, once), role-based fan-out, subject/props, and order number fallback.

<sup>Written for commit 2a963a3ec537d981f7325892de9d42e550c52b4b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12411?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/refund/tasks.py
+++ b/server/polar/refund/tasks.py
@@ -1,0 +1,66 @@
+import uuid
+
+from sqlalchemy.orm import joinedload
+
+from polar.email.schemas import (
+    ChargebackPreventionRefundEmail,
+    ChargebackPreventionRefundProps,
+)
+from polar.email.sender import enqueue_email_template
+from polar.models import Order, Refund
+from polar.models.user_organization import OrganizationRole
+from polar.user_organization.service import (
+    user_organization as user_organization_service,
+)
+from polar.worker import AsyncSessionMaker, TaskPriority, actor
+
+from .repository import RefundRepository
+
+_NOTICE_RECIPIENT_ROLES = (OrganizationRole.owner, OrganizationRole.admin)
+
+
+@actor(
+    actor_name="refund.send_chargeback_prevention_notice",
+    priority=TaskPriority.LOW,
+)
+async def send_chargeback_prevention_notice(refund_id: uuid.UUID) -> None:
+    async with AsyncSessionMaker() as session:
+        repository = RefundRepository.from_session(session)
+        refund = await repository.get_by_id(
+            refund_id,
+            options=(joinedload(Refund.order).joinedload(Order.customer),),
+        )
+        if refund is None:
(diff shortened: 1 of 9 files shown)


server/polar/refund/tasks.py
```diff
@@ -0,0 +1,66 @@
...
+
+        order_number = order.invoice_number or str(order.id)
+        customer_name = order.customer.display_name
+        subject = f"Chargeback prevented for order {order_number}"
+
+        members = await user_organization_service.list_by_org(
+            session, order.organization_id
+        )
+        for member in members:
+            if member.role not in _NOTICE_RECIPIENT_ROLES:
```
frankie567, line 48: Should we add a notification setting to toggle that?

frankie567 (approved):
Good as is, just wondering if we should introduce a notification setting for those emails.

server/polar/refund/tasks.py
pieterbeulque: I was wondering if we should do it, store it as the notification settings like new sale & new subscriber, but then not tie it into the notification system itself? Decided not to do it to keep both flows isolated, but maybe there's no harm in having the toggle anyway.

frankie567: Yeah, good call. I guess the settings won't change even if we revamp the notification system, so I would say it can be interesting to introduce it now.

pieterbeulque: Merged this + the notification section toggle in #12442

pieterbeulque:
#12442

PR #12412, opened 2026-06-16
ci: Deploy by tag instead of digest

Yopi:
It looks like the image source treats digests as a tag. Given a digest (e.g. `image@sha256:abc`) the non-mutually exclusive if statements will mark it as both digest (image, sha256:abc) and tag (image@sha256, abc). 

https://github.com/render-oss/terraform-provider-render/blob/b24a8b46ff4821f5bec08ad3db3ea402b8a97d3e/internal/provider/common/models.go#L262-L272

Given the preference of tags over digests in the runtimesource 
https://github.com/render-oss/terraform-provider-render/blob/v1.8.0/internal/provider/common/runtimesource.go#L175-L183

The terraform provider will try to use the invalid reference as a tag.

This causes all deploys that comes from Terraform to fail since we are referencing an image tag that does not actually exist.

---
## Summary by cubic
Switch deployment to image tags instead of digests to prevent the Terraform provider from misparsing `@sha256` and breaking Render deploys. Cleaned up workflows by removing digest plumbing and using the same tag for Sentry `dist`.

- **Bug Fixes**
  - Use `TAG="sha-${GITHUB_SHA::7}"` in `deploy-environment.yml`, pass it to `deploy_server.sh`, and set Sentry `dist` to this tag.
  - Update `deploy_server.sh` to accept tags and use `ghcr.io/polarsource/polar:${TAG}` and `ghcr.io/polarsource/polar-playwright:${TAG}`.
  - Remove unused digest inputs/outputs and related steps from `deploy.yml` and `deploy-environment.yml`.

<sup>Written for commit 1600d2533798efa79d70a79a632e31ef19541267. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12412?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/deploy-environment.yml
+++ b/.github/workflows/deploy-environment.yml
@@ -108,12 +108,13 @@ jobs:
 
       - name: Deploy Backend to Render
         run: |
+          TAG="sha-${GITHUB_SHA::7}"
           ./.github/workflows/deploy_server.sh \
-            "${{ inputs.docker-digest }}" \
+            "$TAG" \
             "${{ inputs.has-migrations }}" \
             "${{ inputs.render-api-service-id }}" \
             "${{ inputs.render-worker-service-ids }}" \
-            "${{ inputs.docker-digest-playwright }}" \
+            "$TAG" \
             "${{ inputs.render-playwright-worker-service-ids }}"
         env:
           RENDER_API_TOKEN: ${{ secrets.RENDER_API_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-environment.yml
joebon: Nit: this one doesn't look like it's being used anymore so can be dropped

joebon: Not: let's use tag here aswell to make correlations easier?

joebon (approved):
Good find!
Two non-blocking comments/suggestions.

.github/workflows/deploy-environment.yml
Yopi: Good idea!

PR #12361, opened 2026-06-12
oauth2: Only allow issuing oauth tokens for permission

Yopi:
We only allow you to use the Oauth2 flow to issue an token when the user conducting the flow has the organization manage role. This is basically only admins today.

---
## Summary by cubic
Restrict OAuth2 org-scoped tokens to users with the `organization_manage` permission. Non-managers can’t see organizations during authorize and can’t obtain org tokens via auth code or web grants.

- **Bug Fixes**
  - Authorize filters organizations to those the user can manage via `get_accessible_org_ids`.
  - Authorization Code and Web grants validate the org sub with `select_user_org_ids(..., permission=OrganizationPermission.organization_manage)`.
  - Tests cover: member excluded in authorize, consent/web grant rejected for members or managers of other orgs, and admin success path.

<sup>Written for commit 6484af52d3a5d49109ef77fde8352e633e55b3c4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12361?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #12415, opened 2026-06-16
fix(oauth2): default dynamically-registered clients to user-scoped tokens

joebon:
## Problem

Merchants reported that the public MCP server (`mcp.polar.sh`) stopped working around the deploy of #12361. That PR restricted **organization-scoped** OAuth2 token issuance to users with the `organization_manage` permission — in both the authorization_code and web grants, plus the `/authorize` org listing.

External MCP clients authenticate via standard OAuth2 + Dynamic Client Registration. DCR clients default to **organization-scoped** tokens, so non-admin members can no longer complete the OAuth flow (they see no organizations at consent / the grant is rejected).

## Fix

Default Dynamically-Registered clients to **user-scoped** tokens, which don't hit the `organization_manage` gate (it only applies to the organization sub_type):

- **`server/polar/oauth2/schemas.py`** — `OAuth2ClientConfiguration.default_sub_type` defaults to `user` instead of `organization`. Clients that want org scope can still set `default_sub_type` explicitly.
- **`clients/packages/client/src/v1.ts`** — regenerated (`@default` annotation updated).

## Notes

- **Admins were not broken** by #12361 (they have `organization_manage`, so the org-scoped flow still worked). This change unblocks **non-admin members** and, for everyone else on DCR, switches the default token scope from org to user.
- **Only affects new registrations.** Existing DCR clients have `organization` baked into their metadata and stay broken until they re-register. If the public `mcp.polar.sh` server uses a single fixed pre-registered client rather than per-user DCR, it needs its metadata updated directly — to be confirmed. A targeted data migration can flip existing clients if desired (caveat: it would affect all org-default DCR clients, not just MCP).
- The in-dashboard AI product assistant (web grant) is intentionally **not** changed here.

### DIFF
--- a/clients/packages/client/src/v1.ts
+++ b/clients/packages/client/src/v1.ts
@@ -22602,7 +22602,7 @@ export interface components {
       tos_uri?: string | null
       /** Policy Uri */
       policy_uri?: string | null
-      /** @default organization */
+      /** @default user */
       default_sub_type: components['schemas']['SubType']
       /**
        * Created At
@@ -22667,7 +22667,7 @@ export interface components {
       tos_uri?: string | null
       /** Policy Uri */
       policy_uri?: string | null
-      /** @default organization */
+      /** @default user */
       default_sub_type: components['schemas']['SubType']
     }
     /** OAuth2ClientConfigurationUpdate */
@@ -22713,7 +22713,7 @@ export interface components {
       tos_uri?: string | null
       /** Policy Uri */
       policy_uri?: string | null
-      /** @default organization */
+      /** @default user */
       default_sub_type: components['schemas']['SubType']
       /** Client Id */
       client_id: string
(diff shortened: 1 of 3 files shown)


frankie567 (approved):
Yep, makes sense. If we can update the doc accordingly, would be awesome: https://github.com/polarsource/polar/blob/4e087bb2f44078f42616781e02a4bd07a6bb80e2/docs/integrate/oauth2/connect.mdx#L83-L88

joebon:
> Yep, makes sense. If we can update the doc accordingly, would be awesome:
> 
> https://github.com/polarsource/polar/blob/4e087bb2f44078f42616781e02a4bd07a6bb80e2/docs/integrate/oauth2/connect.mdx#L83-L88

Good catch! I will

pieterbeulque (approved):

PR #12430, opened 2026-06-17
feat(member): make member email updatable

strandhvilliam:
Make member email updatable

Allows a member's email to be edited via the member update API, with the necessary safety guards and seat-snapshot sync.

Backend
- Added email to MemberUpdate; plumbed through both update_member and update_member_by_external_id endpoints.
- member_service.update() validates the new email before applying it:
- Uniqueness — rejects an email already used by another active member of the same customer (case-insensitive) with a 422, instead of letting the partial unique index throw a 500.
  - Owner guard — blocks editing the email of an individual customer's owner, since sync_owner_email mirrors customer.email onto that member and a direct edit would be silently reverted (and could spawn a duplicate owner on portal sign-in).
- Seat email sync — when a member's email changes, the snapshot on its connected active seats (member_id set, pending/claimed) is updated. Seats with no member connected (member_id NULL, e.g. legacy seats) are left untouched.

Frontend
- Removed email edit block on EditMemberModal, wired to backend endpoint

| Edit Member Modal with Email |
|---|
| Screenshot 2026-06-17 at 10 44 43|

---
## Summary by cubic
Make member email editable in the update APIs and the Edit Member modal. Enforces safe updates, trims input, and syncs active seat email snapshots.

- **New Features**
  - Backend: add `email` to `MemberUpdate`; support in `update_member` and `update_member_by_external_id`. Validate by trimming, rejecting duplicates within the same customer (case-insensitive), and blocking edits for an individual customer’s owner.
  - Seat sync: when a member’s email changes, update email snapshots on connected active seats (`pending`/`claimed`); seats without a member link are left untouched.
  - Frontend: enable email editing in `EditMemberModal` for team customers; disable with a note for individual owners. Only team customers can open the modal from `MembersSection`. Update generated client to include `MemberUpdate.email` and reflect that `name`, `email`, and `role` are updatable.

<sup>Written for commit 7cecf6f818cfc4e0ad840836629fc1b363ed3f8c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12430?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


clients/apps/web/src/components/Customer/EditMemberModal.tsx
```diff
@@ -110,13 +131,11 @@ export const EditMemberModal = ({
...
-          <Button
-            type="submit"
-            className="self-end"
-            loading={updateMember.isPending}
-          >
-            Save Member
-          </Button>
+          <Box justifyContent="end">
+            <Button type="submit" loading={updateMember.isPending}>
+              Save Member
```
pieterbeulque: Nit pick but I think _Update Member_ is more in line with other microcopy

pieterbeulque (approved):

server/polar/member/endpoints.py
```diff
@@ -213,7 +213,7 @@ async def update_member(
     """
     Update a member.
 
-    Only name and role can be updated.
+    Only name, email and role can be updated.
```
pieterbeulque, line 216:
```suggestion
```

Can leave it out since now everything is updatable

strandhvilliam: Isn't `external_id` still immutable?

clients/apps/web/src/components/Customer/EditMemberModal.tsx
strandhvilliam: Nice one! Will fix 🚀

server/polar/member/endpoints.py
pieterbeulque: Fair, yes

server/polar/member/endpoints.py
```diff
@@ -226,6 +226,7 @@ async def update_member(
         member,
         name=member_update.name,
         role=member_update.role,
+        email=member_update.email,
```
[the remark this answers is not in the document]
strandhvilliam: **LLM battle** 🍿

Root facts

- The email-change failures raise PolarRequestValidationError (service.py:658, :677), which the global handler renders as 422 — the exact same path as FastAPI's built-in RequestValidationError (exception_handlers.py:29-33, :64-66).
- update_member already takes a request body (member_update: MemberUpdate), so FastAPI already auto-documents a 422 HTTPValidationError in the OpenAPI schema for this route by default.
- This isn't a new contract: the same update() method already raised PolarRequestValidationError for role/ownership rules before this change (service.py:737, :767 — "Only the owner can transfer ownership", "Cannot change role…"). Those were never declared in responses= either, because they don't need to be. The email validation adds zero new status codes.

Why "fixing" it would be wrong

The project convention in server/AGENTS.md is explicit:

▎ 422 is for request-payload validation only … (Don't add a content-less 422: {"description": …} override — it clobbers FastAPI's default HTTPValidationError and breaks the generated client.)

So declaring an explicit 422 here would actively regress the generated client by overwriting the default HTTPValidationError schema. The reviewer's instinct ("declare non-2xx responses") is right in general — but it applies to status-coded business errors (404/409, which carry a custom PolarError.schema()), not to 422 payload-validation, which is the framework default.

/ claude

frankie567 (approved):

server/polar/member/service.py
```diff
@@ -641,6 +642,49 @@ async def create(
...
+                            "type": "value_error",
+                            "loc": ("body", "email"),
+                            "msg": (
+                                "Cannot change the email of an individual customer's "
+                                "owner. Update the customer's email instead."
+                            ),
+                            "input": new_email,
+                        }
+                    ]
+                )
```
psincraian: nitpick - 

sync_owner_email as you mention only syncs the email of customer.type == individiaul.

So, I think this can be changed to:

```python
customer = await customer_repository.get_by_id(member.customer_id)
if customer.type == CustomerType.individual:
   raise PolarRequestValidation(....)
```

psincraian (approved):

server/polar/member/service.py
strandhvilliam: Makes sense! Changing it.

PR #12435, opened 2026-06-17
feat(backoffice): support cases handles all cases instead of just appeals

maximevast:
## Summary

https://github.com/user-attachments/assets/84269dab-9069-4fbe-9818-ad88c7caca3c

## What

Instead of directly displaying the appeal case in the org page, displays a list of cases and add a specific case page.
Customize UI depending on case type (i.e. dispute has no "accept" / "deny" button)

## Checklist

---
## Summary by cubic
Backoffice now handles all support case types (appeals and disputes) end to end with a shared case detail page and unified lists. The org page gets a “Support Cases” tab that links into the new detail, and actions preserve the user’s origin via `return_to`.

- **New Features**
  - New case detail page at `/backoffice/support-cases/{case_id}` renders any case thread, supports assignment and staff replies; approve/deny actions show only for appeals, with a smart back link to the org or case list.
  - Organization “Support Cases” tab lists all cases (appeals + disputes) with status, assignee, “awaiting reply,” and opened date; each links to the detail page. The nav dot lights up when any case is open.
  - Global Cases list adds a Type filter (All/Appeals/Disputes) and type badges; Tier/Recency sort retained; rows link to the detail page.

- **Refactors**
  - Added polymorphic query `cases_statement` and `open_case_organization_ids`; used by org pages, lists, and open-case counts.
  - Replaced org-scoped reply route with generic `support_cases:reply`; actions/dialogs use `return_to` via `case_detail_url`/`append_return_to` with an open-redirect guard.
  - Service resolves the owning organization for any case via `get_organization_id`; dispute notification emails are suppressed until merchant UI exists. Added tests for queries, URLs, and tasks.

<sup>Written for commit 4efc432173da3896e6fb40cef544a63a86427fb3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12435?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/support_case/service.py
+++ b/server/polar/support_case/service.py
@@ -1,7 +1,12 @@
 from collections.abc import Sequence
+from uuid import UUID
 
-from polar.models import Customer, File, Organization, User
+from sqlalchemy.orm import joinedload
+
+from polar.models import Customer, Dispute, File, Organization, User
 from polar.models.support_case import (
+    DisputeSupportCase,
+    ReviewAppealSupportCase,
     SupportCase,
     SupportCaseAttachment,
     SupportCaseAudience,
@@ -10,10 +15,13 @@ from polar.models.support_case import (
     SupportCaseMessageType,
     SupportCaseParticipant,
     SupportCaseParticipantKind,
+    SupportCaseType,
 )
-from polar.postgres import AsyncSession
+from polar.postgres import AsyncReadSession, AsyncSession
 
 from .repository import (
+    DisputeSupportCaseRepository,
+    ReviewAppealSupportCaseRepository,
     SupportCaseAttachmentRepository,
     SupportCaseMessageRepository,
     SupportCaseParticipantRepository,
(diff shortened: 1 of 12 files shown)


server/polar/support_case/service.py
```diff
@@ -44,6 +52,30 @@ async def create[CaseT: SupportCase](
         )
         return case
 
+    async def get_organization_id(
+        self, session: AsyncSession | AsyncReadSession, case: SupportCase
+    ) -> UUID | None:
+        """The organization that owns a case, resolved through its domain object."""
```
frankie567, line 58: Now that I see it, I'm wondering if we should denormalize `organization_id` on SupportCase. We did it already for a few models (like Subscription), I think it'll save us this kind of pain.

frankie567 (approved):
Looks good, just a suggestion that we should consider denormalize `organization_id` on SupportCase

server/polar/support_case/service.py
maximevast: I keep that in my mind for another PR :) 
Thanks for the review!

PR #12449, opened 2026-06-17
RFC: Tasks on lambda

Yopi:
Introduce an RFC to move the current Dramatiq tasks to SQS+lambda instead.

Work in progress

---
## Summary by cubic
Adds a draft RFC to migrate background tasks from `Dramatiq` to per‑task `SQS` + `Lambda` for elastic, task‑level scaling and faster catch‑up under load. Adds the design doc to the engineering handbook and updates the nav.

- **New Features**
  - Design doc: per‑task `SQS` queue + `Lambda`, independent concurrency, retries via SQS visibility/redrive; includes enqueue split code and message format.
  - Rollout: split enqueueing so selected actors go to `SQS`, others stay on `Dramatiq`; per‑task infra via `Terraform` module; rollback via a fallback lambda (SQS→`Redis`).
  - Infra/ops: static egress IPs allow‑listed to Render DB; shared SG/VPC to reuse Hyperplane ENIs; logs/traces to `Logfire`; separate `Redis`; local dev via `localstack`.
  - Open items/risks: DB connections and `pgBouncer`; cron scheduling (keep on Render for now, consider `EventBridge`); 15‑min `Lambda` timeout.

<sup>Written for commit ad3b2ed8d538be7d78fc8e305b1aa6cf24f5c8fb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12449?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/handbook/engineering/design-documents/tasks-on-lambda.mdx
+++ b/handbook/engineering/design-documents/tasks-on-lambda.mdx
@@ -1,0 +1,176 @@
+---
+title: "Tasks on Lambda"
+description: "Proposal to move task processing away from dramatiq and worker services to be dynamically scale Lambda functions"
+---
+
+<Info>
+**Status**: Draft
+
+**Created**: 2026-06-17
+
+</Info>
+
+## Abstract
+
+The async workflow that we have today relies on worker services using Dramatiq to read from one or multiple specific queues. We specify up front how many workers, how many processes and how many threads each process have.
+
+This proposal is to migrate this workflow to instead use _n_ SQS queues and _n_ Lambda functions (one per task, n tasks). This will allow us to scale the concurrency of each task worker independently of each other. It will allow us
+to automatically increase our capacity if we need to catch up on a queue we are lagging behind on.
+
+This will change our architecture and infrastructure needs, and move us one step closer to running on AWS.
+
+## Motivation
+
+### Problems with Current Approach
+
+The current approach to workers that we have works fine for stable workloads. It is very easy to add new tasks (just specify which queue they should be put on) and they will automatically get picked up. The main issues we have today is:
+
+* When we get overwhelmed with tasks on one of the queues the other workers will not pick up and help with the tasks on the queues they are not listening to.
+* We have to pre-emptively decide which queue a task should go on (high prio, medium prio, etc). This makes it tricky to give more resources to a task that has overwhelmed a queue if we realise that it was more important than previously thought.
+* When the queues are empty we have a bunch of service workers that are running without doing any work.
+
+```mermaid
+flowchart LR
+      producer["Producer<br/>(API, workers)"]
+      nothing["Nothing more to be done"]
+
+      producer -->|"enqueue_job()<br/>flushed after DB commit"| high["high_priority"]
+      producer --> med["medium_priority"]
+      producer --> low["low_priority"]
+      producer --> web["webhooks"]
+      producer --> tb["tinybird"]
+      producer --> inv["invoices_and_receipts"]
+
+      high --> worker
+      med --> worker
+      low --> worker
+      web --> worker
+      tb --> worker
+      inv --> worker
+
+      worker["Pick up task → run actor"]
+
+      worker -->|"failure: exponential backoff"| worker
+      worker -.->|"enqueue new job"| producer
+      worker -->|"completed"| nothing
+```
+
+### Desired Outcomes
+
+We use async workers for many different things, but oftentimes we require the work to be conducted in fairly short succession from when it was scheduled. This proposal aims to ensure that we don't get overloaded by a sudden influx of tasks, self created or otherwise.
+Additionally it should stay easy to reason about the code running and our infrastructure, as well as to know what code is running where.
+
+## Proposal
+
+The proposal is to move the worker workloads to AWS, and set up SQS with Lambda functions acting on the messages from SQS.
+
+During roll out we will have a list of which workers run in Lambda, and which run in Dramatiq. In the enqueue_job we will use this to determine whether to enqueue the task via Dramatiq onto Redis, or onto SQS.
+
+```python
+if settings.WORKER_SQS_ENABLED and settings.WORKER_SQS_ACTORS:
+    sqs_jobs = [
+        job for job in self._enqueued_jobs
+        if job[0] in settings.WORKER_SQS_ACTORS
+    ]
+    redis_jobs = [
+        job for job in self._enqueued_jobs
+        if job[0] not in settings.WORKER_SQS_ACTORS
+    ]
+else:
+    sqs_jobs = []
+    redis_jobs = self._enqueued_jobs
+```
+
+Each task that is handled by Lambda will get its own SQS queue. This means that the infrastructure per task is 1 Lambda, 1 SQS queue, and all other resources needed to handle this (IAM policies, etc). The logs and traces from the Lambda will be forwarded to Logfire, similar to how we handle logs today. The infrastructure will be encapsulated with a terraform module to make it easy to spin up (and down) new tasks as we add them.
+
+The tasks will be enqueued somewhat similar to how we enqueue tasks to Dramatiq today.
+
+Dramatiq
+```python
+message = fn.message_with_options(
+    args=args,
+    kwargs=kwargs,
+    redis_message_id=redis_message_id,
+    source_correlation_id=correlation_id,
+)
+encoded_message = message.encode()
+```
+
+SQS
+```python
+json.dumps(
+    {
+        "actor": actor,
+        "args": args,
+        "kwargs": kwargs,
+        "correlation_id": correlation_id,
+    },
+    separators=(",", ":"),
+    default=_json_obj_serializer,
+)
+```
+
+Ending up with something like the following in SQS:
+```json
+{
+    "messageId": "d5f8…",
+    "receiptHandle": "AQEB…",
+    "body": "{\"actor\":\"event.ingested\",\"args\":[[\"0f9c…\"]],\"kwargs\":{},\"correlation_id\":\"01J8…\"}",
+    "attributes": {
+      "ApproximateReceiveCount": "1",
+      "SentTimestamp": "1718…",
+      "ApproximateFirstReceiveTimestamp": "1718…"
+    },
+    "messageAttributes": {},
+    "eventSource": "aws:sqs",
+    "eventSourceARN": "arn:aws:sqs:us-east-2:…:polar-tasks-event-ingested"
+  }
+```
+
+```mermaid
+flowchart LR
+     producer["Producer<br/>(API, workers)"]
+
+     producer -->|"enqueue_job()<br/>flushed after DB commit"| sqs_a["SQS: polar-tasks-event-ingested"]
+     producer --> sqs_b["SQS: polar-tasks-email-send"]
+     producer --> sqs_c["SQS: polar-tasks-task_c"]
+     producer --> sqs_more["..."]
+
+     sqs_a --> lambda_a["Lambda: event.ingested"]
+     sqs_b --> lambda_b["Lambda: email.send"]
+     sqs_c --> lambda_c["Lambda: task_c"]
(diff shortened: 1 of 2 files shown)


handbook/engineering/design-documents/tasks-on-lambda.mdx
```diff
@@ -0,0 +1,176 @@
...
+     lambda_c -.->|"enqueue new job"| producer
+
+     note_conc["variable concurrency<br/>(scales with load)"]
+     lambda_b -.- note_conc
+
+     classDef note fill:#fff3cd,stroke:#d39e00,color:#000;
+     class note_conc note;
+```
+
+We preserve the same retry/backoff behavior, implemented through SQS visibility/redrive instead of via the `Retry` exception handler in Dramatiq.
```
frankie567, line 158: What about other kinds of exceptions? Will they get automatic retries as well? We rely on that behavior quite a lot

handbook/engineering/design-documents/tasks-on-lambda.mdx
```diff
@@ -0,0 +1,176 @@
...
+
+## Alternatives considered
+
+N/A
+
+## Open questions
+
+### Database Connections
+
+Given that we are aiming for scaling the concurrency per lambda instead of per queue (as we have done previously) we have the risk of running out of database connections. The number of database connections is constrained by the database type we use. Here we might need to calculate what the max scaling we can afford, as well as look into introducing something like [pgBouncer](https://planetscale.com/blog/scaling-postgres-connections-with-pgbouncer) to allow us to scale up our database connections without choking the database.
```
frankie567: Yes, and more generally inter-blocking queries. I've seen during the Saturday's incident the low priority queue completely stuck because N tasks were running in parallel competing for resources. The solution was to reduce the number of processes to unblock the situation. I'm wondering if we may observe that kind of problems more often if we scale Lambdas aggressively.

frankie567 (commented):
That's super interesting, really looking forward to see this in action.

One thing that's not covered here for which I would like to have some insights is **deployment**. How does it work? Can we just deploy the bundle once and AWS will automatically replace old instances by new ones?

handbook/engineering/design-documents/tasks-on-lambda.mdx
pieterbeulque: Should we set up pgBouncer regardless to have one fewer moving part when evaluating workers vs lambdas?

Yopi: Contention is a very good point. I think it begs the question if we should consider something to allow us to partition or order messages more clearly, instead of just delaying messages to get them in the right order

handbook/engineering/design-documents/tasks-on-lambda.mdx
Yopi: This might have been a bit vaguely written. I am hoping we can achieve the exact same backoff and retry mechanisms as we have today, but via SQS's visibility function instead. So we set it to invisible for retryCount * backoff seconds

Yopi:
> One thing that's not covered here for which I would like to have some insights is **deployment**. How does it work? Can we just deploy the bundle once and AWS will automatically replace old instances by new ones?

I can add a section on deployment. In paralell I'm prototyping some code to show it more clearly.

Basically my thinking is:
* We'll build a docker image in GHA like we do for Render, and push it to ECR instead of GHCR
* We'll trigger a deployment script that will update the code running in the lambdas

frankie567:
> > One thing that's not covered here for which I would like to have some insights is **deployment**. How does it work? Can we just deploy the bundle once and AWS will automatically replace old instances by new ones?
> 
> I can add a section on deployment. In paralell I'm prototyping some code to show it more clearly.
> 
> Basically my thinking is:
> 
> * We'll build a docker image in GHA like we do for Render, and push it to ECR instead of GHCR
> * We'll trigger a deployment script that will update the code running in the lambdas

Sounds very sensible!

handbook/engineering/design-documents/tasks-on-lambda.mdx
frankie567: Hmm, not sure I see how we would partition messages? 🤔

handbook/engineering/design-documents/tasks-on-lambda.mdx
```diff
@@ -0,0 +1,176 @@
...
+
+### Rollout and rollback plan
+
+Rollout will be done on a task-by-task basis. We can consider having a fall-back lambda that reads from SQS and writes to Redis to allow us to move back in the event that something does not work out with the lambda solution.
+
+## Alternatives considered
+
+N/A
+
+## Open questions
```
psincraian, line 197: Aside of DB connections, we also have Redis connections that's much higher but we also limit on those. 

And maybe, one thing to consider is having two Redis one for the API and another one internal in AWS VPC so we reduce the outbound traffic and latency.

handbook/engineering/design-documents/tasks-on-lambda.mdx
psincraian: SQS supports FIFO queues if we want deterministic processing

handbook/engineering/design-documents/tasks-on-lambda.mdx
```diff
@@ -0,0 +1,176 @@
...
+      worker -.->|"enqueue new job"| producer
+      worker -->|"completed"| nothing
+```
+
+### Desired Outcomes
+
+We use async workers for many different things, but oftentimes we require the work to be conducted in fairly short succession from when it was scheduled. This proposal aims to ensure that we don't get overloaded by a sudden influx of tasks, self created or otherwise.
+Additionally it should stay easy to reason about the code running and our infrastructure, as well as to know what code is running where.
+
+## Proposal
```
psincraian, line 64: One thing that I would like to discuss as a followup would be dead letter queues or how to store messages that we cannot process.

Now, we ignore them. But I think having a quick thought on them now will make this easier to configure in the future

handbook/engineering/design-documents/tasks-on-lambda.mdx
psincraian, line 197: I will add some problems that I still got when working with lambdas at scale:

- We can hit ENI limits soon. Those can be raised easily, but the errors are loud. 
- Hard limit timeout is 15m. I don't think we have any tasks that takes that long

psincraian, line 197: How do you plan to implement cron jobs? In the past I have used EventBridge for that

handbook/engineering/design-documents/tasks-on-lambda.mdx
```diff
@@ -0,0 +1,176 @@
...
+    redis_jobs = [
+        job for job in self._enqueued_jobs
+        if job[0] not in settings.WORKER_SQS_ACTORS
+    ]
+else:
+    sqs_jobs = []
+    redis_jobs = self._enqueued_jobs
+```
+
+Each task that is handled by Lambda will get its own SQS queue. This means that the infrastructure per task is 1 Lambda, 1 SQS queue, and all other resources needed to handle this (IAM policies, etc). The logs and traces from the Lambda will be forwarded to Logfire, similar to how we handle logs today. The infrastructure will be encapsulated with a terraform module to make it easy to spin up (and down) new tasks as we add them.
```
psincraian: keeping things frugal and to make things easier, I think we can use only IAM policy as those count as resources in terraform

handbook/engineering/design-documents/tasks-on-lambda.mdx
```diff
@@ -0,0 +1,176 @@
...
+
+### Local development
+
+Local development will utilise [localstack](https://github.com/localstack/localstack) which has support for both SQS and Lambda, and will be encapsulated in the `dev` commands used by the engineering team.
+
+### Rollout and rollback plan
+
+Rollout will be done on a task-by-task basis. We can consider having a fall-back lambda that reads from SQS and writes to Redis to allow us to move back in the event that something does not work out with the lambda solution.
+
+## Alternatives considered
```
psincraian, line 186: Maybe as an alternative approach we can consider ECS that reads from SQS. I think by default it can scale based on 1 sqs, but you can create custom metrics to be able to scale based on different ones. 

This move will be more similar to our current setup but on AWS.

psincraian (commented):
Really nice job! I only have a couple of important questions that I would love your thoughs on them

handbook/engineering/design-documents/tasks-on-lambda.mdx
Yopi: That's fair, yes. All of queues and lambdas can use the same policies, since they should basically be the same as well

handbook/engineering/design-documents/tasks-on-lambda.mdx
Yopi: I was thinking to postpone that and wait with moving the Scheduler out from workers, but it's a good question!

 I think EventBridge is pretty standard for that type of usage, but it would be nice to not split out the cron definitions from the python code.

Yopi: Yes, it would be very nice to segregate Redis to be honest. The only thing that we have to be mindful of is if we have things that span across the lambda workers and the regular workers / API.

In the case for locks, most of them are probably within the same task, so that should be fine.

handbook/engineering/design-documents/tasks-on-lambda.mdx
Yopi: Exactly, I'm also not sure what we want to do with them. Usually we handle broken things fairly quickly, but our current set up lets them live for quite a long time with retries. It's ideal if we never lose messages, even if they are broken in one way or another.

handbook/engineering/design-documents/tasks-on-lambda.mdx
Yopi: I also don't think we have any that runs for that long. I think a minute is where we are aiming at, so that should be fine.

If we reuse the same security groups and VPC subnets for all of the lambdas, we will be able to reuse the same [Hyperplane ENI](https://docs.aws.amazon.com/lambda/latest/dg/configuration-vpc.html#configuration-vpc-best-practice). This will give us 65k connections per subnet. These will scale automatically behind the scenes, so even if we exhaust them we will get a new Hyperplane ENI. That takes some time though, so the task requiring it will be delayed.
 
> Hyperplane ENIs are associated with a particular combination of security groups and VPC subnets. If you’ve already attached one function to a VPC, specifying the same subnets and security groups when you attach another function means that Lambda can share the network resources and avoid the need to create a new Hyperplane ENI

It looks like the default VPC ENI limit is 500, so if we were to run 3 subnets * security group per lambda, we could probably hit that sooner. 

But both of those things are good points that we can cover in the RFC.

handbook/engineering/design-documents/tasks-on-lambda.mdx
Yopi: Yes, I can add ECS as an alternative. Personally I think that it's very close to the existing setup already, so it would be a question if that would help us in any way (other than it would move us off Render). Are you seeing any additional benefits that I might be missing?

Yopi:
Thank you @frankie567 and @psincraian. Very good feedback and questions!

handbook/engineering/design-documents/tasks-on-lambda.mdx
Yopi: I was thinking if we were partioning e.g. by organization, customer then there wouldn't be any competition for resources and we would not run into that issue. Deterministic processing or ordering is another solution, but if they the competing tasks come in close succession they would still get spun up concurrently and run concurrently

PR #12503, opened 2026-06-18
feat(worker): run tinybird worker on a DB-less broker

joebon:
The tinybird worker (i.e. the tasks scheduled on the queue it consumes) never use the db, so we shouldn't set up an idle db connection pool for it.

---
## Summary by cubic
Run the Tinybird worker on a DB‑less `dramatiq` broker to avoid creating a Postgres engine/pool. Other workers keep using the default DB-backed broker.

- **Refactors**
  - `get_broker(*, database: bool = True)` now conditionally adds `SQLAlchemyMiddleware`.
  - Added `polar.worker.run_without_db` that sets a DB‑less broker before importing `polar.tasks`.
  - Updated production and sandbox Terraform to start the Tinybird worker with `polar.worker.run_without_db` (queues/flags unchanged).

<sup>Written for commit 3a737453451f19cd1f2a780a916a28cc114d249a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12503?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/worker/run_without_db.py
+++ b/server/polar/worker/run_without_db.py
@@ -1,0 +1,26 @@
+import dramatiq
+
+from polar.worker._broker import get_broker
+from polar.worker._encoder import JSONEncoder
+
+# Entrypoint for workers whose queues never touch PostgreSQL. Building the broker
+# without the SQLAlchemy middleware avoids creating a database engine/pool the
+# worker would never use. The broker must be set as the global broker before
+# `polar.tasks` is imported, so actors are declared against it rather than the
+# default (database-backed) broker created by `polar.worker`.
+broker = get_broker(database=False)
+dramatiq.set_broker(broker)
+dramatiq.set_encoder(JSONEncoder(broker))
+
+from polar import tasks  # noqa: E402
+from polar.logfire import configure_logfire  # noqa: E402
+from polar.logging import configure as configure_logging  # noqa: E402
+from polar.posthog import configure_posthog  # noqa: E402
+from polar.sentry import configure_sentry  # noqa: E402
+
+configure_sentry()
+configure_logfire("worker")
+configure_logging(logfire=True)
+configure_posthog()
+
+__all__ = ["broker", "tasks"]
--- a/server/polar/worker/_broker.py
+++ b/server/polar/worker/_broker.py
@@ -175,7 +175,7 @@ class LogfireMiddleware(dramatiq.Middleware):
         return self.after_process_message(broker, message)
 
 
-def get_broker() -> dramatiq.Broker:
+def get_broker(*, database: bool = True) -> dramatiq.Broker:
     redis_pool = redis.ConnectionPool.from_url(
         settings.redis_url,
         client_name=f"{settings.ENV.value}.worker.dramatiq",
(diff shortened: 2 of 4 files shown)


Yopi (approved):
Cool. Did you try it out locally?

server/polar/worker/run_without_db.py
```diff
@@ -0,0 +1,26 @@
...
+
+from polar.worker._broker import get_broker
+from polar.worker._encoder import JSONEncoder
+
+# Entrypoint for workers whose queues never touch PostgreSQL. Building the broker
+# without the SQLAlchemy middleware avoids creating a database engine/pool the
+# worker would never use. The broker must be set as the global broker before
+# `polar.tasks` is imported, so actors are declared against it rather than the
+# default (database-backed) broker created by `polar.worker`.
+broker = get_broker(database=False)
```
frankie567, line 11: To avoid having to duplicate the run script, I think we can port that approach (i.e. declare broker here) to `run.py` and have a dedicated environment variable to enable/disable the database connection. WDYT?

joebon: I like the explicitness of this approach tbh. And I think it makes sense to make the distinction in the code rather than in environment config, since the code knows of the requirements of the tasks, queues and workers. So duplicating the run script is a fairly small price to pay imho.
But I'm not fanatic about it 😅

joebon:
> Cool. Did you try it out locally?

@Yopi yes!

The worker runs fine locally and picks up messages - checked like this:
```
server % uv run python -c "
import dramatiq
import polar.worker.run_without_db
dramatiq.get_broker().get_actor('tinybird.ingest').send([])
"
```

server/polar/worker/run_without_db.py
frankie567: Fair enough

frankie567:
Did you check that this worker alone actually opens up connections? If I understand the SQLAlchemy docs correctly, the pool remains empty until a connection is requested: https://docs.sqlalchemy.org/en/21/core/pooling.html#sqlalchemy.pool.QueuePool.params.pool_size

So, even if we instantiate the engine, this worker should not open connections

joebon:
> Did you check that this worker alone actually opens up connections? If I understand the SQLAlchemy docs correctly, the pool remains empty until a connection is requested: https://docs.sqlalchemy.org/en/21/core/pooling.html#sqlalchemy.pool.QueuePool.params.pool_size
> 
> So, even if we instantiate the engine, this worker should not open connections

Good point! 😅 
I'd argue this could make sense nonetheless. As a follow-up I want to add a health check to workers that _need_ database connections so they don't get marked as healthy if they cannot reach the db.
That would prevent old workers from being shut down if we accidentally try to deploy new workers that for some reason have broken db access - which happened recently.

I'd consider this a nice thing to fix before, so that health/liveness check doesn't include the tinybird worker since it doesn't need a db.

WDYT @frankie567 ?

frankie567:
Yes the healthcheck definitely makes sense. I'm still not 100% bought on the added complexity of disabling but DB, but well, let's see.

frankie567 (approved):

PR #12508, opened 2026-06-18
refactor(enums): introduce RecurringInterval as the canonical interval type

joebon:
Part 1 of 3 of the **decouple meter cycle from billing cycle** stack (polarsource/feedback#154).

## What

- Rename enum `SubscriptionRecurringInterval` to the more general `RecurringInterval`
- Add alias `SubscriptionRecurringInterval` => `RecurringInterval`
- Add alias `MeterInterval` => `RecurringInterval`

## Why

A preparation for introducing a "meter cycle" for subsciptions, where meters (and meter credits) are benefits connected to a subscription that need to cycle on a different cadence than the subscription billing.

## Stack

1. **→ this PR** - `recurring-interval-rename` (base: `main`)
2.  #12509 - product/subscription `meter_interval` fields + divisibility validator
3. #12510 - two-clock scheduler + carried-overage meter-cycle settlement

### DIFF
--- a/server/polar/enums.py
+++ b/server/polar/enums.py
@@ -55,7 +55,7 @@ class PayoutAccountType(StrEnum):
         }[self]
 
 
-class SubscriptionRecurringInterval(StrEnum):
+class RecurringInterval(StrEnum):
     day = "day"
     week = "week"
     month = "month"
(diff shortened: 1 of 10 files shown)


server/polar/enums.py
```diff
@@ -66,24 +66,34 @@ def as_literal(self) -> Literal["day", "week", "month", "year"]:
...
 
 
+# `RecurringInterval` is the canonical interval-unit enum (day/week/month/year),
+# shared by the billing interval and the usage interval.
+#
+# Backwards-compatible alias: the historical name stays valid for existing imports.
+SubscriptionRecurringInterval = RecurringInterval
+# Semantic alias for the product/subscription usage cycle. Same underlying values;
+# the distinct name documents intent at usage sites.
+UsageInterval = RecurringInterval
```
frankie567: I'm wondering if we should either:

* Replace imports everywhere to be `RecurringInterval`
* Create `NewType` for each one so we can have some type safety to not shuffle them — but might be a darling

joebon: Good point! 

```
SubscriptionRecurringInterval = NewType('SubscriptionRecurringInterval`, RecurringInterval)
```

This unfortunately leads to member accesses like `SubscriptionRecurringInterval.month` raising `RuntimeError`s - not even caught by type checks.
So in that case we'd also need to switch all member accesses to go to `RecurringInterval.<member>` directly which defeats the purpose imho.

What's the clear advantage of doing `RecurringInterval` directly everywhere @frankie567 ?

frankie567: Ok, let's forget the newtype.

> What's the clear advantage of doing RecurringInterval directly everywhere @frankie567 ?

Well, just consistency 😄 I've no technical argument :p

joebon: Makes sense! I agree with you that I'd want the type safety in place, that's the whole point with the rename. But aliasing def isn't the same as inheritance so it kinda breaks. Should I perhaps just go back and define a new explicit type `UsageInterval` that's a carbon copy of `SubscriptionRecurringInterval`?

frankie567: Hmm 🤔 I think we might overthink this. Yes, there is no type safety in the current state, but is it such a big deal? We just store those values and use them to compute the next cycle; I'm not really sure there is a strong advantage to constraint that much more

joebon: Fair enough, but I'm confused about what that means for this PR: good to go as is?

frankie567: Yes, go

frankie567 (approved):

sebastianekstrom (approved):

PR #12509, opened 2026-06-18
feat(product): add meter interval foundation for decoupled meter cycling

joebon:
Part 2 of 3 of the **decouple meter cycle from billing cycle** stack (polarsource/feedback#154).

> Stacked on #12508 — review/merge that first.

## What

Adds the data foundation for a meter cycle independent of the billing cycle:

- `meter_interval` / `meter_interval_count` on **Product** (nullable, immutable like `recurring_interval`) + schema wiring.
- `current_meter_period_start` / `current_meter_period_end` on **Subscription**, snapshotted from the product at creation/checkout and re-initialised at trial conversion and product change.
- The **divisibility validator**: the meter interval must evenly divide the billing interval (the full interval matrix from the design doc), validated once at product creation so a misaligned pair can never exist at runtime.
- Alembic migration for the new columns (`sa.String()`, no app-enum import).

## Why

A merchant wants to bill yearly but cycle credits monthly. The meter interval expresses that cadence; the divisibility rule guarantees meter and billing boundaries re-align at every renewal.

## Notes

- No behavior change yet — nothing cycles on the meter interval until Part 3. Data + validation only.
- Default behavior unchanged: products without `meter_interval` behave exactly as today.

## Stack

1. #12508  (base: `main`)
2. #12509 - product/subscription `meter_interval` fields + divisibility validator. **this PR**
3. #12510 - two-clock scheduler + carried-overage meter-cycle settlement


server/polar/product/schemas.py
```diff
@@ -474,6 +480,45 @@ class ProductCreateRecurring(TrialConfigurationInputMixin, ProductCreateBase):
             "if set to 2 it will be every other month, and so on."
         ),
     )
+    meter_interval: MeterInterval | None = Field(
```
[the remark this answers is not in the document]
frankie567: That's already the case for all the other fields in that schema. We're fine with it.

server/polar/product/schemas.py
```diff
@@ -813,6 +858,17 @@ class ProductBase(TrialConfigurationOutputMixin, TimestampedSchema, IDSchema):
             "None for one-time products."
         )
     )
+    meter_interval: MeterInterval | None = Field(
+        default=None,
```
frankie567: Don't add a `default` for output schemas: they _can_ be None but we _need_ to provide a value. It's particularly impactful for OpenAPI schemas: with a default, this will mark the field as **not required**. Therefore, for TS SDK, it'll generate an optional field like this:

```ts
interface Product {
    meter_interval?: MeterInterval | null
}
```

Which is not semantically correct for an output schema.

@cubic-dev-ai Please add this to your learnings.

server/polar/enums.py
```diff
@@ -94,6 +94,52 @@ def get_next_period(self, d: datetime, anchor_day: int, leap: int = 1) -> dateti
 MeterInterval = RecurringInterval
 
 
+def meter_interval_divides_billing_interval(
```
frankie567: Maybe we should put that in a dedicated module?

server/polar/enums.py
```diff
@@ -94,6 +94,52 @@ def get_next_period(self, d: datetime, anchor_day: int, leap: int = 1) -> dateti
...
+        case (RecurringInterval.month, RecurringInterval.month) | (
+            RecurringInterval.year,
+            RecurringInterval.year,
+        ):
+            return n % m == 0
+        case (RecurringInterval.month, RecurringInterval.year):
+            return (n * 12) % m == 0
+        # Anything else is impossible (meter interval coarser than billing) or would drift.
+        case _:
+            return False
```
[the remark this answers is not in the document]
joebon: Nope, that's deliberately defined as invalid, we don't allow for instance having a monthly subscription with yearly meter cycling

server/polar/product/schemas.py
joebon: Great, thanks! Addressed

server/polar/enums.py
joebon: Makes sense

frankie567 (approved):

sebastianekstrom (approved):

PR #12510, opened 2026-06-18
feat(subscription): settle meter cycles independently of billing

joebon:
## Summary

**Related Issue**: polarsource/feedback#154

Run metered concerns — overage settlement, meter resets, meter-credit grants — on a product's meter interval, independently of the billing interval (e.g. bill yearly but meter monthly). Builds on the merged `meter_interval` fields (#12509). Default behavior is unchanged when no meter interval is set.

## What

- Two-clock scheduler: wakes on `current_period_end` or `current_meter_period_end`, whichever fires first.
- Each meter boundary settles the period's usage into a `subscription_meter_cycle` order, then resets meters and re-grants credits.
- Settlement sweeps the pending metered billing entries (same path as the billing cycle); sub-minimum amounts go to the customer balance; a boundary coinciding with billing folds into the renewal invoice (one order).
- Multi-period lag halts + pages instead of replaying; failed meter charges dun retry-only without changing subscription status.

## Why

Credits and overage were tied to the billing interval — an annual "100 credits/month" plan granted all 1,200 at once and only settled overage yearly. Merchants want to bill yearly but pace usage monthly.

## How

- Snapshot the meter interval + second clock onto the subscription; add the two-clock scan.
- Settle by sweeping pending metered entries (`create_metered_order_items_from_pending`), keeping overage on the same basis as rollover.
- At the billing boundary, skip separate settlement and let the renewal sweep carry the usage.
- Coverage: `tests/e2e/lifecycle/test_meter_cycle.py` + service/order/subscription unit tests.

## Checklist

- [x] This PR addresses a single concern (decoupling the meter cycle from billing)


server/polar/models/order.py
```diff
@@ -57,13 +57,15 @@ class OrderBillingReasonInternal(StrEnum):
     subscription_cycle_after_trial = "subscription_cycle_after_trial"
     subscription_cancel = "subscription_cancel"
     subscription_update = "subscription_update"
+    subscription_meter_cycle = "subscription_meter_cycle"
```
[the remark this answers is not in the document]
joebon: @stilla-bot create an issue for this and assign me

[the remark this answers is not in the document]
joebon: Will be fixed separately

server/polar/order/service.py
```diff
@@ -1508,6 +1526,101 @@ async def _create_order(
...
+        for subscription_meter in subscription.meters:
+            price = self._active_metered_price(
+                subscription, subscription_meter.meter_id
+            )
+            # Without an active metered price there's nothing to bill against, so
+            # don't accumulate overage we could never charge — it would grow
+            # unbounded and surface as a surprise lump charge if a price reappears.
+            if price is None:
+                continue
+            overage = await customer_meter_service.get_overage_units(
```
frankie567: We shouldn't rely on customer meter calculations to invoice metered usage. Billing Entries are the source of truth and should be used. Otherwise, we lose the ability to track billing down to the events.

server/polar/subscription/service.py
```diff
@@ -887,6 +887,15 @@ async def cycle(
...
             subscription.status = SubscriptionStatus.active
 
+        # Sweep the closing meter period before re-arming the clock: settle carried
+        # overage and consume the metered entries (force=True — terminal boundary),
+        # so the renewal order enqueued below bills only static prices. A trial has
+        # no meter clock, so this is a no-op until conversion.
+        if subscription.current_meter_period_end is not None:
+            await self._settle_meter_cycle(
+                session, subscription, subscription.current_meter_period_end, force=True
+            )
```
frankie567: I would expect the overages to be billed on the same invoice as the cycle ones. It's a bit unexpected to have two orders and two payments IMO.

server/polar/subscription/service.py
```diff
@@ -1311,6 +1416,17 @@ async def update_product(
...
                         session.add(entry)
 
+                # Settle the old product's carried overage before the switch: its
+                # metered prices and meters disappear once apply_update() runs.
+                # force=True sweeps any remainder (terminal for the old product).
+                if (
+                    not was_trialing
+                    and subscription.current_meter_period_end is not None
+                ):
+                    await self._settle_meter_cycle(
```
frankie567: Billing entries should have been written for the consumed units; so they should be naturally billed for the next time we issue an invoice.

frankie567 (commented):
I think we have the right behavior, but I'm not aligned with the approach here.

Whenever an event is ingested, we create the corresponding billing entry. Then, when we want to issue an invoice, we only have to look at the pending entries — we shouldn't make any calculations or rely on pre-aggregated numbers.

I think we can simply have the same logic to generate a subscription cycle order and meter cycle order; only the reason changes.

server/polar/order/service.py
joebon: Thanks! I'll align with that

joebon:
> I think we have the right behavior, but I'm not aligned with the approach here.

Great! Let's align on the approach and make sure all gotcha's are at least surfaced 

> Whenever an event is ingested, we create the corresponding billing entry. Then, when we want to issue an invoice, we only have to look at the pending entries — we shouldn't make any calculations or rely on pre-aggregated numbers.

Yes, that mechanic is great! It does result in the "evaporation" case though:
1. A customer over-consumes towards a meter worth $0.30 during a meter cycle - below the threshold for issuing an invoice. We leave the corresponding billing entries as pending
2. Next period, we sweep all pending billing entries again. That sweeping swallows the ones left from the earlier cycle period and if that period's usage is less than the meter credit then the last periods overage is gone.

The point with `carried_units` was to cater for that specific scenario. That or some other bucketed construction for overage usage per meter cycle, when we finally exceed the invoice amount threshold we create an invoice with line items like this:
```
- [Meter name] - April 2026       $0.30
- [Meter name] - June 2026        $0.45
```

It's an extra hassle for sure compared to the "just sweep all pending" approach, but it's part of DoD in https://github.com/polarsource/feedback/issues/154 as I understand it.

Do you have an alternative take on it @frankie567 ?

server/polar/subscription/service.py
```diff
@@ -953,6 +962,102 @@ async def reset_meters(
...
+        ):
+            # No meter cycle to run, but the scheduler locked the row on dispatch —
+            # clear the lock, or the subscription is never picked up again (billing
+            # renewal included).
+            repository = SubscriptionRepository.from_session(session)
+            return await repository.update(
+                subscription, update_dict={"scheduler_locked_at": None}
+            )
+
+        # Advance the clock first, looping in case the worker lagged several
```
psincraian: I think this while can be a problem on the computation. If the cycle don't run for 2 cycles, we have a problem with credits computation on this case as we missed the grants for the inbetween cycles. 


Maybe instead of this looping, we should have some proactive alert that the cycles weren't run.

server/polar/order/service.py
```diff
@@ -2853,6 +2975,14 @@ async def handle_payment_failure(
             )
             return order
 
+        # Meter-cycle settlement failures retry the payment but never escalate the
```
psincraian: And what happens if the order never gets paid? I think we should notify the merchant somehow

psincraian (commented):
This is a nice and complex tax! Good job Joel

server/polar/order/service.py
joebon: Yeah, tbh I've left it out of scope of this PR. Thought it was challenging enough as is. But yeah, there should be a notification that we've given up charging for a usage invoice.

psincraian: Makes sense to for another PR 👍

server/polar/subscription/service.py
joebon: I hear ya and see your point!
But I'm not entirely sure I agree - or at least I don't see it as crystal clear.
With different cadences for billing vs meter cycling there might regardless be usage invoices (depending on meter credits vs usage) in-between billing cycling, so this is only relevant when billing and usage cycles coincide.
Is it then a better experience to _always_ invoice meter usage separately or is it better to merge the invoices for example every 12 months because then we also charge for the subscription itself?

server/polar/order/service.py
joebon: @stilla-bot create an issue for this and assign me to it

server/polar/subscription/service.py
joebon: Ouch, nice catch! Will rework to make sure we catch up period by period instead of squashing all missed periods into one

frankie567:
> Yes, that mechanic is great! It does result in the "evaporation" case though:
> 
> 1. A customer over-consumes towards a meter worth $0.30 during a meter cycle - below the threshold for issuing an invoice. We leave the corresponding billing entries as pending
> 2. Next period, we sweep all pending billing entries again. That sweeping swallows the ones left from the earlier cycle period and if that period's usage is less than the meter credit then the last periods overage is gone.

Maybe @pieterbeulque won't like it but I think in this case we should rely on the builtin Order and Customer Balance mechanism for that: too small orders are already handled and do the following:

* If the amount is below <0.50$, mark it as paid immediately, don't trigger payments
* Add that to the customer balance; it's automatically added to the next Order we generate for the customer

This way, we don't have to care about that problem at that level, it's handled as any other small amount

pieterbeulque:
@frankie567 Totally fine! I think it's the right call to hook it into the existing behavior for handling too small charges. If we automatically add the customer balance to the next order, then it's effectively the same behavior from the customer's perspective, right?

One question: do we consider the outstanding customer balance when deciding to charge an order, or not? In other words, do we need a standalone order to be more than $0.50 to trigger the charge (and then add the customer balance on top of that order), or do we already look at "this order amount + outstanding customer balance" when deciding to charge? Not sure if this phrasing makes sense.

server/polar/subscription/service.py
pieterbeulque: If a subscription cycles on July 15, it'd be cool if we could apply the overages of June 15 – July 15 on the invoice of the renewal on July 15.

Assuming it's as "simple" as skipping the meter cycle if it happens to coincide with the billing cycle, and everything else (billing entries generation) is unaffected. Put differently, in my mental model, any new order creation will pick up the billing entries created by the meter cycle, so it's only the invoicing part that we have to skip on this `meter_cycle coincides with billing_cycle` conditional.

If it's a big increase in complexity, then I'm fine with a double invoice in a first version.

frankie567:
> @frankie567 Totally fine! I think it's the right call to hook it into the existing behavior for handling too small charges. If we automatically add the customer balance to the next order, then it's effectively the same behavior from the customer's perspective, right?

From a customer perspective, they receive an invoice with a total of `0.30$`, but with an "Amount to paid" of "$0". So something happens.

> One question: do we consider the outstanding customer balance when deciding to charge an order, or not? In other words, do we need a standalone order to be more than $0.50 to trigger the charge (and then add the customer balance on top of that order), or do we already look at "this order amount + outstanding customer balance" when deciding to charge? Not sure if this phrasing makes sense.

Yes, the total amount with outstanding balance is considered. Because it also needs to work on the reverse side: if we owe money to the customer, we should use that money to cover the new charge.

pieterbeulque:
Sweet! Let's use that then, sorry for not having considered that in my original spec, I wasn't aware.

server/polar/subscription/service.py
joebon: @psincraian no more while-loop and try to catch up, I switched to checking if we've exceeded a lag threshold and in that case we bail so the cycling stops and then we must fix it manually. Makes sense?

server/polar/subscription/service.py
joebon: @pieterbeulque shouldn't be a problem!

joebon:
@frankie567 thanks a gazillion for the clarifications - _man_ was I out on a goose chase! Reverted now and hopefully way more in line with the original reasoning.

server/polar/subscription/service.py
```diff
@@ -1147,6 +1164,94 @@ async def reset_meter(
...
+    ) -> None:
+        """
+        Settle a meter period's usage into a meter-cycle order. The order import
+        is local because the order service imports this one at module load.
+        """
+        from polar.order.service import order as order_service
+
+        await order_service.settle_meter_cycle_order(session, subscription)
+
+    async def cycle_meters(
```
psincraian, line 1178: I see that in `cycle` we do check the discount expiration date but not here. If the discount has expired we should also check this. Maybe we should extract the logic to be the same on both places.

server/polar/order/service.py
```diff
@@ -3084,6 +3127,35 @@ async def _handle_consecutive_dunning_attempts(
 
         return order
 
+    async def _handle_meter_cycle_dunning_attempt(
```
psincraian, line 3130: So, if there is a yearly subscription with monthly meters and the card is expired. The subscription is still kept to active and nothing stops it, right? That can be problematic for merchants

server/polar/subscription/service.py
joebon: Ah, good catch - thanks! Will fix

server/polar/order/service.py
joebon: Yeah, it's mentioned in the [underlying feature request discussion](https://github.com/polarsource/feedback/issues/154#issuecomment-4688944109) but not clearly decided on:
> *Payment-failure semantics for usage orders (group discussion)*
> …
> I feel like a small failed overage charge shouldn't lead to cancelling a prepaid annual subscription, e.g. $6 of overage cannot kill a $1,200 contract.

and 

[this comment](https://github.com/polarsource/feedback/issues/154#issuecomment-4726632144) by me:
> I think we should start with option 2. If a specific customer/subscription becomes a problem for a merchant the merchant will reasonably know. And if it happens for many subsciptions/customer the merchants will reasonably let us know that it's a problem and then we can re-evaluate if there's a pattern that should lead to subscription termination aswell.

What's your take on it given this context @psincraian ?

server/polar/subscription/service.py
joebon: Actually, it turns out that discounts never expire mid billing cycle so this wasn't actually problem. Nevertheless the expiration check logic could change in the future of course so added it anyway.

server/polar/subscription/service.py
```diff
@@ -1147,6 +1164,99 @@ async def reset_meter(
...
+        if next_period_end <= now:
+            # >1 period elapsed. Replaying per period is unsafe — async grants would
+            # race the next settlement — so halt (lock left set) and page for a
+            # manual catch-up.
+            raise SubscriptionMeterCycleLag(subscription, boundary, now)
+
+        # Drop a lapsed discount before settling: the meter-cycle order applies
+        # `subscription.discount`, so mirror the billing cycle's expiry check
+        # instead of relying on it to have run first.
+        self._clear_expired_discount(subscription)
```
joebon, line 1242: @psincraian here!

server/polar/order/service.py
psincraian: Sounds reasonable, I will say that this will become a problem and the merchant will want to have some way of cancelling subscriptions if the meter orders are not fullfilled. But we can see

psincraian (approved):
Nice job! This looks fine now

PR #12511, opened 2026-06-18
tf/aws: Set up dummy lambda

Yopi:
This is the first worker lambda that we have running. We'll use it as testing grounds

---
## Summary by cubic
Set up a sandbox dummy worker Lambda using a Lambda-ready Docker image with Tailscale, running inside a sandbox VPC with a static NAT egress IP for end-to-end testing. Enabled Postgres TLS and connected the worker to the Render external DB host; CI and IAM updated so images push and deploy to sandbox.

- **New Features**
  - Provisions ECR repo `polar-sandbox-lambda-worker` and a `dummy` worker bound to `polar-sandbox-tasks-dummy`; configures env/secret vars (incl. `TAILSCALE_AUTHKEY`); disabled by default.
  - Adds a sandbox VPC with private subnets, NAT gateway, and a static egress IP; creates a shared Lambda security group; passes subnet/SG IDs to the worker and outputs the egress IP for DB allow lists.
  - Updates the Lambda image to `public.ecr.aws/lambda/python:3.14` with a Tailscale bootstrap and handler `polar.worker.aws_lambda.handler`; adds Postgres TLS support (`POSTGRES_SSL`) and points the worker to the public DB host.

- **Bug Fixes**
  - Grants `ecr:BatchGetImage`, includes the sandbox repo in ECR push permissions, updates Lambda deploy policy ARNs to `polar-*-worker-*`, adds AZ validation to the VPC module, and fixes parsing of the Render external DB host via regex.

<sup>Written for commit 2e82fce7703550a36b3a82b3ed570253efafe05b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12511?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -130,9 +130,86 @@ FROM production AS playwright
 RUN uv run playwright install --with-deps chromium
 
 # Stage 7: AWS Lambda image (SQS-driven background tasks, POC)
-FROM production AS lambda
+FROM public.ecr.aws/lambda/python:3.14 AS lambda-build
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
 
-RUN uv pip install awslambdaric
+ENV PYTHONUNBUFFERED=1
+ENV UV_COMPILE_BYTECODE=1
+ENV UV_LINK_MODE=copy
+ENV UV_NO_CACHE=1
+ENV UV_NO_SYNC=1
 
-ENTRYPOINT ["uv", "run", "python", "-m", "awslambdaric"]
+WORKDIR /app/server
+
+RUN --mount=type=bind,source=uv.lock,target=uv.lock \
+  --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+  dnf install -y gcc gcc-c++ git make \
+  && uv sync --no-group dev --frozen \
+  && dnf clean all
+
+ADD . /app/server/
+
+COPY --from=build-emails-bin /app/bin/react-email-pkg /app/server/
+ENV POLAR_EMAIL_RENDERER_BINARY_PATH=/app/server/react-email-pkg
+
+COPY --from=build-backoffice /app/static /app/server/polar/backoffice/static
+COPY --from=download-ipinfo /data /data
+COPY --from=download-cjk-fonts /fonts/ /app/server/polar/invoice/fonts/
+
+ARG RELEASE_VERSION
+ENV RELEASE_VERSION=${RELEASE_VERSION}
+
+FROM public.ecr.aws/lambda/python:3.14 AS lambda
+
+ARG RELEASE_VERSION
+ENV RELEASE_VERSION=${RELEASE_VERSION}
+
+COPY --from=lambda-build /app /app
+COPY --from=lambda-build /data /data
+COPY --from=docker.io/tailscale/tailscale:stable /usr/local/bin/tailscaled /var/runtime/tailscaled
+COPY --from=docker.io/tailscale/tailscale:stable /usr/local/bin/tailscale /var/runtime/tailscale
+
+ENV LAMBDA_TASK_ROOT=/app/server
+ENV PATH="/app/server/.venv/bin:$PATH"
+ENV PYTHONPATH="/app/server/.venv/lib/python3.14/site-packages:/app/server"
+ENV POLAR_EMAIL_RENDERER_BINARY_PATH=/app/server/react-email-pkg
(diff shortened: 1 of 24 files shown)


Yopi:
Actually, lets put it in sandbox. Since we're not deploying actively to test we might need to roll out python changes as well

server/Dockerfile
```diff
@@ -130,9 +130,86 @@ FROM production AS playwright
...
+  export ALL_PROXY="${ALL_PROXY:-socks5://localhost:1055/}"
+fi
+
+exec /lambda-entrypoint.sh "$@"
+BOOTSTRAP
+chmod +x /var/runtime/polar-bootstrap
+EOF
 
-ENTRYPOINT ["uv", "run", "python", "-m", "awslambdaric"]
+ENTRYPOINT ["/var/runtime/polar-bootstrap"]
```
frankie567, line 214: So I guess it works for the API and current workers because we override the ENTRYPOINT/CMD, is that it? Feels a bit scary at first read :D

terraform/sandbox/aws.tf
```diff
@@ -21,6 +21,57 @@ import {
...
+  environment_variables = {
+    POLAR_ENV                     = "sandbox"
+    POLAR_BASE_URL                = "https://sandbox-api.polar.sh"
+    POLAR_FRONTEND_BASE_URL       = "https://sandbox.polar.sh"
+    POLAR_CHECKOUT_BASE_URL       = "https://sandbox-api.polar.sh/v1/checkout-links/{client_secret}/redirect"
+    POLAR_JWKS                    = "/tmp/jwks.json"
+    POLAR_LOG_LEVEL               = "INFO"
+    POLAR_TESTING                 = "0"
+    POLAR_POSTGRES_DATABASE       = "polar_sandbox"
+    POLAR_POSTGRES_HOST           = local.db_internal_host
```
frankie567: I don't think internal host will work? Lambdas will run outside the Render VPC?

Yopi: It won't work, but not for this reason 😄. It will be available via Tailscale, but we need to add something that allows is directly connectable in the lambda to the DB. Something like [proxychains](https://github.com/haad/proxychains) possibly, or if Tailscale supports it out of the box. I'll have a look

server/Dockerfile
Yopi: Exactly, each of the named stages (AS production, AS playwright, AS lambda) has their own ENTRYPOINT/CMD and that is what ends up as the pushed build when we do e.g.
```
      - uses: useblacksmith/build-push-action@v2
        id: push
        with:
          push: true
          tags: ${{ steps.meta.outputs.tags }}
          labels: ${{ steps.meta.outputs.labels }}
          context: ./server
          target: production
          platforms: linux/amd64
```

But I also agree that it feels a bit scary (and big) to put all of these into the same Dockerfile. 

This specific change doesn't impact anything though since we were already doing it from before

terraform/sandbox/aws.tf
Yopi: Switched it up now. So instead we have

(AWS) -> (Render)
lambda -> NAT gateway -> Internet Gateway (static IP) -> Render (DB external host)

terraform/modules/vpc/main.tf
```diff
@@ -0,0 +1,76 @@
...
+  count = length(var.availability_zones)
+
+  vpc_id            = aws_vpc.this.id
+  cidr_block        = cidrsubnet(var.cidr_block, 4, count.index + 1)
+  availability_zone = var.availability_zones[count.index]
+
+  tags = merge(var.tags, { Name = "${var.name}-private-${var.availability_zones[count.index]}" })
+}
+
+resource "aws_nat_gateway" "this" {
```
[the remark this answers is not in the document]
Yopi: Yes, that's fine for now.

terraform/modules/vpc/main.tf
```diff
@@ -0,0 +1,76 @@
+resource "aws_vpc" "this" {
```
frankie567, line 1: VPC here we are!

frankie567 (approved):

PR #12536, opened 2026-06-19
Manage subscription.net_amount from order or checkout

pieterbeulque:
This is a weird PR, but exploring a fix to polarsource/feedback#179.

The reason we also have to look at a checkout is that that's the only place where we have tax information for a trial subscription (the order `amount` for a starting trial is `0`), otherwise we could've worked from an order only.

I don't really like how this is a side-effect that has to be called from the three order creation paths.

Other than that, it's pretty straight-forward, we calculate the tax rate (either from the order's net_amount vs taxable base), or we re-apply the same tax rate on `subscription.amount` vs `subscription.net_amount` when a subscription is updated to keep the same tax rate.

---
## Summary by cubic
Derives `subscription.net_amount` from the tax behavior of the originating order/checkout and keeps the effective tax ratio when amounts change. Adds a backfill script to fix existing tax-inclusive subscriptions.

- **Bug Fixes**
  - Added `Subscription.update_net_amount_from(charge)` to apply the charge’s tax behavior: back out tax for inclusive; keep gross for exclusive; ignore $0/credits/failed tax.
  - `update_amount_and_currency` preserves the previous amount/net_amount ratio; falls back to gross when no prior ratio exists; TODO to reset when a plan change modifies `tax_behavior`.
  - Applied in checkout, subscription order, and trial order flows so `net_amount` is correct from day one.

- **Migration**
  - Added `server/scripts/backfill_subscription_net_amount_inclusive_tax.py` to recompute `net_amount` for inclusive-tax subscriptions using the latest qualifying order; batched and idempotent. Run after deploy to correct existing data.

<sup>Written for commit 3ed4abf9eb667024e8dfd5c2027032de3f2bcd06. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12536?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/tests/order/test_service.py
+++ b/server/tests/order/test_service.py
@@ -1047,6 +1047,15 @@ class TestCreateSubscriptionOrder:
         assert order.billing_reason == OrderBillingReasonInternal.subscription_cycle
         assert order.subscription == subscription
 
+        if tax_behavior == TaxBehavior.inclusive:
+            assert subscription.net_amount == round(
+                subscription.amount
+                * order.net_amount
+                / (order.net_amount + order.tax_amount)
+            )
+        else:
+            assert subscription.net_amount == subscription.amount
+
         calculate_tax_mock.assert_called_once_with(
             str(order.id),
             subscription.currency,
--- a/server/polar/models/subscription.py
+++ b/server/polar/models/subscription.py
@@ -46,6 +46,7 @@ if TYPE_CHECKING:
         CustomerSeat,
         Discount,
         Meter,
+        Order,
         Organization,
         PaymentMethod,
         Product,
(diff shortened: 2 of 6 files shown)


server/tests/order/test_service.py
```diff
@@ -1047,6 +1047,15 @@ async def test_cycle_fixed_price(
         assert order.billing_reason == OrderBillingReasonInternal.subscription_cycle
         assert order.subscription == subscription
 
+        if tax_behavior == TaxBehavior.inclusive:
+            assert subscription.net_amount == round(
+                subscription.amount
+                * order.net_amount
+                / (order.net_amount + order.tax_amount)
+            )
```
Yopi, line 1055: Why isn't this just subscription.net_amount = order.net_amount?

Yopi (approved):
I don't have any better suggestions right now, but it feels like we will end up in an inconsistent state initially and then it will (usually?) resolve itself as subscriptions renew. 

I wonder if we should make a tax calculation for $0 (trials, discounts) for the full amount as well and then use that for the subscription net amount initially?

server/tests/order/test_service.py
pieterbeulque: An order can have prorations and overage charges, so we can't work with the order amount directly, but I think things got a bit messy for Claude writing the unit tests then.

frankie567 (approved):
That's definitely hacky, but I can't propose something really better 😅

The """right""" fix would be to issue a tax calculation we never record (hmm, already starts to smell) at start and each cycle to have a clean amount.

PR #12540, opened 2026-06-19
Make Organization Access Tokens endpoints web only

frankie567:
Allowing to mint tokens from the API doesn't really make sense and will always be a security risk. This PR forces the endpoints to be web only, and hides the corresponding scope from the API.

<a href="https://cubic.dev/pr/polarsource/polar/pull/12540?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/packages/client/src/v1.ts
+++ b/clients/packages/client/src/v1.ts
@@ -5040,15 +5040,10 @@ export interface paths {
     /**
      * List
      * @description List organization access tokens.
-     *
-     *     **Scopes**: `organization_access_tokens:read` `organization_access_tokens:write`
      */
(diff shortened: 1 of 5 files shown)


pieterbeulque (approved):
This was a dead end and you can remove as much as you want/can from it again 😊

PR #12554, opened 2026-06-22
Resolve tax behavior when changing subscription product

pieterbeulque:
## Summary

**Related Issue**: 

When a subscription's product changes, the stored tax behavior must be recomputed from the new product's price configuration to ensure recurring charges are calculated correctly. Previously, the tax behavior remained stale after a product change, causing incorrect tax calculations.

## What

- Added `resolve_subscription_tax_behavior()` function to compute the correct tax behavior for a subscription based on its current product, organization defaults, and customer billing address
- Updated `cycle()` to recompute tax behavior when applying a pending product change
- Updated `update_product()` to recompute tax behavior after changing the product
- Updated `calculate_charge_preview()` to recompute tax behavior when previewing charges with a pending product change
- Exported `get_tax_behavior_from_option()` from tax calculation module for use in resolution logic
- Added comprehensive test coverage for tax behavior updates across product changes

## Why

When a subscription's product changes (either immediately or via a pending update), the new product may have different tax behavior configuration. The subscription's stored `tax_behavior` field must be updated to reflect this, otherwise:
1. Recurring charges are calculated with stale tax behavior
2. Charge previews show incorrect tax amounts
3. The behavior diverges from what the checkout would have set for a new subscription

This mirrors the checkout's behavior of resolving tax behavior from the product configuration at subscription creation time.

## How

The solution introduces a `resolve_subscription_tax_behavior()` function that:
1. Checks the new product's price for an explicit tax behavior override
2. Falls back to the organization's default tax behavior
3. Resolves location-based behavior using the customer's billing address (if available)
4. Returns the appropriate `TaxBehavior` enum value

This function is called in three places where product changes occur:
- When cycling and applying a pending product update
- When immediately updating the product
- When previewing charges with a pending product update

## Checklist

- [x] This PR addresses a single concern (fixing tax behavior staleness on product changes)
- [x] New functionality is covered by tests (3 new test cases added)

https://claude.ai/code/session_01M26ovQDegfCrHLev22EErp

<a href="https://cubic.dev/pr/polarsource/polar/pull/12554?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/subscription/service.py
+++ b/server/polar/subscription/service.py
@@ -34,6 +34,7 @@ from polar.enums import (
     SubscriptionProrationBehavior,
     SubscriptionRecurringInterval,
     TaxBehavior,
+    TaxBehaviorOption,
 )
 from polar.event.service import event as event_service
 from polar.event.system import (
@@ -101,7 +102,10 @@ from polar.product.guard import (
 from polar.product.price_set import NoPricesForCurrencies, PriceSet
 from polar.product.repository import ProductRepository
 from polar.product.service import product as product_service
-from polar.tax.calculation import TaxCalculationLogicalError
+from polar.tax.calculation import (
+    TaxCalculationLogicalError,
+    get_tax_behavior_from_option,
+)
 from polar.tax.calculation import tax_calculation as tax_calculation_service
 from polar.webhook.service import webhook as webhook_service
 from polar.worker import enqueue_job, make_bulk_job_delay_calculator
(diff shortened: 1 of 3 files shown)


frankie567:
Hmm, actually, it was on purpose the tax behavior was frozen on subscription. The idea was to lock the tax behavior once and for all so there is no surprise for the customer in case of product update. But maybe that was the wrong call.

server/polar/subscription/service.py
```diff
@@ -228,6 +232,43 @@ def _from_timestamp(t: int | None) -> datetime | None:
...
+    organization-default and product-config changes. When the product itself
+    changes, though, it must be recomputed from the new product's price (falling
+    back to the organization default), mirroring what the checkout does. Otherwise
+    a cross-tax_behavior product change keeps a stale behavior and the recurring
+    tax is computed wrong.
+
+    Requires `subscription.prices`, `subscription.customer` and
+    `subscription.organization` to be loaded.
+    """
+    tax_behavior_option = subscription.organization.default_tax_behavior
```
frankie567, line 250: So, if the merchant changes their default on the organization, tax behavior will change on next product update? That really feels touchy IMO

pieterbeulque: Yeah, that's probably not the right call.

pieterbeulque:
>Hmm, actually, it was on purpose the tax behavior was frozen on subscription. The idea was to lock the tax behavior once and for all so there is no surprise for the customer in case of product update. But maybe that was the wrong call.

I can get behind that! Thanks for clarifying.

Eventually, we could expose options to manage this for existing customers / subscriptions then, I guess, but let's fix that when it presents itself 😁

PR #12561, opened 2026-06-22
feat: add backoffice button to complete offboarding

psincraian:
## Summary

Adds a **Complete Offboarding** action on the backoffice organization detail page 

## What

- New `organizations:offboarded_dialog` backoffice endpoint + a button in the offboarding action group.
- The confirmation modal warns that organization members will be emailed and shows the wind-down anchor data: last paid order, offboarding-since date, computed completion date, and days remaining (rounded up).
- New `OrganizationService.set_organization_offboarded` (guards that status is `offboarding`) and a shared `_transition_to_offboarded` helper now used by both the manual path and the auto-offboard cron.
- Repository helper `get_last_paid_order_at` plus a shared `PAID_ORDER_STATUSES` constant so the cron and the modal's anchor logic can't drift.

## Why

The auto-offboard cron only completes offboarding after the full wind-down period. Admins need a way to complete it on demand, with clear visibility into whether they are doing so early.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/12561?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/organization/repository.py
+++ b/server/polar/organization/repository.py
@@ -48,11 +48,6 @@ UNSNOOZE_EXPIRED_BATCH_SIZE = 500
 # Maximum orgs the auto-offboard cron processes per run.
 OFFBOARD_EXPIRED_BATCH_SIZE = 500
 
-# Order statuses that count as a completed (not fully refunded) payment when
-# anchoring the offboarding chargeback-risk window. Shared so the auto-offboard
-# cron and the backoffice "remaining period" display can't drift.
-PAID_ORDER_STATUSES = (OrderStatus.paid, OrderStatus.partially_refunded)
-
 
 class OrganizationRepository(
     RepositorySortingMixin[Organization, OrganizationSortProperty],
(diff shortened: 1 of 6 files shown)


server/polar/organization/repository.py
```diff
@@ -49,6 +48,11 @@
 # Maximum orgs the auto-offboard cron processes per run.
 OFFBOARD_EXPIRED_BATCH_SIZE = 500
 
+# Order statuses that count as a completed (not fully refunded) payment when
+# anchoring the offboarding chargeback-risk window. Shared so the auto-offboard
+# cron and the backoffice "remaining period" display can't drift.
+PAID_ORDER_STATUSES = (OrderStatus.paid, OrderStatus.partially_refunded)
```
frankie567: Maybe we could add this directly to the OrderStatus enum, similar to what we do for SubscriptionStatus:

https://github.com/polarsource/polar/blob/bba68b8cc3513a9d16731fb0a4a5c6fc16355d8d/server/polar/models/subscription.py#L61-L72

frankie567 (changes requested):

server/polar/organization/repository.py
psincraian: Good point!

frankie567 (approved):

PR #12582, opened 2026-06-22
tf/lambda: Set up specific Redis instance for AWS workers

Yopi:
Instead of opening up and making the Redis on Render available to the lambda workers, we set up a specific one for the lambda workers. This ensures that we don't run out of connections due to an bump in concurrency, and that the two paralell infrastructures cant impact each other.

<a href="https://cubic.dev/pr/polarsource/polar/pull/12582?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -295,29 +295,29 @@ jobs:
           secrets: |
             IPINFO_ACCESS_TOKEN=${{ secrets.IPINFO_ACCESS_TOKEN }}
 
-  # deploy-lambda-sandbox:
-  #   name: "Deploy Lambda Worker to Sandbox 🧪"
-  #   needs: [changes, build-lambda-worker]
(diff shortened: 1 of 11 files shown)


frankie567 (approved):
Yes, however, we'll need to be careful when migrating some tasks because Redis is used for shared states: typically, customer state cache or event streams.

Yopi:
> Yes, however, we'll need to be careful when migrating some tasks because Redis is used for shared states: typically, customer state cache or event streams.

Absoutlely, there needs to be a care. This also means that we can't enqueue a Render task from AWS since it's a different redis instance.

Overall I think it's worth it though

PR #12585, opened 2026-06-22
feat(auth): add user_session_organizations table

maximevast:
## Summary

Join table to down-scope a UserSession to one or several organizations.
Migration only — no relationship or reads yet; the consuming code lands in a follow-up.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/12585?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


frankie567 (approved):
Yes! Before merging, what would you think of a JSONB column instead of a hard M2M table? I don't have a strong preference, but curious to hear your thoughts

maximevast:
> Yes! Before merging, what would you think of a JSONB column instead of a hard M2M table? I don't have a strong preference, but curious to hear your thoughts

@frankie567 I assumed you had a strong opinion since you said M2M in your request. I found the dedicated table cleaner and safer (FK) but the problem is that we will need more tables if we ever implement per org personal_access_token + oauth2_token. (so 3 tables instead of just adding 1 jsonb column to each entity)

I'm super confortable switching to jsonb if you are okay with that.

frankie567:
@maximevast Yeah, after some thoughts, let's keep the M2M. A bit heavier, but it's a decision we won't regret in the long run in the event we need to do more complex stuff.

PR #12586, opened 2026-06-22
feat(auth): enforce UserSession organization down-scope

maximevast:
## Summary

Until we implement SSO, nothing is writing in the `UserSessionOrganization` table so this is transparent as every callers explicitly pass `auth_subject.organization_ids = None`

## Checklist

---
## Summary by cubic
Enforces organization down-scope on user sessions across auth-aware queries and reads. Down-scoped sessions only see data for scoped organizations; unscoped sessions behave as before.

- **New Features**
  - Add `organization_ids` to `AuthSubject`, populated from `UserSession.organization_scopes` (selectin-loaded). `None` = unrestricted; empty set = no access; always intersected with membership.
  - Introduce `select_accessible_org_ids(auth_subject, permission=...)` for scope-aware org filtering; keep `select_user_org_ids(user_id, ...)` for membership-only checks. `AuthzRepository.get_user_org_ids` now takes `AuthSubject[User]` and applies down-scope.
  - Apply scoping in metrics, search, orders, subscriptions, discounts, custom fields, events (customer/member checks), and transactions. Transaction endpoints and service now accept `AuthSubject`.

- **Migration**
  - Use `select_accessible_org_ids(auth_subject, ...)` in auth-aware statements instead of `select_user_org_ids(user_id, ...)`.
  - Update `AuthzRepository.get_user_org_ids(...)` to pass the `AuthSubject[User]`.
  - Update `TransactionService.search/lookup(...)` and endpoints to pass the `AuthSubject` rather than a `User`.

<sup>Written for commit cbb14a76ff15d17dc023494d2fbe8ac0bc9fe735. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12586?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/authz/repository.py
+++ b/server/polar/authz/repository.py
@@ -13,6 +13,7 @@ def select_user_org_ids(
     user_id: UUID,
     *,
     permission: OrganizationPermission | None = None,
+    scoped_to: frozenset[UUID] | None,
 ) -> Select[tuple[UUID]]:
     """SQL `SELECT` of organization IDs the user is a member of.
 
@@ -24,6 +25,12 @@ def select_user_org_ids(
     when building auth-aware repository statements. When ``permission`` is
     provided, results are restricted to organizations where the user's role
     grants that permission.
+
+    ``scoped_to`` is the session/token down-scope (``AuthSubject.organization_ids``)
+    and is **required** — a caller must consciously forward the subject's scope,
+    or pass ``None`` to opt out (unrestricted). When a set is given, results are
+    intersected with it. Making it required keeps the filter fail-closed: a
+    forgotten argument is a type error, not a silent all-orgs leak.
     """
     stmt = (
         select(UserOrganization.organization_id)
@@ -36,6 +43,8 @@ def select_user_org_ids(
     )
     if permission is not None:
         stmt = stmt.where(UserOrganization.role.in_(roles_with_permission(permission)))
+    if scoped_to is not None:
+        stmt = stmt.where(UserOrganization.organization_id.in_(scoped_to))
     return stmt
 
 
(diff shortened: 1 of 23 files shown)


server/polar/authz/repository.py
```diff
@@ -13,6 +13,7 @@ def select_user_org_ids(
     user_id: UUID,
     *,
     permission: OrganizationPermission | None = None,
+    scoped_to: frozenset[UUID] | None,
```
frankie567: I'm wondering if it would be less error-prone and cumbersome to actually pass the full `AuthSubject` to this function:

```py
def select_user_org_ids(
    subject: AuthSubject[User],
    *,
    permission: OrganizationPermission | None = None,
) -> Select[tuple[UUID]]:
    ...
```

Advantages:
* One argument to pass, containing all the auth data directly
* A bit more type safe: we can only pass a User subject vs a random UUID

maximevast:
@frankie567 it's indeed way better like this, thanks!

server/tests/transaction/service/test_transaction.py
```diff
@@ -14,6 +15,10 @@
 from tests.transaction.conftest import create_transaction
 
 
+def _auth(user: User) -> AuthSubject[User]:
```
frankie567: That's typical of Claude, it puts those private helper functions everywhere!

maximevast: don't you find if more readable this way?

server/tests/transaction/service/test_transaction.py
```diff
@@ -14,6 +15,10 @@
...
+    return AuthSubject(user, set(), None)
+
+
 @pytest.mark.asyncio
 class TestSearch:
     async def test_no_access(
```
frankie567, line 22: Instead of having this ugly wrapper, you could add the `@pytest.mark.auth` decorator to the test, and then invoke the `auth_subject: AuthSubject[User]` fixture in your tests:

```py
@pytest.mark.auth(AuthSubjectFixture(subject="user_second"))
async def test_no_access(
        self,
        session: AsyncSession,
        auth_subject: AuthSubject[User],
        all_transactions: list[Transaction],
    ) -> None:
        results, count = await transaction_service.search(
            session, auth_subject, pagination=PaginationParams(1, 10)
        )
```

```py
@pytest.mark.auth() # Default is `user`
async def test_no_filter(
        self,
        session: AsyncSession,
        auth_subject: AuthSubject[User],
        user_organization: UserOrganization,
        readable_user_transactions: list[Transaction],
        all_transactions: list[Transaction],
    ) -> None:
```

frankie567 (approved):
Good for me.

server/tests/transaction/service/test_transaction.py
frankie567: See my follow up comments ;)

Yopi (approved):

PR #12600, opened 2026-06-23
tf: create a dedicated Redis for sandbox

frankie567:
## Summary by cubic
Provision a dedicated Redis for the sandbox on Render and point sandbox workers to it, isolating sandbox queues from production. Adds a temporary drain worker to read from the old shared Redis for a clean cutover.

- **New Features**
  - Create `render_redis.redis_sandbox` (`redis-sandbox`, standard, Ohio) with private networking only.
  - Wire sandbox services to the new Redis via `POLAR_REDIS_HOST`, `POLAR_REDIS_PORT`, `POLAR_REDIS_DB` (DB set to `0`), and expose `redis_sandbox_id`.
  - `render_service` module now supports per-worker `redis_host`, `redis_port`, and `redis_db` env vars.

- **Migration**
  - Add `worker-sandbox-drain` to consume from production Redis (`DB 1`) and drain in-flight tasks; remove once queues are empty.

<sup>Written for commit c7b3772356210d777a2ffb8fbbe4e82fa38877f4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12600?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/terraform/modules/render_service/main.tf
+++ b/terraform/modules/render_service/main.tf
@@ -378,11 +378,22 @@ resource "render_web_service" "worker" {
 
   custom_domains = length(each.value.custom_domains) > 0 ? each.value.custom_domains : null
 
-  env_vars = {
-    SERVICE_NAME             = { value = each.key }
-    dramatiq_prom_port       = { value = each.value.dramatiq_prom_port }
-    POLAR_DATABASE_POOL_SIZE = { value = each.value.database_pool_size }
-  }
+  env_vars = merge(
+    {
+      SERVICE_NAME             = { value = each.key }
+      dramatiq_prom_port       = { value = each.value.dramatiq_prom_port }
+      POLAR_DATABASE_POOL_SIZE = { value = each.value.database_pool_size }
+    },
+    each.value.redis_host != null ? {
+      POLAR_REDIS_HOST = { value = each.value.redis_host }
+    } : {},
+    each.value.redis_port != null ? {
+      POLAR_REDIS_PORT = { value = each.value.redis_port }
+    } : {},
+    each.value.redis_db != null ? {
+      POLAR_REDIS_DB = { value = each.value.redis_db }
+    } : {}
+  )
 }
 
 resource "render_cron_job" "cron" {
--- a/terraform/sandbox/aws.tf
+++ b/terraform/sandbox/aws.tf
@@ -63,7 +63,7 @@ module "dummy_lambda_worker" {
     POLAR_POSTGRES_SSL            = "true"
     POLAR_REDIS_HOST              = module.redis.host
     POLAR_REDIS_PORT              = tostring(module.redis.port)
-    POLAR_REDIS_DB                = "1"
+    POLAR_REDIS_DB                = "0"
     POLAR_AWS_REGION              = "us-east-2"
     POLAR_WORKER_SQS_ENABLED      = "true"
     POLAR_WORKER_SQS_QUEUE_PREFIX = "polar-sandbox-tasks"
--- a/terraform/sandbox/render.tf
+++ b/terraform/sandbox/render.tf
@@ -28,6 +28,23 @@ data "render_redis" "redis" {
 }
 
 # =============================================================================
+# Sandbox Redis Instance
+# =============================================================================
+
+resource "render_redis" "redis_sandbox" {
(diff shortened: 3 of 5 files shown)


frankie567:
@Yopi Maybe we could limit the risk by splitting into three phases:

1. Create the Redis
2. Add the drain worker (which would start working alongside others)
3. Move API/workers to the new Redis

Yopi:
> @Yopi Maybe we could limit the risk by splitting into three phases:
> 
>     1. Create the Redis
> 
>     2. Add the drain worker (which would start working alongside others)
> 
>     3. Move API/workers to the new Redis

That was going to be my suggestion as well 😄

terraform/sandbox/render.tf
```diff
@@ -129,6 +150,17 @@ module "sandbox" {
...
+    # Temporary drain worker - listens to ALL queues on production Redis
+    # This allows draining in-flight tasks from the current shared Redis
+    worker-sandbox-drain = {
+      start_command      = "uv run dramatiq polar.worker.run -p 2 -t 4 --queues high_priority,medium_priority,low_priority,webhooks,tinybird,invoices_and_receipts"
+      dramatiq_prom_port = "10004"
+      plan               = "standard"
+      redis_host         = local.production_redis_host
+      redis_port         = local.production_redis_port
+      redis_db           = "1"
+    }
```
Yopi, line 163: This is smart

terraform/sandbox/aws.tf
```diff
@@ -63,7 +63,7 @@ module "dummy_lambda_worker" {
     POLAR_POSTGRES_SSL            = "true"
     POLAR_REDIS_HOST              = module.redis.host
     POLAR_REDIS_PORT              = tostring(module.redis.port)
-    POLAR_REDIS_DB                = "1"
+    POLAR_REDIS_DB                = "0"
```
Yopi, line 66: That's fair to keep them in sync, but this is a different Redis instance. Maybe it should even be `2` just to be clear about it

terraform/modules/render_service/main.tf
```diff
@@ -378,11 +378,22 @@ resource "render_web_service" "worker" {
...
-    dramatiq_prom_port       = { value = each.value.dramatiq_prom_port }
-    POLAR_DATABASE_POOL_SIZE = { value = each.value.database_pool_size }
-  }
+  env_vars = merge(
+    {
+      SERVICE_NAME             = { value = each.key }
+      dramatiq_prom_port       = { value = each.value.dramatiq_prom_port }
+      POLAR_DATABASE_POOL_SIZE = { value = each.value.database_pool_size }
+    },
+    each.value.redis_host != null ? {
```
Yopi, line 387: Is this here to ensure that they overwrite the other variables?

Yopi: Otherwise it should go in the redis env block right?

terraform/sandbox/render.tf
frankie567: ❤️

terraform/sandbox/aws.tf
frankie567: Oh right! I'll just discard the change here, too eager search/replace

terraform/modules/render_service/main.tf
frankie567: So it's a convenience to declare that temporary drain worker that still reads from the production Redis: if `redis_host`/`redis_port`/`redis_db` is passed explicitly in the config, then set env vars at the service level.

Render is very explicit that env vars at service level take precedence over environment groups: https://render.com/docs/configure-environment-variables#linking-a-group-to-a-service

It has no impact on other services, they'll take the value normally from the env group.

frankie567:
Closing in favor of stacked PR.

terraform/modules/render_service/main.tf
Yopi: Ah, makes sense!

PR #12602, opened 2026-06-23
Make auto reply messages ignore unread in backoffice

sebastianekstrom:
## Summary

We've added an auto reply to the first appeal message, telling the user we will respond in 1-2 business days. However this meant that we no longer have an `Awaiting reply` state in Backoffice, since we've technically have replied (but automatically). This makes it hard for support to know which tickets they are needed in.

## Screenshot

Monosnap Cases · Polar Backoffice 2026-06-23 13-39-53

<a href="https://cubic.dev/pr/polarsource/polar/pull/12602?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/organization_review/tasks.py
+++ b/server/polar/organization_review/tasks.py
@@ -395,14 +395,20 @@ async def post_appeal_greeting(case_id: uuid.UUID) -> None:
         existing = await message_repository.list_by_case(case_id, visible_to=None)
         if any(
             message.author_kind == SupportCaseMessageAuthorKind.platform
+            or (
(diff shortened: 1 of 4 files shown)


sebastianekstrom:
Closing this in favor of another solution

PR #12605, opened 2026-06-23
tf: wire sandbox workers to their dedicated Redis instance

frankie567:
That's the touchy one :D Services will now be wired to the sandbox Redis instance. The drain worker has an overload config so it's able to drain in-flight tasks from the production Redis instance.

<a href="https://cubic.dev/pr/polarsource/polar/pull/12605?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/terraform/modules/render_service/main.tf
+++ b/terraform/modules/render_service/main.tf
@@ -378,11 +378,18 @@ resource "render_web_service" "worker" {
 
   custom_domains = length(each.value.custom_domains) > 0 ? each.value.custom_domains : null
 
-  env_vars = {
-    SERVICE_NAME             = { value = each.key }
-    dramatiq_prom_port       = { value = each.value.dramatiq_prom_port }
-    POLAR_DATABASE_POOL_SIZE = { value = each.value.database_pool_size }
-  }
+  env_vars = merge(
+    {
+      SERVICE_NAME             = { value = each.key }
+      dramatiq_prom_port       = { value = each.value.dramatiq_prom_port }
+      POLAR_DATABASE_POOL_SIZE = { value = each.value.database_pool_size }
+    },
+    (each.value.redis_host != null && each.value.redis_port != null && each.value.redis_db) ? {
+      POLAR_REDIS_HOST = { value = each.value.redis_host }
+      POLAR_REDIS_PORT = { value = each.value.redis_port }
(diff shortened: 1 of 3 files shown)


terraform/modules/render_service/main.tf
```diff
@@ -378,11 +378,22 @@ resource "render_web_service" "worker" {
...
+    each.value.redis_host != null ? {
+      POLAR_REDIS_HOST = { value = each.value.redis_host }
+    } : {},
+    each.value.redis_port != null ? {
+      POLAR_REDIS_PORT = { value = each.value.redis_port }
+    } : {},
+    each.value.redis_db != null ? {
+      POLAR_REDIS_DB = { value = each.value.redis_db }
+    } : {}
+  )
```
Yopi: I don't know if it would be nicer to do something like this (not sure if this compiles)
```suggestion
    (each.value.redis_host != null && each.value.redis_port != null && each.value.redis_db  ? {
      POLAR_REDIS_HOST = { value = each.value.redis_host }
      POLAR_REDIS_PORT = { value = each.value.redis_port }
      POLAR_REDIS_DB = { value = each.value.redis_db }
    } : {},
  )
```

terraform/sandbox/render.tf
```diff
@@ -123,7 +127,7 @@ module "sandbox" {
     database_pool_size     = "10"
     postgres_database      = "polar_sandbox"
     postgres_read_database = "polar_sandbox"
-    redis_db               = "1"
+    redis_db               = "0"
```
Yopi: I feel like this is a scary change. In a way it's nice to have this separate even though they are on separate instances to prevent accidentally reading from one to another

Yopi (approved):
Just two small comments

terraform/sandbox/render.tf
frankie567: Fair enough 👍

terraform/modules/render_service/main.tf
frankie567: Yep, it worked (with minor tweaks) :D

PR #12610, opened 2026-06-23
docs: add RFC for encrypting secrets at rest with AWS KMS

psincraian:
## Summary

Adds a design document (RFC) proposing how Polar should encrypt secrets at rest in the database.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/12610?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


strandhvilliam (dismissed):

handbook/engineering/design-documents/secrets-encryption.mdx
```diff
@@ -0,0 +1,250 @@
...
+- KMS keeps every old version of the material for as long as the key exists.
+- The wrapped data key records which version wrapped it, so KMS always picks the right one to unwrap.
+
+So existing data still decrypts after rotation, with no re-encryption and no code change.
+
+Example: written in 2026, read in 2027.
+
+1. June 2026: we encrypt a GitHub token. KMS wraps its data key with material version A. We store `v1.<wrapped-with-A>.<nonce>.<ciphertext>`.
+2. January 2027: KMS rotates the key automatically. It creates material version B and keeps version A.
+3. March 2027: we read that 2026 token. We send the wrapped data key to KMS. KMS sees it was wrapped with version A, uses it, and returns the data key. We decrypt locally, having rewritten nothing.
```
frankie567, line 253: But we need to make sure to encrypt again with the data key returned by version B?

handbook/engineering/design-documents/secrets-encryption.mdx
```diff
@@ -0,0 +1,250 @@
...
+        return await asyncio.to_thread(self._decrypt_sync, blob, context)
+```
+
+We decrypt one secret, only when it is about to be used:
+
+```python
+token = await secret_service.decrypt(
+    slack_app.bot_token_encrypted, context={"table": "slack_apps", "id": str(slack_app.id)}
+)
+await slack_client.conversations_list(token)
```
frankie567: I agree with the premise that having an implicit conversion at SQLAlchemy type level is a waste of resources. However, having to summon this service manually is a bit cumbersome.

I think we can make this a bit more DX and type safe by wrapping this in a specific type. POC:

```py
# Wrapper class
class EncryptedString:
    def __init__(self, encrypted_value: str, context: dict[str, str] | None = None) -> None:
        self.encrypted_value = encrypted_value
        self.context = context or {}
    
    async def decrypt(self, **context: str) -> str:
        return await secret_service.decrypt(
            self.encrypted_value, context={**self.context, **context}
        )

    @classmethod
    async def encrypt(cls, value: str) -> typing.Self:
        return await secret_service.encrypt(value)
```

```py
# SQLAlchemy type
class EncryptedStringType(TypeDecorator):
    def __init__(self, context: dict[str, str], **kwargs: Any) -> None:
        super().__init__(**kwargs)
        self.context = context

    def process_bind_param(self, value: Any, dialect: Dialect) -> Any:
        if isinstance(value, EncryptedString):
            return value.encrypted_value
        raise ValueError("Value must be an instance of EncryptedString")

    def process_result_value(self, value: Any, dialect: Dialect) -> Any:
        if value is not None:
            return EncryptedString(value, self.context)
        return None
```

```py
# Model declaration
class MyModel(RecordModel):
    foo: Mapped[EncryptedString] = mapped_column(EncryptedStringType({"table": "my_models"}), nullable=False)
```

```py
# Usage
value = "SuperSecret"
m = MyModel(foo=await EncryptedString.encrypt(value))
decrypted_value = await m.foo.decrypt()
```

handbook/engineering/design-documents/secrets-encryption.mdx
```diff
@@ -0,0 +1,250 @@
...
+```
+
```
frankie567, line 207: We'll also need to think about the cases where we serialize those values into a Pydantic schema. Typically OAuth Clients are serialized in the API output. So we'll probably need a wrapper for that as well.

joebon (approved):
I like this!
There'll reasonably be DX to fine tune down the line but the overall approach is solid IMHO.

handbook/engineering/design-documents/secrets-encryption.mdx
psincraian: that's not needed, we can keep using the old data key for as long as we want

frankie567: Ok, so the interest of rotation is to limit the number of data encrypted with the same key, that's the point? And if one is compromised, then we fall back to the re-encryption procedure?

psincraian: Yes, that's one of the reasons. The other reason is that we don't need to re-encrypt all the old data once we release a new key.

handbook/engineering/design-documents/secrets-encryption.mdx
psincraian: That's a good point. Now, I only have a doubt if we should expose the secret directly in the get/list endpoints or have a separate endpoint or query parameter for that 🤔


For now, I wouldn't say it's needed.

handbook/engineering/design-documents/secrets-encryption.mdx
psincraian: That's a much better DX! Will add this

handbook/engineering/design-documents/secrets-encryption.mdx
frankie567: In the case of OAuth, the DCR spec requires the `client_secret` to be returned in the payload when creating/updating/getting a client. So at least for those, we'll need to manually unwrap the value before feeding it to Pydantic.

psincraian: Fixed that

PR #12614, opened 2026-06-23
lambda: Implement exponential backoff

Yopi:
Similar to Dramatiq we have an exponential backoff, this is done via the message visiblity API. SQS has a limitation on max 12 hours hidden visibility, as such we clamp the backoff to that.

<a href="https://cubic.dev/pr/polarsource/polar/pull/12614?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/worker/_sqs.py
+++ b/server/polar/worker/_sqs.py
@@ -1,5 +1,4 @@
 import asyncio
-import functools
 import itertools
 import json
 from collections import defaultdict
@@ -35,7 +34,6 @@ class SQSSendError(Exception):
         super().__init__(f"Failed to send {len(failed)} message(s) to {queue_name}")
 
 
-@functools.lru_cache(maxsize=1)
 def get_sqs_client() -> "SQSClient":
     return boto3.client(
         "sqs",
@@ -56,7 +54,6 @@ def get_sqs_client() -> "SQSClient":
     )
 
 
-@functools.lru_cache(maxsize=1)
 def get_consumer_sqs_client() -> "SQSClient":
     """SQS client authenticated by the Lambda execution role (no static keys)."""
     return boto3.client(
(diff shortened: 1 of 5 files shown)


server/polar/worker/_sqs.py
```diff
@@ -55,19 +56,36 @@ def get_sqs_client() -> "SQSClient":
     )
 
 
+@functools.lru_cache(maxsize=1)
+def get_consumer_sqs_client() -> "SQSClient":
```
frankie567, line 57: Why a cached function vs a simple module variable?

server/polar/worker/_sqs.py
```diff
@@ -55,19 +56,36 @@ def get_sqs_client() -> "SQSClient":
...
+    """SQS client authenticated by the Lambda execution role (no static keys)."""
+    return boto3.client(
+        "sqs",
+        endpoint_url=settings.SQS_ENDPOINT_URL,
+        config=Config(
+            region_name=settings.AWS_REGION,
+            signature_version=settings.AWS_SIGNATURE_VERSION,
+            connect_timeout=2,
+            read_timeout=5,
+            retries={"max_attempts": 2, "mode": "standard"},
```
frankie567, line 67: What those retries refer to in this context? Only the enqueue/dequeue calls?

frankie567 (approved):

server/polar/worker/_sqs.py
Yopi: There's no good reason. I'll switch both of them

server/polar/worker/_sqs.py
Yopi: Exactly, this is for the AWS SDK, so it's how we treat errors for those operations. E.g. enqueue/dequeue

PR #12617, opened 2026-06-23
refactor(cases): Migrate the appeal chat to be generic to any case type

maximevast:
## Summary

Migrate the appeal chat onto a generic, case-id-addressed /v1/support-cases/{id} API that any case type (appeal, dispute, future) reuses, and drop the appeal-specific chat endpoints.

ℹ️ read commit by commit it's gonna be easier!

## What

- New merchant-facing /v1/support-cases/{id} surface (APITag.private): get thread, post reply, download attachment.
- Discovery fields so the frontend can find a case id: appeal_case_id on review-status, case_id on disputes.
- Appeal chat frontend rewired onto the generic endpoints, keyed by the discovered case id.
- Removed the legacy /appeal/case* endpoints and the now-dead appeal_case_service read methods; regenerated the client.

## Why

The SupportCase primitive (messages/attachments/participants) is already generic, but the HTTP surface was appeal-only. Making it reusable lets disputes and future chats hook in without minting a new endpoint per domain.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/12617?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/polar/support_case/endpoints.py
```diff
@@ -0,0 +1,158 @@
...
+) -> SupportCaseMessage:
+    """Post a merchant reply to a support case.
+
+    The reply may carry free text, attachments, or both. Attachments must first
+    be uploaded through the files API with service ``support_case_attachment``.
+    """
+    case = await support_case_service.get(session, auth_subject, id)
+    if case is None:
+        raise ResourceNotFound()
+    if not isinstance(case, ReviewAppealSupportCase):
```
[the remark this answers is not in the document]
maximevast: `raise CaseRepliesNotSupportedError(case)` is pretty explicit about the intent. At the time of writing, no UI exists for replying to disputes. This will be relaxed when implementing the merchant / customer facing chat

server/polar/support_case/endpoints.py
```diff
@@ -0,0 +1,158 @@
...
+        }
+    )
+
+
+@router.post(
+    "/{id}/messages",
+    summary="Reply to Support Case",
+    response_model=SupportCaseMessageSchema,
+    responses={
+        200: {"description": "Reply posted."},
```
frankie567: Nitpick because it's a private API: `201` status code

server/polar/models/dispute.py
```diff
@@ -126,6 +127,24 @@ def order(cls) -> Mapped["Order"]:
     def payment(cls) -> Mapped["Payment"]:
         return relationship("Payment", lazy="raise")
 
+    @declared_attr
+    def support_case(cls) -> Mapped["DisputeSupportCase | None"]:
```
frankie567: Since you're only interested in the ID here, I think `column_property` is more straightforward: https://docs.sqlalchemy.org/en/21/orm/mapped_sql_expr.html#using-column-property

server/polar/organization/endpoints.py
```diff
@@ -955,14 +949,8 @@ async def get_review_status(
...
 
-    return OrganizationReviewStatus(
-        verdict=review.verdict,  # type: ignore[arg-type]
-        reason=review.reason,
-        appeal_submitted_at=review.appeal_submitted_at,
-        appeal_reason=review.appeal_reason,
-        appeal_decision=review.appeal_decision,
-        appeal_reviewed_at=review.appeal_reviewed_at,
-    )
+    case = await appeal_case_service.get_case(session, review)
```
frankie567, line 952: Why not adopting the same approach as for Dispute.case_id with a SQLAlchemy trick?

server/polar/organization/endpoints.py
frankie567, line 1: I would put that commit (0ebe56188701ff6b882bc608022a07d6e4ecfc38) on a separate PR: in a single operation, it's possible (esp. with latest CI changes) the old frontend will still be live when this is deployed and will try to reach the old endpoints.

frankie567 (commented):
Nice! Main thing: put the old endpoints deletion in a separate PR.

server/polar/organization/endpoints.py
maximevast: That's what I did at first but the blast radius was larger and it feels cleaner to do this. I can implement it if you want

frankie567: Definitely not a must-have :)

frankie567 (approved):
Merci 🙏

frankie567 (approved):

PR #12619, opened 2026-06-23
tf: Set up AWS organizations and accounts

Yopi:
Configures organization units, accounts and sets up an organization folder for the management account

<a href="https://cubic.dev/pr/polarsource/polar/pull/12619?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/terraform/organization/main.tf
+++ b/terraform/organization/main.tf
@@ -1,0 +1,113 @@
+provider "aws" {
+  region = "us-east-1"
+}
+
+data "aws_organizations_organization" "current" {}
+
+locals {
+  management_account = {
+    id = "975049931254"
+  }
+
+  workload_organizational_units = {
+    production = {
+      name = "Production"
+    }
+    sandbox = {
+      name = "Sandbox"
+    }
+    test = {
+      name = "Test"
+    }
+  }
+
+  workload_accounts = {
+    production = {
+      id                  = "538043300756"
+      name                = "production"
+      email               = var.production_account_email
+      organizational_unit = "production"
+    }
+    sandbox = {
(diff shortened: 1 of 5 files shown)


terraform/organization/main.tf
```diff
@@ -0,0 +1,114 @@
...
+}
+
+check "management_account" {
+  assert {
+    condition     = data.aws_organizations_organization.current.master_account_id == local.management_account.id
+    error_message = "Terraform must run from the Polar management account (${local.management_account.id})."
+  }
+
+  assert {
+    condition     = data.aws_organizations_organization.current.master_account_email == var.management_account_email
```
joebon: What's the point with asserting this, why doesn't just verifiying the id suffice?

joebon (approved):
Looks good! Just a non-blocking question

terraform/organization/main.tf
Yopi: I think it's interesting to use assertions like this which is why I kept it in. Might be something we want to do more of.

But in this specific case it's a bit LLM-overzealousness. The idea is to ensure that you're not trying to accidentally change the master account (since you can't) so Terraform will scream before the apply breaks

frankie567 (commented):
Will we want to move the S3 buckets to those workload accounts? :)

PR #12626, opened 2026-06-23
Return 410 on expired checkouts even post deletion

pieterbeulque:
We throw a server error instead of a not found error if a checkout page throws a 403, cluttering our Vercel SSR logs with errors. By checking the checkout expiration first, we can work around that.

<a href="https://cubic.dev/pr/polarsource/polar/pull/12626?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -1475,11 +1475,11 @@ class CheckoutService:
         if checkout is None:
             raise ResourceNotFound()
 
-        if not checkout.organization.can_authenticate:
-            raise NotPermitted()
-
         if checkout.is_expired:
             raise ExpiredCheckoutError()
+
+        if not checkout.organization.can_authenticate:
+            raise NotPermitted()
         return checkout
 
     async def mark_opened(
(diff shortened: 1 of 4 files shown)


joebon (approved):

server/polar/checkout/service.py
```diff
@@ -1475,11 +1475,11 @@ async def get_by_client_secret(
...
             raise ResourceNotFound()
 
-        if not checkout.organization.can_authenticate:
-            raise NotPermitted()
-
         if checkout.is_expired:
             raise ExpiredCheckoutError()
+
+        if not checkout.organization.can_authenticate:
+            raise NotPermitted()
```
frankie567, line 1482: Since we're now checking both status codes on the frontend, I don't really see why we swapped the checks. From a pure API perspective, I think it semantically makes more sense to have the 403 first.

pieterbeulque: Yeah, I'll revert it, I think the front-end only fix is cleaner.

PR #12637, opened 2026-06-24
docker: Optimize docker build

Yopi:
- Pin uv version
- Add more caches (uv dependencies, npm dependencies)
- Remove debug step

<a href="https://cubic.dev/pr/polarsource/polar/pull/12637?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -8,7 +8,8 @@ WORKDIR /app
 
 COPY ./emails/pnpm-lock.yaml .
 COPY ./emails/package.json .
-RUN pnpm install --frozen-lockfile
+RUN --mount=type=cache,target=/pnpm/store \
+  pnpm install --frozen-lockfile --store-dir=/pnpm/store
 
 COPY ./emails/src src/
 COPY ./emails/tsconfig.json .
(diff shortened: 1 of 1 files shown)


server/Dockerfile
```diff
@@ -83,24 +85,24 @@ FROM --platform=$BUILDPLATFORM python:3.14-slim AS production
...
 WORKDIR /app/server
 
 RUN --mount=type=bind,source=uv.lock,target=uv.lock \
   --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+  --mount=type=cache,target=/root/.cache/uv \
   apt-get update \
   && apt-get install -y \
   git \
   build-essential \
-  redis \
```
frankie567, line 102: `redis` is not needed?

frankie567 (approved):

joebon (approved):

server/Dockerfile
joebon: That's the redis server package I believe, which isn't needed. And the redis client in python doesn't need any redis system libraries afaik

Yopi: Exactly 😄

frankie567: As long as there are compiled wheels I guess, but UV is quite good at that

PR #12640, opened 2026-06-24
ci: Run frontend and backend deploys in parallel per env

Yopi:
We are still gating production on sandbox, but this allows the two builds to go out at the same time

<a href="https://cubic.dev/pr/polarsource/polar/pull/12640?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/deploy-environment.yml
+++ b/.github/workflows/deploy-environment.yml
@@ -85,7 +85,6 @@ jobs:
     concurrency:
       group: deploy-backend-${{ inputs.environment }}
       cancel-in-progress: false
-    needs: build-frontend
     if: always() && inputs.skip-backend != true
     runs-on: ubuntu-latest
     environment: ${{ inputs.environment }}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -221,6 +221,8 @@ jobs:
           context: ./server
           target: production
           platforms: linux/amd64
+          cache-from: type=registry,ref=ghcr.io/polarsource/polar:buildcache
+          cache-to: type=registry,ref=ghcr.io/polarsource/polar:buildcache,mode=max
           build-args: |
             RELEASE_VERSION=${{ github.sha }}
           secrets: |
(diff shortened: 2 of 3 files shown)


pieterbeulque:
Thinking of why this was added in the first place, are there risks tied to the unpredictable deploy order of front-end & back-end?

Yopi:
> Thinking of why this was added in the first place, are there risks tied to the unpredictable deploy order of front-end & back-end?

It's less about the predictable build order, but more about not deploying the backend with a broken frontend. The previous order was:
- Build backend
- Build frontend
- Deploy backend
- Deploy frontend

As can be seen by:
```
  promote-frontend:
    name: "Frontend: Promote to Production"
    concurrency:
      group: deploy-frontend-promotion-${{ inputs.environment }}
      cancel-in-progress: true
    needs: [build-frontend, deploy-backend]
```

What this change does is:
- Build backend
- Deploy backend | Build frontend
- Deploy frontend

What could happen then is that the frontend build fails, and you end up deploying the backend.

I _think_ that it's an OK tradeoff, given that the backend doesn't break the previous frontends, but it does loosen our guarantees slightly.

I was also considering whether we should do these two fully in parallel:

1.
- Build backend
- Deploy backend to sandbox
- Deploy backend to production

2.
- Build frontend
- Deploy frontend to sandbox
- Deploy frontend to production

However that would remove the guarantees that you point out (i.e. frontend might go out before backend, backend will go out in prod while frontend is still deploying to sandbox, etc).

It's a discussion I'm happy for us to have, because while it is riskier in one sense (if you don't know about it), it would also speed up deploys overall I believe

frankie567:
The previous order was introduced to avoid deploying a frontend that relies on API that are not deployed yet.

But I agree with you relaxing the reverse case (i.e. outdated frontend with new backend) should be okay.

frankie567 (approved):

PR #12648, opened 2026-06-24
Allow iOS widget to refresh token

sebastianekstrom:
## Summary by cubic
Enables the iOS widget to share and refresh OAuth tokens with the app via a shared Keychain, keeping widget metrics current and preventing 401s. The app can adopt a rotated session so users stay signed in when the widget refreshes tokens.

- New Features
  - Added WidgetTokenStore.swift to read/write tokens in the shared Keychain and refresh via the OAuth token endpoint; uses a file lock to serialize refreshes.
  - Enabled shared Keychain entitlements for both app and widget; configured `expo-secure-store` with the shared access group and AfterFirstUnlock, with group-first reads and fallback; exported unified session keys for reuse.
  - Widget fetches with WidgetTokenStore.apiToken() and retries once with forceRefresh() on 401; treats missing token as unauthorized and keeps legacy `widget_api_token` as fallback.
  - Auth refresher adopts a session rotated by the widget on invalid_grant (from shared storage) instead of clearing; otherwise clears when tokens match.

- Migration
  - Add Keychain Access Group `$(AppIdentifierPrefix)com.polarsource.Polar` to both app and widget in the Apple Developer portal and update provisioning profiles.
  - Ensure the team prefix matches `55U3YA3QTA`; if not, update the `accessGroup` in Swift and SecureStore options.

<sup>Written for commit 6dffe15725a925e740ad996d1e769751b726fdea. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12648?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/apps/app/app.config.js
+++ b/clients/apps/app/app.config.js
@@ -64,6 +64,9 @@ module.exports = {
         'com.apple.security.application-groups': [
           'group.com.polarsource.Polar',
         ],
+        'keychain-access-groups': [
+          '$(AppIdentifierPrefix)com.polarsource.Polar',
+        ],
       },
       associatedDomains: ['applinks:polar.godetour.link'],
     },
--- a/clients/apps/app/auth/refresher.test.ts
+++ b/clients/apps/app/auth/refresher.test.ts
@@ -16,6 +16,12 @@ jest.mock('expo-auth-session', () => {
   }
 })
 
+const mockGetStorageItemAsync = jest.fn<Promise<string | null>, [string]>()
+
+jest.mock('@/hooks/storage', () => ({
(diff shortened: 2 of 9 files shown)


strandhvilliam (approved):
Looks good to me!

PR #12674, opened 2026-06-25
aws: Set up roles and permission boundaries

Yopi:
This maps groups to roles to permissions, and sets up a structure of access for: 

- Admin
- Engineering
- Read only

We also introduce a permission boundary to prevent privilege escalation via role creation

<a href="https://cubic.dev/pr/polarsource/polar/pull/12674?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/terraform/modules/github_oidc/main.tf
+++ b/terraform/modules/github_oidc/main.tf
@@ -36,6 +36,12 @@ variable "policy_arns" {
   default     = {}
 }
 
+variable "permissions_boundary_arn" {
+  description = "Optional permissions boundary ARN to attach to the role."
+  type        = string
+  default     = null
+}
+
 resource "aws_iam_openid_connect_provider" "github" {
   url             = "https://token.actions.githubusercontent.com"
   client_id_list  = ["sts.amazonaws.com"]
@@ -43,7 +49,8 @@ resource "aws_iam_openid_connect_provider" "github" {
 }
 
 resource "aws_iam_role" "github_actions" {
-  name = var.role_name
+  name                 = var.role_name
+  permissions_boundary = var.permissions_boundary_arn
 
   assume_role_policy = jsonencode({
     Version = "2012-10-17"
--- a/terraform/modules/permission_boundary/main.tf
+++ b/terraform/modules/permission_boundary/main.tf
@@ -1,0 +1,135 @@
+terraform {
+  required_version = ">= 1.2"
+
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 5.0"
+    }
+  }
+}
+
+variable "policy_name" {
+  description = "Name of the permission boundary managed policy."
+  type        = string
+  default     = "PolarPermissionBoundary"
+}
+
+data "aws_caller_identity" "current" {}
+
+data "aws_partition" "current" {}
+
+locals {
+  boundary_arn = "arn:${data.aws_partition.current.partition}:iam::${data.aws_caller_identity.current.account_id}:policy/${var.policy_name}"
+}
+
+data "aws_iam_policy_document" "boundary" {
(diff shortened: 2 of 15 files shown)


.github/workflows/deploy.yml
frankie567, line 252: All ARN have access to the ECR? Doesn't matter if it's always sandbox that pushes the build?

frankie567 (approved):

.github/workflows/deploy.yml
Yopi: We'll have to push to both sandbox and prod when we deploy this on prod. The alternative would be that we split out ECR into a third account that both have access to

terraform/modules/github_oidc/main.tf
```diff
@@ -36,14 +36,21 @@ variable "policy_arns" {
   default     = {}
 }
 
+variable "permissions_boundary_arn" {
+  description = "Optional permissions boundary ARN to attach to the role."
+  type        = string
+  default     = null
```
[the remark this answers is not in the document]
Yopi: That's fine for now. We can tighten this when it's rolled out

terraform/organization/identity_center.tf
```diff
@@ -0,0 +1,137 @@
...
+      base_name          = "PolarAdmin"
+      description        = "Unrestricted administrator access."
+      managed_policy_arn = "arn:aws:iam::aws:policy/AdministratorAccess"
+      boundary           = false
+      groups             = ["awsadmins"]
+    }
+    engineering = {
+      base_name          = "PolarEngineering"
+      description        = "Administrator access confined by the Polar permission boundary."
+      managed_policy_arn = "arn:aws:iam::aws:policy/AdministratorAccess"
```
psincraian: I woudn't grant all engineering admin access by default. Call me paranoiac, but this has too broad roles.

psincraian (commented):
This looks nice! 

Only two things:
1. I think Engineering should have a new role that don't have access to Org or Account Management.
2. I would have a breaking glass to be able to access Admin that don't depend on Google IdP just in case it's down, something is not well configured, etc. To not get locked out.

terraform/organization/identity_center.tf
Yopi: Yeah you're right. The permission boundary protects a little bit, but probably not enough. It's better we set up a specific policy for this role

terraform/modules/permission_boundary/main.tf
```diff
@@ -0,0 +1,137 @@
...
+      values   = [local.boundary_arn]
+    }
+  }
+
+  statement {
+    sid    = "DenyAssumingOrPassingNonPolarRoles"
+    effect = "Deny"
+    actions = [
+      "sts:AssumeRole",
+      "iam:PassRole",
```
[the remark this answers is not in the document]
Yopi: iam:PassRole doesn't support the iam:PermissionsBoundary condition key. There are no legacy roles

terraform/organization/identity_center.tf
Yopi: What do you think about using the PowerUserAccess role for now? https://docs.aws.amazon.com/aws-managed-policy/latest/reference/PowerUserAccess.html

Yopi:
> 2. I would have a breaking glass to be able to access Admin that don't depend on Google IdP just in case it's down, something is not well configured, etc. To not get locked out.

I think we technically can log in with the root account per account

PR #12698, opened 2026-06-26
Display disputes in the dashboard

sebastianekstrom:
## Summary

This PR will add a new dashboard page where the merchant can see all their disputes, and also detail about the dispute. There's no current functionality to actually accept/counter a dispute, that will be added in follow up PR's.

**The functionality is not yet done**, so to keep the PR size down I've hidden it behind the `disputes_enabled` flag

## Screenshots

| List view | Details view  |
|--------|--------|
| Screenshot 2026-06-26 at 12 50 05 | Screenshot 2026-06-26 at 12 50 07 |

<a href="https://cubic.dev/pr/polarsource/polar/pull/12698?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/dispute/repository.py
+++ b/server/polar/dispute/repository.py
@@ -13,7 +13,7 @@ from polar.kit.repository import (
     RepositorySortingMixin,
     SortingClause,
 )
-from polar.models import Dispute, Payment
+from polar.models import Dispute, Order, Payment
 from polar.models.dispute import DisputeAlertProcessor
 
 from .sorting import DisputeSortProperty
@@ -87,7 +87,10 @@ class DisputeRepository(
         statement = (
             self.get_base_statement()
             .join(Dispute.payment)
-            .options(contains_eager(Dispute.payment))
+            .options(
+                contains_eager(Dispute.payment),
+                joinedload(Dispute.order).joinedload(Order.customer),
+            )
         )
         statement = statement.where(Payment.organization_id.in_(org_ids))
         return statement
--- a/server/polar/dispute/repository.py
+++ b/server/polar/dispute/repository.py
@@ -1,7 +1,7 @@
 from uuid import UUID
 
 from sqlalchemy import Select
-from sqlalchemy.orm import contains_eager, joinedload
+from sqlalchemy.orm import joinedload
 
 from polar.authz.types import AccessibleOrganizationID
 from polar.enums import PaymentProcessor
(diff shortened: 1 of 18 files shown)


pieterbeulque (approved):

server/polar/dispute/schemas.py
```diff
@@ -8,12 +9,24 @@
...
     IDSchema,
+    Schema,
     TimestampedSchema,
 )
 from polar.models.dispute import DisputeStatus
 
 DisputeID = Annotated[UUID4, Path(description="The dispute ID.")]
 
+
+class DisputeCustomer(Schema):
```
frankie567: For consistency with other endpoints, like Orders, we can do:

```py
from polar.customer.schemas.customer import CustomerBase

class DisputeCustomer(CustomerBase): ...
```

This will output all the base fields of a Customer, which are already loaded anyway

server/polar/dispute/schemas.py
```diff
@@ -55,6 +68,33 @@ class DisputeBase(IDSchema, TimestampedSchema):
     currency: Annotated[
         str, Field(description="Currency code of the dispute.", examples=["usd"])
     ]
+    reason: Annotated[
+        str | None,
+        Field(
+            description=(
+                "The reason for the dispute as reported by the card network "
+                "(e.g. `fraudulent`, `product_not_received`). "
+                "`null` until the processor reports it."
```
frankie567: We use `None` usually vs `null` in description. Not the best thing now that I think of it, but want to keep it consistent and change them all in a single effort.

server/polar/dispute/schemas.py
```diff
@@ -55,6 +68,33 @@ class DisputeBase(IDSchema, TimestampedSchema):
...
+            ),
+            examples=["fraudulent"],
+        ),
+    ]
+    evidence_due_by: Annotated[
+        datetime | None,
+        Field(
+            description=(
+                "Deadline to submit evidence in response to the dispute. "
+                "`null` when no response is required."
```
frankie567: See https://github.com/polarsource/polar/pull/12698/changes#r3481102651

server/polar/dispute/repository.py
```diff
@@ -87,7 +87,10 @@ def get_statement_by_org_ids(
         statement = (
             self.get_base_statement()
             .join(Dispute.payment)
-            .options(contains_eager(Dispute.payment))
+            .options(
+                contains_eager(Dispute.payment),
+                joinedload(Dispute.order).joinedload(Order.customer),
```
frankie567: This should go in `get_eager_options` instead. Then, call this method to the statements built inside `DisputeService.list` and `DisputeService.get`

frankie567 (changes requested):

server/polar/dispute/schemas.py
sebastianekstrom: Makes sense!

server/polar/dispute/schemas.py
sebastianekstrom: Fixed!

server/polar/dispute/schemas.py
sebastianekstrom: Fixed!

server/polar/dispute/repository.py
sebastianekstrom: Fixed. All fixes are made in https://github.com/polarsource/polar/pull/12698/changes/f4aa7ca304faf8accb4ec28049d165fe1714d452 (will rebase before merge)

frankie567 (approved):

PR #12702, opened 2026-06-26
lambda: Add ability to add more lambdas to TF and CI

Yopi:
## Summary by cubic
Switch sandbox to a single Lambda worker with one SQS queue (`*-default`). All allowlisted `dramatiq` actors now route to this queue; Terraform and CI target this one worker.

- **New Features**
  - Route all allowed actors to `polar-<env>-tasks-default`; added test to confirm mapping.
  - Terraform defines one worker named `default`; shared env/secrets locals; added Resend email config.
  - CI deploys `polar-sandbox-worker-default`.

- **Migration**
  - Terraform: replace `module "dummy_lambda_worker"` with `module "lambda_worker"` (name `default`); state moved via `moved` blocks.
  - Outputs renamed to singular: use `lambda_worker_function_name` and `lambda_worker_queue_url`.
  - IAM policies updated to reference `module.lambda_worker.*`.

<sup>Written for commit dd917797934e0f0161d12d00bd9391ea29d50679. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12702?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/terraform/sandbox/aws.tf
+++ b/terraform/sandbox/aws.tf
@@ -24,15 +24,27 @@ resource "aws_vpc_security_group_ingress_rule" "redis_lambda" {
   ip_protocol                  = "tcp"
 }
 
-module "dummy_lambda_worker" {
-  source = "../modules/aws_task_worker"
+locals {
+  lambda_workers = {
+    dummy = {
+      reserved_concurrency = null
+    }
+    "email-send" = {
+      reserved_concurrency = null
+    }
+  }
+}
+
+module "lambda_worker" {
+  source   = "../modules/aws_task_worker"
+  for_each = local.lambda_workers
 
   environment              = "sandbox"
-  name                     = "dummy"
-  queue_name               = "polar-sandbox-tasks-dummy"
+  name                     = each.key
+  queue_name               = "polar-sandbox-tasks-${each.key}"
   image_uri                = "${module.lambda_worker_ecr.repository_url}:latest"
   enabled                  = true
-  reserved_concurrency     = null
+  reserved_concurrency     = each.value.reserved_concurrency
   subnet_ids               = local.lambda_subnet_ids
   security_group_ids       = local.lambda_security_group_ids
   permissions_boundary_arn = data.aws_iam_policy.permission_boundary.arn
@@ -69,6 +81,11 @@ module "dummy_lambda_worker" {
   }
 }
 
+moved {
+  from = module.dummy_lambda_worker
+  to   = module.lambda_worker["dummy"]
+}
+
 # =============================================================================
 # Task producer IAM user (SQS send-only, used by the Render backend)
 # =============================================================================
@@ -88,7 +105,7 @@ data "aws_iam_policy_document" "tasks_producer" {
       "sqs:SendMessage",
       "sqs:GetQueueUrl",
     ]
-    resources = [module.dummy_lambda_worker.queue_arn]
+    resources = [for worker in module.lambda_worker : worker.queue_arn]
   }
 }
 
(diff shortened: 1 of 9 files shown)


terraform/sandbox/aws.tf
```diff
@@ -67,6 +55,48 @@ module "dummy_lambda_worker" {
     POLAR_SENTRY_DSN      = var.backend_sentry_dsn_sandbox
     TAILSCALE_AUTHKEY     = var.lambda_worker_tailscale_token
   }
+
+  lambda_workers = {
```
frankie567: Maybe it's because we're gradually adopting Lambdas, but will there be a way to declare only one lambdas and have ways to "tune" the message routing in terms of priority/other parameters? Because I foresee it'll be quite heavy and cumbersome to have to manage 1 Lambda per task.

frankie567 (approved):

terraform/sandbox/aws.tf
Yopi: The idea was to use one lambda per task to be able to set independent limits on them, but it's a good point that the flip side is that we get a lambda to manage per task. 

There are of course other ways we could do it, like you say. We could do it like we route the tasks today by priority, we could route them by theme, etc.

I think there are two aspects at hand: one is the maintainability in terms of we need to make updates to env variables or similar changes to the infra. This is handled by Terraform, but with 130+ tasks it might still take a while to apply and handle.

The other aspect is individual config changes, deploys, etc, in which looping through and deploying that many lambdas might be a bit slow or unnecessary. I think the deploy step itself will be fast, but there might be a point in not going that route

frankie567: Yeah, I really think one lambda per task will quickly become unmanageable. Quickly chatted with an LLM, and it recommended the following pattern:

> **Components**
> 
> * SQS Queues: Three queues for task prioritization:
>   * polar-tasks-high: High-priority tasks (e.g., user-facing actions)
>   * polar-tasks-default: Standard tasks
>   * polar-tasks-low: Low-priority tasks (e.g., analytics, cleanup)
> * Lambda Function: Single worker function (polar-task-worker) deployed once. Written in Python, compatible with existing task logic.
> * Event Source Mappings: Lambda polls polar-tasks-high first. If empty, it checks polar-tasks-default, then polar-tasks-low. Configured via AWS Lambda event source mappings with batch size and concurrency limits.
> * Scaling: Lambda instances scale automatically based on the number of messages in each queue. Concurrency limits can be set per queue to control resource allocation.

So, if I understand correctly, we could have our own priority controls based on how we poll the queues.

Yopi: I don't think that that's entirely the correct pattern either. We want to get away from one queue type fully cannibalizing others, so a spike in high prio tasks shouldn't make us stop processing lower prio tasks.

I think we can mirror what we have for the service workers today, and then play with the reserved concurrency, so we reserve (more?) concurrency for high prio tasks, while still processing in all of the queues. I pushed a suggestion for it

terraform/sandbox/aws.tf
```diff
@@ -54,19 +42,73 @@ module "dummy_lambda_worker" {
...
     POLAR_WORKER_SQS_QUEUE_PREFIX = "polar-sandbox-tasks"
   }
 
-  secret_environment_variables = {
+  lambda_worker_secrets = {
     POLAR_CURRENT_JWK_KID = var.backend_current_jwk_kid_sandbox
     POLAR_JWKS_CONTENT    = var.backend_jwks_sandbox
     POLAR_LOGFIRE_TOKEN   = var.logfire_token
     POLAR_POSTGRES_PWD    = local.db_password
+    POLAR_RESEND_API_KEY  = var.backend_resend_api_key_sandbox
```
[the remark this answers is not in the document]
Yopi: This is OK. We'll run multiple tasks in the same lambda now

frankie567 (approved):

terraform/sandbox/aws.tf
Yopi: Actually, thinking about this more. During the roll out, lets just start with a single monofunction and see where it takes us

PR #12706, opened 2026-06-26
feat: add nested customer member endpoints under /v1/customers/{id}/members

psincraian:
## Summary

Adds member endpoints nested under `/v1/customers/{id}/members` and external variants. The current endpoints are unchanged, for now. 

SDK will look like:

```python
polar_sdk.customers.members.create()
polar_sdk.customers.members.update()
# ...
```

## What

New public endpoints, defined in the member module and registered under the `/customers` prefix:

- `POST /v1/customers/{id}/members` and `POST /v1/customers/external/{external_id}/members`
- `GET|PATCH|DELETE /v1/customers/{id}/members/{member_id}`
- `GET|PATCH|DELETE /v1/customers/external/{external_id}/members/external/{member_external_id}`

Supporting changes: a `MemberCreateFromCustomer` body schema (customer comes from the path), `member_service.create` now resolves the customer by internal or external id, and a new auth-scoped `member_service.get_for_customer` helper. The TS client is regenerated.

## How

- The route carries `tags=["customers", "members"]`, which writes both the `operationId` (`customers:members:*`) and the `x-speakeasy-group` (`customers.members`), so the SDKs nest the methods automatically. 
- New routes are set on members/endpoints.py to avoid many recursive imports issues

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/12706?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/member/endpoints.py
+++ b/server/polar/member/endpoints.py
@@ -2,8 +2,8 @@ from uuid import UUID
 
 from fastapi import Depends, Query
 
-from polar.customer.schemas.customer import ExternalCustomerID
-from polar.exceptions import PolarRequestValidationError, ResourceNotFound
+from polar.customer.schemas.customer import CustomerID, ExternalCustomerID
+from polar.exceptions import NotPermitted, PolarRequestValidationError, ResourceNotFound
 from polar.kit.pagination import ListResource, PaginationParamsQuery
 from polar.models.member import MemberRole
 from polar.openapi import APITag
@@ -16,7 +16,13 @@ from polar.postgres import (
 from polar.routing import APIRouter
 
 from . import auth, sorting
-from .schemas import ExternalMemberID, Member, MemberCreate, MemberUpdate
+from .schemas import (
+    ExternalMemberID,
+    Member,
+    MemberCreate,
+    MemberCreateFromCustomer,
+    MemberUpdate,
+)
 from .service import member_service
 
 router = APIRouter(
@@ -24,9 +30,27 @@ router = APIRouter(
     tags=["members", APITag.public],
 )
 
+# Nested router: the new customer-scoped member CRUD, mounted under
+# /v1/customers/{id}/members. Lives in the member module (logic ownership) but
+# is grouped under `customers.members` in the SDK via its tags.
+customer_members_router = APIRouter(
+    prefix="/customers",
+    tags=["customers", "members", APITag.public],
+)
+
 MemberNotFound = {
     "description": "Member not found.",
     "model": ResourceNotFound.schema(),
+}
+
+CustomerNotFound = {
+    "description": "Customer not found.",
+    "model": ResourceNotFound.schema(),
+}
+
+NotPermittedToAddMembers = {
+    "description": "Not permitted to add members.",
+    "model": NotPermitted.schema(),
 }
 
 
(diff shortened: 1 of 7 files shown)


server/polar/member/endpoints.py
```diff
@@ -98,6 +128,260 @@ async def list_members(
...
+    """Get a member of a customer by its ID."""
+    member = await member_service.get_for_customer(session, auth_subject, id, member_id)
+    if member is None:
+        raise ResourceNotFound("Member not found")
+
+    return Member.model_validate(member)
+
+
+@customer_members_router.get(
+    "/external/{external_id}/members/external/{member_external_id}",
```
frankie567: I think we can get rid of the second `/external` sub-path: we're already in the `/external` sub-path of customers. And since we don't plan to support external customer /internal member, I think it's ok to simplify that.

server/polar/member/endpoints.py
```diff
@@ -16,19 +16,49 @@
...
+    Member,
+    MemberCreate,
+    MemberCreateFromCustomer,
+    MemberUpdate,
+)
+from .service import AmbiguousExternalCustomerID, member_service
 
 router = APIRouter(
     prefix="/members",
     tags=["members", APITag.public],
```
frankie567: Let's remove the `APITag.public` from the main router and add:

* `APITag.public` to the list endpoint (we'll still support it)
* `APITag.private` to the now deprecated members endpoint

frankie567 (changes requested):

server/polar/member/service.py
```diff
@@ -95,14 +123,24 @@ async def get_by_external_id(
         """Get a member by external ID if the auth subject has access to it."""
         repository = MemberRepository.from_session(session)
         org_ids = await get_accessible_org_ids(session, auth_subject)
+
+        # Resolve the customer first so an ambiguous external customer ID always
+        # surfaces as a 409, regardless of how many members happen to match.
+        if external_customer_id is not None:
```
psincraian, line 130: @frankie567 I have a doubt on how to fix those... I think it's a valid case when we use OAuth and we have access to multiple organizations, the same external_id can be scoped to two different customers, of different orgs, but the same external id. 


I think this affects all external endpoints 🥲

psincraian: If that ever happens, the user will be stuck.

frankie567: `Customer.external_id` is unique per organization, there is an index for that: https://github.com/polarsource/polar/blob/c4a7db126d82c5663370f8135bdd120b7a88a1ca/server/polar/models/customer.py#L147

Fortunately, otherwise it would have been a nightmare :D

frankie567: Oh, well, _across_ accessible organizations! Hmm... 🤔

frankie567: Feels like an edge case, to be honest. This problem has always been around. I'm fine with having an error and recommend the user to either:

* Use an organization-scoped token
* Use the Polar ID
* Disambiguate their external ID

server/polar/member/endpoints.py
psincraian: done

server/polar/member/endpoints.py
psincraian: Looking again, that makes sense

frankie567 (approved):

PR #12713, opened 2026-06-26
Add the ability for merchants to accept a dispute

sebastianekstrom:
## Summary

Add the ability to accept a dispute (we still need to handle this manually in Stripe). I've also added a `accepted_at` column which can come in handy in the future (will do similar to counter disputes if we think it's a good pattern).

As a reminder, this whole feature is behind a flag, so it's not exposed anywhere just yet.

## Screenshots
| Dashboard | Backoffice (list view) | Backoffice (details view) |
|--------|--------|--------|
| Screenshot 2026-06-26 at 17 10 01 | Screenshot 2026-06-26 at 17 09 49 | Screenshot 2026-06-26 at 17 09 52 |

<a href="https://cubic.dev/pr/polarsource/polar/pull/12713?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/migrations/versions/2026-06-26-1610_add_accepted_at_to_dispute.py
```diff
@@ -0,0 +1,36 @@
+"""add accepted_at to dispute
```
sebastianekstrom: Will move migration to its own PR once this PR is reviewed

clients/apps/web/src/components/Disputes/DisputeBanner.tsx
```diff
@@ -24,14 +30,18 @@ export const DisputeBanner = ({ dispute }: { dispute: schemas['Dispute'] }) => {
...
         <Text variant="heading-xxs" as="h3">
-          The customer disputed this payment
+          {accepted
+            ? 'You accepted this dispute'
+            : 'The customer disputed this payment'}
+        </Text>
+        <Text color="muted">
+          {accepted
+            ? 'The disputed amount and the dispute fee will be deducted from your balance. No further action is needed.'
+            : getDisputeReasonExplanation(dispute.reason)}
```
maximevast: Still wondering if we want this wording to be that "final", maybe support will decide to counter it ?

server/polar/models/dispute.py
```diff
@@ -117,6 +118,12 @@ class Dispute(RecordModel):
     past_due: Mapped[bool] = mapped_column(Boolean, nullable=False, default=False)
     submission_count: Mapped[int] = mapped_column(Integer, nullable=False, default=0)
 
+    # When the merchant conceded the chargeback. `status` flips
+    # to `accepted` at the same time
+    accepted_at: Mapped[datetime | None] = mapped_column(
```
maximevast: I would rather use a dedicated `SupportCaseMessageType` (`merchant_accept` | `merchant_deny` ?) as I think that the thread should contain the lifecycle (with timestamps)

server/polar/dispute/dispute_case.py
```diff
@@ -102,6 +102,31 @@ async def mark_under_review(
...
+        """Record that the merchant chose to accept (concede) the dispute.
+
+        Surfaces on the support thread so staff see the merchant's decision; the
+        chargeback itself is still settled out-of-band with the processor.
+        """
+        await self._assert_open(session, case)
+        return await support_case_service.post_message(
+            session,
+            case,
+            author_kind=SupportCaseMessageAuthorKind.merchant,
```
maximevast: Don't you think we should use a system type message and keep merchant / platform / customer for chat ?

maximevast (commented):
I think we need to clarify how we use the SupportCase service (ie message types / lifecycle events). 
Happy to discuss it live if you want!

server/polar/models/dispute.py
```diff
@@ -39,6 +39,7 @@ class DisputeStatus(StrEnum):
     under_review = "under_review"
     lost = "lost"
     won = "won"
+    accepted = "accepted"
```
frankie567: `accepted` is not good for a dispute status. In Stripe, if you accept a dispute, the status is `lost`. No matter if you tried to counter it or not. Besides, it'll break the synchronization with Stripe state.

server/polar/dispute/service.py
```diff
@@ -106,6 +112,48 @@ async def get(
...
+                "accepted_at": datetime.now(UTC),
+            },
+        )
+
+        case = await dispute_case_service.get_case(session, dispute)
+        if case is not None:
+            actor = auth_subject.subject if is_user(auth_subject) else None
+            await dispute_case_service.accept(session, case, actor=actor)
+
+        return await repository.get_one_or_none(statement)
```
frankie567: We should look into adding the call to Stripe in this case. The risk is way too high to forget to make the corresponding action in Stripe and let the dispute expire. Basically, it's calling the following endpoint: https://docs.stripe.com/api/disputes/close

server/polar/dispute/service.py
```diff
@@ -106,6 +112,48 @@ async def get(
...
+            session, auth_subject, permission=OrganizationPermission.sales_manage
+        )
+        statement = (
+            repository.get_statement_by_org_ids(org_ids)
+            .options(*repository.get_eager_options())
+            .where(Dispute.id == id)
+        )
+        dispute = await repository.get_one_or_none(statement)
+        if dispute is None:
+            return None
```
frankie567: The usual pattern we adopt on update endpoints is the following:

1. We call the `get` method to try to retrieve the object with the subject, and raise 404 if None.
2. We call the actual update method with the retrieved object.

So the endpoint would look like:


```py
async def accept(
    id: DisputeID,
    auth_subject: auth.DisputesWrite,
    session: AsyncSession = Depends(get_db_session),
) -> Dispute:
    """Accept a dispute, conceding the chargeback.

    Records the merchant's decision on the dispute's support case so staff can
    settle the chargeback with the processor.
    """
    dispute = await dispute_service.get(session, auth_subject, id)

    if dispute is None:
        raise ResourceNotFound()

    return await dispute_service.accept(session, dispute)
```

frankie567 (changes requested):

clients/apps/web/src/components/Disputes/DisputeBanner.tsx
sebastianekstrom: Good question! Chatted with Francois about this and we'll automatically accept the dispute on Stripe's end, updating the copy to reflect that.

server/polar/dispute/service.py
```diff
@@ -106,6 +112,34 @@ async def get(
...
+            dispute.payment_processor_id
+        )
+        await self.upsert_from_stripe(session, stripe_dispute)
+
+        repository = DisputeRepository.from_session(session)
+        reloaded = await repository.get_by_id(
+            dispute.id, options=repository.get_eager_options()
+        )
+        assert reloaded is not None
+        return reloaded
```
frankie567, line 141: `upsert_from_stripe` returns the Dispute directly, so we can do: `return await self.upsert_from_stripe(session, stripe_dispute)`

frankie567 (approved):

server/polar/dispute/service.py
sebastianekstrom: Looked into it and it seems that `upsert_from_stripe` flushes internally which expires the `case_id` column property on the returned instance, and we get a 500. Not familiar with this though so happy to make the change if you think it makes sense!

frankie567: Ah yes, I understand why. Let's keep it like this, we can optimize that later.

PR #12743, opened 2026-06-29
fix(refund): forbid dispute_prevention as a user-supplied refund reason

maximevast:
## Summary

Reject `dispute_prevention` as a user-supplied refund reason

Fixes polarsource/feedback#182

### Problem
`POST /v1/refunds/` (and the backoffice refund form) accept the full`RefundReason` enum, including `dispute_prevention`. That reason is meant to be system-only — set exclusively by `create_from_stripe` (chargeback early-warning alert) and `create_from_dispute`, both of which also link a `Dispute`.

A merchant could pass `reason=dispute_prevention` directly, producing:
1. A semantically invalid refund — `dispute_prevention` with no linked dispute — exposed as-is through the public API.
2. A misleading "Chargeback prevented" email to org owners/admins, claiming an early-warning alert fired when none did.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/12743?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/backoffice/forms.py
+++ b/server/polar/backoffice/forms.py
@@ -3,7 +3,7 @@ import re
 from collections.abc import Generator
 from enum import StrEnum
 from inspect import isclass
-from typing import Any, Self
+from typing import Any, Literal, Self, get_args, get_origin
 
 from fastapi.datastructures import FormData
 from pydantic import AfterValidator, BaseModel, StringConstraints, ValidationError
@@ -521,6 +521,12 @@ def _get_input_field(field: FieldInfo) -> FormField:
             if issubclass(field.annotation, StrEnum):
                 return SelectField(
                     options=[(item.value, item.name) for item in field.annotation]
+                )
+        if get_origin(field.annotation) is Literal:
+            members = get_args(field.annotation)
+            if all(isinstance(member, StrEnum) for member in members):
+                return SelectField(
+                    options=[(member.value, member.name) for member in members]
                 )
 
     return InputField(**_string_constraint_attrs(field))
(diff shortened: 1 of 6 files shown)


pieterbeulque (approved):

server/polar/refund/schemas.py
```diff
@@ -63,3 +63,12 @@ class RefundCreate(MetadataInputMixin, Schema):
             """
         ),
     )
+
+    @field_validator("reason")
+    @classmethod
+    def reason_must_not_be_system_only(cls, value: RefundReason) -> RefundReason:
```
frankie567: I think I would prefer an approach where we simply list the allowed values on the reason field:

```py
reason: typing.Literal[RefundReason.foo, RefundReason.bar]
```

It's a bit more verbose but it automatically leverages Pydantic validation + document OpenAPI properly

frankie567 (changes requested):
Could we have another approach where we

server/polar/backoffice/forms.py
```diff
@@ -522,6 +522,12 @@ def _get_input_field(field: FieldInfo) -> FormField:
                 return SelectField(
                     options=[(item.value, item.name) for item in field.annotation]
                 )
+        if get_origin(field.annotation) is Literal:
```
frankie567, line 525: Ooooh, magic!

frankie567 (approved):

PR #12744, opened 2026-06-29
Fix dunning lifecycle check to use order status instead of paid flag

pieterbeulque:
## Summary

**Related Issue**: polarsource/feedback#224

Fixes the `isOrderInDunningLifecycle` function to correctly identify orders in dunning by checking the order status field instead of relying on the `paid` boolean flag.

## What

Changed the condition in `isOrderInDunningLifecycle` from `!order.paid` to `order.status === 'pending'` to more accurately determine if an order is eligible for dunning lifecycle handling.

## Why

The `paid` boolean flag may not be a reliable indicator of whether an order should enter the dunning lifecycle. Using the explicit `status === 'pending'` check provides clearer intent and more accurate logic for identifying orders that are pending payment and have associated failed payment attempts.

## How

Updated the condition in the `isOrderInDunningLifecycle` function to check the order status field directly rather than the inverse of the paid flag.

## Checklist

- [x] This PR addresses a single concern (one bug fix)

https://claude.ai/code/session_01CGL4Gr3GdydhmvU4Dp7Kap

### DIFF
--- a/clients/apps/web/src/utils/order.ts
+++ b/clients/apps/web/src/utils/order.ts
@@ -42,7 +42,7 @@ export function isOrderInDunningLifecycle(
   payments: schemas['Payment'][],
 ): boolean {
   return (
-    order.status === 'pending' &&
+    (order.status === 'pending' || order.status === 'void') &&
     order.subscription_id !== null &&
     payments.some((payment) => payment.status === 'failed')
   )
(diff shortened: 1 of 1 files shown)


joebon (approved):
Can't really tell if it'll have the expected outcome but afaict looks this reasonable

pieterbeulque:
@joebon Thanks. It's not 100% ideal but at least it excludes draft orders. For void orders, it's harder, as we don't know which were voided because of failed payment recovery or because of a backoffice action, but this should be fine for now.

PR #12749, opened 2026-06-29
feat: auto-cancel subscriptions for wound-down organizations

psincraian:
## Summary

Automatically cancel customers' subscriptions once a merchant organization has been wound down, without emailing those customers.

## What

- New daily cron `organization.cancel_expired_subscriptions` scans for orgs in a terminal-bad status (`denied`, `blocked`, `offboarded`) for longer than `ORGANIZATION_SUBSCRIPTION_CANCELLATION_DELAY` (7 days) that still have billable subscriptions, and fans out a per-org job.
- `subscription.cancel_for_organization` immediately cancels each billable subscription with customer email suppressed (a new `notify_customer` flag threaded through the cancellation path).

## Why

When a merchant is shut down, their customers' subscriptions should stop, but those customers should not receive cancellation emails.

## How

- `notify_customer` defaults to `True`, so all existing cancellation/revocation paths are unchanged; only the org wind-down path passes `False`, which gates solely the customer emails (webhooks, events, benefit revocation and pending-order voiding still fire).
- `offboarded` (the terminal state) triggers cancellation
- The per-org job re-checks the org status at execution time (with `include_blocked=True`) so a reactivated org is a no-op, and is idempotent at the subscription level so duplicate/concurrent runs are safe.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/12749?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/subscription/repository.py
+++ b/server/polar/subscription/repository.py
@@ -106,6 +106,20 @@ class SubscriptionRepository(
         )
         return await self.get_all(statement)
 
+    async def list_billable_by_organization(
+        self, organization_id: UUID, *, options: Options = ()
+    ) -> Sequence[Subscription]:
+        statement = (
+            self.get_base_statement()
+            .where(
+                Subscription.organization_id == organization_id,
+                Subscription.status.in_(SubscriptionStatus.billable_statuses()),
+                Subscription.ended_at.is_(None),
+            )
+            .options(*options)
+        )
+        return await self.get_all(statement)
+
     async def get_all_by_customer(
         self, customer_id: UUID, *, options: Options = ()
     ) -> Sequence[Subscription]:
--- a/server/polar/subscription/repository.py
+++ b/server/polar/subscription/repository.py
@@ -106,23 +106,18 @@ class SubscriptionRepository(
         )
         return await self.get_all(statement)
 
-    async def list_billable_by_organization(
-        self,
-        organization_id: UUID,
-        *,
-        options: Options = (),
-        limit: int | None = None,
-        after_id: UUID | None = None,
-    ) -> Sequence[Subscription]:
-        statement = self.get_base_statement().where(
-            Subscription.organization_id == organization_id,
-            Subscription.status.in_(SubscriptionStatus.billable_statuses()),
-            Subscription.ended_at.is_(None),
+    def get_billable_by_organization_statement(
+        self, organization_id: UUID, *, options: Options = ()
+    ) -> Select[tuple[Subscription]]:
+        return (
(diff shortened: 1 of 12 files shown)


server/polar/subscription/repository.py
```diff
@@ -106,6 +106,24 @@ async def list_billable_by_customer(
...
+        limit: int | None = None,
+        after_id: UUID | None = None,
+    ) -> Sequence[Subscription]:
+        statement = self.get_base_statement().where(
+            Subscription.organization_id == organization_id,
+            Subscription.status.in_(SubscriptionStatus.billable_statuses()),
+            Subscription.ended_at.is_(None),
+        )
+        if after_id is not None:
+            statement = statement.where(Subscription.id > after_id)
```
frankie567: Can we reliably sort UUID4?

server/polar/subscription/service.py
```diff
@@ -1826,6 +1827,62 @@ async def cancel_customer(
...
+            or organization.status not in SUBSCRIPTION_CANCELLATION_STATUSES
+        ):
+            return
+
+        subscription_repository = SubscriptionRepository.from_session(session)
+        # Keyset pagination by id: advances past each batch regardless of whether
+        # a subscription was cancelled, so the loop always terminates (even if a
+        # subscription is skipped) and never re-materializes the whole set.
+        after_id: uuid.UUID | None = None
+        while True:
```
frankie567: I think a better approach would be to use `repository.stream`: batching is implicit, you only have one loop to care about, and you don't have this ID ordering problem.

frankie567 (changes requested):

server/polar/subscription/repository.py
psincraian: shit, I missed that in the last commit 🙈

server/polar/subscription/service.py
psincraian: Done in 38ba7b30e7

server/polar/subscription/repository.py
frankie567: > Postgres does give a stable total order on the native uuid type, so keyset was correct  

Oh, TIL 👍

frankie567 (approved):

PR #12751, opened 2026-06-29
feat(sso): add organization SSO connection model

maximevast:
## Summary

Introduces OrganizationSSOConnection with a typed, discriminated JSONB configuration (OIDC: issuer, client_id, auth_method, conditional client_secret).

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/12751?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/polar/models/organization_sso_connection.py
```diff
@@ -0,0 +1,85 @@
...
+class OrganizationSSOConnectionType(StrEnum):
+    oidc = "oidc"
+
+
+class OIDCAuthMethod(StrEnum):
+    client_secret = "client_secret"
+    private_key_jwt = "private_key_jwt"
+
+
+class OIDCConfiguration(BaseModel):
```
frankie567: I think we should have a simpler `TypedDict` at model level, and keep the Pedantic schema for the API schema validation. The problem with validated models is that it can quickly break data loading if we stumble upon a record that has for some reason an outdated structure. It also requires a bit less code to work (i.e. no TypeDecorator).

frankie567 (changes requested):

server/polar/models/organization_sso_connection.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @declared_attr
+    def organization(cls) -> Mapped["Organization"]:
+        return relationship("Organization", lazy="raise")
+
+    type: Mapped[OrganizationSSOConnectionType] = mapped_column(
+        StringEnum(OrganizationSSOConnectionType), nullable=False
+    )
+    configuration: Mapped[OIDCConfiguration] = mapped_column(JSONB, nullable=False)
+    enabled: Mapped[bool] = mapped_column(Boolean, nullable=False, default=True)
```
frankie567, line 50: Small hesitation about the default: maybe `False` would be safer (typically, you test the flow once before enabling it on your org). But that can be easily changed when we have the actual logic.

frankie567 (approved):

PR #12754, opened 2026-06-29
feat: add envelope-encryption primitives for secrets at rest

psincraian:
## Summary

First step of the secrets-at-rest encryption RFC (`handbook/engineering/design-documents/secrets-encryption.mdx`): the foundational encryption layer. No model columns or migrations change yet, so this is self-contained and unblocks the per-column rollout.

## What

- `KeyProvider` protocol with two implementations: a KMS-backed provider (production/sandbox) and a static-key provider (local/CI, so tests make no cloud calls).
- `EncryptedString`: holds ciphertext and owns the only paths that touch the provider `__str__`/`__repr__` return `<encrypted>` so secrets don't leak into logs.
- `EncryptedStringType`: a wrap-only SQLAlchemy `TypeDecorator` that does no crypto or I/O — it boxes stored ciphertext into an `EncryptedString` on load and unboxes it on save, so listing rows fires zero provider calls.
- Config: `AWS_KMS_KEY_ID` and `ENCRYPTION_LOCAL_KEY`.

## Why

Secrets (OAuth tokens, client secrets, Slack tokens) are stored in plain text today. Envelope encryption means a leaked DB dump, backup, or read replica no longer exposes them, while the master key stays out of the app and database.

## How

Envelope encryption with AES-256-GCM: each secret gets a fresh data key wrapped by the provider; the per-row encryption context is bound as associated data so a ciphertext copied to another row (or decrypted without its id) fails closed. `cache_ok=True` is kept safe by storing the context as a sorted tuple (verified to produce distinct statement-cache keys per column).

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/12754?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/kit/encryption.py
+++ b/server/polar/kit/encryption.py
@@ -1,0 +1,189 @@
+"""Envelope encryption for secrets stored at rest.
+
+Each secret is stored as ciphertext wrapped in an :class:`EncryptedString`. The
+column is mapped with a wrap-only :class:`EncryptedStringType` that does no
+crypto and no I/O: on load it boxes the stored ciphertext into an
+:class:`EncryptedString`; on save it unboxes it. Encryption and decryption are
+explicit ``await`` calls that go through a :class:`KeyProvider` chosen by config.
+
+See the design document for the full rationale:
+``handbook/engineering/design-documents/secrets-encryption.mdx``.
+"""
+
+import asyncio
+import base64
+import functools
+import hashlib
+import json
+import os
+from typing import Any, Protocol
+
+import sqlalchemy as sa
+from cryptography.hazmat.primitives.ciphers.aead import AESGCM
+from sqlalchemy.engine.interfaces import Dialect
+
+from polar.config import settings
+from polar.kit.extensions.sqlalchemy.types import TypeDecorator
+
+VERSION = "v1"
+NONCE_SIZE = 12
+DATA_KEY_SIZE = 32
+
+
+def _encode_context(context: dict[str, str]) -> bytes:
+    return json.dumps(context, sort_keys=True, separators=(",", ":")).encode("utf-8")
+
+
+def _b64encode(value: bytes) -> str:
+    return base64.b64encode(value).decode("ascii")
+
+
+def _b64decode(value: str) -> bytes:
+    return base64.b64decode(value)
+
+
+class KeyProvider(Protocol):
+    """Wraps and unwraps the per-secret data key. The plaintext secret never
+    reaches the provider."""
+
+    async def generate_data_key(self, context: dict[str, str]) -> tuple[bytes, bytes]:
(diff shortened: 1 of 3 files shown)


server/polar/kit/encryption.py
```diff
@@ -0,0 +1,187 @@
...
+
+def _b64decode(value: str) -> bytes:
+    return base64.b64decode(value)
+
+
+class KeyProvider(Protocol):
+    """Wraps and unwraps the per-secret data key. The plaintext secret never
+    reaches the provider."""
+
+    def generate_data_key(self, context: dict[str, str]) -> tuple[bytes, bytes]:
```
frankie567: Would it be more future-proof to make them async methods? Even if boto3 is not async, we may have other procotols that might (or migrate to an [async boto client](https://pypi.org/project/aioboto3/))

server/polar/kit/encryption.py
```diff
@@ -0,0 +1,187 @@
...
+class EncryptedString:
+    """Holds the ciphertext of a secret and owns the only paths that touch the
+    key provider. Immutable: any new value is a fresh instance, so ORM change
+    tracking works without ``sqlalchemy.ext.mutable``."""
+
+    __slots__ = ("context", "encrypted_value")
+
+    def __init__(self, encrypted_value: str, context: dict[str, str]) -> None:
+        self.encrypted_value = encrypted_value
+        self.context = context
```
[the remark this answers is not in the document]
frankie567: True, but quite a nitpick IMHO

server/polar/kit/encryption.py
```diff
@@ -0,0 +1,187 @@
...
+        data_key = await asyncio.to_thread(
+            provider.decrypt_data_key, _b64decode(wrapped), context
+        )
+        plaintext = AESGCM(data_key).decrypt(
+            _b64decode(nonce), _b64decode(ciphertext), _encode_context(context)
+        )
+        return plaintext.decode("utf-8")
+
+    def __repr__(self) -> str:
+        return "<encrypted>"
```
frankie567: For debugging, might be worth to have `__repr__` as

```py
f"{self.__class__.__name__}(\"***\", {self.context!r})"
```

And keep `__str__` like this

psincraian: makes sense! done

server/polar/kit/encryption.py
psincraian: done!

frankie567 (approved):

PR #12786, opened 2026-06-30
lambda: Route both cron and regular actors

Yopi:
This enables us to use the lambda workers for cron jobs as well. All types of actors are routed to either Redis or SQS

<a href="https://cubic.dev/pr/polarsource/polar/pull/12786?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/worker/_broker.py
+++ b/server/polar/worker/_broker.py
@@ -21,9 +21,11 @@ from polar.logfire import instrument_httpx
 from polar.logging import CorrelationID, Logger
 from polar.operational_errors import handle_operational_error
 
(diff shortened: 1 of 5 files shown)


frankie567 (approved):
Ah neat approach 👍

PR #12795, opened 2026-06-30
feat(sso): organization-scoped OIDC SSO login

maximevast:
## Summary

Adds the per-organization OIDC SSO login flow (`/v1/auth/{slug}/sso/{connection_id}/authorize` → IdP → callback → slug-scoped complete) that authenticates an existing member against their organization's identity provider (client_secret or private_key_jwt via reauth 0.2.0) and mints a user session down-scoped to exactly that one organization, rejecting non-members and unverified emails.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/12795?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/polar/auth/sso/factor.py
```diff
@@ -0,0 +1,120 @@
...
+        client_secret: str,
+        state_service: OAuth2StateService,
+    ) -> None:
+        super().__init__(
+            identifier=str(connection_id),
+            client_id=client_id,
+            client_secret=client_secret,
+            state_service=state_service,
+        )
+        self.DISCOVERY_ENDPOINT = _discovery_endpoint(issuer)
```
frankie567, line 82: Now that I see it, it's a bit dumb from `reauth` to have a class var for the discovery endpoint vs a simple object attribute set in `__init__`.  Could be a nice follow-up improvement

server/polar/auth/sso/factor.py
frankie567, line 1: Let's move SSO related stuff into its own package: `polar/auth/sso/...`.

`polar/auth/oauth2/helpers.py` can be moved to `polar/auth/helpers.py`.

server/polar/auth/oauth2/sso_login.py
```diff
@@ -0,0 +1,234 @@
...
+        )
+    except (IdentityNotAttachedException, FactorsRemainingException) as e:
+        raise PolarAuthRedirectionError(
+            "Authentication session cannot be completed"
+        ) from e
+
+    user_repository = UserRepository.from_session(session)
+    user = await user_repository.get_by_id(identity_id)
+    if user is None:
+        raise PolarAuthRedirectionError("User not found for authenticated identity")
```
frankie567: I think we should check again the membership of the user at this point — it might have changed between the callback and complete.

server/polar/auth/oauth2/sso_login.py
```diff
@@ -0,0 +1,234 @@
...
+    session: AsyncSession = Depends(get_db_session),
+) -> RedirectResponse:
+    context = authentication_session.context or {}
+    sso_organization_id = context.get("sso_organization_id")
+    if sso_organization_id is None:
+        raise PolarAuthRedirectionError("Authentication session cannot be completed")
+
+    # An SSO factor alone completes the login, even for members enrolled in other
+    # factors, so we don't enforce remaining factors here.
+    try:
```
frankie567: I'm not sure it's necessary for us to do this. I see two cases:

1. The organization _enforces_ the SSO for this org (follow-up effort). Thus, the only factor available is the SSO connection; so nothing to skip.
2. The organization doesn't enforce it, then, I'm more than fine to challenge users with their 2FA if they have it. Feels a bit more safe to me since we don't have any guarantee the OIDC provider enforced a 2FA on their side. Render has this behavior.

server/polar/auth/authentication_session.py
```diff
@@ -104,6 +107,23 @@ async def delete(
         await self.session.execute(statement)
         await self.session.flush()
 
+    async def complete(
```
frankie567: Given my comment https://github.com/polarsource/polar/pull/12795/changes#r3498243385, this would make unnecessary to overload this method

server/polar/auth/factors.py
```diff
@@ -290,3 +295,23 @@ async def get_factors(
...
+) -> set[FactorBase[typing.Any]]:
+    organization_repository = OrganizationRepository.from_session(session)
+    organization = await organization_repository.get_by_slug(slug)
+    if organization is None:
+        raise ResourceNotFound()
+
+    sso_repository = OrganizationSSOConnectionRepository.from_session(session)
+    connections = await sso_repository.get_enabled_by_organization(organization.id)
+
+    return base_factors | {
```
frankie567, line 315: _Future_: if organization _enforces_ SSO, we don't include base factors 🧠

server/polar/auth/service.py
```diff
@@ -138,6 +142,11 @@ async def _create_user_session(
             scopes=scopes,
             expires_at=utc_now() + expire_in,
         )
+        if organization_ids is not None:
+            user_session.organization_scopes = [
+                UserSessionOrganization(organization_id=organization_id)
+                for organization_id in organization_ids
+            ]
```
frankie567, line 149: Here we are 🔥

server/polar/auth/schemas.py
frankie567, line 51: We'll likely need a proper schema for Factor; since we'll need the SSO Connection ID on the frontend's side. Something like:

```py
class BaseFactor(Schema):
    type: typing.Literal["email_otp", "totp", "backup_codes", "apple", "github", "google"]

class SSOFactor(Schema):
    type: typing.Literal["sso"]
    connection_id: UUID4
    organization_slug: str

Factor = BaseFactor | SSOFactor
```

frankie567 (changes requested):
Yes, we're finally getting there :) A bit of refactor there and there, but that's super solid already 👍

clients/packages/client/src/v1.ts
frankie567, line 1: Doesn't it create some issues to fix in the web app?

frankie567 (approved):

PR #12799, opened 2026-06-30
lambda: Implement longer backoffs for retries

Yopi:
In https://github.com/polarsource/polar/pull/12614 we implemented exponential backoff, but only up until 12 hours. This PR adds an EventBridge connection to allow us to schedule delays for more than 12 hours, which gives us equivalent behaviour as we have with Dramatiq.

<a href="https://cubic.dev/pr/polarsource/polar/pull/12799?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


frankie567 (approved):
Interesting! Can we imagine in the future to move our cron job to this scheduler?

Yopi:
> Interesting! Can we imagine in the future to move our cron job to this scheduler?

@psincraian and I discussed it briefly here https://github.com/polarsource/polar/pull/12449#discussion_r3435510907, and we could definitely do it! It's a bit a question of where it is nicer to have them configured (in Python or in Terraform). Having them in Python is a bit closer to the code so it might be nicer after all, but it's doable either way

PR #12806, opened 2026-06-30
Improve event type selector

pieterbeulque:
image

Allow to select all custom events & all system events, and use the friendly event labels instead of the system names.

<a href="https://cubic.dev/pr/polarsource/polar/pull/12806?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/event/service.py
+++ b/server/polar/event/service.py
@@ -66,7 +66,7 @@ from .schemas import (
     VarianceEvent,
 )
 from .sorting import EventNamesSortProperty, EventSortProperty
-from .system import SystemEvent
+from .system import SYSTEM_EVENT_LABELS, SystemEvent
 
 log: Logger = structlog.get_logger()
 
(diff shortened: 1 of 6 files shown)


server/polar/event/service.py
```diff
@@ -779,9 +779,25 @@ async def list_names(
         end = start + pagination.limit
         paginated_stats = tinybird_stats[start:end]
 
+        event_type_repository = EventTypeRepository.from_session(session)
```
pieterbeulque, line 782: @Yopi Not sure if this will have a performance impact?

Yopi: It might have 😄, but we can try it out

Yopi (approved):
I think the bot might have a point in regards of select all though. Might be worth to treat it "all" separately ?

frankie567 (approved):

pieterbeulque:
@Yopi The bot has a point indeed. I looked into a fix but it's not super straightforward. Let's wing it for now

PR #12846, opened 2026-07-01
Embed organization permissions in auth payload for sync checks

pieterbeulque:
## Summary

Refactor permission checking to use permissions embedded in the authenticated user payload instead of fetching role definitions. This eliminates loading states and enables server-side permission gating during SSR.

This would also fix e.g. https://github.com/polarsource/polar/issues/11865

## What

- Modified `useHasPermission` hook to return a synchronous boolean instead of `boolean | undefined` by reading permissions directly from `userOrganizations`
- Created new `OrganizationPermissionGuard` server component for gating route segments on organization permissions during SSR
- Updated finance layout to use `OrganizationPermissionGuard` for `finance:read` permission
- Removed client-side permission checks and `AccessRestricted` fallbacks from `PayoutsPage` and `IncomePage` components
- Extended `OrganizationWithRole` schema to include a `permissions` field computed from the user's role
- Updated generated client types to reflect the new `permissions` field

## Why

The previous implementation had several issues:
- Permission checks were asynchronous, causing loading states and potential UI flashes
- Role definitions were fetched separately, adding unnecessary network requests
- Client-side checks couldn't prevent unauthorized pages from rendering during SSR
- Permissions needed to be checked in multiple places with duplicated logic

By embedding permissions in the authenticated payload, we get:
- Synchronous permission checks with no loading state
- Server-side gating that prevents unauthorized content from rendering
- Single source of truth (the auth payload) instead of separate role queries
- Cleaner component code without permission check boilerplate

## How

1. Backend: Extended `OrganizationWithRole` to compute and include the user's permissions based on their role
2. Frontend: Updated `useHasPermission` to read from the embedded permissions instead of querying roles
3. Created `OrganizationPermissionGuard` server component to gate entire route segments
4. Applied the guard to the finance routes and removed redundant client-side checks
5. Updated generated client types to match the schema changes

## Checklist

- [x] This PR addresses a single concern (refactoring permission checks for sync evaluation)
- [x] No new tests needed (existing permission logic preserved, just moved)

https://claude.ai/code/session_015P4L1cxHgDy1yGfpXUduyx

<a href="https://cubic.dev/pr/polarsource/polar/pull/12846?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pieterbeulque:
@frankie567 Would love to hear your thoughts on this approach

frankie567 (approved):
Ah yes, that's way better. Looking at permissions directly vs role makes sense since it's actually just an alias for a set of permissions.

maximevast (approved):
Neat refactor!

strandhvilliam:
By order of @pieterbeulque, this PR was split to reduce risk of next cache inconsistencies when deploying. New branches are:

- api: https://github.com/polarsource/polar/pull/13213
- web: https://github.com/polarsource/polar/pull/13214

Also tested a bit more and added a small fix.

pieterbeulque:
Tested & rolled out by @strandhvilliam 🦸

PR #12847, opened 2026-07-01
webhooks: Set max backoff

Yopi:
Without the max backoff we will always come to 2s backoff, which means that all retries fire within ~20s of each other

The docs state
> Delivery Retries
> If we hit an error while trying to reach your endpoint, whether it is a temporary network error or a bug, we’ll retry to send the event up to 10 times with an exponential backoff.

So I believe that this is the correct way to go.

<a href="https://cubic.dev/pr/polarsource/polar/pull/12847?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/webhook/tasks.py
+++ b/server/polar/webhook/tasks.py
@@ -188,6 +188,7 @@ async def _webhook_event_send(
             _, delay = compute_backoff(
                 delivery_count,
                 factor=settings.WORKER_MIN_BACKOFF_MILLISECONDS,
+                max_backoff=20 * 60_000,
             )
             raise Retry(delay=delay) from e
     # Success


server/polar/webhook/tasks.py
```diff
@@ -189,16 +188,9 @@ async def _webhook_event_send(
             _, delay = compute_backoff(
                 delivery_count,
                 factor=settings.WORKER_MIN_BACKOFF_MILLISECONDS,
+                max_backoff=20 * 60_000,
```
pieterbeulque, line 191: Hmmm. This feels silly, as `WORKER_MIN_BACKOFF_MILLISECONDS` is also set to two seconds, we were basically just scheduling a retry two seconds later?

I'm not sure if/how this interplays with the FIFO_GUARD_* variables. Should it be set to `WEBHOOK_FIFO_GUARD_MAX_AGE` or does that have nothing to do with it?

pieterbeulque (approved):

server/polar/webhook/tasks.py
frankie567: FIFO is not guaranteed in case of retries. If there is a delivery attempt, then we consider we fulfilled our ordering guarantee.

frankie567 (approved):
Good catch!

server/polar/webhook/tasks.py
Yopi: Haha this is why I tagged you. I'm also not 100% sure. I think if you are having API issues, letting this be higher than the FIFO GUARD would let your other requests get a try as well even in the incorrect order. Setting it to `WEBHOOK_FIFO_GUARD_MAX_AGE` would cap the exponential backoff to 1m

PR #12853, opened 2026-07-01
fix(checkout): populate customer name from cardholder name when unset

pieterbeulque:
An existing customer without a name (e.g. created through the API without
one, or matched by email) never received a name during checkout, because
customer.name was only written for newly created customers. As a result
billing_name — which falls back to name — stayed null, so orders had no
billing name and invoices could neither auto-generate nor be created via
the API.

Set customer.name from the checkout/cardholder name whenever the customer
has no name yet, regardless of whether it was just created. Existing
customers that already have a name are still protected from being
overwritten by the cardholder name (e.g. a company customer paid for by an
employee), preserving the previous team-name fix.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01UxtYmEsYmo9vkpmm3bNr9C

<a href="https://cubic.dev/pr/polarsource/polar/pull/12853?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/scripts/backfill_customer_name.py
+++ b/server/scripts/backfill_customer_name.py
@@ -1,0 +1,171 @@
+import asyncio
+from typing import Any, cast
+
+import typer
+from rich.progress import (
+    Progress,
+    SpinnerColumn,
+    TextColumn,
+    TimeElapsedColumn,
+)
+from sqlalchemy import (
+    ColumnElement,
+    CursorResult,
+    func,
+    select,
+    update,
+)
+
+from polar.config import settings
+from polar.kit.db.postgres import AsyncSession, create_async_sessionmaker
+from polar.kit.db.postgres import create_async_engine as _create_async_engine
+from polar.models import Checkout, Customer
+from polar.models.checkout import CheckoutStatus
+
+from .helper import configure_script_logging, typer_async
+
+cli = typer.Typer()
+
+
+def _computed_name() -> ColumnElement[str | None]:
+    """
+    Name derived from the customer's earliest succeeded checkout.
+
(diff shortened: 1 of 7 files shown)


Yopi (approved):

frankie567 (approved):

server/scripts/backfill_customer_name.py
```diff
@@ -0,0 +1,171 @@
...
+        ) as progress:
+            task = progress.add_task("[cyan]Batch 0: 0 rows updated", total=None)
+
+            while True:
+                batch = (
+                    select(Customer.id)
+                    .where(
+                        Customer.deleted_at.is_(None),
+                        Customer.name.is_(None),
+                        Customer.name.is_distinct_from(_computed_name()),
```
Yopi: It feels a bit strange to do both is none and is distinct from computed_name, but I guess computed_name has the null check as well to ensure the difference here

Yopi (approved):
A bit scary to update existing orders, but they can't generate invoice right now either way with those preconditions right?

server/scripts/backfill_customer_name.py
pieterbeulque: Yeah, that's an unnecessary duplicate.

PR #12855, opened 2026-07-01
feat(sso): organization SSO login page

maximevast:
## Summary

Adds `/auth/sso/[slug]` — the login page for an organization.

It starts an org-scoped session (`POST /{slug}/start`) and renders one button per returned `available_factors`: the org's SSO connections **plus** the base factors (Apple/GitHub/Google + email OTP). 
It's fully driven by `available_factors`, so once "forced SSO" ships (base factors dropped server-side) the page shows only SSO buttons with no change here.

The SSO callback now redirects to `/auth/sso/{slug}` (instead of the global `/auth`) so the org-scoped session completes via `/{slug}/complete` and stays scoped to the organization.

Base factors on this page do a normal global login (SSO is the only org-scoped path) — acceptable since forced-SSO will hide them.

https://github.com/user-attachments/assets/1a3b8074-fb1b-46a4-b60a-b93c0a542c2b

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/12855?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/auth/endpoints.py
+++ b/server/polar/auth/endpoints.py
@@ -1,4 +1,5 @@
 import typing
+from uuid import UUID
 
 from fastapi import Depends, Request, Response
 from fastapi.responses import RedirectResponse
@@ -34,6 +35,7 @@ from polar.postgres import AsyncSession, get_db_session
 from polar.routing import APIRouter
 from polar.user.repository import UserRepository
 from polar.user.service import user as user_service
+from polar.user_organization.repository import UserOrganizationRepository
 
 from .authentication_session import (
     AuthenticationSessionService,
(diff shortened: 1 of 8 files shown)


server/polar/auth/sso/endpoints.py
```diff
@@ -263,10 +263,11 @@
...
     )
 
-    # Like the social login flow, hand off to the frontend so any remaining
-    # factors (e.g. TOTP) are challenged before the session completes
+    # Hand off to the organization login page so any remaining factors (e.g.
+    # TOTP) are challenged and the org-scoped session is completed via
+    # `/{slug}/complete` rather than the global, unscoped completion.
     response = RedirectResponse(
-        settings.generate_frontend_url("/auth"), status_code=303
+        settings.generate_frontend_url(f"/auth/sso/{slug}"), status_code=303
```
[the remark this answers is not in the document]
frankie567: I think it's not true since we resolved `slug` at this point, right? So it should be safe

server/polar/auth/endpoints.py
```diff
@@ -146,15 +148,32 @@ async def complete(
         raise PolarAuthRedirectionError("User not found for authenticated identity")
 
     context = authentication_session.context or {}
-    factor: LoginMethod = typing.cast(
-        LoginMethod, authentication_session.used_factors[0]
-    )
+
+    # An SSO-authenticated session stays scoped to its organization, whichever
+    # completion path (including the global 2FA pages) it reaches.
```
frankie567, line 153: Good thing that we guard that; that said, could we simply error instead of trying to accommodate? Not certitude here, just want to know what you think; and what could be the downsides.

maximevast: but then when doing SSO -> Authenticator APP; you end up with an error which is not what we want correct?

frankie567: Ah yes, right, didn't think about the TOTP routes. Hmm. Let's keep it like this then.

frankie567 (approved):

PR #12874, opened 2026-07-02
feat(backoffice): tag violated AUP section on human denials and offboarding

IsakPar021:
## Summary

Add a structured, required "Violated AUP section" picker to all four backoffice
human-denial dialogs, so denials can be grouped by policy section in Metabase
instead of relying on free-text reasons.

## What

- Add an `AUPSection` enum covering the 36 AUP "Prohibited Products" sections
  plus an `Other` catch-all, with display labels decoupled from the stored codes
  so policy renumbering never invalidates historical rows.
- Add a nullable `violated_aup_section` column on `organization_review_feedback`
  (+ migration) and thread it through `record_human_decision`.
- Render a required select in the **Deny**, **Set Offboarding**, **Deny Appeal**
  (org detail page, shown for denied orgs with a pending appeal), and
  support-case **Deny Appeal** dialogs, via shared parse/render helpers.
  Submissions without a valid section are rejected server-side and re-render
  the dialog with an inline error.
- Display the chosen section in the review section of the org detail page.

## Why

Human denial reasons are currently free text, which makes it impossible to
answer "how many orgs did we deny per AUP section" in Metabase. A single
enum column on the existing review-feedback row covers all four denial
surfaces with one `GROUP BY`.

The column is nullable so historical rows are untouched; enforcement is
at the dialog layer.

## Checklist


psincraian (approved):
Love this! 

Image

server/polar/organization_review/schemas.py
```diff
@@ -27,6 +27,156 @@ class DecisionType(StrEnum):
     SNOOZE = "SNOOZE"
 
 
+class AUPSection(StrEnum):
+    """Acceptable Use Policy 'Prohibited Products' (1-36), plus a catch-all `other`.
+
+    The stored value is the *code*, decoupled from the doc's display number,
+    so policy renumbering never invalidates historical rows or Metabase queries.
+    """
```
frankie567, line 36: Therefore, I think we should get rid of the numbering in this enum. It doesn't really make sense to hardcode it, especially since it'll likely change with AUP updates.

If the point is to render those values in a certain order, then it's better to have an explicit function or class method that returns them as an explicit ordered list that can be easily updated — but the keys remain stable over time.

frankie567 (changes requested):

server/polar/organization_review/schemas.py
IsakPar021: Agreed, thanks! Dropped the numbering from the enum entirely — keys are now stable descriptive codes (`trading_financial`, `osint`, …), and the display numbering/ordering moved to an `AUP_SECTION_LABELS` dict that can be freely updated when the AUP changes (same pattern as `REVIEW_CONTEXT_LABELS` and friends). Since this never shipped there are no stored rows with the old values, so no data migration needed. Hope this fixes it!

server/migrations/versions/2026-06-28-1147_add_violated_aup_section_to_.py
```diff
@@ -0,0 +1,37 @@
...
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# Polar Custom Imports
+
+# revision identifiers, used by Alembic.
+revision = "94bf1c4552b6"
+down_revision = "31d662178679"
```
frankie567: I think you'll need to rebase against main and rearrange the file name and down revision so the migration chronology is correct.

frankie567 (approved):

server/migrations/versions/2026-06-28-1147_add_violated_aup_section_to_.py
IsakPar021: Rebased against main and moved the migration to the head of the chain — file renamed to today and `down_revision` now points at `b181f0be8208` (the `slack_apps` migration it was forked alongside). Single head verified.

frankie567: Let's go then!

PR #12875, opened 2026-07-02
feat(organization): block organization creation from a scoped session

maximevast:
## Summary

Block organization creation from a scoped session (backend + UI)

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/12875?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/apps/web/src/components/Layout/Dashboard/DashboardSidebar.tsx
+++ b/clients/apps/web/src/components/Layout/Dashboard/DashboardSidebar.tsx
@@ -1,4 +1,5 @@
 import { SupportButton } from '@/components/Feedback/SupportButton'
+import { useAuth } from '@/hooks/auth'
 import { NotificationsPopover } from '@/components/Notifications/NotificationsPopover'
 import { OmniSearch } from '@/components/Search/OmniSearch'
 import { CONFIG } from '@/utils/config'
@@ -50,6 +51,7 @@ export const DashboardSidebar = ({
 }) => {
   const router = useRouter()
   const { state } = useSidebar()
+  const { currentUser } = useAuth()
 
   const isCollapsed = state === 'collapsed'
   const [searchOpen, setSearchOpen] = useState(false)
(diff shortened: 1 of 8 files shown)


clients/apps/web/src/components/Layout/Dashboard/DashboardSidebar.tsx
```diff
@@ -232,15 +234,17 @@ export const DashboardSidebar = ({
...
-                  <DropdownMenuItem
-                    onClick={() =>
-                      CONFIG.IS_SANDBOX
-                        ? router.push('/onboarding/sandbox')
-                        : router.push('/onboarding/business')
-                    }
-                  >
-                    New Organization
-                  </DropdownMenuItem>
+                  {!currentUser?.organization_scoped && (
```
pieterbeulque, line 237: How can they then work around this? Log out and then log in through a non-scoped authentication method?

pieterbeulque (approved):

clients/apps/web/src/components/Layout/Dashboard/DashboardSidebar.tsx
pieterbeulque: Imagine I'm the Polar admin, Polar uses SSO, and I want to create a new organization. How does it work?

maximevast: Yep ! Org scoped sessions cannot access other user orgs by definition; so creating a new one would result in the newly created one being invisible.

maximevast: You would then need to login using non SSO factors. ~The problem will be if you enabled "sso_enforced" on your org.~ Well, no you can still login but will have no organization, and you'll be able to create one

pieterbeulque: Okay! Sounds rough around the edges but we can smoothen that out once we get to it!

maximevast: Happy to discuss a better solution !
Currently, only sso sessions will be down scoped to 1 org; so I think that the behavior is correct. We can discuss it further when @frankie567 gets back.

maximevast: I guess the question is really "do we want sso sessions to only see the 1 org it's linked to". It's a decision we took with @frankie567 but we can reconsider later if needed.

PR #12707, opened 2026-06-26
Allow expired license keys backed by subscriptions to remain valid

pieterbeulque:
## Summary

License keys that are backed by active subscriptions should remain valid even after their TTL expires. Only license keys granted through one-time orders should respect the expiration date.

## What

- Added `is_granted_through_subscription()` method to `LicenseKeyRepository` to check if a license key is backed by an active subscription
- Updated `validate()` and `activate()` methods in `LicenseKeyService` to skip expiration checks for subscription-backed license keys
- Added comprehensive test coverage for both expired order-based and subscription-based license keys in both activation and validation flows

## Why

License keys granted as part of a subscription benefit should be valid as long as the subscription is active, regardless of any TTL set on the key itself. The expiration date should only apply to one-time order-based license keys. This allows organizations to manage license key validity through subscription status rather than manual TTL management.

## How

1. Added a repository method that queries for a `BenefitGrant` with a non-null `subscription_id` for the given license key
2. Modified the expiration check logic in both `validate()` and `activate()` to call this method before rejecting expired keys
3. Added four new test cases:
   - `test_activate_expired_order_license_key_should_fail`: Verifies order-based expired keys are rejected
   - `test_activate_expired_subscription_license_key_succeeds`: Verifies subscription-based expired keys are accepted
   - `test_validate_expired_order_license_key_should_fail`: Verifies order-based expired keys fail validation
   - `test_validate_expired_subscription_license_key_succeeds`: Verifies subscription-based expired keys pass validation
4. Added `create_order_grant()` helper method to `TestLicenseKey` fixture for creating order-based license key grants

## Checklist

- [x] This PR addresses a single concern (distinguishing expiration handling for subscription vs order-based license keys)
- [x] The diff is reasonably sized and easy to review
- [x] New functionality is covered by tests (4 new test cases added)
- [x] No unrelated changes or drive-by fixes are included

https://claude.ai/code/session_012neVzkYXantd74AvaostP2

<a href="https://cubic.dev/pr/polarsource/polar/pull/12707?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #12876, opened 2026-07-02
feat: issue license keys without expiration when granted by subscriptions

joebon:
An alternative approach to https://github.com/polarsource/polar/pull/12707

---
## Summary by cubic
Issue license keys without expiration when granted through subscriptions; one‑time order grants still use TTL. Propagates grant scope to benefit strategies and clears existing expirations on subscription-backed keys.

- **New Features**
  - Threads grant scope (`subscription_id`/`order_id`) into all benefit strategies; `license_keys` skips setting `expires_at` when a subscription is present.
  - On benefit grant updates, unsets `expires_at` for subscription-backed license keys.

- **Migration**
  - Run `python server/scripts/unset_subscription_license_key_expiration.py` to remove `expires_at` from license keys granted through billable subscriptions.

<sup>Written for commit 26576a0545be08870d3a30a263106aaaf98b7b4d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12876?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/benefit/grant/service.py
+++ b/server/polar/benefit/grant/service.py
@@ -238,6 +238,7 @@ class BenefitGrantService(ResourceServiceReader[BenefitGrant]):
                 grant.properties,
                 attempt=attempt,
                 member=member,
+                **scope_to_args(scope),
             )
         except BenefitActionRequiredError as e:
             if e.grant_properties is not None:
@@ -521,6 +522,7 @@ class BenefitGrantService(ResourceServiceReader[BenefitGrant]):
                 update=True,
                 attempt=attempt,
                 member=member,
+                **grant.scope,
             )
         except BenefitActionRequiredError as e:
             if e.grant_properties is not None:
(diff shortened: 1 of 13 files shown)


frankie567 (commented):
I prefer that approach :) That said, since we change the benefit strategy contract, might be worth to take this opportunity to pass the whole `**scope` instead of just the subscription. WDYT?

joebon:
> I prefer that approach :) That said, since we change the benefit strategy contract, might be worth to take this opportunity to pass the whole `**scope` instead of just the subscription. WDYT?

Boy scouting, I like it! 🫡

frankie567 (approved):

PR #12877, opened 2026-07-02
feat(sso): enforce Force SSO on organization access

maximevast:
## Summary

Enforces an organization's Force SSO (`sso_enforced`) setting: 
a user session that wasn't established through the org's SSO connection can no longer access that organization, and the dashboard redirects such users to re-authenticate via SSO.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/12877?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/apps/web/src/app/(main)/dashboard/[organization]/layout.tsx
+++ b/clients/apps/web/src/app/(main)/dashboard/[organization]/layout.tsx
@@ -1,20 +1,25 @@
 import { AccountSetupProvider } from '@/providers/accountSetup'
 import { OrganizationContextProvider } from '@/providers/maintainerOrganization'
 import { getServerSideAPI } from '@/utils/client/serverside'
-import { getOrganizationBySlugOrNotFound } from '@/utils/organization'
-import { getUserOrganizations } from '@/utils/user'
+import {
+  getOrganizationBySlug,
+  getOrganizationBySlugOrNotFound,
+} from '@/utils/organization'
+import { getAuthenticatedUser } from '@/utils/user'
 import { Metadata } from 'next'
-import { redirect } from 'next/navigation'
+import { notFound, redirect } from 'next/navigation'
 
 export async function generateMetadata(props: {
   params: Promise<{ organization: string }>
 }): Promise<Metadata> {
   const params = await props.params
   const api = await getServerSideAPI()
-  const organization = await getOrganizationBySlugOrNotFound(
-    api,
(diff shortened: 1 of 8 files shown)


clients/apps/web/src/app/(main)/dashboard/[organization]/layout.tsx
```diff
@@ -40,6 +40,12 @@ export default async function Layout(props: {
   const userOrganizations = await getUserOrganizations()
 
   if (!userOrganizations.some((org) => org.id === organization.id)) {
+    // An org that enforces SSO is only reachable through an SSO-scoped session,
```
Comment (pieterbeulque) [inline clients/apps/web/src/app/(main)/dashboard/[organization]/layout.tsx]: I have limited understanding but would this "expose" that the organization requires SSO even if the user doesn't belong to the organization at all?

And if so, is that a bad thing?

Put differently, should we redirect differently between:

 - A user tries to access an organization they aren't a member of
 - A user tries to access an organization they are a member of, but they are not in an SSO-scoped session

pieterbeulque (approved):

clients/apps/web/src/app/(main)/dashboard/[organization]/layout.tsx
Comment (maximevast) [reply clients/apps/web/src/app/(main)/dashboard/[organization]/layout.tsx]: Hum, you're probably right, we should not mix those 2 concerns. Let me update this.

Comment (maximevast) [reply clients/apps/web/src/app/(main)/dashboard/[organization]/layout.tsx]: @pieterbeulque I push another version!

pieterbeulque (approved):

server/polar/user/endpoints.py
```diff
@@ -65,19 +67,26 @@ async def get_authenticated(
...
     return UserRead.model_validate(user).model_copy(
         update={
             "organizations": [
                 OrganizationWithRole.from_organization(org, role)
                 for org, role in org_with_roles
+                if org.id in accessible_ids
+            ],
+            "member_organizations": [
+                MemberOrganization(
+                    id=org.id, slug=org.slug, requires_sso=org.sso_enforced
```
frankie567, line 87: Oh, nice!

PR #12898, opened 2026-07-02
Add "Why you should win" and "Merchant attachments" section to disputes

sebastianekstrom:
## Summary

This PR will:

* add a `Why you should win` section to our disputes, similar to what Stripe does, which will be shown in the Backoffice for easier disputes.
* Add a `Merchant submitted files` section in the backoffice, which updates for every new attachment the merchant sends. 
* Add a new endpoint so that we can show a preview of merchant attachments
* add some UI polish to the dashboard (loading state) and Backoffice (border colors).

## Screenshots

| Dashboard | Backoffice  |
|--------|--------|
| Screenshot 2026-07-03 at 11 45 23 | Screenshot 2026-07-03 at 12 06 06 |

### DIFF
--- a/server/polar/backoffice/support_cases/endpoints.py
+++ b/server/polar/backoffice/support_cases/endpoints.py
@@ -370,6 +370,7 @@ async def case_detail(
         author_emails=author_emails,
         current_user_id=user_session.user_id,
         attachments_by_message=attachments_by_message,
+        attachments=attachments,
         dispute=dispute,
         return_to=return_to if is_safe_return_to(return_to) else None,
     )
--- a/server/polar/support_case/schemas.py
+++ b/server/polar/support_case/schemas.py
@@ -6,6 +6,7 @@ from pydantic import UUID4, Field, model_validator
 from polar.exceptions import ResourceNotFound
 from polar.kit.schemas import IDSchema, Schema, TimestampedSchema
 from polar.models.support_case import (
+    DisputeWinReason,
     SupportCaseMessageAuthorKind,
     SupportCaseMessageType,
     SupportCaseType,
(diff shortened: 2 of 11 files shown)


server/polar/backoffice/support_cases/endpoints.py
```diff
@@ -322,6 +324,37 @@ async def download_attachment(
...
+    if attachment is None or not attachment.file.mime_type.startswith("image/"):
+        raise HTTPException(status_code=404, detail="Attachment not found")
+
+    s3_service = S3_SERVICES[attachment.file.service]
+    obj = await run_in_threadpool(s3_service.get_object_or_raise, attachment.file.path)
+    return StreamingResponse(
+        obj["Body"],
+        media_type=attachment.file.mime_type,
+        headers={"Cache-Control": "private, max-age=3600"},
+    )
```
frankie567: Ooooof! Wouldn't it be simpler to relax the CSP to allow our S3 buckets? :D

server/polar/support_case/schemas.py
```diff
@@ -55,6 +56,18 @@ class SupportCaseMessageCreate(Schema):
...
+        description=(
+            "For a dispute case, the merchant's stated grounds for contesting. "
+            "Set on the counter submission; ignored for other case types."
+        ),
+    )
+    dispute_win_reason_other: str | None = Field(
+        default=None,
+        max_length=500,
+        description="Free-text detail when `dispute_win_reason` is `other`.",
+    )
```
frankie567, line 71: I we start to have specific fields depending on the support case; then I think we should have dedicated schemas per type of support case; and pass those directly to the respective `add_reply` service.

Thoughts @maximevast?

server/polar/backoffice/support_cases/endpoints.py
sebastianekstrom: Good point, that might actually be easier 😬 Fixed!

maximevast (approved):
LGTM !!

server/polar/support_case/schemas.py
maximevast: I agree but maybe it can wait a third variant? @frankie567

frankie567: ok then

frankie567 (approved):

PR #12903, opened 2026-07-02
tf: Add pgBouncer

Yopi:
This adds pgbouncer services to all environments, mapped to each replica of the db.

The initial step here is to roll it out, and the follow up is to start using it for the different services.

<a href="https://cubic.dev/pr/polarsource/polar/pull/12903?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/infra/pgbouncer/Dockerfile
+++ b/infra/pgbouncer/Dockerfile
@@ -1,0 +1,23 @@
+FROM debian:trixie-slim
+
+RUN apt-get -qq update \
+  && apt-get -qq install -y --no-install-recommends \
+    ca-certificates \
+    postgresql-common \
+  > /dev/null \
+  && /usr/share/postgresql-common/pgdg/apt.postgresql.org.sh -y \
+  && apt-get -qq install -y --no-install-recommends pgbouncer > /dev/null \
+  && apt-get -qq clean \
+  && rm -rf \
+    /var/lib/apt/lists/* \
+    /tmp/* \
+    /var/tmp/*
+
+COPY --chmod=755 entrypoint.sh /usr/local/bin/pgbouncer-entrypoint
+
+RUN chown -R postgres:postgres /etc/pgbouncer
+
+USER postgres
+EXPOSE 5432
+
+CMD ["/usr/local/bin/pgbouncer-entrypoint"]
--- a/infra/pgbouncer/Dockerfile
+++ b/infra/pgbouncer/Dockerfile
@@ -1,5 +1,5 @@
 FROM debian:trixie-slim
-
+ARG PGBOUNCER_APT_VERSION=1.25.2-1.pgdg13+1
 RUN apt-get -qq update \
   && apt-get -qq install -y --no-install-recommends \
     ca-certificates \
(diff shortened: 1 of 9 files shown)


terraform/modules/pgbouncer/main.tf
```diff
@@ -0,0 +1,32 @@
+locals {
```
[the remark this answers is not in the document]
Yopi: This is built via the `build_infra_images.yml`. Then it needs to get deployed via terraform instead of being deployed via the CI like the services that we currently have.

infra/pgbouncer/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM debian:trixie-slim
+
```
joebon, line 2:
```suggestion
ARG PGBOUNCER_APT_VERSION=1.25.2-1.pgdg13+1
```

infra/pgbouncer/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN apt-get -qq update \
+  && apt-get -qq install -y --no-install-recommends \
+    ca-certificates \
+    postgresql-common \
+  > /dev/null \
+  && /usr/share/postgresql-common/pgdg/apt.postgresql.org.sh -y \
+  && apt-get -qq install -y --no-install-recommends pgbouncer > /dev/null \
```
joebon:
```suggestion
  && apt-get -qq install -y --no-install-recommends pgbouncer="${PGBOUNCER_APT_VERSION}" > /dev/null \
```

infra/pgbouncer/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM debian:trixie-slim
```
joebon, line 3: Any inspiration Dockerfile we can reference here that could help realigning with after drift in the future?

terraform/production/render.tf
```diff
@@ -406,6 +406,69 @@ module "production" {
...
+  pgbouncer_read_pool_configs = {
+    "polar-read" = {
+      max_client_conn   = "5000"
+      default_pool_size = "50"
+      reserve_pool_size = "10"
+    }
+    "polar-replica" = {
+      max_client_conn   = "1000"
+      default_pool_size = "20"
+      reserve_pool_size = "0"
```
joebon, line 446: 👍

joebon (approved):
Nice!
Added an optional suggestion to pin the pgbouncer version up front to avoid accidental drift. That has bitten me before at least (not for pgbouncer though, but still).
Also an optional request for a reference to an inspiration Dockerfile, if any

infra/pgbouncer/Dockerfile
Yopi: Good idea!

Yopi:
> Nice! Added an optional suggestion to pin the pgbouncer version up front to avoid accidental drift. That has bitten me before at least (not for pgbouncer though, but still). Also an optional request for a reference to an inspiration Dockerfile, if any

Thanks! Great suggestions

joebon (approved):

PR #12911, opened 2026-07-02
Auto proration mode

pieterbeulque:
Context: https://github.com/polarsource/feedback/issues/35

Decisions:

 - An upgrade is either when the amount gets more expensive or when the interval period extends (monthly → annual)
 - A downgrade is when the amount gets cheaper or when the interval period shortens
 - Seat changes are excluded and always charged immediately

Open questions:

 - Maybe we should only allow this as an organization setting, but not as the proration behavior value on a `PATCH /v1/subscriptions/:id`? But maybe that's an unnecessary limitation.
 - There's a chance that some people will want more flexibility, so maybe the other fix is to split our current single setting into three different settings: `upgrade_proration` `downgrade_proration` `seat_change_proration`, and let that drive it. Open for suggestions.

---

Upgrades: Prorate + charge now
Downgrades: Schedule at period end

This is the most common requested behavior but impossible to configure if you're relying on the Customer Portal to manage your subscriptions.

<a href="https://cubic.dev/pr/polarsource/polar/pull/12911?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/polar/event/system.py
```diff
@@ -212,6 +212,7 @@ class SubscriptionCreatedEvent(Event):
 class SubscriptionUpdatedMetadataFields(TypedDict, total=False):
     product_id: str
     proration_behavior: SubscriptionProrationBehavior
+    requested_proration_behavior: SubscriptionProrationBehavior
```
pieterbeulque, line 215: I'm not sure about this… I think we can just resolve `auto` in the correct mode and "lose" that information.

PR #12914, opened 2026-07-02
feat(merchant-migration): add pre-check phase

psincraian:
## Summary

Adds the pre-check step of merchant migrations: before importing anything, read the merchant's connected billing source (Stripe), normalize it, and report whether its catalog + subscriptions can be imported into Polar. Nothing is written for now.

Three pieces to review:

1. adapters/stripe.py - reads customers / products+prices / subscriptions via a restricted-key Stripe client and maps them to CanonicalRecords (canonical.py). So the engine never sees Stripe types.
2. precheck.py - PrecheckEngine.run reads the adapter output and aggregates (email counts, product-name map). Each check yields a PrecheckIssue.
3. service.py / endpoints.py - a thin wiring only

There are two categories for checks:
- warning: the migration can proceed but the record is skipped. Most of the errors will be warnings. 
- blocker: the migration cannot be performed. For now, we only check the org status, but later on we will check stripe account country (we cannot import India PAN) or from Connect accounts.

## Checklist

### DIFF
--- a/server/polar/merchant_migration/adapters/base.py
+++ b/server/polar/merchant_migration/adapters/base.py
@@ -1,0 +1,14 @@
+from collections.abc import AsyncIterator
+from typing import Protocol
+
+from ..canonical import CanonicalRecord
+
+
+class SourceAdapter(Protocol):
+    """Reads one billing provider into provider-agnostic CanonicalRecords.
+
+    ``extract`` is an async iterator because source data can be huge and must be
+    streamed, not materialized.
+    """
+
+    def extract(self) -> AsyncIterator[CanonicalRecord]: ...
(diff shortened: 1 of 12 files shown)


server/polar/merchant_migration/adapters/base.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class SourceAdapter(Protocol):
+    """Reads one billing provider into provider-agnostic CanonicalRecords.
+
+    ``extract`` is an async iterator because source data can be huge and must be
+    streamed, not materialized.
+    """
+
+    def extract(self) -> AsyncIterator[CanonicalRecord]: ...
```
frankie567, line 14: The method is not asynchronous contrary to the Stripe implementation

frankie567 (approved):

PR #12934, opened 2026-07-03
Allow reusing slugs of soft-deleted custom fields

pieterbeulque:
## Summary

**Related Issue**: 

Allow custom field slugs to be reused after the original field is soft-deleted, while maintaining uniqueness constraints for active fields.

## What

- Created a new `CustomFieldRepository` with soft-deletion-aware query methods
- Replaced the global unique constraint on `(slug, organization_id)` with a partial unique index that only applies to non-deleted rows
- Updated `create()` and `update()` methods in `CustomFieldService` to use the new repository for slug validation
- Added database migration to implement the partial unique index
- Added comprehensive tests for both create and update operations with soft-deleted fields

## Why

Previously, once a custom field was soft-deleted, its slug remained permanently reserved and couldn't be reused. This is unnecessarily restrictive since soft-deleted records are logically removed from the system. By using a partial unique index that ignores soft-deleted rows, we allow slugs to be recycled while maintaining data integrity for active fields.

## How

1. **Repository Pattern**: Extracted slug lookup logic into `CustomFieldRepository` which respects soft-deletion status through the `RepositorySoftDeletionMixin`
2. **Database Constraint**: Replaced the unconditional unique constraint with a PostgreSQL partial unique index using `postgresql_where=text("deleted_at IS NULL")`
3. **Service Updates**: Modified `create()` and `update()` to use the repository's `get_by_organization_and_slug()` method, which automatically excludes soft-deleted records
4. **Tests**: Added test cases verifying that:
   - Creating a field with a slug of an active field still raises an error
   - Creating a field with a slug of a soft-deleted field succeeds
   - Updating a field to use a slug of a soft-deleted field succeeds

## Checklist

- [x] This PR addresses a single concern (allow slug reuse for soft-deleted fields)
- [x] New functionality is covered by tests (4 new test cases added)

https://claude.ai/code/session_01CDqgCLtZUBuhe63s2CUb7r

<a href="https://cubic.dev/pr/polarsource/polar/pull/12934?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/migrations/versions/2026-07-03-1000_custom_fields_slug_partial_unique_index.py
+++ b/server/migrations/versions/2026-07-03-1000_custom_fields_slug_partial_unique_index.py
@@ -1,0 +1,47 @@
+"""Make custom_fields slug uniqueness ignore soft-deleted rows
+
+Revision ID: d3f5a9c47e21
+Revises: b181f0be8208
+Create Date: 2026-07-03 10:00:00.000000
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# Polar Custom Imports
+
+# revision identifiers, used by Alembic.
+revision = "d3f5a9c47e21"
+down_revision = "b181f0be8208"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
+    # Ensures we don't break app by applying a deadlock-inducing migration
+    op.execute("SET LOCAL lock_timeout = '5s'")
+    op.drop_constraint(
+        "custom_fields_slug_organization_id_key", "custom_fields", type_="unique"
+    )
+    op.create_index(
+        "custom_fields_slug_organization_id_active_key",
+        "custom_fields",
+        ["slug", "organization_id"],
+        unique=True,
+        postgresql_where=sa.text("deleted_at IS NULL"),
+    )
+
+
+def downgrade() -> None:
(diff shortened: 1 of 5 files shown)


server/migrations/versions/2026-07-03-1000_custom_fields_slug_partial_unique_index.py
```diff
@@ -0,0 +1,47 @@
...
+def upgrade() -> None:
+    # Ensures we don't break app by applying a deadlock-inducing migration
+    op.execute("SET LOCAL lock_timeout = '5s'")
+    op.drop_constraint(
+        "custom_fields_slug_organization_id_key", "custom_fields", type_="unique"
+    )
+    op.create_index(
+        "custom_fields_slug_organization_id_active_key",
+        "custom_fields",
+        ["slug", "organization_id"],
```
pieterbeulque, line 30: I didn't want to change this but it's weird that this isn't ordered `organization_id, slug`

frankie567 (approved):

pieterbeulque:
@stilla can you fix the migration head

PR #12937, opened 2026-07-03
Revert "feat(terraform): add Cloudflare IP ranges module for forwarded_allow_ips (#12867)"

frankie567:
Uvicorn actually sees the Render proxy IP, and as such doesn't recognize it as a valid forwarded hosts; which breaks URL generation and HTTPS scheme detection.

Tested on test environment, wasn't applied on production/sandbox.

---
## Summary by cubic
Reverts the Cloudflare IP ranges module and updates `forwarded_allow_ips` to restore correct host and HTTPS scheme detection in Uvicorn behind Render. Fixes broken URL generation.

- **Bug Fixes**
  - Removed `terraform/modules/cloudflare_ips` and all module references.
  - Set `forwarded_allow_ips` to "*" in sandbox and test; removed the setting in production.
  - Uvicorn now sees the Render proxy as allowed, restoring correct URL and HTTPS behavior.

<sup>Written for commit b757416801d808397253f75fe3879de889be1f65. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12937?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Yopi (approved):

Yopi:
> Uvicorn actually sees the Render proxy IP, and as such doesn't recognize it as a valid forwarded hosts; which breaks URL generation and HTTPS scheme detection.

Should we test the same but with their IP?

frankie567:
> > Uvicorn actually sees the Render proxy IP, and as such doesn't recognize it as a valid forwarded hosts; which breaks URL generation and HTTPS scheme detection.
> 
> Should we test the same but with their IP?

That would be the ideal I guess, but I don't think they communicate it. I can ask

Issue #12103, opened 2026-06-02
Canceling a pending payout re-holds the released funds for the full payout delay


pieterbeulque:
When a pending payout is canceled, the transactions that were reserved for it are released but get re-subjected to the account's full `payout_transaction_delay`. As a result, the funds fall under the held balance and are not immediately available for a new payout, even though they already cleared the delay.

This also causes a mismatch between the _Available_ badge on the transactions themselves and the held balance on Finance → Income.

**Steps to reproduce**

1. Use an account with a non-zero payout_transaction_delay (default is 7 days).
2. Have balance transactions that have aged past the delay (i.e. currently available).
3. Create a payout — this reserves those transactions (payout_transaction_id is set).
4. Cancel the payout while it's pending.
5. Check the account's balance summary.

PR #12097, opened 2026-06-02
feat: allow held payouts for organizations under review

psincraian:
## Summary

Phase 2 of the [payout-reviews](https://handbook.polar.sh/engineering/design-documents/payout-reviews). Organizations under review (`REVIEW` or `SNOOZED`) can now request a payout instead of being blocked. The payout is created as `held`, then released automatically when the org is approved, or canceled and fully refunded if the org is denied, blocked, offboarded, or swaps its payout account.

## What

- Capability flip: `payouts` is now allowed for `REVIEW`/`SNOOZED`.
- Backfill script to change the payouts to enabled.
- `PayoutService.create` branches on org status under. Orgs in `Active` are created on `pending`, orgs in `REVIEW`/`SNOOZED` to `held`, everything else raises `OrganizationCannotPayout`.
- `payout.created` is split into a plain event plus a new `payout.transfer` task that performs the Stripe transfer.
- New `organization.release_held_payouts` and `organization.cancel_pending_payouts` jobs, wired into the org status transitions.
- Cancel now reverses the per-payout fees.
- Concurrency: `cancel` and `transfer_stripe` lock and read the row, so we avoid double credits. 
- Repository: the hourly transfer cron only picks `pending`, and held payouts count against a payout account so it can't be deleted with funds reserved.

## Checklist

- [x] This PR addresses a single concern (one bug fix, one feature, one refactor)
- [x] The diff is reasonably sized and easy to review
- [x] New functionality is covered by tests
- [x] Linting and type checking pass (`uv run task lint && uv run task lint_types`)
- [x] No unrelated changes or drive-by fixes are included

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  - Orgs in REVIEW/SNOOZED may request payouts (created as held); held payouts can be released back to pending and processed later. New background tasks support releasing and bulk-canceling by account.

* **Bug Fixes**
  - Held payouts are cancelable and properly reverse reserved fees; transfers skip held/canceled payouts and detect account swaps to cancel instead. Pending/counting and pickup logic clarified to exclude held from Stripe pickup.

* **Tests**
  - Expanded unit/integration coverage for held-payout lifecycle, transfers, bulk cancel/release, and fee reversals.

* **Chores**
  - Deploy-time backfill script to enable payout capability for REVIEW/SNOOZED orgs.

PR #12957, opened 2026-07-06
fix(transaction): count payout reversals as available balance

maximevast:
## Summary

**Related Issue**: #12103

Canceling a pending or held payout returned the reserved gross via a payout_reversal row dated now, which then re-aged behind the full payout_transaction_delay instead of becoming available again — so the funds fell under held balance and couldn't back a new payout, even though they'd already cleared the delay before being reserved. The available-balance filter exempts payout-side rows from aging (the payout debit and its fees always count immediately) but was missing the mirror branch for the payout_reversal credit, so the debit landed instantly while its reversal didn't. 

Fix adds payout_reversal alongside payout in the filter, restoring the symmetry. This gap dates to the delay filter (#11494) never being reconciled with the reversal row from the cancel path; #12097 documented the re-aging as matching pending-cancel but that behavior was itself the same unclosed gap. 
Return-of-funds is unchanged (total balance was always restored); only the availability classification is corrected, for both cancel paths. Tests exercise the real transaction/fee services: held and pending cancel now leave the full gross+fees available, plus a test that a second payout can be created from the released funds (previously raised InsufficientBalance).

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/12957?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/transaction/service/transaction.py
+++ b/server/polar/transaction/service/transaction.py
@@ -181,7 +181,12 @@ class TransactionService(BaseTransactionService):
                     func.coalesce(
                         func.sum(Transaction.amount).filter(
                             or_(
-                                Transaction.type == TransactionType.payout,
+                                Transaction.type.in_(
+                                    (
+                                        TransactionType.payout,
+                                        TransactionType.payout_reversal,
+                                    )
+                                ),
                                 Transaction.platform_fee_type.in_(
                                     PlatformFeeType.payout_fee_types()
                                 ),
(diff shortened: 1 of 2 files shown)


frankie567 (approved):

pieterbeulque:
Fantastic!

PR #12961, opened 2026-07-06
files: Prevent too many parts and too big bodies

Yopi:
By uploading a large number of parts, we parse that into Python objects which are held in memory. S3 limits the number of parts to 10k, which prevents this issue.

<a href="https://cubic.dev/pr/polarsource/polar/pull/12961?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/integrations/aws/s3/schemas.py
+++ b/server/polar/integrations/aws/s3/schemas.py
@@ -3,7 +3,7 @@ import hashlib
 from datetime import datetime
 from typing import Any, Self
 
-from pydantic import UUID4, computed_field
+from pydantic import UUID4, Field, computed_field
 
 from polar.kit.schemas import IDSchema, Schema
 from polar.kit.utils import human_readable_size
@@ -29,7 +29,7 @@ class S3FileCreatePart(Schema):
 
 
 class S3FileCreateMultipart(Schema):
-    parts: list[S3FileCreatePart]
+    parts: list[S3FileCreatePart] = Field(max_length=10_000)
 
 
 class S3FileCreate(Schema):
--- a/server/polar/middlewares.py
+++ b/server/polar/middlewares.py
@@ -1,5 +1,6 @@
 import contextlib
 import functools
+import json
 import re
 
 import dramatiq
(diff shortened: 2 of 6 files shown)


server/polar/integrations/aws/s3/schemas.py
```diff
@@ -29,7 +29,7 @@ def get_boto3_arguments(self) -> dict[str, Any]:
 
 
 class S3FileCreateMultipart(Schema):
-    parts: list[S3FileCreatePart]
```
psincraian, line 32: Does the frontend split the request?

Yopi: Yes it does 😄

Yopi: https://github.com/polarsource/polar/blob/97e9430901a4197876318672cdd4fae99fc49a2b/clients/apps/web/src/components/FileUpload/Upload.ts#L85-L102

psincraian (approved):

server/polar/middlewares.py
```diff
@@ -146,6 +147,81 @@ async def send_wrapper(message: Message) -> None:
...
+                pass
+
+        received = 0
+        response_started = False
+
+        async def receive_wrapper() -> Message:
+            nonlocal received
+            message = await receive()
+            if message["type"] == "http.request":
+                received += len(message.get("body", b""))
```
frankie567: I'm not too comfortable with reading the body systematically in a middleware. Could we simply enforce to have `Content-Length` set and return 411 if not?

server/polar/middlewares.py
```diff
@@ -146,6 +147,81 @@ async def send_wrapper(message: Message) -> None:
...
+    async def _send_too_large(self, send: Send) -> None:
+        body = json.dumps(
+            {
+                "error": "RequestBodyTooLarge",
+                "detail": (
+                    "Request body exceeded the maximum allowed size "
+                    f"of {self.limit} bytes."
+                ),
+            }
+        ).encode("utf-8")
```
frankie567: We could probably cache this as an attribute in __init__

server/polar/middlewares.py
Yopi: I was a bit uncertain about doing this overall, I'm not entirely sure what ramifications it will have.

Let me see if we can verify if it's always set

server/polar/middlewares.py
Yopi: It's no longer possible to see what this refers to, but I think it's this:
```
        body = json.dumps({"error": error, "detail": detail}).encode("utf-8")
```

Running a test in a loop it looks like it's basically instant

frankie567 (approved):

PR #12973, opened 2026-07-06
feat(compass): live insights API over metrics

emilwidlund:
image

**This work is gated behind a feature flag**

## What

Adds **Compass**, a live-computed narration layer on top of the existing metrics service that surfaces prioritized, explained insights about an organization's business. This is the **backend-only** slice; frontend wiring follows in a separate PR.

## How

- **Signals layer** (`signals.py`) builds uniform `MetricSignal` inputs from the metrics service over a window — swappable for dedicated Tinybird pipes later without touching detectors.
- **8 rule-based detectors** (`detectors/`), each pure and deterministic (signals in → `Insight` out):
  - `mrr`, `subscribers`, `arpu` — month-over-month movement (revenue / growth)
  - `churn` — trailing-30d cancellation rate with a critical band (retention)
  - `trials` — trialing-MRR pipeline share (growth / opportunity)
  - `margin`, `cost_per_user` — cost-adjusted unit economics (cost); self-gate to silence when Tinybird cost/revenue data is absent
  - `conversion` — checkout conversion, 30d vs prior 30d (product)
- **`GET /v1/compass/insights`** — reuses the `metrics_read` scope, ordered by severity → priority → confidence. **Fully stateless** (no persistence).
- **Per-organization gating** behind the existing `compass_enabled` feature flag (`Organization.is_compass_enabled`); disabled orgs get an empty feed.
- **Seeding + diagnostics** — backdated seed cohorts (active/churn/trial + checkout history) that trip the SQL-backed detectors locally, plus a `scripts/compass_diag.py` to run every detector against live metrics per org.

## Testing

- 40 unit tests for the detectors + key/confidence helpers (`tests/compass/test_detectors.py`).
- `uv run task lint`, `uv run task lint_types`, and the compass suite all green.
- Verified end-to-end against seeded data: enabled org → insights returned; `compass_enabled=false` → empty feed.

## Notes

- Cost/margin detectors depend on Tinybird-backed metrics that read 0 in local dev; they self-gate and only surface where that data exists.
- No new dependencies; no DB migrations (stateless).

<a href="https://cubic.dev/pr/polarsource/polar/pull/12973?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/polar/compass/signals.py
```diff
@@ -0,0 +1,70 @@
...
+    return values[index]
+
+
+def sum_last_n_periods(response: MetricsResponse, slug: str, n: int) -> float:
+    """Total of a metric over the most recent `n` periods (a rolling window)."""
+    return sum(series(response, slug)[-n:])
+
+
+def format_currency(cents: float) -> str:
+    return f"${cents / 100:,.0f}"
```
Yopi, line 51: I think we have this already, right?

server/polar/compass/signals.py
```diff
@@ -0,0 +1,70 @@
...
+    values = series(response, slug)
+    index = len(values) - 1 - n
+    if index < 0:
+        return None
+    return values[index]
+
+
+def sum_last_n_periods(response: MetricsResponse, slug: str, n: int) -> float:
+    """Total of a metric over the most recent `n` periods (a rolling window)."""
+    return sum(series(response, slug)[-n:])
```
Yopi: It might be neat to let these live with Metrics instead of here, to compartmentalise if they are valuable there as well as here

server/polar/compass/signals.py
```diff
@@ -0,0 +1,70 @@
...
+@dataclass(frozen=True)
+class MetricSignal:
+    """A metric's current value against a baseline, plus a sample size."""
+
+    slug: str
+    current: float
+    baseline: float
+    sample_n: int
+    """Population behind the change (e.g. active subscriptions), for confidence."""
+    drivers_dimension: str | None = None
```
Yopi: Might be an enum

server/polar/compass/signals.py
```diff
@@ -0,0 +1,70 @@
+"""
+Signal layer: turns the metrics API into the uniform inputs detectors reason over.
+
+In the target architecture these come straight from parameterized Tinybird pipes
+that return value + baseline + drivers in one fast columnar query. For the scaffold
+we compute them by reading the existing `metrics` service (which already merges
+Postgres + Tinybird, and handles auth, filtering and caching) over a window, then
+diffing periods in Python. Detectors depend only on `MetricSignal`, so swapping in
+dedicated pipes later is a signals-layer change, not a detector change.
```
Yopi: I'm not sure if I follow this comment. Are you saying that in the future it won't be based on metrics?

server/polar/compass/service.py
```diff
@@ -0,0 +1,162 @@
...
+
+        today = (now or datetime.now(tz=timezone)).astimezone(timezone).date()
+        # One window serves every detector: the union of their slugs over the
+        # longest lookback, fetched once per organization.
+        slugs = sorted({slug for d in detectors for slug in d.metric_slugs})
+        days = max(d.lookback_days for d in detectors) + _LOOKBACK_HEADROOM_DAYS
+
+        # Organizations are processed serially on purpose: the session (and its
+        # `SET LOCAL` connection state inside the metrics service) must not be
+        # used concurrently.
```
Yopi, line 80: This sounds like something we might want to look into

server/polar/compass/keys.py
```diff
@@ -0,0 +1,49 @@
+"""
+The deterministic insight key: `detector_id:organization_id:period_bucket`.
+
+Insights are computed live and never persisted, so this key is the only stable
```
Yopi, line 4: Going forward, we'll probably want to persist them. But this makes sense during the testing period to not have to deal with db migrations back and forth

Yopi (approved):
Very cool! I think this is good enough to start testing things out.

I was thinking a little bit about the detectors, if it would be possible to tighten up their formulation and simplify their definitions somehow. I don't want to say DSL, but it feels like there is some logic in there which makes them difficult to parse as a human. This can also come later though

server/polar/compass/signals.py
emilwidlund: The polarkit one preserves cents - this one drops it for better presentation. But let me see what I can do & if I can reuse.

frankie567:
@stilla-bot Can you review this code against our ADR in handbook/engineering/decisions?

frankie567:
@stilla-bot Can you make a comment with a code commit suggestion to fix the ADR-0007 violation?

frankie567 (approved):
As discussed, let's rollout this while it's experimental. I would recommend to apply Stilla suggestions (see above) so we're at least consistent with our rulings.

PR #12977, opened 2026-07-06
fix(dashboard): invoice download getting stuck and overview widgets reappearing

psincraian:
## Summary

Fixes two small but persistent dashboard bugs.

- **Invoice download** would get stuck on the loading spinner: the invoice was marked as generated but the download never opened, forcing a page refresh and a re-generate before it worked.
- **Overview customization** didn't stick: removing widgets from the overview and refreshing brought them all back.

## What

- Rework the invoice generate → wait-for-event → download flow so the button never hangs: wait for the generation event only when the backend will actually regenerate (invoice missing or a billing field changed), otherwise download the existing document directly. On timeout, reset loading and show a retry message instead of spinning forever, and download via a user gesture. A refetch-only event listener keeps state fresh when generation finishes late or is triggered elsewhere.
- Restore any non-empty saved overview metric selection instead of only a selection of exactly five items, so removed widgets stay removed after a refresh (falling back to defaults only when nothing valid is stored).

## Why

Both were user-facing paper cuts: the stuck invoice download required a manual refresh to recover, and the overview reset silently discarded any customization that wasn't exactly the default count.

## Checklist

- [x] Linting and type checking pass (web lint + type check on the changed files)

> Note: bundles two related dashboard bug fixes (invoice download + overview widgets), so the single-concern box is intentionally left unchecked.

### DIFF
--- a/clients/apps/web/src/components/DashboardOverview/OverviewSection.tsx
+++ b/clients/apps/web/src/components/DashboardOverview/OverviewSection.tsx
@@ -31,13 +31,11 @@ export function OverviewSection({ organization }: OverviewSectionProps) {
   const { isShown, show, hide } = useMetricSelectorModal()
 
   const initialMetrics = React.useMemo<(keyof schemas['Metrics'])[]>(() => {
-    const stored = organization.feature_settings?.overview_metrics
-    if (stored?.length === 5) {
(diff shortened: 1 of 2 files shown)


maximevast (approved):
LGTM!

PR #12980, opened 2026-07-06
2fa: Require code to be able to disable 2fa or regenerate backup codes

Yopi:
Otherwise an attacker who is logged in might delete the already configured 2fa or regenerate the backup codes, loosening up the benefits of the system

---
## Summary by cubic
Require a 2FA code to disable/delete TOTP and to generate/regenerate backup codes, with a consistent OTP input in Settings. Enabling TOTP now returns backup codes immediately so users can save them.

- **New Features**
  - Require a valid TOTP or backup code to disable TOTP, and to generate/regenerate backup codes when TOTP is enabled.
  - Add a reusable TwoFactorCode modal with OTP slots and a backup code toggle, used for disable and backup code actions.
  - TOTP enable returns newly generated backup codes; UI surfaces them.

- **Migration**
  - API changes:
    - DELETE /v1/auth/totp now accepts { code } in the JSON body and returns 204; returns 403 on invalid/missing code and 404 if not enrolled.
    - POST /v1/auth/backup-codes accepts optional { code }; when TOTP is enabled a code is required; returns 201 on success or 403 on invalid/missing code.
    - POST /v1/auth/totp/enable returns 200 with { codes } (backup codes) instead of 202; may return 403 (invalid code) or 409 (already enabled).
  - Update client usage:
    - `useTOTPDelete(code)` and `useBackupCodesEnroll(code?)` now take a code.
    - Handle new 403/409 responses and the backup codes payload from TOTP enable.

<sup>Written for commit 086a9b5e57748af2347b2f1f56d48761c17ef167. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/12980?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/apps/web/src/components/Settings/BackupCodesRegenerateModal.tsx
+++ b/clients/apps/web/src/components/Settings/BackupCodesRegenerateModal.tsx
@@ -1,11 +1,11 @@
 'use client'
 
-import { ConfirmModal } from '@/components/Modal/ConfirmModal'
+import TwoFactorCodeModal from './TwoFactorCodeModal'
 
 interface BackupCodesRegenerateModalProps {
   isShown: boolean
   hide: () => void
-  onRegenerate: () => Promise<{ codes: string[] } | null>
+  onRegenerate: (code: string) => Promise<{ error?: { detail?: unknown } }>
 }
 
 export default function BackupCodesRegenerateModal({
@@ -13,20 +13,15 @@ export default function BackupCodesRegenerateModal({
   hide,
   onRegenerate,
 }: BackupCodesRegenerateModalProps) {
-  const handleConfirm = async () => {
-    await onRegenerate()
-    hide()
-  }
-
   return (
-    <ConfirmModal
+    <TwoFactorCodeModal
       isShown={isShown}
       hide={hide}
       title="Regenerate backup codes?"
       description="Your current backup codes will stop working immediately and be replaced with new backup codes."
       destructive
-      destructiveText="Regenerate"
-      onConfirm={handleConfirm}
+      confirmLabel="Regenerate"
+      onConfirm={onRegenerate}
     />
   )
 }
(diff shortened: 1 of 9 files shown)


frankie567 (commented):
I'm not very satisfied with this approach: it gives too much responsibility to the TOTP endpoints.

I think we should rather explore an approach where we know the "strength" of the UserSession. Basically, have a "freshness" threshold (e.g. <5 minutes) where those operations are allowed; otherwise prompt the user to "refresh" their session through their standard login methods. Effectively, TOTP endpoints would only care about validating they have a fresh session.

I recognize this is significantly more work; so happy to tackle this myself in the coming days/weeks.

Yopi:
> I'm not very satisfied with this approach: it gives too much responsibility to the TOTP endpoints.
> 
> I think we should rather explore an approach where we know the "strength" of the UserSession. Basically, have a "freshness" threshold (e.g. <5 minutes) where those operations are allowed; otherwise prompt the user to "refresh" their session through their standard login methods. Effectively, TOTP endpoints would only care about validating they have a fresh session.
> 
> I recognize this is significantly more work; so happy to tackle this myself in the coming days/weeks.

That makes sense, like GitHub does it then? Ideally we would allow you then to specify if an endpoint needs a fresh session or not per endpoint

Yopi:
I agree with you, but I would meekly argue that for these endpoints in particular it's not too much responsibility for them to cover themselves. I.e. they deal with TOTP and thus they also require a TOTP to make an action.

PR #12983, opened 2026-07-06
feat(merchant-migration): persist canonical records to the ledger

psincraian:
## Summary

Persist normalized canonical records to the `merchant_migration_records` during precheck, so we don't check again and again.

## What

- Add `serialize()` for canonical dataclasses, producing a JSON-safe dict via `jsonable_encoder` (datetimes to ISO strings, enums to values) for the `canonical` JSONB column.
- Add `MerchantMigrationRecordRepository` with an idempotent `upsert`, keyed per organization by `(type, source_id)`.
- `run_precheck` now stages each record as it streams to the precheck engine, in the same single incremental pass.

## How

Re-running a migration refreshes a `pending` snapshot and re-points it at the current migration, while leaving `imported`/`skipped`/`failed` records untouched, so a prior run's results are never re-imported.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/12983?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/merchant_migration/canonical.py
+++ b/server/polar/merchant_migration/canonical.py
@@ -137,49 +137,51 @@ def deserialize(
     """Rebuild a canonical record from a stored ``canonical`` dict, the inverse
     of ``serialize``. Lets consumers work off the staged ledger instead of
     re-reading the source."""
-    if type == MerchantMigrationRecordType.product:
-        return CanonicalProduct(
-            source_id=data["source_id"],
-            product_source_id=data["product_source_id"],
-            name=data["name"],
-            recurring_interval=data["recurring_interval"],
-            recurring_interval_count=data["recurring_interval_count"],
-            prices=[
-                CanonicalPrice(
-                    source_id=price["source_id"],
-                    currency=price["currency"],
-                    amount=price["amount"],
-                    pricing_scheme=CanonicalPricingScheme(price["pricing_scheme"]),
-                )
-                for price in data["prices"]
-            ],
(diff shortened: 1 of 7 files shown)


server/polar/merchant_migration/canonical.py
```diff
@@ -116,3 +119,67 @@ class CanonicalAccount:
...
+    return datetime.fromisoformat(value) if value else None
+
+
+def deserialize(
+    type: MerchantMigrationRecordType, data: dict[str, Any]
+) -> CanonicalRecord:
+    """Rebuild a canonical record from a stored ``canonical`` dict, the inverse
+    of ``serialize``. Lets consumers work off the staged ledger instead of
+    re-reading the source."""
+    if type == MerchantMigrationRecordType.product:
```
Yopi: Would it be neater to do this as a match/case instead of if?

Yopi (approved):
I'm not sure if I have the full context here, but looks good to me!

server/polar/merchant_migration/canonical.py
psincraian: Good call

PR #12989, opened 2026-07-07
lint: Merge server linters + add frontend_url linter

Yopi:
This merges all server linters into a unified runner that allows us to only parse the AST once and apply all rules after that.

It will still allow you to run a single linter (similar as the previous behaviour). 

This also adds a new linter for checking the use of request derived values in generate_frontend_url to prevent open redirect issues.

<a href="https://cubic.dev/pr/polarsource/polar/pull/12989?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/authz/repository.py
+++ b/server/polar/authz/repository.py
@@ -28,7 +28,7 @@ def select_user_org_ids(
     stmt = (
         # The one blessed place that expands user memberships into org ids;
         # every other caller must go through this helper.
-        select(UserOrganization.organization_id)  # noqa: org-scope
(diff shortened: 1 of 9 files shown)


maximevast (approved):
LGTM but maybe worth another pair of eyes?

frankie567 (approved):

PR #12994, opened 2026-07-07
feat: merchant migration pre-check panel + records drawer

psincraian:
## Summary

Adds the **pre-check** section to merchant migrations.

## What

**Backend**
- `PrecheckReport` gains an `entities` summary (per-entity `total` / `importable` / `skipped`), computed from the same classification the drawer shows, so counts always match.
- New `GET /v1/merchant-migrations/{id}/records`: reads the connected source on demand, classifies each record importable/skipped by **reusing the pre-check drop predicates**, filters by `entity` + optional `status`, and paginates.
- `classify_records` / `summarize_records` live alongside the pre-check engine so a record's status can't drift from the report's warnings.

**Frontend**
- `PrecheckPanel` renders on the pre_check timeline step: run/re-run pre-check, ready-vs-blockers status, blocker list, and the summary grid.
- `PrecheckSummary` grid → `PrecheckRecordsDrawer`: paginated, filterable (All / Importing / Stay on Stripe) list per entity.

## How

- Records are classified live from the source (not persisted); the endpoint uses a write session because building the adapter rotates the Stripe token.
- The summary is the classification tallied per entity, guaranteeing the grid and the drawer agree.
- UI authored with the Orbit `<Box />` design system.

## Checklist

- [x] New functionality is covered by tests (classify/summarize unit tests, `list_records` service tests, records endpoint test)
- [x] Linting and type checking pass (`uv run task lint && uv run task lint_types`, `pnpm lint`)

<a href="https://cubic.dev/pr/polarsource/polar/pull/12994?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/merchant_migration/precheck.py
+++ b/server/polar/merchant_migration/precheck.py
@@ -1,12 +1,13 @@
-"""Streams CanonicalRecords into a PrecheckReport of blockers and warnings
-(design Appendices A and D).
+"""Turns CanonicalRecords into a PrecheckReport of blockers, warnings and
+per-entity import counts (design Appendices A and D), and classifies individual
+records for the review drawer.
 
-Consumes the records once and keeps only the aggregates the cross-record checks
-need, so it never holds the whole source catalog in memory.
+The per-record classification reuses the same `_check_*` predicates as the
+report, so a record's importable/skipped status always matches the summary.
 """
 
 from collections import Counter
-from collections.abc import AsyncIterable, Iterable
+from collections.abc import AsyncIterable, Iterable, Sequence
 
 from polar.enums import SubscriptionRecurringInterval
 from polar.kit.currency import (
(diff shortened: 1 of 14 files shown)


server/polar/merchant_migration/precheck.py
```diff
@@ -324,3 +360,212 @@ def _check_subscription(
...
+def _interval_label(product: CanonicalProduct) -> str:
+    if product.recurring_interval is None:
+        return "One-time"
+    count = product.recurring_interval_count
+    if count == 1:
+        return f"Every {product.recurring_interval}"
+    return f"Every {count} {product.recurring_interval}s"
+
+
+def _amount_label(amount: int | None, currency: str) -> str:
```
frankie567: I think we have utilities for that already — that are able to handle zero decimal currencies and such

psincraian: fixed!

frankie567 (approved):

PR #13000, opened 2026-07-07
feat(compass): typed insight actions, product margin + cost concentration detectors

emilwidlund:
First of a 3-PR stack building toward the Compass assistant. This layer is useful standalone: richer detectors and a typed action contract for insights.

## What's in here

**Typed insight actions.** `Insight.primary_action` becomes a discriminated union (`view_metric` | `adjust_price`) instead of a single shape, so the frontend resolves each action type to its own destination and new action types are a schema addition, not a rework.

**ProductMarginDetector.** Per-product margin decline with a concrete price recommendation derived from cost to serve (`adjust_price` action carrying current and suggested price). Product costs are read through the product's active-customer cohort, since cost events attach to customers, not products.

**CostConcentrationDetector** (fills the `risk` category). Fires when one customer exceeds 40% of tracked costs (critical at 60%), fed by a per-customer cost prefetch from the events by-customer statistics. Confidence-gated on at least 5 customers carrying costs, so small orgs aren't told their costs are "concentrated".

**Signal layer + prefetches.** `ProductPricing` and `CustomerCostSignal` dataclasses keep detectors pure (data in, insight out); the service prefetches them only when a selected detector declares the need. Repository helpers for product cohorts and org products with prices.

## Testing

- 57 detector tests, including firing/silent/sample-gate cases for both new detectors and the em-dash copy guard across every registered detector.
- lint + mypy clean.

## Stack

1. **→ this PR** — detectors + typed actions
2. `compass-assistant-backend-2-core` — the assistant core (agent, blocks, SSE endpoint)
3. `compass-assistant-backend-3-tools` — the tool fleet + self-serve inference cost tracking

<a href="https://cubic.dev/pr/polarsource/polar/pull/13000?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/compass/detectors/product_margin.py
+++ b/server/polar/compass/detectors/product_margin.py
@@ -1,6 +1,6 @@
 import math
 
-from polar.kit.currency import get_currency_decimal_factor
+from polar.kit.currency import format_currency, get_currency_decimal_factor
 from polar.metrics.aggregation import latest
 
 from ..schemas import (
@@ -9,7 +9,7 @@ from ..schemas import (
     InsightCategory,
     InsightSeverity,
 )
-from ..signals import ProductPricing, format_currency, format_pct
+from ..signals import ProductPricing, format_pct
 from .base import Detector, DetectorContext, confidence_for_sample
 
 # Below this a product's margin is thin enough to flag; below the critical bar
@@ -95,13 +95,15 @@ class ProductMarginDetector(Detector):
         title = f"{worst.name} margin is down to {margin_str}"
         body = (
             f"{worst.name} keeps {margin_str} of its "
-            f"{format_currency(worst.price_amount)} price after costs. "
-            f"Roughly {format_currency(cost_to_serve)} per customer goes to "
-            f"serving it, across {worst_subs} active subscriptions."
+            f"{format_currency(worst.price_amount, worst.currency)} price "
+            f"after costs. Roughly "
+            f"{format_currency(cost_to_serve, worst.currency)} per customer "
+            f"goes to serving it, across {worst_subs} active subscriptions."
         )
         if suggested is not None:
             body += (
-                f" A list price around {format_currency(suggested)} would "
+                f" A list price around "
+                f"{format_currency(suggested, worst.currency)} would "
                 f"restore a {format_pct(_TARGET_MARGIN)} margin for new "
                 "customers."
             )
--- a/server/polar/compass/detectors/product_margin.py
+++ b/server/polar/compass/detectors/product_margin.py
@@ -1,0 +1,136 @@
+import math
+
+from polar.metrics.aggregation import latest
+
+from ..schemas import (
+    AdjustPriceAction,
+    Insight,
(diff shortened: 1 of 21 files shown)


server/polar/compass/service.py
```diff
@@ -23,13 +32,30 @@
...
+# pipe has no product filter), so per-product cost reads go through the
+# product's active-customer cohort as a `customer_id` filter. The filter is a
+# GET query param, which bounds how many customer UUIDs fit in the URL —
+# products with a larger cohort are skipped (their margin needs a dedicated
+# product-attributed pipe) rather than silently sampled.
+_MAX_COHORT_CUSTOMERS = 150
+
+# Per-customer cost ranking window and depth for concentration detectors.
+_CUSTOMER_COSTS_WINDOW_DAYS = 30
+_MAX_COST_CUSTOMERS = 10
```
psincraian: I think you try to get customer cost concentration. But you also have confidence_for_sample but this never is reached as you have medium that needs 20 samples

server/polar/compass/detectors/product_margin.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def evaluate(self, ctx: DetectorContext) -> Insight | None:
+        worst: ProductPricing | None = None
+        worst_margin = _HEALTHY_MARGIN
+        worst_subs = 0
+        for product in ctx.products:
+            margin = latest(product.metrics, "gross_margin_percentage")
+            # 0 margin means no cost/revenue data for this product, not a
+            # zero-margin business — stay silent rather than mislead.
+            if margin <= 0 or margin >= worst_margin:
```
psincraian, line 82: what is the rationale to exclude the negative products from here?

emilwidlund: With the cohort-based cost attribution (a customer's full cost counts toward every product they hold), negative margins are usually artifacts: a cheap product whose customers also run expensive workloads reads deeply negative, and a price recommendation built on that would be wrong.

emilwidlund: The goal is to identify products that costs more to serve than you're currently making from revenue. But as costs aren't tied to products, we need a workaround. So we look at all customer costs in a product cohort and try to decide from there.

server/polar/compass/service.py
emilwidlund: Fixed in dca48b6bb3: the prefetch depth is now a shared CUSTOMER_COSTS_SAMPLE_LIMIT = 100 (sized to the highest confidence threshold), and the copy says "at least 100 customers" when the sample is capped

psincraian (approved):

frankie567 (approved):

PR #13017, opened 2026-07-07
feat(compass): scope-aware assistant core with generative UI blocks

emilwidlund:
The Compass assistant: an agentic chat endpoint over the merchant's own
organization, built on Pydantic AI + the Pydantic AI Gateway (the
organization_review pattern).

- SSE endpoint POST /v1/compass/assistant streaming text deltas, renderable
  block events and opaque conversation state; gated per organization by the
  compass_enabled feature flag before any model call.
- Scope safety: the agent runs under the caller's auth subject. Its toolset
  is derived from the token's granted scopes, each tool re-checks its scope
  at runtime, and the organization always comes from validated deps, never
  model arguments. Tool exceptions degrade to a polite message, never a
  traceback.
- Generative UI: tools return typed blocks from a closed discriminated
  union that the client maps to predefined components. The model places
  blocks under the claims they support via [block:N] markers parsed out of
  the token stream.
- Core tools: metrics charts, and insights with a fetch/render split so a
  broad question yields an assessment plus one or two supporting cards, not
  a card dump.

Stacked on compass-assistant-backend-1-actions. The wider tool fleet and
inference cost tracking follow in the next PR.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

<a href="https://cubic.dev/pr/polarsource/polar/pull/13017?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/compass/assistant/blocks.py
+++ b/server/polar/compass/assistant/blocks.py
@@ -78,9 +78,7 @@ class DataTableBlock(Schema):
 
     type: Literal[BlockType.data_table] = BlockType.data_table
     entity: str = Field(description="What the rows are, e.g. `subscriptions`.")
-    title: str | None = Field(
-        default=None, description="Heading rendered above the table."
-    )
+    title: str | None = Field(description="Heading rendered above the table.")
     columns: list[DataTableColumn]
     rows: list[dict[str, str | int | float | None]]
     total_count: int
@@ -92,9 +90,7 @@ class EntityListBlock(Schema):
 
     type: Literal[BlockType.entity_list] = BlockType.entity_list
     entity: str = Field(description="What the items are, e.g. `orders`.")
-    title: str | None = Field(
-        default=None, description="Heading rendered above the list."
-    )
+    title: str | None = Field(description="Heading rendered above the list.")
     columns: list[DataTableColumn]
     rows: list[dict[str, str | int | float | None]]
     total_count: int
@@ -105,8 +101,8 @@ class CustomerCardBlock(Schema):
 
     type: Literal[BlockType.customer_card] = BlockType.customer_card
     email: str
-    name: str | None = None
-    avatar_url: str | None = None
+    name: str | None
+    avatar_url: str | None
     created_at: datetime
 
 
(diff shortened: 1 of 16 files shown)


server/polar/compass/assistant/stream.py
```diff
@@ -0,0 +1,260 @@
...
+            result = run.result
+            assert result is not None
+            _track_usage(deps, result.usage, model_provider, model_name)
+            yield _event(
+                "done",
+                {
+                    "message_history": seal_history(
+                        deps,
+                        ModelMessagesTypeAdapter.dump_json(
+                            result.all_messages()
```
psincraian, line 252: what I'm worried here, is that we send all the history to the merchant. I think for now this is safe as we don't expose private information that the merchant don't have, but don't scale well 👀

Maybe some ideas:
- remove the tool result calls (not sure if we can really do that) 
- encrypt it

server/polar/compass/assistant/blocks.py
```diff
@@ -0,0 +1,117 @@
...
+from typing import Annotated, Literal
+
+from pydantic import Discriminator, Field
+
+from polar.kit.schemas import Schema
+
+from ..schemas import Insight
+
+
+class BlockType(StrEnum):
```
psincraian, line 22: love this

psincraian (approved):
We will fix the refactoring the exposition of  messages to a follow up PR

server/polar/compass/assistant/stream.py
emilwidlund: Claude's 2 cents on it:

```
The durable answer is the one both ideas are approximating: stop round-tripping state entirely — store it
  server-side. Redis keyed by a conversation id (we already have redis on AssistantDeps), value = messages JSON bound
  to org + scopes digest, TTL ~1 hour, client round-trips only the opaque id. That solves scale (constant payload),
  privacy (nothing leaves the server, encryption question evaporates), and replay (binding enforced server-side) in
  one move, and it deletes the HMAC seal rather than layering on it. The statelessness of the current design was a
  prototype virtue; a TTL'd Redis key barely compromises it.
```

PR #13023, opened 2026-07-07
feat(stripe): integrate account risk signals into org review

psincraian:
## Summary

Stripe can warn us when a connected account looks risky. Two warnings are live: a fraudulent website and a fraudulent merchant. When Stripe sends one, we store it and, if it is serious, flag the seller for a human.

This is the first step toward a shared "risk signals" inbox. Later our own review agent will add signals to the same table, and humans will triage them.

## What

- New table `organization_risk_signals` (source, type, risk level, reason, raw payload). One row per warning. We store only serious ones (elevated/highest).
- A signed webhook `POST /integrations/stripe/webhook-account-risk` with its own secret. These are "thin" events, so the background job fetches the full event by id to read the details.
- Handles both live signals, which have different payload shapes (website is flat; merchant is nested).
- When an org enters review, we ask Stripe to check its website. The result comes back as one of these webhooks.
- For a serious signal on a live (active) org, we pull it into review so payouts are held and a human looks.

## Why

Surface Stripe's fraud warnings into our review flow, so a risky seller is held for a human instead of paying out.

## Not on yet

The feature is off until we set `POLAR_STRIPE_ACCOUNT_RISK_WEBHOOK_SECRET` and connect the webhook on Stripe's side. Until then the URL returns 404 and we never call Stripe. Safe to merge dark.

Notes: delinquency is not enabled yet (Stripe is still setting it up). The event names and shapes were confirmed against a live sandbox. Terraform prod secret wiring is deferred.

## Checklist


server/polar/integrations/stripe/endpoints.py
```diff
@@ -47,9 +49,11 @@
...
 
 
-async def enqueue(session: AsyncSession, event: stripe.Event) -> None:
-    event_type: str = event["type"]
-    task_name = f"stripe.webhook.{event_type}"
+async def enqueue(
+    session: AsyncSession, event: stripe.Event, task_name: str | None = None
+) -> None:
+    # Default: one actor per event type. Callers with a dedicated actor override.
+    task_name = task_name or f"stripe.webhook.{event['type']}"
```
frankie567, line 56: Might not be worth to add a specific parameter just for that new endpoint, we can just directly call await `external_event_service.enqueue(...)` over there

server/polar/integrations/stripe/endpoints.py
```diff
@@ -100,3 +111,18 @@ async def webhook_connect(
 ) -> None:
     if event["type"] in CONNECT_IMPLEMENTED_WEBHOOKS:
         return await enqueue(session, event)
+
+
+@router.post(
+    "/webhook-account-risk",
```
frankie567, line 117: So, those are defined as separate webhooks on Stripe's side? Or is it that you wanted to specifically split them?

psincraian:
Splitting this into two smaller, reviewable PRs:
- #13218 — model + migration (isolated so the Migration Isolation Check passes)
- #13219 — the integration code (stacked on #13218)

Closing this one in favor of those.

PR #11490, opened 2026-05-07
Preserve license key status on benefit grant updates

pieterbeulque:
## Summary

**Related Issue**: 

Fix license key status being reset during benefit grant updates. The `customer_update_grant` method should preserve the current status (disabled/revoked) instead of overwriting it when updating other properties like activation limits.

## What

Removed `"status"` from the list of attributes updated in `customer_update_grant` method in `license_key/service.py`. This prevents the status field from being reset during benefit updates.

Added comprehensive test coverage (`TestCustomerUpdateGrant.test_preserves_status_on_benefit_update`) that verifies license key status is preserved when updating benefit properties, covering both `disabled` and `revoked` statuses.

## Why

When a benefit's properties are updated (e.g., activation limits), the associated license key's status should not be modified. Previously, the status was being included in the update attributes, causing it to be reset even when it had been explicitly set to `disabled` or `revoked`.

## How

- Removed `"status"` from the `update_attrs` list in `customer_update_grant`
- Added parameterized test that:
  1. Creates a benefit and grant with a license key
  2. Sets the license key status to `disabled` or `revoked`
  3. Updates the benefit's activation properties
  4. Calls `customer_grant` to update the grant
  5. Verifies the status is preserved while other properties are updated

## Checklist

- [x] This PR addresses a single concern (preserve status on benefit updates)
- [x] The diff is reasonably sized and easy to review
- [x] New functionality is covered by tests (parameterized test added)
- [ ] Linting and type checking pass (to be verified)
- [x] No unrelated changes or drive-by fixes are included

https://claude.ai/code/session_01VWFUsyjchHb1CtPX7CxSYh

PR #13025, opened 2026-07-07
fix: preserve license keys across revoke → re-grant

maximevast:
## Problem
Benefit grants are reused across revoke → re-grant (same `BenefitGrant` row, `properties` preserved) — for every benefit type *except* license keys. The license-key strategy only reused the stored `license_key_id` when `grant(update=True)`, but a status-driven re-grant calls `grant(update=False)`, so it minted a brand-new key and left the old one `revoked`. Any flow that revokes then re-grants a license key silently rotated the customer's key.

## Fix
Three coordinated changes so a status-driven re-grant reuses the existing key and un-revokes it:
- `LicenseKey.mark_granted()` — status helper mirroring `mark_revoked()`.
- `customer_grant`/`customer_update_grant` gain an opt-in `regrant` flag that flips `revoked → granted` on the reused key. Off by default, so admin benefit-property edits keep preserving status (honors #11490, which deliberately stopped resetting key status on benefit updates). `disabled` is never resurrected — only `revoked`.
- The license-key strategy reuses `license_key_id` whenever present (not only on `update=True`) and passes `regrant=not update`, so re-grants reuse the key while admin edits don't.

## Blast radius (intended)
The fix lives at the shared benefit-grant choke point, so it applies to every status-driven license-key re-grant, not just one caller:
- Pause → resume (the transparency requirement driving this).
- Dunning recovery (`past_due → active`) — previously rotated the key on every recovery; now fixed.
- Upgrade/downgrade reusing the same license benefit.
This is deliberate: the goal is to make license keys behave like every other benefit. Admin-`disabled` keys stay disabled through all of it.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/13025?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/tests/subscription/test_service.py
```diff
@@ -5282,6 +5282,39 @@ async def test_mark_past_due_skips_ended_subscription(
         enqueue_job_mock.assert_not_called()
 
 
+@pytest.mark.asyncio
+class TestMarkActive:
+    """Test subscription service dunning recovery functionality"""
```
frankie567: Unclear comment in the context

frankie567 (approved):
Great, that was an important gap to fix. Nitpick on a test's comment

PR #13057, opened 2026-07-08
sqs: Handle local credentials

Yopi:
Now that we are relying on no credentials (role via OIDC), then the local tests break since we need the credentials specified there.

---
## Summary by cubic
Use static AWS credentials for SQS in development/testing and when `SQS_ENDPOINT_URL` is set, while keeping the default OIDC/no-credentials path in production. Fixes broken local tests and local `boto3` usage.

- **Bug Fixes**
  - Updated `get_sqs_client()` to use `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` when no worker creds are set and env is dev/test or an endpoint URL is provided; explicit `WORKER_SQS_*` creds still take precedence.
  - Added tests for dev, test, prod, endpoint override, and credential precedence.

<sup>Written for commit bf48c713886e7fe291d78c206e1a7a0466a86d8a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/13057?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/worker/_sqs.py
+++ b/server/polar/worker/_sqs.py
@@ -41,7 +41,11 @@ class SQSSendError(Exception):
 def get_sqs_client() -> "SQSClient":
     access_key_id = settings.WORKER_SQS_AWS_ACCESS_KEY_ID
     secret_access_key = settings.WORKER_SQS_AWS_SECRET_ACCESS_KEY
-    if access_key_id is None and settings.SQS_ENDPOINT_URL is not None:
+    if access_key_id is None and (
+        settings.SQS_ENDPOINT_URL is not None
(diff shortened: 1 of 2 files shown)


psincraian (approved):
True!

PR #13059, opened 2026-07-08
feat(oauth2): dual-write hashed and encrypted OAuth2Client secrets

psincraian:
## Summary

Encrypt `OAuth2Client.client_secret` and `registration_access_token` at rest. This is the additive first phase: the app dual-writes new columns and a backfill fills existing rows, with no behavior change yet.

## What

- Add four nullable columns: a deterministic HMAC (`*_hash`) and an envelope-encrypted copy (`*_encrypted`) for each of the two secrets, plus a migration and a batched backfill script.
- Dual-write on the write paths: registration writes the hash synchronously (authlib's sync flow can't make the async KMS call); rotation and the MCP client write both copies.
- Reads, client authentication, and value lookups are unchanged and still use the plaintext columns.

## Why

Stored in plaintext, these secrets are exposed by any database-level leak. Unlike the other encrypted secrets, they are compared synchronously and looked up by value, so they need a deterministic hash for verification and an encrypted copy to stay revealable.

## How

The hybrid follows the pattern already used for `OAuthAccount` and `SlackApp`, with the hash borrowed from how `OAuth2Token` hashes its tokens. Rollout is per column (dual-write → backfill → cut over → drop plaintext); this PR ships only dual-write and backfill.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/13059?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/oauth2/authorization_server.py
+++ b/server/polar/oauth2/authorization_server.py
@@ -114,8 +114,11 @@ class ClientRegistrationEndpoint(_ClientRegistrationEndpoint):
 
         if request.user is not None:
             oauth2_client.user_id = request.user.id
-        oauth2_client.registration_access_token = generate_token(
-            prefix=CLIENT_REGISTRATION_TOKEN_PREFIX
+
+        # Sync: must run while we hold the plaintext, and authlib can't await.
+        oauth2_client.set_client_secret_sync(oauth2_client.client_secret)
+        oauth2_client.set_registration_access_token_sync(
+            generate_token(prefix=CLIENT_REGISTRATION_TOKEN_PREFIX)
         )
 
         self.server.session.add(oauth2_client)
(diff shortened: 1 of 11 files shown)


server/polar/oauth2/authorization_server.py
```diff
@@ -118,6 +118,15 @@ def save_client(
             prefix=CLIENT_REGISTRATION_TOKEN_PREFIX
         )
 
+        # Only the hashes can be set here: encryption is async and can't run in
+        # authlib's sync flow, so the backfill fills the encrypted columns.
```
frankie567: So, when will we set the encrypted version when creating a new client?

psincraian: sorry, missed to commit 🙈 the idea is to hash and encrypt them here

frankie567 (approved):

PR #13061, opened 2026-07-08
guard duty: Store results from Guard Duty scans on files

Yopi:
This just stores the maliciousness of the files, it doesn't do anything with them yet.

<a href="https://cubic.dev/pr/polarsource/polar/pull/13061?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/file/tasks.py
+++ b/server/polar/file/tasks.py
@@ -1,0 +1,44 @@
+from typing import Any
+
+import structlog
+
+from polar.config import settings
+from polar.worker import AsyncSessionMaker, TaskPriority, actor
+
+from .repository import FileRepository
+from .service import file as file_service
+
+log = structlog.get_logger()
+
+
+@actor(actor_name="file.guardduty_scan_result", priority=TaskPriority.MEDIUM)
+async def guardduty_scan_result(scan_result: dict[str, Any]) -> None:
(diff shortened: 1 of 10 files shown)


server/polar/file/tasks.py
```diff
@@ -0,0 +1,44 @@
...
+from polar.config import settings
+from polar.worker import AsyncSessionMaker, TaskPriority, actor
+
+from .repository import FileRepository
+from .service import file as file_service
+
+log = structlog.get_logger()
+
+
+@actor(actor_name="file.guardduty_scan_result", priority=TaskPriority.HIGH)
```
psincraian: I think high should be only for checkout things, this one fits in medium queue

psincraian (approved):

server/polar/file/tasks.py
Yopi: Yeah, you're probably right. It doesn't entirely matter because this is only called via lambda, but good nonetheless

PR #13062, opened 2026-07-08
files: Hide files that are listed as malicious

Yopi:
We still need to notify the creator somehow, but this hides the files from the API

<a href="https://cubic.dev/pr/polarsource/polar/pull/13062?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


psincraian (commented):
Should we mark it as deleted also?

Yopi:
> Should we mark it as deleted also?

Yeah I'm not sure here to be honest. I'd like something more than just hiding them, maybe not even hiding them at all. 

We shouldn't serve them to users, but I think we should list them and mark them as malicious in the GUI. So probably should just rethink this PR altogether tbh

frankie567 (approved):

PR #13076, opened 2026-07-09
fix(subscription): re-land auto-resume scheduler with descriptor-binding fix

maximevast:
## Summary

Re-lands #12997 (reverted in #13072) with the production bug fixed and a safety net so this class of failure can't fail silently again.

### The bug, in plain terms
The scheduler stashed a database column (`current_period_end` / `resumes_at`) on the job-store object and read it back via `self.trigger_column`. A SQLAlchemy column is a descriptor, so reading it off self made SQLAlchemy think the job store was a database row and crash (`'NoneType' … supports_population`). APScheduler caught that, logged one warning, and kept retrying — so the scheduler stayed "up" while quietly no longer scheduling any subscription renewals. The fix exposes the column through a `@property`, so the lookup happens on the model class (correct) instead of the instance.

### Green → red → green (commit by commit)
A regression test written against the working code (green), then the refactor reintroduced to reproduce the exact crash (red), then the property fix (green) — so the history proves the bug and the fix rather than just asserting them.

### Making the scheduler noisier
Job-store query failures are now captured to Sentry and re-raised, so a broken store pages us instead of disappearing into an APScheduler warning-and-retry loop.

## Checklist

### DIFF
--- a/server/polar/subscription/repository.py
+++ b/server/polar/subscription/repository.py
@@ -412,24 +412,6 @@ class SubscriptionRepository(
         )
         return await self.get_all(statement)
 
-    async def get_subscriptions_to_resume(
-        self,
-        now: datetime,
-        *,
-        options: Options = (),
-    ) -> Sequence[Subscription]:
-        """Find paused subscriptions whose scheduled resume time has passed."""
-        statement = (
-            self.get_base_statement()
-            .where(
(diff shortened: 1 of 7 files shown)


psincraian (approved):

frankie567 (approved):
Thanks 👍

PR #13077, opened 2026-07-09
Adds subscription retries rfc

maximevast:
## Summary

**Related Issue**: polarsource/feedback/issues/131

RFC for stopping subscription payment retries (dunning) when a `past_due` subscription is cancelled. 

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/13077?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


psincraian (approved):
This looks like a nice improvement 👏

stop-retries-on-cancel-plan.md
```diff
@@ -0,0 +1,145 @@
...
+today ("I cancelled, why was I charged?").
+
+## Resolved decisions
+
+1. **Warning signal source** — ✅ derived boolean field on the `Subscription` read schema
+   (Commit 2).
+2. **Warning surface** — ✅ shown in both cancel confirmations, driven by the derived field.
+   `{amount}` = outstanding order total. Strings go to
+   `clients/packages/i18n/src/locales/en.ts`. Proposed starting copy (see open question on tone):
+   - **Merchant** (cancel/revoke confirmation): *"This subscription has a failed payment from the
```
pieterbeulque, line 135: Merchant copy doesn't live in `i18n`

stop-retries-on-cancel-plan.md
```diff
@@ -0,0 +1,145 @@
...
+     current period. Cancelling now ends it immediately and stops retrying that payment — the
+     outstanding {amount} won't be collected."*
+   - **Customer portal** (cancel confirmation): *"Your latest payment didn't go through.
+     Cancelling now ends your subscription immediately and stops any further payment attempts —
+     you won't be charged for it."*
```
pieterbeulque:
```suggestion
   - **Customer portal** (cancel confirmation): *"Your latest payment didn't go through.
     Cancelling now ends your subscription immediately and stops any further payment attempts."*
```

stop-retries-on-cancel-plan.md
```diff
@@ -0,0 +1,145 @@
...
+all.
+
+## The rule (agreed in Slack)
+
+The deciding factor is the **benefit revocation grace period** — i.e. whether the customer still
+had access during the past-due window:
+
+| At cancel time | Behaviour |
+| --- | --- |
+| `past_due` **and grace inactive** (no grace configured, or grace expired → benefits already revoked) | **Cancel immediately + void the pending order → dunning stops.** The customer never got unpaid-for access, so nothing is owed. |
```
pieterbeulque: I'm unsure about the "grace period expired" portion of this logic.

Especially for a first version, I think we can make tighten things a bit more and only allow this new behavior if the grace period configured is 0, maybe? The customer never had access, so it feels fair to drop the payment.

When the grace period is more 0, it starts to feel like a grey zone, where a customer could use it for N more days with no intention to pay, then at N + 1 days, cancel it and never be charged. This is a potential abuse path that we don't have with grace period 0.

pieterbeulque: Not sure how to phrase this correctly but I'm not sure what happens with a failed payment / past due subscription when the grace period organization setting is changed while a subscription is already past due. Do we (re-)grant? Or is the revocation timestamp calculated at the moment of failed payment. Otherwise changing this setting can have some unintended side effects.

maximevast:
https://github.com/polarsource/polar/pull/13188

PR #13083, opened 2026-07-09
feat: merchant-migration review dashboard

psincraian:
## Summary

Adds the dashboard for the merchant-migration assessment step. Merchants see what
will move from Stripe, fix what needs fixing, and start the import from one table.

Table - ideally almost element here will be on Ready.

preview-screenshot

Modal - when there is an error we show more details on what happened

modal

## What

- A review table listing products, customers and subscriptions in one place.
  Columns are Name, Type, Import and Amount.
- Three filters: what needs the merchant's attention, what won't import, and
  everything.
- A row opens a side modal with the full reason and the record's details.
- Row selection, so the merchant can leave records behind. Rows that can't be
  imported show a disabled checkbox.
- The import button and its count sit in the toolbar.
- A stepper showing where the migration is: connect, assessment, card movement,
  cutover.

## Why

The pre-check sorts every record into one of three outcomes: it imports as it
is, it needs a fix first, or it can't move at all. The merchant needs to act on
the middle group and understand the last one, so the table is built around that
question.

## How

- `reviewStatus.ts` maps the server's `status`, `import_status` and
  `reason_level` to one label and one colour, shared by the table and the modal.
- `reviewSelection.ts` models selection the way the API expects it: pick rows, or
  take everything except some rows. Large catalogs never need every id on the
  client.
- `steps.ts` owns which backend steps each visible step covers, so the stepper,
  the card and the detail page agree.

## Checklist

- [x] Linting and type checking pass (`pnpm lint`, `tsc --noEmit`)


clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewTableView.tsx
```diff
@@ -0,0 +1,230 @@
...
+      onPageSizeChange(next.pageSize)
+      return
+    }
+    onPageChange(next.pageIndex + 1)
+  }
+
+  if (!hasCatalog) {
+    return (
+      <Text variant="caption" color="muted">
+        No records staged yet. Use &ldquo;Refresh from Stripe&rdquo; to scan
```
Comment (strandhvilliam) [inline clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewTableView.tsx]: Image

When testing I got to this step without no refresh button showing up? Maybe hat was fixed in your later commit. 😀

clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewTableView.tsx
```diff
@@ -0,0 +1,230 @@
...
+          columnGap="m"
+          rowGap="s"
+          flexWrap="wrap"
+        >
+          <Box maxWidth="100%" overflowX="auto">
+            <ReviewStatusTabs value={filter} onChange={onFilterChange} />
+          </Box>
+          <Box alignItems="center" columnGap="s" rowGap="s" flexWrap="wrap">
+            <Box maxWidth="100%" overflowX="auto">
+              <ReviewEntityTabs
```
Comment (strandhvilliam) [inline clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewTableView.tsx:214]: Just thinking out loud now, but I don't think using `SegmentedControls` as 'Filter' functionality is anything we use above other DataTables in the dashboard so it's kind of an on-off. 🤔 Not saying it looks bad, it actually looks good and makes sense so all options are immediately shown.

So I'd say let's keep it and maybe discuss if we wan't to be consistent with the other tables. I could even see us going towards this kinda look more.

Comment (psincraian) [reply clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewTableView.tsx:214]: yeah, it's one of the questions that I have also 🥲 I wanted to be able to show all the different options but not sure if this is the right approach.

I think a filter would make sense, but the problem that I have is that non importable rows or the ones that may require action are not in the front page and may be missed. One idea is to order by those by default and use standard filters

clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewTableView.tsx
Comment (psincraian) [reply clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewTableView.tsx]: ha! that is a real bug 👏 thanks for taking a look at this

clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewTableView.tsx
Comment (strandhvilliam) [reply clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewTableView.tsx:214]: yeah I agree that we do not want them too be hidden and easily missable. I'd say that we go with this for now and we can always iterate on the UX/UI later if we change our mind 😀

clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewTableView.tsx
```diff
@@ -0,0 +1,260 @@
...
+  return (
+    <Box as="section" flexDirection="column" rowGap="xl">
+      {importResult && (
+        <Alert
+          variant="success"
+          title="Catalog imported"
+          description={importResultText(importResult)}
+        />
+      )}
+      {importError && (
```
Comment (strandhvilliam) [inline clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewTableView.tsx:185]: Image

Maybe we can surface the error here to help the user know if there is something not right?

I needed to look at logs and check the code to find out what went wrong 😀
```
if account.is_connect_platform:
    yield PrecheckIssue(
        level=PrecheckIssueLevel.blocker,
        code="connect_platform_account",
        message=(
            "The source is a Connect platform; only platform-account data "
            "is copyable, so it can't be migrated automatically."
        )
```

clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/reviewRows.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+export function entityLabelSingular(entity: ReviewEntity): string {
+  return SINGULAR_LABELS[entity]
+}
+
+// A row can be picked for import only when the pre-check says it's importable
+// and it isn't already in the ledger as imported.
+export function isSelectable(row: ReviewRow): boolean {
+  return (
```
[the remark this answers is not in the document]
Comment (psincraian) [reply clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/reviewRows.ts:29]: that is expected

clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewRecordModal.tsx
```diff
@@ -0,0 +1,107 @@
...
+          </Field>
+          {amount && (
+            <Field label="Amount">
+              <Text monospace tabularNums>
+                {amount.money}
+                {amount.interval ?? ''}
+              </Text>
+            </Field>
+          )}
+          {row.subtitle && (
```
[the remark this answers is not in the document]
Comment (psincraian) [reply clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewRecordModal.tsx:74]: this in the modal can't happen

clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewTableView.tsx
```diff
@@ -0,0 +1,260 @@
...
+      <Box flexDirection="column" rowGap="m">
+        <Box
+          alignItems="center"
+          justifyContent="between"
+          columnGap="m"
+          rowGap="s"
+          flexWrap="wrap"
+        >
+          <Box maxWidth="100%" overflowX="auto">
+            <ReviewStatusTabs value={filter} onChange={onFilterChange} />
```
Comment (strandhvilliam) [inline clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewTableView.tsx]: Image

So sorry if this is nitpicky, only fix what you want to fix 🙈

I got confused cause I couldn't see my one product, but it was cause I was in Needs Attention. Which is good! But I thought it just didn't transfer. Could we  add a number counter on the StatusTabs as well as the EntityTabs so we know there are X Needs Attention and Y All? Just putting it here in case someone else would be confused 🤷‍♂️

Comment (psincraian) [reply clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewTableView.tsx]: that is a good point also!

strandhvilliam (approved):
This is so cool! I'm super impressed with the flow and how intuitive it feels.

There are small UX/design nitpicks and stuff that I think we should consider and continue iterating on as we go. But better to ship it and test it out 😎

clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewTableView.tsx
Comment (psincraian) [reply clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/migrations/review/ReviewTableView.tsx:185]: true, we lose and we don't save that 🙈

PR #13085, opened 2026-07-10
tf: enable invariants tasks in Lambdas

frankie567:
- tf: enable invariants tasks in Lambdas

<a href="https://cubic.dev/pr/polarsource/polar/pull/13085?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/terraform/global/production.tf
+++ b/terraform/global/production.tf
@@ -581,7 +581,7 @@ resource "tfe_variable" "worker_sqs_actors_production" {
   category        = "terraform"
   description     = "JSON array of Dramatiq actor names routed to the SQS execution engine for production"
   sensitive       = false
-  value           = "[\"dummy\"]"
+  value           = "[\"dummy\",\"observability.invariants.enqueue\",\"observability.invariants.check\"]"
   variable_set_id = tfe_variable_set.production.id
 
   lifecycle {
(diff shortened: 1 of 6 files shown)


psincraian (approved):

terraform/global/production.tf
```diff
@@ -581,7 +581,7 @@ resource "tfe_variable" "worker_sqs_actors_production" {
   category        = "terraform"
   description     = "JSON array of Dramatiq actor names routed to the SQS execution engine for production"
   sensitive       = false
-  value           = "[\"dummy\"]"
+  value           = "[\"dummy\",\"observability.invariants.enqueue\",\"observability.invariants.check\"]"
```
psincraian: this looks a bit ugly but ok

frankie567: I'm afraid we can't really do better 😅 I would advocate to keep the default set in `config.py`, but since Jesper explicitly set a variable for it, I preferred to keep it that way until he returns so we can challenge this

Yopi: These are currently set to ignore value changes, so you have to update it manually in tf cloud, or remove the lifecycle policy.

I did it like this to allow for easier roll out and roll back (instant instead of a build and deploy) but going forward it might indeed make more sense to have it in code.

You should he able to do jsonencode([“regular”, “array”) which looks a bit neater

frankie567:
-

PR #13095, opened 2026-07-10
feat(subscription): preview a subscription change before applying it

maximevast:
## Summary

Adds `POST /subscriptions/{id}/change-preview` and its customer-portal twin.
Given a proposed product or seat change, they return the proration breakdown and the amount due today, without applying anything.

The change is generated and applied inside a savepoint, then rolled back — so the preview prices the exact rows the invoice would bill, rather than reimplementing the calculation. 
Built on `compute_order_amounts`, extracted in #13087.

The portal endpoint mirrors the portal's own update contract: `product_id` alone for a product change, `seats` + `proration_behavior` for seats. A behavior the customer cannot apply must not be previewable.

Closes polarsource/feedback#58.

## Checklist

---
## Summary by cubic
Add subscription change preview to estimate costs before applying changes. Supports product switches or seat changes with exact proration and amount due today; nothing is persisted.

- **New Features**
  - Added `POST /v1/subscriptions/{id}/change-preview` and `/v1/customer-portal/subscriptions/{id}/change-preview`.
  - Request: internal accepts `product_id` or `seats` with optional `proration_behavior`; portal accepts only `product_id` or `seats` (no proration behavior).
  - Returns `SubscriptionChargePreview` with proration lines and total due; errors: 400 `NotASeatBasedSubscription`, 403 `AlreadyCanceledSubscription` (portal also `UpdateSubscriptionPlanNotAllowed`, `UpdateSubscriptionSeatsNotAllowed`), 404, 422.
  - Access: requires `subscriptions:write` and org `sales_manage`; portal requires billing write and follows org portal plan/seat update flags.
  - Updated client SDK with `subscriptions:preview_change` and `customer_portal:subscriptions:preview_change`; removed portal `proration_behavior` fields.

- **Bug Fixes**
  - If a change ends a trial, quote a full period; if the trial continues, quote 0 with no prorations.
  - Non-seat → seat previews now promote seats to the tier minimum and reject `next_period`.
  - Portal seat counts typed as `Int32` to match the column.

<sup>Written for commit cc897e0a421ebdeadf3510d4a2a739bb9b5391e1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/13095?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/customer_portal/schemas/subscription.py
+++ b/server/polar/customer_portal/schemas/subscription.py
@@ -12,7 +12,13 @@ from pydantic import (
 from pydantic.json_schema import SkipJsonSchema
 
 from polar.enums import SubscriptionProrationBehavior
-from polar.kit.schemas import IDSchema, Schema, SetSchemaReference, TimestampedSchema
+from polar.kit.schemas import (
+    IDSchema,
+    Int32,
+    Schema,
+    SetSchemaReference,
+    TimestampedSchema,
+)
 from polar.meter.schemas import NAME_DESCRIPTION as METER_NAME_DESCRIPTION
 from polar.models.subscription import CustomerCancellationReason
 from polar.product.schemas import (
@@ -87,7 +93,7 @@ class CustomerSubscriptionUpdateProduct(Schema):
 
 
 class CustomerSubscriptionUpdateSeats(Schema):
-    seats: int = Field(
+    seats: Int32 = Field(
         description="Update the number of seats for this subscription.",
         ge=1,
     )
@@ -111,7 +117,7 @@ class CustomerSubscriptionChangePreviewProduct(Schema):
 class CustomerSubscriptionChangePreviewSeats(Schema):
     model_config = ConfigDict(extra="forbid")
 
-    seats: int = Field(
+    seats: Int32 = Field(
         description="Preview a change of the subscription to this number of seats.",
         ge=1,
     )
--- a/server/polar/customer_portal/schemas/subscription.py
+++ b/server/polar/customer_portal/schemas/subscription.py
@@ -11,7 +11,6 @@ from pydantic import (
 )
 from pydantic.json_schema import SkipJsonSchema
 
-from polar.enums import SubscriptionProrationBehavior
 from polar.kit.schemas import (
     IDSchema,
     Int32,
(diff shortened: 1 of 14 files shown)


server/polar/customer_portal/schemas/subscription.py
```diff
@@ -80,7 +93,7 @@ class CustomerSubscriptionUpdateProduct(Schema):
 
 
 class CustomerSubscriptionUpdateSeats(Schema):
-    seats: int = Field(
+    seats: Int32 = Field(
```
frankie567, line 95: 🧠

server/polar/customer_portal/schemas/subscription.py
```diff
@@ -93,6 +106,36 @@ class CustomerSubscriptionUpdateSeats(Schema):
...
+
+
+class CustomerSubscriptionChangePreviewSeats(Schema):
+    model_config = ConfigDict(extra="forbid")
+
+    seats: Int32 = Field(
+        description="Preview a change of the subscription to this number of seats.",
+        ge=1,
+    )
+    proration_behavior: SubscriptionProrationBehavior | None = Field(
```
frankie567: Customers can't decide the proration behavior themselves, so I don't think we should expose it in the customer portal API

server/polar/order/amounts.py
```diff
@@ -16,6 +16,19 @@
 log: Logger = structlog.get_logger()
 
 
+@dataclass(frozen=True)
+class LineItem:
```
frankie567: Not sure to get why we need this proxy object?

frankie567 (changes requested):

server/polar/order/amounts.py
maximevast: that was to avoid orm objects in the read path, but it will be removed when doing the refactor you asked (shared read + write path).

Do you want me to get rid of it now or do we wait for the big refactor ?

frankie567: What would it take to remove it in this PR? Does it add lot of complexity with the session management?

frankie567 (approved):

PR #13112, opened 2026-07-13
feat: add invariants alerting on stalled order and subscription creation

psincraian:
## Summary

Add two observability invariants that alert when order or subscription creation stalls.

In production, orders and subscriptions are created all the time. If they stop, something in the checkout, payment, or subscription pipeline is broken. Today nothing tells us. These invariants catch it.

## What

- `NoRecentOrdersInvariant`: alerts if no new order in the last 15 minutes.
- `NoRecentSubscriptionsInvariant`: alerts if no new subscription in the last 45 minutes.

They run on the existing invariants cron (every 15 minutes) and post to Slack on failure, like the other invariants.

They run in production only. Sandbox, staging, and development have too little traffic for a quiet period to mean anything.

## Why

- Orders and subscriptions are a good health signal for the whole payment flow.
- A long silence is a strong sign of an outage.
- We want to know fast, not from a customer report.

## How

- Each invariant reads the most recent `created_at` from its table.
- If the newest one is older than the threshold, it raises and Slack is notified.
- The query uses the existing `created_at` btree index, so it reads a single row.
- A new `ENVIRONMENTS` attribute on the base `Invariant` gates where an invariant runs. `None` means all environments (the default). The service skips an invariant when the current environment is not in its set.

Thresholds are set above the largest natural gap seen in production, so normal quiet periods do not trigger a false alarm.

## Checklist

### DIFF
--- a/server/polar/observability/invariants/rules/base.py
+++ b/server/polar/observability/invariants/rules/base.py
@@ -1,6 +1,7 @@
 import abc
 import typing
 
+from polar.config import Environment
 from polar.postgres import AsyncReadSession
 
 
@@ -25,6 +26,9 @@ class Invariant(abc.ABC):
 
     Invariants are conditions that must always hold true for the system to be considered in a valid state.
     """
+
+    ENVIRONMENTS: typing.ClassVar[set[Environment] | None] = None
+    """Environments where this invariant runs. `None` means all environments."""
 
     def __init__(self, session: AsyncReadSession) -> None:
         self.session = session
--- a/server/polar/observability/invariants/rules/no_recent_subscriptions.py
+++ b/server/polar/observability/invariants/rules/no_recent_subscriptions.py
@@ -21,7 +21,12 @@ class NoRecentSubscriptionsInvariantError(InvariantError):
         super().__init__(
             NoRecentSubscriptionsInvariant,
             message,
-            {"threshold": str(threshold), "last_subscription_at": last_subscription_at},
+            {
+                "threshold": str(threshold),
+                "last_subscription_at": (
+                    last_subscription_at.isoformat() if last_subscription_at else None
+                ),
+            },
         )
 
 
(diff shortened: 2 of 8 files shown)


server/polar/observability/invariants/rules/base.py
```diff
@@ -26,6 +27,9 @@ class Invariant(abc.ABC):
     Invariants are conditions that must always hold true for the system to be considered in a valid state.
     """
 
+    ENVIRONMENTS: typing.ClassVar[set[Environment] | None] = None
+    """Environments where this invariant runs. `None` means all environments."""
```
frankie567, line 31: Instead of having a setting per invariant, what about a config boolean? I guess what we want is to run this only for production.

server/polar/observability/invariants/rules/no_recent_subscriptions.py
```diff
@@ -0,0 +1,59 @@
...
+
+    In production, subscriptions are created continuously. A prolonged silence does
+    not happen under normal traffic and indicates the checkout or subscription
+    pipeline is stalled. Only runs in production, where traffic is continuous; other
+    environments have too little traffic for the absence of subscriptions to be
+    meaningful.
+    """
+
+    ENVIRONMENTS = {Environment.production}
+    THRESHOLD = timedelta(minutes=45)
```
frankie567, line 46: How did you determine that threshold?

frankie567 (commented):
Nice idea! I have just a few interrogations :)

server/polar/observability/invariants/rules/no_recent_subscriptions.py
psincraian: answered privately

server/polar/observability/invariants/rules/base.py
psincraian: I think it provides the same value but more flexibility

frankie567 (approved):

PR #13114, opened 2026-07-13
refactor: remove seat_based_pricing_enabled feature flag

psincraian:
## Summary

The `seat_based_pricing_enabled` org feature flag is now `true` for every organization. It is a constant. This PR removes it.

The change is behavior-preserving. Every gate had the form `not member_model and not seat_based`. Since `seat_based` is always true, those branches were already dead in production.

## What

- Remove the field from the org feature settings schemas (read + update).
- Remove the two seat-based validation errors from organization update.
- Remove the dead guards in the member, customer_seat, and product services, and the `prepare_members` task.
- Remove the seat feature-check helper (`check_seat_feature_enabled`) that became a redundant org-existence check, and its now-unused `FeatureNotEnabled` error.
- Simplify the frontend to always offer seat-based pricing. Regenerate the API client.
- Delete tests that only checked the removed gating.

## Why

The flag was an opt-in product gate. It has been enabled everywhere, so keeping it adds dead code and needless branches.

## How

Treat the flag as a constant `true` and simplify each site. The DB JSONB key is not deleted, and the migration scripts are left untouched, so the separate `member_model` migration of remaining orgs still works.

## Checklist


pieterbeulque:
What a beautiful diff.

PR #13116, opened 2026-07-13
feat(observability): add PayoutTransactionsAmountInvariant

frankie567:
Add new invariant rule that validates payout transaction amounts correspond
to the sum of transactions that reference them via payout_transaction_id.

- PayoutTransactionsAmountInvariant: checks if payout amounts match sum
  of paid transactions
- Excludes reversed payouts (by design, when reversed payout_transaction_id
  is reset to NULL on paid transactions)
- Uses abs(diff) > 0 to catch mismatches in both directions
- Follows existing invariant pattern with LIMIT=10 and has_more flag

Tests cover:

- Success cases (no payouts, matching amounts, zero amount)
- Failure cases (amount mismatch, missing transactions, sum exceeds payout,
  over limit)
- Edge case (reversed payouts excluded)

<a href="https://cubic.dev/pr/polarsource/polar/pull/13116?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/observability/invariants/rules/payout_transactions_amount_invariant.py
+++ b/server/polar/observability/invariants/rules/payout_transactions_amount_invariant.py
@@ -1,0 +1,106 @@
+import uuid
+from datetime import timedelta
+
+from sqlalchemy import func, over, select
+
+from polar.models import Transaction
+from polar.models.transaction import TransactionType
+
+from .base import Invariant, InvariantError
+
+
+class PayoutTransactionsAmountInvariantError(InvariantError):
+    """Exception raised when the PayoutTransactionsAmountInvariant check fails."""
+
+    def __init__(
+        self, count: int, payout_transactions: list[uuid.UUID], differences: list[int]
+    ) -> None:
+        message = (
+            f"Found {count} payout transactions with amount mismatch "
+            "between payout and sum of paid transactions."
+        )
+        super().__init__(
+            PayoutTransactionsAmountInvariant,
+            message,
+            {
+                "count": count,
+                "payout_transactions": {
+                    "ids": payout_transactions,
+                    "differences": differences,
+                    "has_more": count > len(payout_transactions),
+                },
+            },
+        )
+
+
+class PayoutTransactionsAmountInvariant(Invariant):
+    """
+    Invariant that checks if payout transaction amounts match the sum of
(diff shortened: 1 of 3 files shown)


server/polar/observability/invariants/rules/payout_transactions_amount_invariant.py
```diff
@@ -0,0 +1,103 @@
...
+
+    LIMIT = 10
+
+    async def check(self) -> None:
+        # Payout transactions without reversal (same payout_id)
+        payout_subq = (
+            select(Transaction.id, Transaction.amount)
+            .where(
+                Transaction.type == TransactionType.payout,
+                ~Transaction.payout_id.in_(
```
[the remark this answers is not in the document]
frankie567: @cubic-dev-ai Payout transactions _always_ have a Payout associated, so this is a nitpick.

server/polar/observability/invariants/rules/payout_transactions_amount_invariant.py
```diff
@@ -0,0 +1,103 @@
...
+            .where(Transaction.payout_transaction_id.isnot(None))
+            .group_by(Transaction.payout_transaction_id)
+            .subquery()
+        )
+
+        # Find mismatches
+        diff_expr = func.abs(payout_subq.c.amount) - func.coalesce(
+            paid_subq.c.total_amount, 0
+        )
+        statement = (
```
psincraian, line 82: Should we limit the data to the last X days only? I mean we are mainly interested in new payouts so I don't see the benefit of checking older ones than 30d for example

psincraian (approved):

server/polar/observability/invariants/rules/payout_transactions_amount_invariant.py
frankie567: Hmm, true, but I'm concerned we may have some bugs/issues that deteriorate existing data. But that may be a bit paranoid.

PR #13125, opened 2026-07-13
  feat(subscription): stop dunning retries when a past-due subscription is cancelled

maximevast:
## Summary

**Related Issue**: polarsource/feedback#131

Cancelling a `past_due` subscription now stops the failed-payment retries (dunning) instead of letting them run to a surprise charge days later.

**Customer portal (video):**

https://github.com/user-attachments/assets/5861d1c4-1cef-4cf4-831f-ec7f28aaef13

**Merchant dashboard (screenshot):**

Screenshot 2026-07-13 at 14 52 31

## What

- Cancelling a `past_due` subscription whose organization has **no benefit revocation grace period** (`benefit_revocation_grace_period == 0`) now cancels it **immediately** and **voids the pending order**, so dunning stops. Non-past-due subs, and past-due subs with a grace period configured, are unchanged.
- New `cancel-preview` endpoint on both the merchant API and the customer portal, returning whether cancelling stops collection and the outstanding amount.
- Both cancel confirmations warn the user when this applies. The merchant modal names the outstanding amount and forces immediate cancellation (the "end of period" option is disabled).

Also included, surfaced while building this:
- **fix(organization):** the default `subscription_settings` was a shared dict; made it a per-instance factory so orgs can't mutate each other's defaults.
- **fix(customer-portal):** the "Manage subscription" panel rendered a stale state snapshot, so a cancellation didn't reflect until reload; it now derives from the live list.

## Why

Before this, cancelling a past-due subscription left the pending order in dunning. A retry could succeed days after the customer or merchant believed the subscription was cancelled, leading to refunds and chargebacks. The only way to stop retries was an immediate revoke (merchant/backoffice only) or a staffer manually voiding the order — and customers had no lever at all.

We gate strictly on grace period `== 0`: with no grace, the customer never retained access during the past-due window, so dropping the failed payment is unambiguously fair. A configured grace period keeps collecting, to avoid a "use it through the grace window, then cancel and never pay" abuse path.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/13125?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/polar/subscription/service.py
```diff
@@ -1885,12 +1887,14 @@ async def cancel(
         customer_reason: CustomerCancellationReason | None = None,
         customer_comment: str | None = None,
     ) -> Subscription:
+        immediately = await self._cancel_stops_collection(session, subscription)
```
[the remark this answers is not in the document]
maximevast: Not really new, `cancel_customer` already voids pending orders the same async way (via `_on_subscription_revoked`), so this just reuses existing behavior. And in practice `next_payment_attempt_at` is always in the future at cancel time, so the order isn't due and the job clears it seconds later. You'd need a retry due at the exact cancel moment and the dunning cron firing in that gap.

Can make it sync if we prefer, but I'd keep it consistent. @frankie567 wdyt?

server/tests/subscription/test_endpoints.py
```diff
@@ -668,7 +671,10 @@ async def test_valid_past_due(
 
         assert response.status_code == 200
         updated_subscription = response.json()
-        assert updated_subscription["status"] == SubscriptionStatus.past_due
+        # No revocation grace period (org default 0): cancelling a past-due
+        # subscription ends it immediately instead of at period end.
+        assert updated_subscription["status"] == SubscriptionStatus.canceled
+        assert updated_subscription["cancel_at_period_end"] is False
```
pieterbeulque: This is unexpected to me, since we explicitly pass `cancel_at_period_end` `True` in the `PATCH` call.

clients/apps/web/src/components/Subscriptions/CancelSubscriptionModal.tsx
```diff
@@ -158,7 +196,10 @@ const CancelSubscriptionModal = ({
                         </SelectTrigger>
                         <SelectContent>
                           <SelectItem value="revoke">Immediately</SelectItem>
-                          <SelectItem value="cancel_at_period_end">
+                          <SelectItem
+                            value="cancel_at_period_end"
+                            disabled={stopsCollection}
```
pieterbeulque: This disables the current behavior, right?

server/polar/subscription/service.py
pieterbeulque: Would it be cleaner to have the Customer Portal front-end chose automatically between `cancel` and `revoke` depending on the `stops_collection` value? I'm not fond of this muddying of the difference between `cancel` and `revoke`.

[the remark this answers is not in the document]
frankie567: Yeah, it's nitpick

frankie567:
I'm not really comfortable with that. It bundles lot of hidden rules and side effects into the cancellation endpoint; it feels like it'll do more harm than good.

I'm wondering if we could just surface it as a warning in the customer portal or create a specific support case "Customer asks for immediate cancellation and collection stop" that has to be manually handled by the merchant.

server/polar/subscription/service.py
maximevast: @pieterbeulque do you want it to be only semantic (the label of the CTA label change depending on the boolean, the backend keeps doing it's magic) or should we add revoke to the customer api ?

clients/apps/web/src/components/Subscriptions/CancelSubscriptionModal.tsx
maximevast: Yeah, on purpose: for a past-due sub with no grace period, cancelling always ends immediately, so "End of current period" would just get overridden by the backend.

maximevast:
Closing in favor of https://github.com/polarsource/polar/pull/13188

PR #13128, opened 2026-07-13
fix order callout banner handling unrecoverable decline codes

strandhvilliam:
Adds a distinct banner state for subscription payments that fail with a permanent Stripe decline code (e.g. stolen_card, expired_card, do_not_honor) — cases where automatic retries will never succeed and the customer must update their payment method.

And some refactoring.

| New non recoverable state |
|---|
| Screenshot 2026-07-13 at 17 41 33 |

<a href="https://cubic.dev/pr/polarsource/polar/pull/13128?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


clients/apps/web/src/components/Orders/OrderCalloutBanner.tsx
```diff
@@ -225,26 +184,132 @@ export const OrderCalloutBanner = ({
...
-            nextAttemptAt={order.next_payment_attempt_at ?? null}
-            revocationDeadline={revocationDeadline}
+        </BannerColumn>
+      </BannerFrame>
+    )
+  }
+
+  if (nonRecoverable) {
+    return (
+      <BannerFrame heading="Payment method can't be charged">
```
strandhvilliam, line 227: Better copy for this?

Thinking here is that it shows the card is the problem.

But maybe we want to show the non-retryable aspect? Like "Payment won't be retried"

clients/apps/web/src/components/Orders/OrderCalloutBanner.tsx
```diff
@@ -225,26 +184,132 @@ export const OrderCalloutBanner = ({
...
+          <ColumnHeading
+            label={
+              revocationDeadline
+                ? 'Subscription will be canceled on'
+                : 'Subscription will be canceled'
+            }
+            datetime={revocationDeadline}
+          />
+          <Text>
+            This card can&apos;t be charged again. The subscription will be
```
strandhvilliam: Maybe 'card' is to specific? Maybe 'payment method' is better?

Also the 'subscription will be cancelled unless the customer updates their payment method' maybe is unnecessary filler?

clients/apps/web/src/components/Orders/OrderCalloutBanner.tsx
pieterbeulque: I think this is great as a title.

clients/apps/web/src/components/Orders/OrderCalloutBanner.tsx
pieterbeulque: Card & payment method is basically the same thing, for now, and I think card reads smoother.

The only copy that's not 100% is that we say "This card can't be charged again" but we don't explain why. Maybe we need an "Unrecoverable" badge or callout in/near the bank decline reason, that explains that reason XYZ is not recoverable (every attempt will throw the same error), to tie it together.

> subscription will be cancelled unless the customer updates their payment method

I think it's good to explain to merchants what the way out is (i.e. work with the customer to add a new card)

clients/apps/web/src/utils/order.ts
```diff
@@ -70,3 +71,64 @@ export function isOrderDunningFailed(
     subscription.customer_cancellation_reason === null
   )
 }
+
+// Mirror of UNRECOVERABLE_DECLINE_CODES in server/polar/models/payment.py.
+// Keep in sync — Stripe decline codes that will never succeed on retry.
+const UNRECOVERABLE_DECLINE_CODES = new Set([
```
pieterbeulque, line 77: Since it's mostly for aesthetics, I'm fine with the duplication, but it'd be great if we could work around that (computed `recoverable` field on the payment API?). Though I'm also not a fan of cluttering our schema with these internals.

strandhvilliam: Yeah I was thinking about this also, can easily to get out of sync 🙈

Same issue in other parts are for example: `VISIBILITY_CONFIGURABLE_BENEFIT_TYPES`, `BLOCKED_WORDS` etc.

Can't think of another way than exposing computed fields though to solve it right now. But something for the future maybe

clients/apps/web/src/components/Orders/OrderCalloutBanner.tsx
strandhvilliam: This is good feedback, will experiment with that badge/callout idea

strandhvilliam: Maybe something like this? (em-dashes looks kinda AI-ish but we seem to be using it a lot for copy in the codebase)

Screenshot 2026-07-14 at 09 42 50

pieterbeulque: Maybe just repeat the unrecoverable.

> This decline reason is unrecoverable. Every retry will fail the same way, so we won't charge the card again.

clients/apps/web/src/utils/order.ts
pieterbeulque: Yeah, don't worry about it

pieterbeulque (approved):

PR #13143, opened 2026-07-14
fix(sso): allow clearing the SSO connection name

pieterbeulque:
Fixes https://github.com/polarsource/feedback/issues/247

## Summary

**Related Issue**: polarsource/feedback#247

Users could not clear the human-friendly `name` of an SSO connection once it had been set. Both ways of clearing it were dropped by the backend, so the name was stuck at whatever it was first assigned.

## What

- `OrganizationSSOConnectionService.update` now dumps the patch with `exclude_unset=True` instead of `exclude_none=True`, so an explicit `name: null` is preserved and applied (clearing the column) instead of being silently discarded.
- The `name` field on the create/update schemas now normalizes a blank or whitespace-only string to `null` (via a `BeforeValidator`), so `name: ""` clears the label instead of failing validation with a 422.

## Why

The `name` column is nullable and is meant to be clearable — the dashboard already sends `name: values.name || null` when the field is emptied. Two independent issues prevented that from working:

1. `model_dump(exclude_none=True)` removed the explicit `null`, producing an empty update that changed nothing.
2. `NonEmptyStr` (`min_length=1`) rejected `""` with a `422`.

`exclude_unset=True` is also the PATCH convention already used by the other update services (account, customer, product), so this brings SSO in line with them.

## How

- Preserve explicitly-provided values (including `null`) by switching to `exclude_unset=True`; fields omitted from the request body are still left untouched.
- Add an `OptionalName` type that maps blank/whitespace input to `None` before validation, applied to both the create and update schemas. The generated OpenAPI schema is unchanged, so no client regeneration is required, and the dashboard needs no changes (it already sends `null`).
- Cover the update flow with tests: setting the name, clearing via `null`, clearing via `""`, and preserving the name when it is not part of the request.

## Checklist

- [ ] Linting and type checking pass (`uv run task lint && uv run task lint_types`) — `ruff check`/`ruff format` verified on the changed files; the full suite / mypy will run in CI

---
_Generated by [Claude Code](https://claude.ai/code/session_0138Kq5PDRGxhuwLCzbErXPE)_

<a href="https://cubic.dev/pr/polarsource/polar/pull/13143?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/sso/schemas.py
+++ b/server/polar/sso/schemas.py
@@ -1,6 +1,12 @@
 from typing import Annotated, Literal
 
-from pydantic import UUID4, Discriminator, Field, StringConstraints
+from pydantic import (
+    UUID4,
+    BeforeValidator,
+    Discriminator,
+    Field,
+    StringConstraints,
+)
 
 from polar.kit.schemas import HttpsUrl, IDSchema, Schema, TimestampedSchema
 from polar.models.organization_sso_connection import (
@@ -9,6 +15,15 @@ from polar.models.organization_sso_connection import (
 )
 
 NonEmptyStr = Annotated[str, StringConstraints(strip_whitespace=True, min_length=1)]
+
+
+def _blank_to_none(value: object) -> object:
+    if isinstance(value, str) and not value.strip():
+        return None
+    return value
+
+
+OptionalName = Annotated[NonEmptyStr | None, BeforeValidator(_blank_to_none)]
 
 
 class OIDCConfigurationBase(Schema):
@@ -66,7 +81,7 @@ class OrganizationSSOConnection(IDSchema, TimestampedSchema):
 
 
 class OrganizationSSOConnectionCreate(Schema):
-    name: NonEmptyStr | None = Field(
+    name: OptionalName = Field(
         default=None,
         description="Human-friendly label for the connection, shown on the login page.",
     )
@@ -83,7 +98,7 @@ class OrganizationSSOConnectionCreate(Schema):
 
 
 class OrganizationSSOConnectionUpdate(Schema):
-    name: NonEmptyStr | None = Field(
+    name: OptionalName = Field(
         default=None,
         description="Human-friendly label for the connection, shown on the login page.",
     )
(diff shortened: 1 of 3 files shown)


server/polar/sso/schemas.py
```diff
@@ -11,6 +17,15 @@
 NonEmptyStr = Annotated[str, StringConstraints(strip_whitespace=True, min_length=1)]
 
 
+def _blank_to_none(value: object) -> object:
+    if isinstance(value, str) and not value.strip():
+        return None
+    return value
+
+
+OptionalName = Annotated[NonEmptyStr | None, BeforeValidator(_blank_to_none)]
```
frankie567: We already have `EmptyStrToNone` type.

frankie567 (changes requested):

server/polar/sso/schemas.py
pieterbeulque: Thanks! I tried finding it but my ⌘ + F skills were lacking

server/polar/sso/service.py
```diff
@@ -89,7 +89,7 @@ async def update(
         ):
             raise LastSSOConnectionRequired()
         repository = OrganizationSSOConnectionRepository.from_session(session)
-        update_dict = update.model_dump(exclude_none=True)
+        update_dict = update.model_dump(exclude_unset=True)
```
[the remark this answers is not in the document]
pieterbeulque: Good catch — fixed in d913936. Since `name` is the only nullable column, the service now preserves an explicit `null` for `name` (to clear it) and drops explicit nulls for `configuration`/`enabled`, so they retain their current value instead of failing the write. Added `test_null_configuration_is_ignored` and `test_null_enabled_is_ignored` to cover it.

---

server/polar/sso/schemas.py
pieterbeulque: @cubic-dev-ai can you store that as a learning?

maximevast (approved):
LGTM

frankie567 (approved):

PR #13189, opened 2026-07-15
feat: merchant-migration review listing and classification

psincraian:
## Summary

The read side of merchant migration: it classifies a merchant's staged Stripe catalog and serves the data the review dashboard shows before anything is imported. 

## What

- Classifies each staged record as **importable** or **skipped**, each with a reason and price detail (amount, currency, interval).
- `/records` gains an **"all"** view across products, customers and subscriptions, a **"needs attention"** filter (importable rows that carry a warning), and per-record warnings: no billing country, on trial, and one-time or discounted items that won't import.
- Warns when a source product name already matches an existing Polar product.
- Stable record ordering so paginated listing stays consistent.

## Why

Merchants need to see exactly what will and won't come over, and why, before they run the import. Getting the classification right up front avoids surprises during import.

## How

There is one classification plan per entity, shared by the report and (in the follow-up PR) the importer, so the review can never promise something the importer would skip.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/13189?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/merchant_migration/adapters/stripe.py
+++ b/server/polar/merchant_migration/adapters/stripe.py
@@ -224,6 +224,8 @@ class StripeAdapter:
             line_item_count=len(items),
             quantity=first_item.get("quantity") or 1,
             payment_method=self._resolve_payment_method(subscription),
+            has_discount=bool(subscription.get("discounts"))
+            or subscription.get("discount") is not None,
         )
 
     def _map_customer(self, customer: stripe_lib.Customer) -> CanonicalCustomer:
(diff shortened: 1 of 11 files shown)


server/polar/merchant_migration/adapters/stripe.py
```diff
@@ -224,6 +224,8 @@ def _map_subscription(
             line_item_count=len(items),
             quantity=first_item.get("quantity") or 1,
             payment_method=self._resolve_payment_method(subscription),
+            has_discount=bool(subscription.get("discounts"))
```
psincraian, line 227: Discounts are gonna be handled later on. Many things to take into account

PR #13216, opened 2026-07-16
feat: add one-off script to reject Stripe accounts of denied and blocked orgs

psincraian:
## Summary

Add a one-off script that rejects the Stripe connected account of every organization that is currently `denied` or `blocked`.

## Why

The backoffice deny/block dialog now has an opt-in "Disable Stripe account" box. It rejects the org's Stripe account when a reviewer denies or blocks them. Orgs denied or blocked *before* that box existed still have live Stripe accounts. This script syncs Stripe with our side so those accounts get rejected too.

Rejecting on Stripe blocks payouts there, stops account reuse, and makes us look better to Stripe by flagging bad accounts.

## What

- Finds non-deleted `denied` and `blocked` orgs with a live Stripe payout account.
- Rejects each account through the same service the product uses.
- Dry-run by default. Pass `--execute` to actually reject.
- `--reason` picks the Stripe reject reason (default `terms_of_service`).

## How

- Reuses `payout_account_service.reject_stripe_account`, so it cannot drift from the product flow.
- Skips any Stripe account linked to more than one org. Those are printed in a table for manual handling, since rejecting is permanent and would hit the other orgs too.
- Per-account Stripe errors are caught and logged, so one failure does not stop the run.

## Checklist

### DIFF
--- a/server/scripts/reject_denied_orgs_stripe_accounts.py
+++ b/server/scripts/reject_denied_orgs_stripe_accounts.py
@@ -1,10 +1,11 @@
 """
-Reject the Stripe connected accounts of organizations that are already denied.
+Reject the Stripe accounts of organizations that are already denied or blocked.
 
 The backoffice deny/block dialog now has an opt-in "Disable Stripe account" box
-that rejects the org's Stripe connected account. Orgs denied *before* that box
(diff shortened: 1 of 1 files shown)


maximevast (approved):
LGTM !

PR #13224, opened 2026-07-16
feat(models): add subscription_pauses table

maximevast:
## Context

`paused_at` is cleared on resume, so a subscription row only knows about its *current* pause — past pause windows are unreconstructable, which blocks accurate paused-count / paused-MRR metrics. Since pause has ~zero adoption, we're remodeling it as interval data now rather than building metrics on top of the flat columns (see #subscriptions-pause discussion).

## Changes

- New `SubscriptionPause` model (`subscription_pauses`): one row per pause   period — `started_at` set when the pause takes effect, `ended_at` on resume,   never cleared. Partial unique index guarantees at most one open pause per
  subscription. 
- Migration creates the table

`pause_at_period_end` and `resumes_at` stay on `subscription`: they're future intent, not history — same split as `cancel_at_period_end`/`ends_at` vs `ended_at`.

## Follow-ups

1. Rewire pause/resume/cycle to open/close pause rows; derive `paused_at` from    the open row and drop the column. API schema unchanged (flat fields derived).
2. Paused Subscriptions / Paused MRR metrics; exclude paused from MRR & active.

Model + migration only, per the deploy-migrations-first convention.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/13224?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


maximevast:
https://polar-sh.slack.com/archives/C0B3HPC2T5G/p1784219210431989?thread_ts=1784215978.152649&cid=C0B3HPC2T5G

PR #13254, opened 2026-07-17
Add route-level permission guards to dashboard sections

strandhvilliam:
## Summary

Adds OrganizationPermissionGuard to all pages in dashboard, replacing some client side permission checks and gating some pages without permission guards to prepare for future possible RBAC roles. (see https://github.com/polarsource/polar/compare/claude/finance-role-prototype-kli3zj etc)

## What

Added a small layout.tsx guard per section:

- analytics, compass → analytics:read
- customers → customers:read
- products → products:read
- sales → sales:read
- settings/members → members:read
- settings/custom-fields → custom_fields:manage
- settings/migrations, sso, webhooks, billing → organization:manage

Removed the old in-page checks in webhooks, sso, and billing since the guard replaces them.

## Why

Most pages had no access check, and the few that did each rolled their own. In the layout it's one consistent "no access" screen, no loading flicker (permissions come from the auth payload), and pages you can't see never load their data.

## How

Each section is guarded on its read permission, what most people on a team have. most manage actions (refunds, editing) still check for themselves for now keeping view and manage separate is what lets a future role view without editing.

Gating is intentionally general, and In future could be more fine-tuned (for example analytics/ route has one guard since same permission, but easy to add specific guards for costs/ metrics/ etc.

Left out for now: the home dashboard (mixed widgets, needs per-widget handling), onboarding, and hiding manage buttons for read-only users as this is a UX gate, not real security as API enforces it. In future we probably also want the guard to accept multiple permissions if page requires multiple permission, but for now it's the 'loosest' option.

## Checklist

- [x] Linting and type checking pass (uv run task lint && uv run task lint_types)

<a href="https://cubic.dev/pr/polarsource/polar/pull/13254?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/analytics/layout.tsx
+++ b/clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/analytics/layout.tsx
@@ -1,0 +1,21 @@
+import OrganizationPermissionGuard from '@/components/Auth/OrganizationPermissionGuard'
+
+export default async function Layout({
+  children,
+  params,
+}: {
+  children: React.ReactNode
+  params: Promise<{ organization: string }>
+}) {
+  const { organization } = await params
+
+  return (
+    <OrganizationPermissionGuard
+      organizationSlug={organization}
+      permission="analytics:read"
+      standalone
+    >
+      {children}
+    </OrganizationPermissionGuard>
+  )
+}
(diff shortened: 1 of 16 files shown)


pieterbeulque (approved):
Really nice, thanks! Let's maybe ship this on Monday because y'know… weekends.

PR #13256, opened 2026-07-17
sdk: add webhook signature verification helpers

### LINKED ISSUE #13252: sdk: add a webhook signature verification helper
Similar to the one we added manually in Speakeasy SDK:

* https://github.com/polarsource/polar-python/blob/f069b317d1402eb8db3bf8c8065ba58f8ddc0b54/src/polar_sdk/_webhooks/__init__.py#L119-L135
* https://github.com/polarsource/polar-js/blob/63510677499a41e9694be83a96aafbafc750d7f6/src/webhooks.ts#L135-L151

They're both using the standardwebhooks dependency. It should be added directly in their respective `webhooks.{py|ts}` module below the payload definitions.

frankie567:
## Summary

- generate typed webhook verification helpers for the Python and TypeScript SDKs
- verify Standard Webhooks signatures with Python and Node standard-library primitives in a shared, unversioned module
- keep Standard Webhooks test-only while binding each API version to its known event types and payload union
- add root-level compatibility tests with dummy payload models and signatures generated by the reference library

## Why

The generated SDKs expose webhook payload types but did not provide a lightweight way to authenticate incoming webhook requests and return the matching typed payload. Keeping the verification core unversioned avoids duplicating cryptographic logic across API versions and avoids adding Standard Webhooks' transitive dependencies to SDK users.

## Validation

- `cd sdk/generator && just lint && just test` — 61 tests passed
- Python SDK: Ruff and ty passed; 21 tests passed
- TypeScript SDK: formatting, lint, type-check, build passed; 17 tests passed
- verified production dependency trees exclude `standardwebhooks`

Closes #13252

<a href="https://cubic.dev/pr/polarsource/polar/pull/13256?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
 clients/packages/client/src/v1.ts linguist-generated=true
+sdk/python/** linguist-generated=true
+sdk/typescript/** linguist-generated=true
--- a/sdk/generator/generator/emitter.py
+++ b/sdk/generator/generator/emitter.py
@@ -8,7 +8,7 @@ from typing import Literal
 
 from jinja2 import Environment, FileSystemLoader, Template
 
-from generator.ir import APIIR, APIVersion
+from generator.ir import APIIR, APIVersion, LiteralType
 
 _LOOP_DIRECTORY_PATTERN = re.compile(r"^\[\[([\w\.]+)\]\]$")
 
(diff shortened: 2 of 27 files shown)


.gitattributes
```diff
@@ -1 +1,3 @@
 clients/packages/client/src/v1.ts linguist-generated=true
+sdk/python/** linguist-generated=true
```
maximevast, line 2: <3

maximevast (approved):
Not an easy read to be honest but I haven't seen anything suspicious

PR #13268, opened 2026-07-20
fix: send subscription renewal confirmation only after payment succeeds

maximevast:
A failed subscription renewal sent two contradictory emails: "your subscription has been renewed", then "your payment failed".

The confirmation was queued when the order was created, before the charge was attempted. It is now queued when the order becomes paid.

## What

- Move the confirmation email to the paid transition.
- Fix a duplicate paid transition. A charge below the currency minimum, such as a small proration, was marked paid twice. That duplicated the confirmation email, the receipt and the `order.paid` webhook.
- Reword the past-due email to state the failure plainly.

  Screenshot 2026-07-20 at 11 34 22

## Behaviour changes ⚠️ 

- `order.created` now reports `status: pending` for zero-due and sync-settled subscription orders. It previously reported `paid`. All cycle paths are now consistent, and `order.created` always precedes `order.paid`.
- During the deploy, an order created before it and paid after it may send a second confirmation. This is one-time, and the second email is the correct one.
- A renewal with no payment method now sends `order.created` followed by `order.updated`. The retry date and past-due status are in the second one. Before, they were in `order.created`. (⚠️ are we breaking webhooks here?)

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/13268?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/emails/src/emails/subscription_past_due.tsx
+++ b/server/emails/src/emails/subscription_past_due.tsx
@@ -6,6 +6,7 @@ import {
   Text,
   WrapperOrganization,
 } from '../components/foundation'
+import InfoBox from '../components/InfoBox'
 import { organization, product } from '../preview'
 import type { schemas } from '../types'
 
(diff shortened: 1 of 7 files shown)


frankie567 (approved):
Yes, that makes sense. I was reluctant to do the `order.created` with the `pending` state for free orders, but now that I see it, it definitely makes sense. Let's go.

PR #13272, opened 2026-07-20
Fix order payment lock wedging forever, breaking dunning retries

### LINKED ISSUE #13124: Order payment lock can wedge forever, breaking dunning retries
`payment_lock_acquired_at` is only cleared on payment success/failure. If a `PaymentIntent` stalls (e.g. customer never completes 3DS and Stripe never sends a terminal webhook), the lock never releases.

With the order stuck locked:

* `OrderService.trigger_payment` raises `PaymentAlreadyInProgress` on every subsequent attempt, including scheduled dunning.
* `order.trigger_payment` (`server/polar/order/tasks.py`) doesn't catch `PaymentAlreadyInProgress` (only `PaymentFailed` and Stripe connectivity errors are caught), so it dead-letters via Dramatiq's retry middleware.
* `process_dunning`/`process_dunning_order` don't check the lock when selecting orders, so the same order is re-enqueued every hour, failing identically forever with no progression of `next_payment_attempt_at`.

Reproduced on a real order: lock acquired at retry time, PI went to `requires_action`, no resolving webhook ever arrived, and dunning has retried hourly without success since.

## Proposed fix

Simply releasing the lock on `requires_action` isn't safe: it would let a second payment attempt fire while the first PaymentIntent's 3DS challenge is still outstanding, and the first PI can still resolve to `succeeded` later (e.g. via Stripe's off-session SCA recovery email), causing a double charge. The lock must only ever be released in response to a Stripe-confirmed terminal state.

1. Add a scheduled cleanup that finds orders whose `payment_lock_acquired_at` has been held past a stale threshold and explicitly cancels the associated Stripe `PaymentIntent` (new `StripeService.cancel_payment_intent`; no cancellation support exists today). The cleanup does not release the lock directly.
2. Handle the `payment_intent.canceled` webhook (currently unhandled entirely: not in `DIRECT_IMPLEMENTED_WEBHOOKS`, no actor in `stripe/tasks.py`) to release `payment_lock_acquired_at` and advance `next_payment_attempt_at`, the same way `payment_intent.payment_failed` does today. Since the lock is only released once cancellation is confirmed over webhook, the lock and the intent's fate can't drift out of sync, and a stale intent can't sneak in a late success after we've moved on.
3. Handle `PaymentAlreadyInProgress` gracefully in `order.trigger_payment` (similar to the existing catch in `void_pending_orders_for_subscription`) instead of letting it dead-letter every hour.

Related prior art: #12020 (analogous stuck-lock pattern for `scheduler_locked_at` on subscriptions).

<sub>Sent by @frankie567 from </sub>[<sub>Stripe payment pi_3Trn5EDG1jUQrXwC0178qPdx missing</sub>](https://app.stilla.ai/m/memo_01kxdsd07rems9gzszxbwway2s)<sub>.</sub>

maximevast:
Fixes #13124. Supersedes #13242, which tracked the PaymentIntent in a new column — this records it in `payments` instead, so no migration and no backfill.

An off-session intent parked in `requires_action` never gets a resolving webhook, so the payment lock is held forever and dunning dead-letters on the order every hour. Such an intent has no charge and no `last_payment_error`, so it leaves no `Payment` row at all today.

A PaymentIntent confirmed without a charge now records a `pending` payment holding its intent id, re-keyed onto the charge once one exists. This covers checkout and wallet top-ups too, not just renewals. `pending` rather than a status of its own, per @frankie567 — a dedicated one is a follow-up.

An hourly cleanup cancels the intents behind stale locks, read from that record. The lock is released only once `payment_intent.canceled` confirms it, and only while the lock is still stale so a live attempt isn't cut short.
Releasing on a timer instead would risk a double charge: a stuck intent can still settle via Stripe's SCA recovery email.

Orders already wedged have no record, so the cleanup skips them and logs it. Cancel those intents in the Stripe dashboard; the webhook does the rest.

Requires `payment_intent.canceled` enabled on the Stripe webhook endpoints before deploy.

Screenshot 2026-07-20 at 15 23 40

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/13272?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/polar/order/service.py
```diff
@@ -135,6 +135,23 @@
...
+# and can therefore be cancelled. `processing` is excluded: Stripe is settling
+# it and will resolve it on its own.
+AWAITING_PAYMENT_INTENT_STATUSES = {
+    "requires_payment_method",
+    "requires_confirmation",
+    "requires_action",
+}
+
+
+def _payment_lock_is_stale(order: Order) -> bool:
```
frankie567: Let's define that as a method on Order model

frankie567 (changes requested):
A nitpick. High level, I think the right way to handle that would be to actually record the payments in the requires_action state rather than relying on a Stripe query, but that's a bigger effort

frankie567 (approved):

PR #13273, opened 2026-07-20
feat(sso): validate discovery on save, extra authorization parameters, provider presets

maximevast:
## Summary

Three refinements to organization SSO.

- **Validate on save.** `create`, a configuration update, and enabling a connection now fetch `{issuer}/.well-known/openid-configuration` and reject a config that can't sign in — issuer mismatch, missing metadata, or `private_key_jwt` not advertised. 422 with the error pinned to the offending field. Renames and disables stay network-free. Pasting the discovery URL as the issuer is accepted.
- **Extra authorization parameters.** `configuration.authorization_parameters` is appended to the authorize URL, e.g. `hd` to pin a Google Workspace domain. Reserved keys (`state`, `redirect_uri`, `nonce`, …) are rejected: reauth merges extras last.
- **Provider presets.** Google Workspace preset fills the issuer, pins the auth method, and surfaces `hd` as a Workspace domain field. Both modals now attach validation errors to their inputs instead of a generic toast.

https://github.com/user-attachments/assets/ec00b784-d159-4aed-acae-f54576cfa59d

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/13273?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/tests/sso/test_endpoints.py
```diff
@@ -256,6 +291,267 @@
...
+                "configuration": {
+                    "issuer": "https://idp.example.com",
+                    "client_id": "client-id",
+                    "auth_method": "client_secret",
+                    "client_secret": "secret",
+                }
+            },
+        )
+        assert response.status_code == 422
+        assert "https://other.example.com" in response.json()["detail"][0]["msg"]
```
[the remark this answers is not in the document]
maximevast: On a test file ?

server/polar/sso/discovery.py
```diff
@@ -0,0 +1,112 @@
+from typing import Any
```
frankie567: Isn't a bit too much? I mean, if they configure a wrong discovery document, they'll quickly notice it when testing the connection. Given the amount of code involved, I'm not sure it's worth the effort

server/polar/sso/schemas.py
```diff
@@ -14,14 +21,62 @@
...
+    authorization_parameters: AuthorizationParameters = Field(
+        default_factory=dict, description=AUTHORIZATION_PARAMETERS_DESCRIPTION
+    )
+
+    @field_validator("authorization_parameters")
+    @classmethod
+    def validate_authorization_parameters(cls, value: dict[str, str]) -> dict[str, str]:
+        for key in value:
+            if key in RESERVED_AUTHORIZATION_PARAMETERS:
+                raise ValueError(f"`{key}` is set by Polar and can't be overridden.")
```
frankie567: This could be set as an AfterValidator on the key type of AuthorizationParameters

server/polar/sso/schemas.py
```diff
@@ -53,6 +108,9 @@ class OIDCConfigurationRead(Schema):
     auth_method: OIDCAuthMethod = Field(
         description="Authentication method used against the identity provider."
     )
+    authorization_parameters: dict[str, str] = Field(
+        default_factory=dict, description=AUTHORIZATION_PARAMETERS_DESCRIPTION
```
frankie567: No default on an output schema

maximevast: we need a backfill then? 😔

server/polar/sso/discovery.py
maximevast: I understand your point. I'll drop it

server/polar/sso/schemas.py
frankie567: Yes, but since the table is tiny, it can be a simple Alembic SQL migration in the same PR :)

frankie567 (approved):
Nice! Extra mile always :)

PR #13288, opened 2026-07-21
checkout: require full billing address when UPI is selected

strandhvilliam:
## Summary

**Related Issue**: #13280 and two Plain issues inside where customers recorded being blocked in checkout.

This PR makes checkout require a full billing address when the customer selects UPI as their payment method.

The alternative may be to enforce billing address requirement for India as whole on backend in `checkout.py`, similar as for US. But this would also make the normal card payment in India require billing address instead of only UPI. Therefore the change is in client checkout form, which also is where the payment method state lives so it kinda makes sense.

If we want to enforce billing address on all checkouts in india, this could be a small fix in backend instead.

The major con is that this somewhat diludes the state for billing_address visibility to not only be on the checkout in backend, but also handled by frontend too. Tests added to make sure it never loosens the requirements though.

## What

When UPI is selected, checkout now treats `line1`, `city`, and `postal_code` as required (and `line2` as optional), overriding the default `billing_address_fields` from the checkout session. Other payment methods are unchanged.

## Why

UPI needs a complete billing address for payment collection. Without forcing those fields, customers can submit with an incomplete address and fail at confirmation.

## How

- Added a `FULL_BILLING_ADDRESS_PAYMENT_METHODS` list (`upi`) in `CheckoutForm`
- Derived `requireFullBillingAddress` from the selected Stripe payment method (same pattern as wallet methods)
- Passed it into `BaseCheckoutForm`, which merges required address fields when set and drives field visibility/validation from that merged config

## Recording
https://github.com/user-attachments/assets/315a936b-ed56-43a8-9b5d-26127b2d0ea0

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/13288?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/packages/checkout/src/components/CheckoutForm.tsx
+++ b/clients/packages/checkout/src/components/CheckoutForm.tsx
@@ -45,6 +45,7 @@ import PolarLogo from './PolarLogo'
 import { CheckoutBanner } from './CheckoutBanner'
 
 const WALLET_PAYMENT_METHODS = ['apple_pay', 'google_pay', 'link']
+const FULL_BILLING_ADDRESS_PAYMENT_METHODS = ['upi']
 
 const XIcon = ({ className }: { className?: string }) => {
   return (
@@ -81,6 +82,7 @@ interface BaseCheckoutFormProps {
   isUpdatePending?: boolean
   locale?: AcceptedLocale
   isWalletPayment?: boolean
+  requireFullBillingAddress?: boolean
   beforeSubmit?: React.ReactNode
   embed?: boolean
 }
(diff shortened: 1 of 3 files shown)


frankie567:
I'm not comfortable with adding some frontend logic that overrides what the backend says about billing fields visibility. 

I think we should start tracking the `payment_method` on the Checkout model. The client could then keep it sync with the API when the customer selects another method, and we can then apply logic to control the billing address fields.

strandhvilliam:
I suspected so @frankie567, closing this. Sounds like a better way

PR #13296, opened 2026-07-21
add payment_method column to checkouts

strandhvilliam:
## Summary

Migration + model PR *(1/3)* for tracking the payment method selected by the
customer on the checkout form.

Better version than initial bad version: https://github.com/polarsource/polar/pull/13288

## What

- New nullable `payment_method` column on `checkouts` (Stripe payment method
  type string, e.g. `card`, `apple_pay`, `upi`), with migration.
- `FULL_BILLING_ADDRESS_PAYMENT_METHODS` (currently `{"upi"}`): payment
  methods that require a full billing address regardless of country.
- New `Checkout.is_full_billing_address_required` property combining the
  condition previously duplicated inline

## Why

Whether the billing address form is required currently depends on country
and merchant config only. Payment processors require a full billing address
for specific payment methods, so the checkout needs to know which method is
selected to show (and later enforce) the right fields.

## How

Column is nullable with no default and nothing writes it yet, so behavior is
provably unchanged (`payment_method IS NULL` for every row → the new
condition is always false).

## Checklist

- [x] The diff is reasonably sized and easy to re
- [x] New functionality is covered by tests — behavior is inert here; the
      update/confirm tests exercising it land in PR 2 (all 425 existing
      checkout tests pass on this branch)
- [x] Linting and type checking pass (`uv run task_types`)

<a href="https://cubic.dev/pr/polarsource/polar/pull/13296?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/models/checkout.py
+++ b/server/polar/models/checkout.py
@@ -391,7 +391,7 @@ class Checkout(
     ] = association_proxy("product", "attached_custom_fields")
 
     @property
-    def is_full_billing_address_required(self) -> bool:
+    def is_billing_address_required(self) -> bool:
         return (
             self.require_billing_address
             or self.is_business_customer
@@ -403,7 +403,7 @@ class Checkout(
         address = self.customer_billing_address
         country = address.country if address else None
         is_us = country == "US"
-        require_billing_address = self.is_full_billing_address_required or is_us
+        require_billing_address = self.is_billing_address_required or is_us
         return {
             "country": True,
             "state": country in {"US", "CA"},
@@ -418,7 +418,7 @@ class Checkout(
         address = self.customer_billing_address
         country = address.country if address else None
         is_us = country == "US"
-        require_billing_address = self.is_full_billing_address_required or is_us
+        require_billing_address = self.is_billing_address_required or is_us
         return {
             "country": BillingAddressFieldMode.required,
             "state": BillingAddressFieldMode.required
--- a/server/polar/models/checkout.py
+++ b/server/polar/models/checkout.py
@@ -121,7 +121,7 @@ class Checkout(
     payment_processor_metadata: Mapped[dict[str, Any]] = mapped_column(
         JSONB, nullable=False, default=dict
     )
-    payment_method: Mapped[str | None] = mapped_column(
+    payment_method_type: Mapped[str | None] = mapped_column(
         String, nullable=True, default=None
     )
     return_url: Mapped[str | None] = mapped_column(String, nullable=True, default=None)
@@ -395,7 +395,7 @@ class Checkout(
         return (
             self.require_billing_address
             or self.is_business_customer
-            or self.payment_method in FULL_BILLING_ADDRESS_PAYMENT_METHODS
+            or self.payment_method_type in FULL_BILLING_ADDRESS_PAYMENT_METHODS
         )
 
     @property
--- a/server/polar/models/checkout.py
+++ b/server/polar/models/checkout.py
@@ -96,6 +96,11 @@ class CheckoutAnalyticsMetadata(TypedDict, total=False):
     distinct_id: str | None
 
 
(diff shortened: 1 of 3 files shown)


server/polar/models/checkout.py
```diff
@@ -382,14 +390,20 @@ def customer_session_token(self, value: str | None) -> None:
         Sequence["AttachedCustomFieldMixin"] | None
     ] = association_proxy("product", "attached_custom_fields")
 
+    @property
+    def is_full_billing_address_required(self) -> bool:
```
strandhvilliam: I'm wondering about this one and if this is the right pattern. Reason added is to reduce duplicate condition needed to be written to check the multiple reasons for a full address, but maybe this is the wrong layer to add this? It is used in this file though so maybe it makes sense.

server/polar/models/checkout.py
```diff
@@ -96,6 +96,11 @@ class CheckoutAnalyticsMetadata(TypedDict, total=False):
     distinct_id: str | None
 
 
+# Payment methods for which the processor requires a full billing address
+# to process the payment, regardless of the customer's country.
+FULL_BILLING_ADDRESS_PAYMENT_METHODS: frozenset[str] = frozenset({"upi"})
```
strandhvilliam, line 101: Is this the correct domain? `payments.py` maybe? Though probably makes sense having it here since it shouldn't be mixed up with the actual payment_methods model and is only used in checkout for now.

server/polar/models/checkout.py
```diff
@@ -116,6 +121,9 @@ class Checkout(
     payment_processor_metadata: Mapped[dict[str, Any]] = mapped_column(
         JSONB, nullable=False, default=dict
     )
+    payment_method: Mapped[str | None] = mapped_column(
```
strandhvilliam: Ok name? Just a string which will be whatever stripe names it, but just making sure so it doesn't look like it's directly related to the actual payments_method model as it obviously isn't a relation and shouldn't be.

server/polar/models/checkout.py
frankie567: I think it's fine for now, if we need more complex or shared logic, we'll always be able to move to another module.

server/polar/models/checkout.py
frankie567: Yes, fair point. Let's say `payment_method_type` to be extra clear

server/polar/models/checkout.py
frankie567: Hmm... I don't think we need that: it should be contained in `billing_address_fields`. The reason it's duplicated is because of the legacy `customer_billing_address_fields`, but it should go away at some point, so I'm fine if we duplicate it for now.

frankie567: Following https://github.com/polarsource/polar/pull/13297#discussion_r3628032762

Let's then keep it but name it `is_billing_address_required` (no "full").

server/polar/models/checkout.py
strandhvilliam: Renamed in [bd591de](https://github.com/polarsource/polar/pull/13296/commits/bd591de9fe534025d66aac457495e8cf811f8a1a) 🫡

server/polar/models/checkout.py
strandhvilliam: Alright :P , renamed it in [7c448c3](https://github.com/polarsource/polar/pull/13296/commits/7c448c3798150a6df951c12d1654677d1c450504)

frankie567 (approved):

PR #13297, opened 2026-07-21
accept and enforce payment_method on checkout API

strandhvilliam:
## Summary

PR *(2/3)* (based on #13296) for tracking the payment method selected by the customer on the checkout form. Wires the `payment_method` column into the API: the checkout form can now sync the customer's selected payment method, `billing_address_fields` reacts to it, and confirm-time validation enforces it.

## What

- `payment_method_type` accepted on checkout updates (`CheckoutUpdatePublicBase`, so
  both the client-secret and authenticated endpoints)
- `payment_method_type` exposed on the checkout read schemas.
- `payment_method_type` is reset whenever the payment form is no longer required
  (e.g. a 100% discount), mirroring the existing `is_business_customer`
  reset.
- Tests: update round-trip via the public endpoint, `billing_address_fields`
  flipping with the selection, confirm-time 422 for UPI without a full
  address, and the discount-reset behavior.

## Why

Some payment methods require a full billing address regardless of country.
The server needs to know the current selection to render the right form
fields and to reject confirmation when required address data is missing,
today this is only enforced by country/merchant config.

## How

The field flows through the existing generic update path (no special-case
service code for persistence). Enforcement and display share a single
source (`is_full_billing_address_required`), so they cannot drift.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/13297?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/checkout/schemas.py
+++ b/server/polar/checkout/schemas.py
@@ -40,6 +40,7 @@ from polar.kit.metadata import (
     MetadataOutputMixin,
 )
 from polar.kit.schemas import (
+    EmptyStrToNone,
     EmptyStrToNoneValidator,
     HttpUrlToStr,
     IDSchema,
@@ -123,6 +124,20 @@ EmbedOrigin = Annotated[
             "set this to the Origin of the embedding page. "
             "It'll allow the Polar iframe to communicate with the parent page."
         ),
+    ),
+]
+
+_payment_method_description = (
+    "Payment method type selected by the customer in the checkout form, "
+    "e.g. `card`, `apple_pay` or `upi`. "
+    "Some payment methods require a full billing address: keeping this value "
+    "in sync allows `billing_address_fields` to reflect the fields to display."
+)
+PaymentMethodInput = Annotated[
+    EmptyStrToNone,
+    Field(
+        description=_payment_method_description,
+        serialization_alias="payment_method_type",
     ),
 ]
 
@@ -375,6 +390,7 @@ class CheckoutUpdateBase(CustomFieldDataInputMixin, Schema):
     customer_billing_address: CustomerBillingAddressInput | None = None
     customer_tax_id: Annotated[str | None, EmptyStrToNoneValidator] = None
     locale: Locale | None = None
+    payment_method: PaymentMethodInput = None
 
 
 class CheckoutUpdate(
@@ -595,6 +611,10 @@ class CheckoutBase(CustomFieldDataOutputMixin, TimestampedSchema, IDSchema):
         validation_alias=AliasChoices("customer_tax_id_number", "customer_tax_id")
     )
     locale: str | None = None
+    payment_method: str | None = Field(
+        validation_alias=AliasChoices("payment_method_type", "payment_method"),
+        description=_payment_method_description,
+    )
 
     payment_processor_metadata: dict[str, str]
 
--- a/server/polar/checkout/schemas.py
+++ b/server/polar/checkout/schemas.py
@@ -385,7 +385,6 @@ class CheckoutUpdateBase(CustomFieldDataInputMixin, Schema):
     customer_billing_address: CustomerBillingAddressInput | None = None
     customer_tax_id: Annotated[str | None, EmptyStrToNoneValidator] = None
     locale: Locale | None = None
-    payment_method_type: PaymentMethodTypeInput = None
 
 
 class CheckoutUpdate(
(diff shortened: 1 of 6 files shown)


server/polar/checkout/schemas.py
```diff
@@ -126,6 +127,19 @@
     ),
 ]
 
+_payment_method_description = (
+    "Payment method type selected by the customer in the checkout form, "
+    "e.g. `card`, `apple_pay` or `upi`. "
+    "Some payment methods require a full billing address: keeping this value "
+    "in sync allows `billing_address_fields` to reflect the fields to display."
+)
+PaymentMethodInput = Annotated[
```
strandhvilliam: Been thinking back and forth on this, looking at existing patterns and the Stripe spec. Seems like stripe payment methods will always be lowercase strings? Don't want this to break on bad validators so open if a better way 😅

server/polar/checkout/schemas.py
```diff
@@ -375,6 +389,7 @@ class CheckoutUpdateBase(CustomFieldDataInputMixin, Schema):
     customer_billing_address: CustomerBillingAddressInput | None = None
     customer_tax_id: Annotated[str | None, EmptyStrToNoneValidator] = None
     locale: Locale | None = None
+    payment_method: PaymentMethodInput | None = None
```
strandhvilliam: I put it on the CheckoutUpdateBase since I assume it will always be needed, but maybe we want it on the CheckoutUpdatePublic? Seems to be the one used in the useCheckout hook. 🤔

server/polar/checkout/schemas.py
```diff
@@ -126,6 +127,19 @@
     ),
 ]
 
+_payment_method_description = (
+    "Payment method type selected by the customer in the checkout form, "
+    "e.g. `card`, `apple_pay` or `upi`. "
+    "Some payment methods require a full billing address: keeping this value "
+    "in sync allows `billing_address_fields` to reflect the fields to display."
```
frankie567: This is a user-facing description, so the second sentence should go away as it's an implementation detail.

server/polar/checkout/schemas.py
frankie567: Honestly, I would even keep it more wide open and just annotate it `EmptyStrOrNone`, with no specific constraints. We'll never know what Stripe or other payment processor will send us. And we don't really care at this point.

server/polar/checkout/schemas.py
frankie567: Let's put it on `CheckoutUpdatePublic`. In the current state, it'll be only something set by the client. Setting it programmatically from the API would be somewhat undefined behavior.

server/polar/checkout/service.py
```diff
@@ -1075,7 +1074,7 @@ async def _confirm_inner(
                     }
                 )
 
-        if checkout.require_billing_address or checkout.is_business_customer:
+        if checkout.is_full_billing_address_required:
```
frankie567: Ah damn 😅 Let me revisit the comment in the previous PR :p

frankie567: https://github.com/polarsource/polar/pull/13296/changes#r3628038474

server/polar/checkout/schemas.py
strandhvilliam: Of course 🤦‍♂️ Fixed in later commit!

server/polar/checkout/schemas.py
strandhvilliam: Makes sense! Fixed 🥳

server/polar/checkout/schemas.py
strandhvilliam: Alright, makes sense. Moved to `CheckoutUpdatePublic` in later commit.

frankie567:
You should rebase because right now it contains the migration and model changes

strandhvilliam:
Yep already on it. Fixed.

strandhvilliam:
@frankie567 any chance to review this? 🙇😀

frankie567 (approved):
👏

PR #13298, opened 2026-07-21
sync selected payment method to the checkout API

strandhvilliam:
## Summary

PR *(3/3)* for tracking the payment method selected by the customer on the checkout form (based on
#13297. The checkout form keeps the customer's selected payment method in sync with the API so the billing address fields react to the selection, so selecting UPI reveals the full address form, switching back to card hides it.

## What

- `CheckoutForm` PATCHes `{ payment_method }` from the `PaymentElement`
  change event whenever the selection differs from the checkout's stored
  value. The has recomputed `billing_address_fields` from serverside, so the
  address inputs show/hide as the customer switches methods.

## Why

The server-side rules from PRs 1–2 are not used until the client reports the
selection. This closes the loop: select UPI → full address fields render →
confirm enforces them.

## How

We listen to the Stripe PaymentElement's change event and send the selected
payment method to the checkout API. The updated checkout comes back with
recomputed `billing_address_fields`, and the form re-renders the address
inputs from it. On submit, we include the selection in the confirm call so
the server validates against what the customer actually chose.

## Checklist

- [x] New functionality is covered by tests — the PaymentElement is mocked
      in vitest, so the sync path is exercised end-to-end by the stack's
      server tests (update round-trip, field flipping, confirm enforcement)

<a href="https://cubic.dev/pr/polarsource/polar/pull/13298?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/packages/checkout/src/components/CheckoutForm.tsx
+++ b/clients/packages/checkout/src/components/CheckoutForm.tsx
@@ -861,16 +861,7 @@ const StripeCheckoutForm = (props: CheckoutFormProps) => {
           <BaseCheckoutForm
             {...props}
             checkout={checkout}
-            confirm={(data) =>
-              confirm(
-                // Not in form state, comes from Stripe PaymentElement.
-                selectedPaymentMethod
-                  ? { ...data, payment_method_type: selectedPaymentMethod }
-                  : data,
-                stripe,
-                elements,
-              )
-            }
+            confirm={(data) => confirm(data, stripe, elements)}
             isWalletPayment={isWalletPayment}
           >
             {checkout.is_payment_form_required && (
--- a/clients/packages/checkout/src/components/CheckoutForm.tsx
+++ b/clients/packages/checkout/src/components/CheckoutForm.tsx
@@ -795,15 +795,15 @@ const StripeCheckoutForm = (props: CheckoutFormProps) => {
   const onPaymentElementChange = useCallback(
     async (event: StripePaymentElementChangeEvent) => {
       setSelectedPaymentMethod(event.value.type)
-      if (event.value.type !== checkout.payment_method) {
+      if (event.value.type !== checkout.payment_method_type) {
         try {
-          await update({ payment_method: event.value.type })
+          await update({ payment_method_type: event.value.type })
         } catch {
           /* API errors handled by provider */
         }
       }
     },
-    [checkout.payment_method, update],
+    [checkout.payment_method_type, update],
   )
 
   const elementsOptions = useMemo<StripeElementsOptions>(() => {
@@ -865,7 +865,7 @@ const StripeCheckoutForm = (props: CheckoutFormProps) => {
               confirm(
                 // Not in form state, comes from Stripe PaymentElement.
                 selectedPaymentMethod
-                  ? { ...data, payment_method: selectedPaymentMethod }
+                  ? { ...data, payment_method_type: selectedPaymentMethod }
                   : data,
                 stripe,
                 elements,
--- a/clients/packages/checkout/src/components/CheckoutForm.tsx
+++ b/clients/packages/checkout/src/components/CheckoutForm.tsx
@@ -767,7 +767,13 @@ interface CheckoutFormProps {
 }
 
 const StripeCheckoutForm = (props: CheckoutFormProps) => {
-  const { checkout, confirm, themePreset: themePresetProps, locale } = props
+  const {
+    checkout,
+    confirm,
+    update,
+    themePreset: themePresetProps,
+    locale,
+  } = props
   const {
     payment_processor_metadata: { publishable_key },
   } = checkout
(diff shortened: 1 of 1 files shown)


clients/packages/checkout/src/components/CheckoutForm.tsx
```diff
@@ -841,7 +861,16 @@ const StripeCheckoutForm = (props: CheckoutFormProps) => {
           <BaseCheckoutForm
             {...props}
             checkout={checkout}
-            confirm={(data) => confirm(data, stripe, elements)}
+            confirm={(data) =>
+              confirm(
+                // Not in form state, comes from Stripe PaymentElement.
```
strandhvilliam: Bit unsure about this one. If I'm correct the data is the react hook form values which the payment method is not a part of client side (Stripe element prop). Now we just use update to sync to api and here we inject it when confirming to get the fresh state in final request. But maybe we should sync it with RHF even though it's a stripe specific field? Would be more plumbing though. What do you think?

clients/packages/checkout/src/components/CheckoutForm.tsx
```diff
@@ -786,6 +792,20 @@ const StripeCheckoutForm = (props: CheckoutFormProps) => {
     ? WALLET_PAYMENT_METHODS.includes(selectedPaymentMethod)
     : false
 
+  const onPaymentElementChange = useCallback(
+    async (event: StripePaymentElementChangeEvent) => {
+      setSelectedPaymentMethod(event.value.type)
```
strandhvilliam, line 797: I fear it could be edge cases where the api is slow/failing and the Stripe elements show a payment method that is not synced to the backend and they are out of sync. Not 100% if this should be something to handle or which approach in that case. If the api is failing, the checkout probably would have other errors too. Would like a second opinion if this is a case to handle or not 🤔

frankie567: Honestly, we could have this problem for a lot of stuff in the checkout: discount, tax id, product selection, etc... I think it's fine.

clients/packages/checkout/src/components/CheckoutForm.tsx
frankie567: But at this stage, the payment method is already synced, no? There is no way for the customer to both change the payment method and submit the checkout?

clients/packages/checkout/src/components/CheckoutForm.tsx
```diff
@@ -871,9 +900,7 @@ const StripeCheckoutForm = (props: CheckoutFormProps) => {
                     usBankAccount: 'never',
                   },
                 }}
-                onChange={(event: StripePaymentElementChangeEvent) => {
-                  setSelectedPaymentMethod(event.value.type)
-                }}
+                onChange={onPaymentElementChange}
```
frankie567, line 894: I guess Stripe Elements doesn't support controlled state? i.e. allow us to programmatically set the payment method based on the checkout data?

strandhvilliam: This would be best, but couldn't find something like this after some research, only tracking it's state 😢

clients/packages/checkout/src/components/CheckoutForm.tsx
strandhvilliam: Alright!

clients/packages/checkout/src/components/CheckoutForm.tsx
strandhvilliam: lol, think I overanalysed this 😅 was in the 'what-if-client-is-out-of-sync-from-api' and wanted to guard more for it. But in that case this wouldn't matter anyway. Simplified it in newer commit 🚀

clients/packages/checkout/src/components/CheckoutForm.tsx
```diff
@@ -786,6 +792,20 @@ const StripeCheckoutForm = (props: CheckoutFormProps) => {
     ? WALLET_PAYMENT_METHODS.includes(selectedPaymentMethod)
     : false
 
+  const onPaymentElementChange = useCallback(
+    async (event: StripePaymentElementChangeEvent) => {
+      setSelectedPaymentMethod(event.value.type)
+      if (event.value.type !== checkout.payment_method_type) {
+        try {
+          await update({ payment_method_type: event.value.type })
```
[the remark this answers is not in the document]
strandhvilliam: May be valid, but this may be the case for multiple things in the checkout (see: https://github.com/polarsource/polar/pull/13298#discussion_r3628051156), and is fine for now.

frankie567 (approved):

PR #11993, opened 2026-05-29
Add per-customer discount redemption limits

pieterbeulque:
## Summary

Adds support for limiting the number of times a discount can be redeemed by a single customer, with customer identification using email, account ID, and payment method fingerprint (similar to trial abuse detection).

## What

- Added `max_redemptions_per_customer` field to the `Discount` model
- Implemented `check_per_customer_limit_reached()` service method to enforce per-customer limits
- Added `count_redemptions_by_customer()` repository method to count past redemptions by customer identity signals
- Integrated limit enforcement into checkout confirmation flow before payment intent creation
- Updated API schemas, OpenAPI documentation, and TypeScript client types
- Added comprehensive test coverage for limit checking with various customer identification scenarios

## Why

Prevents discount abuse by allowing merchants to limit how many times each customer can redeem a discount. Uses multiple identity signals (customer ID, email with alias stripping, payment fingerprint) to catch attempts to circumvent limits using different accounts or email aliases.

## How

1. **Database**: Added nullable `max_redemptions_per_customer` integer column to discounts table
2. **Repository**: Implemented `count_redemptions_by_customer()` with SQL logic to:
   - Match customers by ID, unaliased email (strips `+alias` suffixes), or card fingerprint
   - Exclude the current in-progress checkout from counts
   - Use OR logic across identity signals (any match counts as same customer)
3. **Service**: Added `check_per_customer_limit_reached()` that queries the repository and compares count against limit
4. **Checkout**: Integrated limit check in `_confirm_inner()` before creating payment intent, extracting fingerprint from Stripe confirmation token
5. **API**: Added field to create/update schemas with validation (minimum 1 when set)
6. **Tests**: Added 6 test cases covering no limit, under limit, limit reached by ID/email/fingerprint, and current checkout exclusion

## Checklist

- [x] This PR addresses a single concern (per-customer discount limits)
- [x] The diff is reasonably sized and focused
- [x] New functionality is covered by tests (6 new test cases in `TestCheckPerCustomerLimitReached`)
- [x] Linting and type checking should pass
- [x] No unrelated changes included

https://claude.ai/code/session_01FXLr4zCCefQQgQSmUf63eH

PR #13328, opened 2026-07-22
feat(discount): per-customer redemption limits

maximevast:
## Summary

Adds `max_redemptions_per_customer`, so a code can be shared publicly with unlimited total uses but capped per buyer.

⚠️  Dashboard field and docs follow in the next PR; until then the limit is API-only.

Closes polarsource/feedback#86.

Builds on @pieterbeulque's #11993 — the schemas, service logic, error class and most tests come from there.

Identity reuses the trial-abuse signals: customer ID, unaliased email, card fingerprint, OR-matched. The check runs before the Stripe intent is created, so a rejected buyer is never charged. Applying an exhausted code is rejected at apply time too, rather than after the payment form.

Redemptions from a discount applied to an existing subscription count toward the limit, but merchants applying one are not blocked.

Column shipped separately in #13322.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/13328?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -19,6 +19,7 @@ from polar.authz.service import assert_resource_permission, get_accessible_org_i
 from polar.checkout.guard import has_product_checkout
 from polar.checkout.schemas import (
     CheckoutConfirm,
+    CheckoutConfirmBase,
     CheckoutConfirmStripe,
     CheckoutCreate,
     CheckoutPriceCreate,
@@ -2284,6 +2285,28 @@ class CheckoutService:
         checkout = await self._update_trial_end(checkout)
 
         session.add(checkout)
+
+        # Tell the buyer as soon as they apply the code, instead of letting them fill in
+        # the payment form first. Confirmation re-checks with the card fingerprint, which
+        # isn't available yet, so it stays the authoritative gate.
+        if (
+            isinstance(checkout_update, CheckoutUpdatePublic)
+            and not isinstance(checkout_update, CheckoutConfirmBase)
+            and checkout.discount is not None
+            and await discount_service.check_per_customer_limit_reached(
+                session, checkout.discount, checkout=checkout
+            )
+        ):
+            raise PolarRequestValidationError(
+                [
+                    {
+                        "type": "value_error",
+                        "loc": ("body", "discount_code"),
+                        "msg": "You have already redeemed this discount.",
+                        "input": checkout_update.discount_code,
+                    }
+                ]
+            )
 
         await self._validate_subscription_uniqueness(session, checkout)
 
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -1035,8 +1035,7 @@ class CheckoutService:
             if checkout.is_payment_setup_required:
                 raise PaymentNotReady()
 
-        # For wallet payments (Apple Pay, Google Pay, Link), we hide the customer name
-        # field for better UX and instead extract the name from Stripe's confirmation token.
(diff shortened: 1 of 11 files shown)


server/polar/checkout/service.py
```diff
@@ -2240,6 +2286,28 @@ async def _update_checkout(
...
+        # isn't available yet, so it stays the authoritative gate.
+        if (
+            isinstance(checkout_update, CheckoutUpdatePublic)
+            and not isinstance(checkout_update, CheckoutConfirmBase)
+            and checkout.discount is not None
+            and await discount_service.check_per_customer_limit_reached(
+                session, checkout.discount, checkout=checkout
+            )
+        ):
+            raise PolarRequestValidationError(
```
pieterbeulque: Why does this not raise `DiscountRedemptionLimitReached`?

server/polar/discount/service.py
pieterbeulque: Unrelated to your changes, but couldn't this introduce issues when two people use the same code at the same time, even if the discount's `max_redemptions` or `max_redemptions_per_customer` are none and we don't care about concurrent usage of the discount code?

server/polar/discount/repository.py
```diff
@@ -45,3 +60,78 @@ async def set_subscription_by_checkout(
...
+                # NULL-safe: subscription-only redemptions have no checkout.
+                DiscountRedemption.checkout_id.is_distinct_from(exclude_checkout_id),
+                or_(*clauses),
+            )
+        )
+
+        if payment_method_fingerprint is not None:
+            statement = statement.join(
+                Payment,
+                Payment.checkout_id == DiscountRedemption.checkout_id,
```
pieterbeulque: Do we need to check for successful payment status to not consider a failed payment a discount redemption?

Similarly, what do we do if a discount code was used on a 100% refunded order? Can they use it again?

maximevast: > what do we do if a discount code was used on a 100% refunded order? Can they use it again?

no but it pre existing right ?

server/polar/checkout/service.py
maximevast: True, fixing this

server/polar/discount/repository.py
maximevast: > Do we need to check for successful payment status to not consider a failed payment a discount redemption?

Yes!

server/polar/discount/service.py
maximevast: I'll fix that

server/polar/discount/repository.py
pieterbeulque: >no but it pre existing right ?

The max redemptions per customer is new, so I'm not sure what you mean with pre-existing?

I guess previously, you only couldn't redeem the discount code on your second purchase (after refunding the first) if purchase A was before the discount's max redemptions was reached but purchase B was after the discount's max redemptions is reached. (holy shit, what a sentence, sorry)

So max redemptions _per customer_ increases the possibility of bumping into it quite a bit.

maximevast: Yes true, it's more likely now. Let me look at how we can handle it

maximevast:
Closing in favor of: https://github.com/polarsource/polar/pull/13394

PR #13342, opened 2026-07-23
server/order: always clear payment lock on payment failure

frankie567:
In the current state, `handle_failure` is short-circuited if the Order is already paid, not touching at the payment lock. However, we have in the wild some Orders that are both paid and have a payment lock, which is a state that should not exist. This PR makes sure to always clear the payment lock on payment failure, even if the Order is already paid.

<a href="https://cubic.dev/pr/polarsource/polar/pull/13342?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/order/service.py
+++ b/server/polar/order/service.py
@@ -2800,14 +2800,6 @@ class OrderService:
             skip_dunning: If True, skip dunning progression (e.g., for manual retries
                 where the user can try again without advancing the dunning state machine)
         """
-        # Don't process payment failure if the order is already paid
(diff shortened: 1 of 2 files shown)


maximevast (approved):
Great!

PR #13351, opened 2026-07-23
chore(rfc): merging customer

maximevast:
## Why

A buyer purchases with their work email, then asks the merchant to move it to their
personal address. That address already exists as a customer in the same organization,
so the merchant gets `A customer with this email address already exists.` and stops.

Their only workaround is a 100% discount code and a second purchase. That leaves a
duplicate order, inflated order counts, and the original purchase stranded on an address
the buyer has abandoned. Support can't help either — backoffice `edit_email` fails on
exactly this case.

We have avoided merging customers for a long time. This RFC argues that most of the
danger sits in cases we can refuse rather than solve (as we do for migration), and proposes the refusals.

## What to review

Two decisions carry the rest:

- **The survivor is derived, not chosen.** Whichever customer holds the billable
  subscription survives, so no live billing relationship is ever rewritten.
- **The refusals.** Both subscribed, both carrying an `external_id`, team customers,
  seat holders.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/13351?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/handbook/docs.json
+++ b/handbook/docs.json
@@ -90,7 +90,8 @@
               "engineering/design-documents/api-versioning",
               "engineering/design-documents/payout-reviews",
               "engineering/design-documents/secrets-encryption",
-              "engineering/design-documents/merchant-migrations"
+              "engineering/design-documents/merchant-migrations",
+              "engineering/design-documents/customer-merge"
             ]
           },
           {
--- a/handbook/engineering/design-documents/customer-merge.mdx
+++ b/handbook/engineering/design-documents/customer-merge.mdx
@@ -1,0 +1,288 @@
+import { OpenQuestion } from "/snippets/open-question.jsx";
+
+<Info>
+  **Status**: Draft
+
+**Created**: July 2026
+
+**Last Updated**: July 23rd, 2026
+
+</Info>
+
+## Briefing
+
+A buyer purchases with their work email. Later they ask the merchant to move the purchase to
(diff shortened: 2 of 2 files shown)


pieterbeulque (commented):
It's not there yet, I think.

First,

>This is the load-bearing decision, and everything else follows from it.
>
>The merchant says which email to keep. They do not pick which customer row survives. We derive it: whichever customer holds the billable subscription survives.

What if the merchant wants control over what customer survives because they rely on the customer A's ID or other customer metadata on customer A?

Second, I also think not doing team type customers is too limiting since the member model is enabled for almost all organizations, so we'll have to consider that.

Third, I think rebuilding every meter from the event stream makes sense, but it feels too important to be a footnote in this RFC. It'd be good to think through the edge cases a bit more (e.g. A & B are on different cycles, or would this not be allowed to merge).

Fourth, what happens with the invoice history? Are any invoices rewritten (I don't think they can be from a legal POV). Are there other accounting or "immutable" artifacts we have to consider that assume a customer doesn't change?

PR #13353, opened 2026-07-23
chore(rfc): multi products checkout

maximevast:
## Summary

feedback#110 asks for add-ons: a base product plus extras, bought together.

That looks expensive. A checkout and a subscription each point at one product today, so selling several at once looks like a change to the billing engine.

It isn't. A renewal never asks which product a subscription is for. It reads the subscription's price list, writes one billing entry per price, turns each entry into an order item, and sums them. Every step keys on product_price_id. product_id appears in no filter, no join condition and no group-by along that path. The query selects it with the row and never reads it.

So a subscription holding prices from two products would already bill both onto one invoice. Nothing in the engine depends on those prices coming from one product. That limit sits at the entry point: checkout lets you pick one product, and create_or_update_from_checkout fills the list from checkout.product alone.

That changes what this feature costs.

The doc names two silent failures. A free base with a paid add-on skips payment, because four Checkout properties resolve to the base product alone. And removing an add-on can revoke the base product's benefits, because list_outdated_grants takes one product and revokes everything not on it.

One decision before the first migration: whether the new quantity columns are nullable. Every existing row has no quantity. The alternative is a three-PR backfill across the highest-volume tables.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/13353?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/handbook/engineering/design-documents/multi-product-checkout.mdx
+++ b/handbook/engineering/design-documents/multi-product-checkout.mdx
@@ -249,6 +249,11 @@ The last rule follows from `Subscription.meter_interval`, which is snapshotted f
 creation. An add-on metered on a different cadence has nowhere to store it, so validation rejects it
 rather than silently rebasing it.
 
+An add-on matching none of the bases on the link is a mistake, and creation returns a 422. Say the link
+offers a monthly base and the merchant lists `<analytics_yearly>` beside it. No buyer can ever see that
+add-on: there is one base, it is monthly, and the intervals never match. Better to reject the link than
+to hand back one that quietly drops an add-on the merchant configured.
+
 ## API
 
 ### Merchant: offering the menu
@@ -261,11 +266,27 @@ POST /v1/checkout-links
 {
   "products": ["<base_monthly>", "<base_yearly>"],
   "addons": [
-    { "product_id": "<analytics>", "max_quantity": 1 },
-    { "product_id": "<extra_credits>", "min_quantity": 0, "max_quantity": 10 }
+    { "product_id": "<analytics_monthly>", "max_quantity": 1 },
+    { "product_id": "<analytics_yearly>", "max_quantity": 1 },
+    { "product_id": "<extra_credits_monthly>", "min_quantity": 0, "max_quantity": 10 },
+    { "product_id": "<extra_credits_yearly>", "min_quantity": 0, "max_quantity": 10 }
   ]
 }
 ```
+
+`addons` is one flat pool for the link, not a per-base mapping. A product carries a single interval
+(`Product.recurring_interval` and `recurring_interval_count`, `server/polar/models/product.py:73-81`),
+so monthly and yearly analytics are two products — exactly as the two bases are two products. The
+interval match **is** the mapping: the checkout offers only the add-ons eligible for the base the buyer
+has selected through `product_id` (`server/polar/checkout/schemas.py:356-361`), and the read schema
+returns that subset rather than the whole pool. Nothing declares which add-on belongs to which base, so
+no configuration can contradict the interval.
+
+Two costs come with that. Each add-on is duplicated per interval offered. And switching the base drops
+the selection instead of carrying it across: `<analytics_monthly>` and `<analytics_yearly>` are
+unrelated rows, and nothing in the model says they are one offer at two cadences. The buyer re-picks.
+Carrying the selection over needs a notion of product equivalence that does not exist today, and it is
+out of scope here.
 
 `PATCH /v1/checkout-links/{id}` takes `addons` as a full replacement set, matching how it already
 handles `products`. `POST /v1/checkouts` takes the same field. The dashboard's checkout-link editor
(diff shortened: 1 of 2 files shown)


handbook/engineering/design-documents/multi-product-checkout.mdx
```diff
@@ -0,0 +1,233 @@
+<Info>
```
maximevast, line 1: @frankie567 told me we have deprecated the endpoint using a single product id and now must pass an array which might substantially change this plan

pieterbeulque (commented):
I don't think this covers the full solution, I'm mostly missing in detail how merchants create add-ons (I assume they're considered regular products here), and then checkouts (session API + checkout links) to bundle base prices with add-ons. Merchants will want to choose whether an add-on is boolean (you buy it or you don't) or quantitative (you buy 0 → N), I guess, so we'd have to solve for that too. And so on, and so on.

It also feels very brittle to only have `subscription.product_id` be linked to the base product, and going through `subscription_product_prices` as a proxy. That feels error-prone. You mention there is deliberately no `subscription_products` table. Why?

The two unaddressed issues also feel important to address. Combining a free subscription with paid add-ons is a realistic use case. And an add-on for "extra meter credits" where the base subscription also carries a "base meter credits" feels likely, so stopping to pay for the extra credits shouldn't also revoke the base credits.

handbook/engineering/design-documents/multi-product-checkout.mdx
```diff
@@ -178,56 +251,111 @@ rather than silently rebasing it.
...
 ```http
-POST /v1/checkouts
+POST /v1/checkout-links
 {
-  "product_id": "...",
+  "products": ["<base_monthly>", "<base_yearly>"],
   "addons": [
-    { "product_id": "...", "quantity": 2 }
+    { "product_id": "<analytics>", "max_quantity": 1 },
+    { "product_id": "<extra_credits>", "min_quantity": 0, "max_quantity": 10 }
```
pieterbeulque: How does this work if the interval has to match? then `<analytics_yearly>` and `<analytics_monthly>` should exist and be mapped accordingly.

maximevast: Yes good catch

handbook/engineering/design-documents/multi-product-checkout.mdx
```diff
@@ -261,12 +266,28 @@ POST /v1/checkout-links
...
+interval match **is** the mapping: the checkout offers only the add-ons eligible for the base the buyer
+has selected through `product_id` (`server/polar/checkout/schemas.py:356-361`), and the read schema
+returns that subset rather than the whole pool. Nothing declares which add-on belongs to which base, so
+no configuration can contradict the interval.
+
+Two costs come with that. Each add-on is duplicated per interval offered. And switching the base drops
+the selection instead of carrying it across: `<analytics_monthly>` and `<analytics_yearly>` are
+unrelated rows, and nothing in the model says they are one offer at two cadences. The buyer re-picks.
+Carrying the selection over needs a notion of product equivalence that does not exist today, and it is
+out of scope here.
```
pieterbeulque, line 289: That does make me think… We're going to have people ask about a yearly base subscription with a monthly add-on subscription (extra meter credits, or sth), so we should account for that.

It's related to what @joebon has been doing.

pieterbeulque:
Hey @maximevast, I'd love for us to explore add-ons (and thus bundled checkouts) a bit more, but I don't think we can incrementally improve on this RFC. Preserving the single-product assumptions rather than modeling add-ons as a first-class concept is an interesting idea worth exploring, but from this RFC I think too many essentials are not covered.

Some things I'm missing top of mind:

 - There is no real catalog-level relationship between base products and add-on products, both in terms of product management (a merchant should be able to define what products are add-ons and can be combined with which other products) and from a customer perspective (not sure if needed)
 - The subscription model has conflicting sources of truth between `product_id`, `subscription_products` and `subscription_product_prices`, which feels fragile.
 - Post-purchase management isn't covered at all: managing add-ons (over API and from the Customer Portal), buying add-ons on an existing subscription (+ proration/cancellation/scheduled updates/…) is missing
 - Competing benefits between add-ons are not covered (do they add up? does one add-on "win"?)
 - I'm missing some confidence in the downstream impact of these changes, The single-product assumption is super common throughout (checkouts, orders, subscriptions, …) so how will this behave in something like a webhook payload, customer portal, … For example, analytics/orders/subscriptions can be filtered by product, how does that work if you select an add-on (can you select an add-on?)

I know polarsource/feedback#110 is pretty broad and I'd like for us to focus on (subscription) add-ons first, if that's a helpful limitation.

PR #13368, opened 2026-07-24
 feat(organization): custom purchase-access link for confirmation emails

strandhvilliam:
## Summary

Lets an organization point customer email links at its own billing page instead of the Polar customer portal, self-serve from organization settings, gated behind a feature flag.

Caveat: this also overrides the confirmation email's "Access purchase" link, which may be non-optimal UX depending on the merchant. Product specific urls might be preferred for this in the future.

## What

- New setting — customer_portal_settings.portal_url. Must be a valid https URL; supports {EMAIL}, {EXTERNAL_ID}, {ORDER_ID} and {SUBSCRIPTION_ID} placeholders, replaced per recipient (URL-encoded, empty when not applicable).
- New feature flag — feature_settings.portal_url_override_enabled (staff-managed). The URL is only used and editable while the flag is on.
- Applies to order and subscription lifecycle emails. Sign-in code and seat invitation emails keep linking to Polar.
- When the override is used, no customer session token is minted — auth is the merchant's responsibility.

## Why

Replaces the CUSTOMER_PORTAL_URL_OVERRIDES env var, which required a deploy and only staff could set. The env var is kept as a fallback until the one organization using it is migrated.
  
## Screenshots
| Preference Control |
|---|
| Screenshot 2026-07-28 at 09 16 56 |

## Checklist

- [x] The diff is reasonably sized and easy to re
- [x] Linting and type checking pass (`uv run tas_types`)
- [x] No unrelated changes or drive-by fixes are

---
<a href="https://app.blacksmith.sh/polarsource/codesmith/polar/pr/13368"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787483936&installation_model_id=13063&pr_number=13368&repository=polarsource%2Fpolar&return_to=https%3A%2F%2Fgithub.com%2Fpolarsource%2Fpolar%2Fpull%2F13368&signature=91fddf60a9640ce0f9a23c22829673324051d1499cc9c90aae8e928006bb27e2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

<a href="https://cubic.dev/pr/polarsource/polar/pull/13368?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/polar/models/organization.py
```diff
@@ -162,6 +163,7 @@ class OrganizationCustomerPortalSettings(TypedDict):
     usage: CustomerPortalUsageSettings
     subscription: CustomerPortalSubscriptionSettings
     customer: NotRequired[CustomerPortalCustomerSettings]
+    custom_url: NotRequired[str | None]
```
strandhvilliam: Is even `custom_url` the best name? `portal_url` maybe?

pieterbeulque: Yeah, it's not a good name 😁  `portal_url` is slightly better, all other things I can think of are longer (and thus, by definition, worse, haha)

server/polar/models/organization.py
```diff
@@ -830,6 +832,51 @@ def customer_portal_subscription_update_plan(self) -> bool:
...
 
+    @property
+    def is_custom_customer_portal_url_enabled(self) -> bool:
+        return self.feature_settings.get("custom_customer_portal_url_enabled", False)
+
+    @property
+    def customer_portal_custom_url(self) -> str | None:
+        return self.customer_portal_settings.get("custom_url") or None
+
+    def get_custom_portal_url(
```
pieterbeulque: Why can't we just always route through a `get_customer_portal_url` that either uses the Polar-hosted or the customer URL?

server/polar/subscription/service.py
```diff
@@ -3381,22 +3381,30 @@ async def _send_customer_email(
 
         async def send_to_recipients(recipients: Sequence[str]) -> None:
             for recipient_email in recipients:
-                token = await customer_service.create_session_token_for_recipient(
```
pieterbeulque, line 3492: `order/service.py` and `subscription/service.py` have some overlapping logic for the session token as well, so I'd put all this in a shared `generate_portal_url` helper called by both, I think.

server/polar/organization/schemas.py
```diff
@@ -195,6 +237,13 @@ class OrganizationFeatureSettings(Schema):
             "provider (e.g. Stripe) to Polar."
         ),
     )
+    custom_customer_portal_url_enabled: bool = Field(
```
pieterbeulque: We have a thing where we define what organization flags are user-settable, not sure if we need to do anything for that to avoid the merchant being able to enable this over API.

server/polar/organization/schemas.py
```diff
@@ -134,6 +136,46 @@ def _coerce_overview_metrics(value: Any) -> Any:
...
+        return None
+    try:
+        url = _http_url_adapter.validate_python(value)
+    except ValidationError:
+        raise ValueError("The custom customer portal URL must be a valid URL.")
+    if url.scheme != "https":
+        raise ValueError("The custom customer portal URL must use HTTPS.")
+    if url.query or url.fragment:
+        raise ValueError(
+            "The custom customer portal URL must not contain query parameters. "
```
pieterbeulque: This feels like an unnecessary restriction, there's probably a way to merge querystrings. Just make sure in case of conflicting keys, we set ours and override pre-existing ones and/or throw a 422 if those keys are predefined.

Alternatively, we should use a template-like syntax like we do on checkout links with `{CHECKOUT_ID}`, so that we don't force the querystring parameter names, and people could do e.g. `?e={EMAIL}&id={EXTERNAL_ID}`

server/polar/models/organization.py
strandhvilliam: haha, `custom_url is` such a bad name since we are dealing with customers 🤦‍♂️. Changed to `portal_url` and sprinkled some `_override` where it made sense

server/polar/models/organization.py
strandhvilliam: Good point. Refactored a bit in new version so it's cleaner, but still have `get_customer_portal_url_override` in `organization.py` while the actual `get_customer_portal_email_url` in customer service, since the customer service owns the customer session token generation, which is used on non-overridden urls.

Don't know if the split is good. Open for suggestions

server/polar/subscription/service.py
strandhvilliam: nice callout, now uses the same `generate_email_portal_url` method from customer service

server/polar/organization/schemas.py
strandhvilliam: I think we should be fine. If I'm not wrong thats the `OrganizationFeatureSettingsUpdate`, so unless we add that feature flag there I think we're in the clear

server/polar/organization/schemas.py
strandhvilliam: Totally! Got stuck in the way we override it with the previous env var thing.

I think the template-like syntax is super cool, so updated with that 🚀

server/polar/order/service.py
```diff
@@ -2389,31 +2329,17 @@ async def send_confirmation_email(
             ]
 
         for recipient_email in recipients:
-            token = await customer_service.create_session_token_for_recipient(
-                session, customer, recipient_email
+            url = await customer_service.generate_email_portal_url(
```
strandhvilliam, line 2368: Currently only overriding the email / customer facing portal_urls. I think overriding the `CustomerSession.customer_url`would be a larger undertaking and made just a bit out of scope?

If I'm correct that is used by the actual merchants integration and API response, so I assume they can already route to their own page if they want.

server/polar/models/organization.py
```diff
@@ -830,6 +832,62 @@ def customer_portal_subscription_update_plan(self) -> bool:
...
+        # The DB-configured URL is gated by the feature flag, so disabling the
+        # flag stops using it.
+        if self.is_portal_url_override_enabled:
+            configured_url = self.customer_portal_url_override
+            if configured_url is not None:
+                values = {
+                    "{EMAIL}": recipient_email,
+                    "{EXTERNAL_ID}": customer.external_id or "",
+                    "{ORDER_ID}": str(order_id) if order_id is not None else "",
+                    "{SUBSCRIPTION_ID}": str(subscription_id)
```
strandhvilliam, line 912: Does the values make sense? I'm thinking `ORDER_ID` and `SUBSCRIPTION_ID` can be good for the merchant to identify what order/sub it originates from.

Question is do we also want to include the `customer_session_token` as an option?

server/polar/organization/schemas.py
```diff
@@ -195,6 +242,13 @@ class OrganizationFeatureSettings(Schema):
             "provider (e.g. Stripe) to Polar."
         ),
     )
+    portal_url_override_enabled: bool = Field(
```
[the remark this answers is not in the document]
strandhvilliam: The documented flag you're referring to is outdated.

server/polar/organization/schemas.py
```diff
@@ -134,6 +136,51 @@ def _coerce_overview_metrics(value: Any) -> Any:
...
+    if value is None:
+        return None
+    value = value.strip()
+    if not value:
+        return None
+    try:
+        url = _http_url_adapter.validate_python(value)
+    except ValidationError:
+        raise ValueError("The customer portal URL must be a valid URL.")
+    if url.scheme != "https":
```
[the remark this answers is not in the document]
strandhvilliam: You're referring to outdated contracts

clients/apps/web/src/components/Settings/CustomerPortalUrlSetting.tsx
```diff
@@ -0,0 +1,191 @@
...
+              return
+            }
+            if (url.trim()) {
+              void validateURL(withoutPlaceholders(url.trim()))
+            }
+          }}
+          disabled={readOnly}
+        />
+      </SettingsGroupItem>
+      <div
```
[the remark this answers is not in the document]
strandhvilliam: This `<div>` is fine for now. 🤫

clients/apps/web/src/components/Settings/CustomerPortalUrlSetting.tsx
```diff
@@ -0,0 +1,191 @@
...
+                    setUrl(e.target.value)
+                    if (urlStatus !== 'idle') {
+                      resetValidation()
+                    }
+                  }}
+                  onBlur={() => {
+                    const trimmed = url.trim()
+                    const portalUrl = trimmed === '' ? null : trimmed
+                    if (portalUrl !== value) {
+                      onChange(portalUrl)
```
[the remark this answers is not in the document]
strandhvilliam: It's fine. There is a toast message on server validation errors, and the `useURLValidation` handles inline errors

server/polar/customer/service.py
```diff
@@ -757,6 +758,54 @@ async def create_session_token_for_recipient(
...
+            return None
+
+        # The Polar portal deep-links to the subscription when the email
+        # concerns one, otherwise to the order.
+        link_id = subscription_id or order_id
+        assert link_id is not None, "order_id or subscription_id is required"
+        query_string = urlencode(
+            {
+                "customer_session_token": token,
+                "id": str(link_id),
```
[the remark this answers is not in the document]
strandhvilliam: Bot is confused about what changed. The behavior is identical to what main already does.

frankie567:
Is it still a thing?

strandhvilliam:
Idk, you think this is still relevant @pieterbeulque ?

Also this is the branch where Stilla made som really weird unrelated stuff that I spoke to them about. Need to clear this up if this is something we still want.

pieterbeulque:
Yes, let's revive this based on what we learned building this.

If you don't mind, we can close this branch (especially since Stilla went rogue on it) and do it over?

strandhvilliam:
Sounds good, closing!

Let me know what you want to do differently from this implementation!

PR #13378, opened 2026-07-27
fix: stop orphaning Stripe identity verification sessions

maximevast:
## Summary

Identity verification webhooks are piling up unhandled in `/external-events`.

`user.identity_verification_id` is a one-slot pointer, and it's the only key the webhook handlers use. When a user starts a second verification session the pointer moves, and every webhook still in flight for the first one finds no user
and raises — a 500, so the event never gets marked handled. A `verified` lost this way leaves the user unverified for good.

Two fixes:

- Webhooks resolve the user from the session's `user_id` metadata when the id  lookup misses, so a late `verified` still lands. `processing`, `requires_input` and `canceled` from a session the user has moved on from are now ignored
  instead of raising.
- `create_identity_verification` branches on the status Stripe reports rather than our own, which lags it. It refuses while a session is `processing` or `verified`, reuses one that `requires_input`, and only replaces a `canceled` one.

## Checklist

---
<a href="https://app.blacksmith.sh/polarsource/codesmith/polar/pr/13378"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787729539&installation_model_id=13063&pr_number=13378&repository=polarsource%2Fpolar&return_to=https%3A%2F%2Fgithub.com%2Fpolarsource%2Fpolar%2Fpull%2F13378&signature=18c9a493fd325ef90267baef933bd9bc8d38205d5a3d9152a335330a247885e9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

<a href="https://cubic.dev/pr/polarsource/polar/pull/13378?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/user/service.py
+++ b/server/polar/user/service.py
@@ -47,16 +47,6 @@ class IdentityVerificationProcessing(UserError):
         super().__init__(message, 403)
 
 
-class IdentityVerificationDoesNotExist(UserError):
-    def __init__(self, identity_verification_id: str) -> None:
-        self.identity_verification_id = identity_verification_id
-        message = (
-            f"Received identity verification {identity_verification_id} from Stripe, "
-            "but no associated User exists."
-        )
-        super().__init__(message)
-
-
 class InvalidAccount(UserError):
     def __init__(self, account_id: UUID) -> None:
         self.account_id = account_id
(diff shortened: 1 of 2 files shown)


server/polar/user/service.py
```diff
@@ -176,23 +176,55 @@ async def create_identity_verification(
...
+    ) -> User | None:
         repository = UserRepository.from_session(session)
-        user = await repository.get_by_identity_verification_id(
-            verification_session.id, for_update=True
+        user = await self._get_verification_session_user(
+            repository, verification_session
         )
         if user is None:
-            raise IdentityVerificationDoesNotExist(verification_session.id)
+            return None
```
pieterbeulque: I'm not sure why we change these signatures too, it feels unnecessary widening?

pieterbeulque: I mean, `_get_verification_session_user` will match more often (because we don't rely on the `user.verification_id` match anymore), but `if user is None` should still raise, in my opinion, instead of silently exiting, maybe?

The error just changes from `IdentityVerificationDoesNotExist` to something like `IdentityVerificationForUnknownUser` or something.

pieterbeulque (approved):

server/polar/user/service.py
maximevast: Yeah that's better this way

PR #13394, opened 2026-07-27
feat(discount): per-customer redemption limits

maximevast:
## Summary

Adds `max_redemptions_per_customer`, so a code can be shared publicly with unlimited total uses but capped per buyer.

Dashboard field and docs follow in the next PR; until then the limit is API-only.

Closes polarsource/feedback#86. Replaces #13328, rebuilt on main with a clean history.

Builds on @pieterbeulque's #11993 — the schemas, service logic, error class and most tests come from there. The column shipped in #13322.

## How it works

Identity reuses the trial-abuse signals: customer ID, unaliased email, card fingerprint, OR-matched. The check runs before the Stripe intent is created, so a rejected buyer is never charged. Applying an exhausted code is rejected at apply time too, rather than after the payment form.

The card fingerprint is an extra signal, not a requirement. Free products and 100%-forever subscriptions never produce a confirmation token, so the check falls back to customer ID and email.

A redemption stops counting once all its orders are fully refunded, so a refunded buyer gets their slot back. Refunds issued to head off a chargeback are the exception: they keep the slot, since buy-refund-repeat is the loop this limit exists to close. A lost dispute never marks the order refunded, so it was never in scope. The global `max_redemptions` budget is unaffected either way: the code was still handed out.

Redemptions from a discount applied to an existing subscription count toward the limit, but merchants applying one are not blocked.

## Commits

1. `refactor(checkout): fetch the confirmation token once` — no behavior change.
2. `fix(discount): only lock the discount row when it has a global cap` — predates this work. An uncapped code made concurrent buyers fail each other.
3. `feat(discount): add max_redemptions_per_customer to the API`
4. `feat(discount): count redemptions per customer` — including the refund and chargeback rules.
5. `feat(checkout): enforce the per-customer discount limit` — takes a customer-row lock, so a per-customer code serializes per buyer instead of per code.

## Checklist

### DIFF
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -1034,8 +1034,7 @@ class CheckoutService:
             if checkout.is_payment_setup_required:
                 raise PaymentNotReady()
 
-        # For wallet payments (Apple Pay, Google Pay, Link), we hide the customer name
-        # field for better UX and instead extract the name from Stripe's confirmation token.
+        confirmation_token: stripe_lib.ConfirmationToken | None = None
         if (
             checkout.payment_processor == PaymentProcessor.stripe
             and checkout_confirm.confirmation_token_id is not None
@@ -1045,19 +1044,21 @@ class CheckoutService:
                 confirmation_token = await stripe_service.get_confirmation_token(
                     checkout_confirm.confirmation_token_id
                 )
-                if (
-                    confirmation_token.payment_method_preview is not None
-                    and confirmation_token.payment_method_preview.billing_details
-                    is not None
-                ):
-                    wallet_name = (
-                        confirmation_token.payment_method_preview.billing_details.name
-                    )
-                    if wallet_name:
-                        checkout.customer_name = wallet_name
-                        session.add(checkout)
             except stripe_lib.StripeError:
-                pass
+                confirmation_token = None
+
+        # For wallet payments (Apple Pay, Google Pay, Link), we hide the customer name
+        # field for better UX and instead extract the name from Stripe's confirmation token.
+        if confirmation_token is not None and checkout.customer_name is None:
+            payment_method_preview = confirmation_token.payment_method_preview
+            if (
+                payment_method_preview is not None
+                and payment_method_preview.billing_details is not None
+            ):
+                wallet_name = payment_method_preview.billing_details.name
+                if wallet_name:
+                    checkout.customer_name = wallet_name
+                    session.add(checkout)
 
         required_fields = self._get_required_confirm_fields(checkout)
         for required_field in required_fields:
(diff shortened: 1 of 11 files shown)


server/polar/checkout/service.py
```diff
@@ -1116,6 +1134,39 @@ async def _confirm_inner(
...
+                # reject. The card fingerprint is read from the confirmation token's
+                # payment method preview (no intent/charge required yet).
+                if (
+                    checkout.discount is not None
+                    and checkout.discount.max_redemptions_per_customer is not None
+                ):
+                    # Serialize concurrent confirmations of the same buyer, so two
+                    # checkouts can't both read the count before either redeems.
+                    customer_repository = CustomerRepository.from_session(session)
+                    await customer_repository.get_by_id(customer.id, for_update=True)
```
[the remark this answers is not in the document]
maximevast: It's a really narrow window, where the customer would need to buy the actual thing twice in order to use a discount twice. Seems really edge case; and locking would impact all legit transaction. I think it's a wont fix.

server/polar/discount/service.py
```diff
@@ -435,25 +440,63 @@ async def is_redeemable_discount(
...
         self, session: AsyncSession, discount: Discount
     ) -> AsyncIterator[DiscountRedemption]:
         """
-        Redeem a discount with FOR UPDATE lock to prevent concurrent redemptions.
-
-        Uses PostgreSQL row-level locking instead of Redis distributed locks.
-        The lock is held until the parent transaction commits.
+        Redeem a discount, locking its row when globally capped so the count can't be
+        read stale. Without that cap the lock would make concurrent buyers of the same
+        code fail each other; the per-customer limit serializes on the customer instead.
```
joebon: For brevity/clarity, let's be explicit here that it's the caller's responsibility to acquire the customer lock.

joebon (approved):
Nice work! Added a non-blocking suggestion on making customer lock responsibility clear in comments

server/polar/discount/service.py
```diff
@@ -435,25 +440,63 @@ async def is_redeemable_discount(
...
         repository = DiscountRepository.from_session(session)
 
-        # Acquire FOR UPDATE lock (we're already inside checkout's transaction)
-        try:
-            await repository.get_by_id(discount.id, for_update=True, nowait=True)
-        except DBAPIError as e:
-            if is_lock_not_available_error(e):
-                raise DiscountNotRedeemableError(discount) from e
-            raise
+        if discount.max_redemptions is not None:
```
pieterbeulque, line 496:
```suggestion
        if discount.max_redemptions is not None or discount.max_redemptions_per_customer is not None:
```

No?

pieterbeulque: Ah, no, I read the comment now 😁

server/polar/discount/service.py
maximevast: Done !

PR #13395, opened 2026-07-27
fix(checkout): reset payment form fields on confirm

strandhvilliam:
## Summary

**Related Issue**: https://github.com/polarsource/feedback/issues/312

A checkout that becomes free during confirmation was still validated as if it needed a payment form, blocking the customer with a "Full billing address is required." error. This rarely would happen since applying a discount in checkout uses the _update_. Still worth fixing.

## What

Extract the payment-form reset that `update()` already performed into a shared helper and run it on confirm too. Adds two confirm tests.

## Why

When a 100% discount makes a checkout free, the payment and billing address sections drop out of the form. `update()` handled that; `confirm()` didn't. So when the discount is applied for the first time in the confirm request, confirm validates against a payment method and business-customer flag the customer can no longer see or change, and rejects a checkout that needs neither.

## How

The reset runs after the tax pass on both paths, so the discount is applied and the checkout's final state is what confirm validates against. Both new tests fail on `main` and pass here.

## Checklist

---
<a href="https://app.blacksmith.sh/polarsource/codesmith/polar/pr/13395"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787758428&installation_model_id=13063&pr_number=13395&repository=polarsource%2Fpolar&return_to=https%3A%2F%2Fgithub.com%2Fpolarsource%2Fpolar%2Fpull%2F13395&signature=72354b751ffe11506bc9d60659db5d6cfd7801e5cd157875e796129fdf37d4dc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

<a href="https://cubic.dev/pr/polarsource/polar/pull/13395?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -940,13 +940,7 @@ class CheckoutService:
         except TaxCalculationLogicalError:
             pass
 
-        # Reset payment form fields if it's no longer required
-        # This handles the case where a 100% discount is applied and the
-        # payment and billing address sections disappear from the frontend
-        if not checkout.is_payment_form_required:
-            checkout.payment_method_type = None
-            if not checkout.require_billing_address:
-                checkout.is_business_customer = False
+        self._reset_payment_form_fields(checkout)
 
         await self._after_checkout_updated(session, checkout)
         return checkout
@@ -1015,6 +1009,8 @@ class CheckoutService:
                     "input": None,
                 }
             )
+
+        self._reset_payment_form_fields(checkout)
 
         # Case where the price was archived after the checkout was created
         if has_product_checkout(checkout) and checkout.product_price.is_archived:
@@ -2242,6 +2238,15 @@ class CheckoutService:
         await self._validate_subscription_uniqueness(session, checkout)
 
         return checkout
+
+    def _reset_payment_form_fields(self, checkout: Checkout) -> None:
+        # Reset payment form fields if it's no longer required
+        # This handles the case where a 100% discount is applied and the
+        # payment and billing address sections disappear from the frontend
+        if not checkout.is_payment_form_required:
+            checkout.payment_method_type = None
+            if not checkout.require_billing_address:
+                checkout.is_business_customer = False
 
     async def _update_price(
         self,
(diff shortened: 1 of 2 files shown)


server/polar/checkout/service.py
```diff
@@ -940,13 +940,7 @@ async def update(
...
             pass
 
-        # Reset payment form fields if it's no longer required
-        # This handles the case where a 100% discount is applied and the
-        # payment and billing address sections disappear from the frontend
-        if not checkout.is_payment_form_required:
-            checkout.payment_method_type = None
-            if not checkout.require_billing_address:
-                checkout.is_business_customer = False
+        self._reset_payment_form_fields(checkout)
```
pieterbeulque: Nit-picky but maybe `_reset_payment_form_fields_if_needed` is a better method name? It reads a bit weird in the flow of the method ("huh, why would you reset the form")

pieterbeulque (approved):

server/polar/checkout/service.py
strandhvilliam: Yeah makes sense. Fixed

PR #13405, opened 2026-07-28
feat(organization): embed hosts allowlist

maximevast:
Related Issue: polarsource/feedback#195

Adds an organization-level allowlist of hosts permitted to embed checkout. Nothing is enforced yet.

## What

Organization.embed_hosts readable and writable over the API, under org-write scope. Entries are hosts with an optional scheme and port: example.com, *.example.com, localhost:3000, http://192.168.1.43:5500. HTTPS is implied, and *.example.com matches any subdomain but not the apex.

Organization.embed_hosts_enforced computed from the list being non-empty, or the organization being created on or after EMBED_HOSTS_ENFORCED_FROM.

embed_origin reduced to its origin on the checkout-link redirect instead of stored verbatim.

## Why

embed_origin arrives from a query parameter and becomes the postMessage targetOrigin carrying a customer session token. The allowlist is what will let merchants constrain it. This PR is the API and the matching; enforcement follows separately so merchants can populate their list before anything breaks.

## How

Entries and origins share one URL parser, so both normalise identically — *.exämple.com is stored as *.xn--exmple-cua.com and matches https://a.xn--exmple-cua.com.

parse_origin reduces a URL to its origin and never errors. 90 days of production shows a strict check would have refused something from 122 of 1,379 embedding organizations, mostly full page URLs like https://www.makercase.com/laserHinge, plus plain HTTP and chrome-extension://. All of those embed correctly, because postMessage compares only the origin. Values with no origin (null, file://) store nothing rather than raising.

No cap on list length: one organization runs 92 distinct apex domains. Wildcards inside a label are not supported, so Vercel previews like website-<hash>-tabler-io.vercel.app can only be covered by *.vercel.app — worth deciding before enforcement.

Merges after #13408.

Checklist

- [x] Linting and type checking pass (uv run task lint && uv run task lint_types)

### DIFF
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -91,6 +91,7 @@ from polar.observability.checkout_metrics import (
     CHECKOUT_SUCCEEDED_TOTAL,
 )
 from polar.order.service import order as order_service
+from polar.organization.embed_hosts import parse_origin
 from polar.postgres import AsyncReadSession, AsyncSession
 from polar.posthog import posthog
 from polar.product.custom_price import validate_custom_price_amount
@@ -698,6 +699,12 @@ class CheckoutService:
         query_prefill: dict[str, str | UUID4 | dict[str, str] | None] | None = None,
         **query_metadata: str | None,
     ) -> Checkout:
+        if embed_origin is not None:
+            parsed_embed_origin = parse_origin(embed_origin)
+            embed_origin = (
+                str(parsed_embed_origin) if parsed_embed_origin is not None else None
+            )
+
         products: list[Product] = []
         for product in checkout_link.products:
             if not product.is_archived and product.visibility != Visibility.draft:
--- a/server/polar/organization/embed_hosts.py
+++ b/server/polar/organization/embed_hosts.py
@@ -1,0 +1,138 @@
+from dataclasses import dataclass
+
+from pydantic import AnyUrl, TypeAdapter, ValidationError
+
+DEFAULT_PORTS = {"http": 80, "https": 443}
+DEFAULT_SCHEME = "https"
+
+MAX_HOST_LENGTH = 253
+
+WILDCARD_PREFIX = "*."
+# Stands in for the wildcard so an entry can go through the origin parser.
+_WILDCARD_LABEL = "wildcard"
+
+_url_adapter = TypeAdapter(AnyUrl)
+
+
+class InvalidEmbedHost(ValueError):
+    def __init__(self, value: str, reason: str) -> None:
+        self.value = value
+        super().__init__(f"{value!r} is not a valid embed host: {reason}")
+
+
+@dataclass(frozen=True, slots=True)
+class ParsedOrigin:
+    scheme: str
+    host: str
+    port: int | None
+
+    def __str__(self) -> str:
+        if self.port is None or self.port == DEFAULT_PORTS.get(self.scheme):
+            return f"{self.scheme}://{self.host}"
+        return f"{self.scheme}://{self.host}:{self.port}"
+
+
+@dataclass(frozen=True, slots=True)
+class HostPattern:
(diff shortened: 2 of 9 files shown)


server/polar/checkout/service.py
```diff
@@ -698,6 +710,12 @@ async def checkout_link_create(
         query_prefill: dict[str, str | UUID4 | dict[str, str] | None] | None = None,
         **query_metadata: str | None,
     ) -> Checkout:
+        if embed_origin is not None:
+            parsed_embed_origin = parse_origin(embed_origin)
+            if parsed_embed_origin is None:
+                raise InvalidEmbedOrigin(embed_origin)
```
pieterbeulque: Hmm, if you look at production data, this will be a breaking change for some merchants. I don't think we need to reject invalid `embed_origin` querystring values right now for the rest of this to work, do we?

server/polar/organization/embed_hosts.py
```diff
@@ -0,0 +1,48 @@
...
+    if url.username is not None or url.password is not None:
+        return None
+
+    if url.path not in (None, "", "/") or url.query is not None:
+        return None
+
+    if url.fragment is not None:
+        return None
+
+    if url.host is None or url.port is None or url.port == 0:
```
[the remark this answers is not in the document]
pieterbeulque: `*.example.com` should be allowed

server/polar/checkout/service.py
maximevast: You are right, converting to draft for now

server/polar/organization/schemas.py
```diff
@@ -71,6 +72,35 @@ def validate_blocked_words(value: str) -> str:
...
+        except InvalidEmbedHost as e:
+            raise ValueError(str(e)) from e
+        if host not in hosts:
+            hosts.append(host)
+    return hosts
+
+
+_embed_hosts_description = (
+    "Hosts allowed to embed this organization's checkout. "
+    "An entry is a host, optionally prefixed by a scheme and suffixed by a port. "
```
joebon, line 89: Hrm, I think we should avoid including a scheme in here. As per the issue, the only scheme we should allow is `https`, with `http` as a dev/local variant.

joebon (changes requested):
It looks like this implementation permits pretty much any scheme, but the only scheme we _should_ be allowing is `https` (with `http` support in dev/local mode). Can we drop the scheme part of the entries entirely?

server/polar/organization/schemas.py
```diff
@@ -476,6 +506,14 @@ class Organization(OrganizationBase):
     customer_portal_settings: OrganizationCustomerPortalSettings = Field(
         description="Settings related to the customer portal",
     )
+    embed_hosts: list[str] = Field(description=_embed_hosts_description)
```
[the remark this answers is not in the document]
maximevast: Column is not nullable anymore: https://github.com/polarsource/polar/pull/13408

server/polar/organization/schemas.py
maximevast: @joebon 

Ran 90 days of embed_origin:

https 1259 orgs / local http 333 / app schemes 87 / public http 21.

Local http at 333 is why we cannot hardcode localhost + 127.0.0.1: 
xxx.localhost, .local, 192.168.x and 10.x are all in there. Those should keep working over http, and I can detect them structurally as they contain no scheme.

Public http is 21 orgs. We can break those and use the grace period to push them to https, wdyt?

87 orgs embed from app schemes, (chrome-extension://<id> and friends). We have to support them, and there's no host to allowlist, what do you suggest for them?

joebon: Thanks for the clarification! I digress, we obviously can't be that strict and rigid. Let's keep as is

joebon (approved):

PR #13424, opened 2026-07-29
feat(payment_method): remind customers before their card expires

joebon:
Related issue: https://github.com/polarsource/feedback/issues/225

Example reminder email:
image

---
<a href="https://app.blacksmith.sh/polarsource/codesmith/polar/pr/13424"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787910528&installation_model_id=13063&pr_number=13424&repository=polarsource%2Fpolar&return_to=https%3A%2F%2Fgithub.com%2Fpolarsource%2Fpolar%2Fpull%2F13424&signature=2b3c8ae1d5e835158081979e35b99c3581cbc73d1d8a95b596aa00adf28d9067"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Send expiring-card reminder emails about 20 days before a card backing an active subscription expires. An hourly scan finds affected cards and emails a portal link; organizations can disable the email, and for existing orgs the default inherits from their subscription cycle setting.

- **New Features**
  - Added `payment_method_expiration_reminder` email (shows brand/last4, expiry, product names) with a secure portal link.
  - Hourly cron `payment_method.scan_expiration_reminders` finds cards expiring within ~20 days and enqueues `payment_method.send_expiration_reminder`; only cards with billable subscriptions are considered. Sends are deduped via `email_logs` by card + expiry month/year. New org setting `payment_method_expiration_reminder` (exposed in Organization Customer Email settings) defaults to the org’s `subscription_cycled` preference for existing orgs; client/OpenAPI types updated.

- **Bug Fixes**
  - Scan now ignores soft-deleted subscriptions to avoid enqueueing reminders for cards with only deleted billable subscriptions.

<sup>Written for commit 7b5a7885ed7aa7dbfca884dc4acaaa8def3d0396. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/13424?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/payment_method/tasks.py
+++ b/server/polar/payment_method/tasks.py
@@ -1,0 +1,52 @@
+import uuid
+from datetime import timedelta
+
+from polar.kit.utils import utc_now
+from polar.worker import (
+    AsyncSessionMaker,
+    CronTrigger,
+    TaskPriority,
+    actor,
+    enqueue_job,
+)
+
+from .repository import PaymentMethodRepository
+from .service import PaymentMethodDoesNotExist
+from .service import payment_method as payment_method_service
+
+EXPIRATION_REMINDER_WINDOW = timedelta(days=30)
+
+
+@actor(
+    actor_name="payment_method.scan_expiration_reminders",
+    cron_trigger=CronTrigger.from_crontab("45 * * * *"),
+    priority=TaskPriority.LOW,
+)
+async def scan_expiration_reminders() -> None:
+    """Scan for soon-to-expire cards needing a reminder and fan out."""
+    now = utc_now()
+    window_end = now + EXPIRATION_REMINDER_WINDOW
+
+    async with AsyncSessionMaker() as session:
+        repository = PaymentMethodRepository.from_session(session)
+        payment_methods = await repository.get_cards_needing_expiration_reminder(
+            now, window_end
+        )
+
+    for payment_method in payment_methods:
+        enqueue_job("payment_method.send_expiration_reminder", payment_method.id)
+
+
+@actor(actor_name="payment_method.send_expiration_reminder", priority=TaskPriority.LOW)
+async def send_expiration_reminder(payment_method_id: uuid.UUID) -> None:
+    async with AsyncSessionMaker() as session:
+        repository = PaymentMethodRepository.from_session(session)
+        payment_method = await repository.get_by_id(
+            payment_method_id, options=repository.get_eager_options()
+        )
+        if payment_method is None:
+            raise PaymentMethodDoesNotExist(payment_method_id)
+
+        await payment_method_service.send_expiration_reminder_email(
+            session, payment_method
+        )
--- a/server/polar/payment_method/tasks.py
+++ b/server/polar/payment_method/tasks.py
@@ -14,7 +14,9 @@ from .repository import PaymentMethodRepository
 from .service import PaymentMethodDoesNotExist
 from .service import payment_method as payment_method_service
 
-EXPIRATION_REMINDER_WINDOW = timedelta(days=30)
+# Cards expire at the end of their month, so a 30 day window would always fire on
+# the 1st — where the reminder competes with everyone's monthly billing emails.
+EXPIRATION_REMINDER_WINDOW = timedelta(days=20)
 
 
 @actor(
--- a/clients/apps/web/src/components/Settings/OrganizationCustomerEmailSettings.tsx
+++ b/clients/apps/web/src/components/Settings/OrganizationCustomerEmailSettings.tsx
@@ -72,6 +72,11 @@ const customerEmails: {
       'Sent when a canceled subscription permanently ends at the end of the billing cycle',
   },
   {
+    key: 'payment_method_expiration_reminder',
+    title: 'Card expiring',
+    description: 'Sent before a card backing an active subscription expires',
+  },
+  {
     key: 'subscription_past_due',
     title: 'Subscription past due',
     description: 'Sent when a subscription payment fails and becomes overdue',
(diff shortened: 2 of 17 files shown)


server/polar/payment_method/tasks.py
```diff
@@ -0,0 +1,52 @@
...
+    priority=TaskPriority.LOW,
+)
+async def scan_expiration_reminders() -> None:
+    """Scan for soon-to-expire cards needing a reminder and fan out."""
+    now = utc_now()
+    window_end = now + EXPIRATION_REMINDER_WINDOW
+
+    async with AsyncSessionMaker() as session:
+        repository = PaymentMethodRepository.from_session(session)
+        payment_methods = await repository.get_cards_needing_expiration_reminder(
```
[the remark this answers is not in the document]
joebon: That's fine, we already do the same elsewhere

server/polar/payment_method/tasks.py
```diff
@@ -0,0 +1,52 @@
...
+            now, window_end
+        )
+
+    for payment_method in payment_methods:
+        enqueue_job("payment_method.send_expiration_reminder", payment_method.id)
```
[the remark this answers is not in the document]
joebon: Acceptable risk, we already do the same elsewhere

server/polar/payment_method/tasks.py
```diff
@@ -0,0 +1,52 @@
...
+    TaskPriority,
+    actor,
+    enqueue_job,
+)
+
+from .repository import PaymentMethodRepository
+from .service import PaymentMethodDoesNotExist
+from .service import payment_method as payment_method_service
+
+EXPIRATION_REMINDER_WINDOW = timedelta(days=30)
```
strandhvilliam: Nice, can really see this being helpful!

Was just thinking if we should frame it differently depending on what subscription period it is? Most are monthly I guess so it makes sense, but I'm just imagining a situation where a yearly subscriber would get a reminder that their card expires 9 months before the actual subscription renews. Which is no problem technically, just maybe a bit confusion or premature. If I understand the code correctly.

One idea is framing it maybe like "a charge is due after your card expires"? Not saying this is what we should do just bringing it up. Otherwise lgtm 😅

joebon: Good point and reminder re: different cadences, thanks @strandhvilliam ! I guess it comes down to whether we want to send the email when the card is about to expire or when the subscription is about to renew and the underlying card is or will soon expire. What's your take?

joebon: I think I would still argue that it'd be more natural to send it close to the actual expiration date. It's a bit more contextual that way imho, especially since I would assume other services where the customer use the card in question would do the same about the same time.

strandhvilliam: Yeah I think you have a point! Updating an expired card is actionable even if your subscription renews much later, so close to expiry probably makes most sense.

strandhvilliam (approved):

pieterbeulque (changes requested):
Love it!

Two things:

1. I'd do a smart default instead of enabling it for anyone. When I added trial conversion & annual renewal reminders, I used the value of the "subscription cycle" email setting as the default. Some merchants have all emails disabled and I don't like enabling a single email for them silently, as they probably don't want us to send any emails at all.

2. Since card expiry is MM/YY, wouldn't this mean we always send on the first or second of the month? People get a lot of reminders or monthly around those days and this email may get buried. What if we do 20 or 15 days? I just ⌘ + F in my mailbox and it looks like the 15th of the month is when GitHub / AWS / ... send them out. 

I'm also wondering if we should be worried about triggering too many emails at once. Ideally we can spread it out across a few days (e.g. with a deterministic offset derived of customer ID or sth), but maybe that's not a problem at our current scale.

joebon:
> Love it!
> 
> Two things:
> 
> 1. I'd do a smart default instead of enabling it for anyone. When I added trial conversion & annual renewal reminders, I used the value of the "subscription cycle" email setting as the default. Some merchants have all emails disabled and I don't like enabling a single email for them silently, as they probably don't want us to send any emails at all.
> 2. Since card expiry is MM/YY, wouldn't this mean we always send on the first or second of the month? People get a lot of reminders or monthly around those days and this email may get buried. What if we do 20 or 15 days? I just ⌘ + F in my mailbox and it looks like the 15th of the month is when GitHub / AWS / ... send them out.
> 
> I'm also wondering if we should be worried about triggering too many emails at once. Ideally we can spread it out across a few days (e.g. with a deterministic offset derived of customer ID or sth), but maybe that's not a problem at our current scale.

Two solid improvements, thanks @pieterbeulque !

I couldn't find any existing pattern for falling back to a pre-existing email setting so I built one myself. And I opted for 20 days.

pieterbeulque (approved):
Beautiful!

Issue #4816, opened 2025-01-12
Customers: Soft deletion does not free up the email


birkjernstrom:
### Description

We soft delete customers, but the unique index on `organization_id` and `email` prevents the email from being freed up for a new customer.

### Expected Behavior

We could probably solve this by modifying the index to include the `deleted_at` timestamp. Causing a unique violation in case of creating a customer with the same email as an existing – non-deleted – customer, but not in case others have a unique deleted_at timestamp.

birkjernstrom:
Moving to new issue to be fixed. We've started using Linear internally for better coordination, but the syncing only works for new issues so creating a new one to ensure this is prioritized.
https://github.com/polarsource/polar/issues/5142

PR #13453, opened 2026-07-30
fix(customer): free external_id and emit customer.deleted when deleting with anonymize

pieterbeulque:
## Summary

**Related Issue**: #4816 (the `external_id` counterpart of the email fix)

`DELETE /v1/customers/{id}?anonymize=true` permanently locks the customer's `external_id` and never emits a `customer.deleted` webhook. A plain delete does neither.

## What

- `CustomerService.anonymize()` now scrubs PII only; `delete()` soft-deletes and then anonymizes.
- That reaches the external-ID clearing and the deletion signals `soft_delete()` already had — no new clearing logic anywhere.
- `CustomerRepository.update()` no longer announces a write to a deleted customer as `customer.updated`, so a delete emits exactly one webhook and one event.
- Customer system events report `saved_external_id`, so `customer.deleted` names the ID it just released instead of `null`.
- `scripts/backfill_customer_deleted_external_id.py` releases the external IDs of customers already stuck in this state.
- Extracted the metadata key as `EXTERNAL_ID_METADATA_KEY` on the model, shared by the model, repository, and script.

## Why

Both bugs have one root cause: `delete(anonymize=True)` returned `anonymize()`, which set `deleted_at` itself via `repository.update()` and never reached `soft_delete()`. That's the only place that frees the external ID *and* the only place that enqueues `WebhookEventType.customer_deleted` / `SystemEvent.customer_deleted`.

**External ID.** The unique constraint `(organization_id, external_id)` doesn't include `deleted_at` (unlike the email index, fixed in #4816), so a deleted-but-not-cleared row holds its external ID forever. It fails badly, too: the conflict check in `CustomerService.create()` uses `get_by_external_id_and_organization`, which filters `deleted_at IS NULL`, so it never sees the deleted row — the collision surfaces only as an `IntegrityError`, a 500 rather than the intended 422, with no self-serve way to release the value. This is reaching support: `external_id` is immutable, so an integration keying Polar customers on its own user id can't recreate a customer for that user after an anonymized delete.

**Deletion signals.** Integrators were told about an anonymized delete only through a `customer.updated` event, leaving them unable to reconcile. Notably `PolarClient.delete_customer()` calls `delete_external(anonymize=True)`, so our own integration path was affected.

There was a third, quieter consequence: `anonymize()` returns early when the customer is already anonymized, so `delete(anonymize=True)` on such a customer skipped the soft delete entirely and silently did nothing.

**Event payload.** Separately, the customer event tasks re-read the customer when the job runs, which is after the request transaction commits. Deletion clears `external_id`, so by then the column is null and every event reported `customer_external_id: null` — including `customer.deleted`, the one event an integrator needs to reconcile a deletion against its own records.

## How

`anonymize()` no longer writes `deleted_at`. `delete()` soft-deletes first, then anonymizes when asked — scrubbing PII is cleanup that follows the deletion, and running it on an already-deleted customer keeps that write from being announced as an update. `soft_delete()` already freed the external ID and emitted the deletion signals; it's now reachable from both paths.

Freeing the external ID stays exactly where it was — at the deletion boundary in `soft_delete()`, not in `anonymize()`. `anonymize()` keeps preserving `external_id` as it always did; since it no longer deletes, clearing there would release a *live* customer's external ID. That failure would be quiet, too: `saved_external_id` falls back to metadata, so the API would keep reporting the ID while `get_by_external_id_and_organization`, which queries the column, stopped finding the customer.

`Customer.saved_external_id` reads the column and falls back to metadata, and the read schema resolves `external_id` via `AliasChoices("saved_external_id", "external_id")`, so **API responses and webhook payloads for deleted customers are unchanged**. The event tasks now read `saved_external_id` too; for a live customer it returns the column verbatim, so only deleted customers report differently.

Existing rows are handled by a backfill script rather than a migration — maintenance work on a busy table, not a schema change. It uses `run_batched_update` from `scripts/helper.py` for the batching loop, per-batch commits, inter-batch sleep, and progress output:

```bash
uv run python -m scripts.backfill_customer_deleted_external_id
```

The predicate (`deleted_at IS NOT NULL AND external_id IS NOT NULL`) excludes rows already processed, which gives the loop its termination condition and makes reruns safe. It also picks up rows predating the `soft_delete()` clearing. Clearing a column can't violate a unique constraint, and a live customer can't already hold the same value (the constraint guarantees it), so there is no conflict risk.

Worth sizing before running:

```sql
SELECT count(*) FROM customers WHERE deleted_at IS NOT NULL AND external_id IS NOT NULL;
```

## Webhook behavior

A delete now emits exactly one webhook and one event, on both paths:

| | before | after |
|---|---|---|
| `delete()` | `customer.updated`, `customer.deleted` | `customer.deleted` |
| `delete(anonymize=True)` | `customer.updated` | `customer.deleted` |

The stray `customer.updated` came from `CustomerRepository.update()`, which enqueued one on every call — and a soft delete is an update, since `RepositorySoftDeletionMixin.soft_delete()` is implemented as `self.update(update_dict={"deleted_at": ...})`. The enqueue now sits inside the existing `if not customer.deleted_at` guard, which already suppressed the matching system event.

Two consequences worth checking in review:

- **Restoring a customer still notifies.** The guard reads `deleted_at` *after* the update is applied, so clearing it emits `customer.updated` as before. Covered by `test_update_restoring_customer_emits_updated`.
- **Updating a customer that stays deleted is now silent** — e.g. GDPR-anonymizing one deleted earlier. It previously sent a `customer.updated` webhook with no accompanying event. Covered by `test_update_deleted_customer_is_silent`.

## Checklist

## Notes for reviewers

**The constraint itself is the deeper issue**, and it's deliberately not touched here. Mirroring the #4816 email fix — a unique index on `(organization_id, external_id, deleted_at)`, `NULLS NOT DISTINCT`, `WHERE external_id IS NOT NULL` — would let deleted customers genuinely keep their external ID while still freeing it for reuse, retiring the metadata round-trip entirely. That's a load-bearing change worth an ADR rather than a drive-by.


pieterbeulque:
I'll address the _Notes for reviewers_ because I think we can do a better job here.

joebon (approved):
Nice catch and fix! Taken note of the follow-up fixes? I.e. removing the extra web hooks and fix the index

PR #13476, opened 2026-07-30
fix(user): handle concurrent creation in get_by_email_or_create

pieterbeulque:
## Summary

**Related Issue**: Fixes polarsource/feedback#302

`user_service.get_by_email_or_create` did a check-then-create with no handling for the duplicate-key error, so two concurrent requests for the same new email could both attempt an insert and the loser got a 500.

## What

- `get_by_email_or_create` now wraps the insert in a savepoint. On `IntegrityError` it rolls back to the savepoint, re-reads the row the winning request created, and returns it with `created=False`. If no row is found, the error is re-raised so unrelated integrity failures still surface.
- Added `TestGetByEmailOrCreate` in `server/tests/user/test_service.py` covering the existing-user path, the create path, the concurrent-creation recovery, and the re-raise path.

## Why

`POST /v1/organizations/{id}/members/invite` calls `get_by_email_or_create` before `organization_service.add_user`. When two admins invite the same new member email simultaneously, both requests can observe `get_by_email() -> None` and both try to create the user. The second insert violates `ix_users_email_case_insensitive` and raises `IntegrityError` during `session.flush()`, which propagated out of the endpoint as a 500 — the request never reached `add_user`.

Both invites should succeed: one creates the user, the other finds it. The same TOCTOU window exists on the other callers of this method (`auth/endpoints.py`, `auth/sso/endpoints.py`, `auth/oauth2/router.py`), so the fix is in the service rather than at the invite call site.

## How

Mirrors the recovery pattern already used elsewhere in the codebase — `EventTypeRepository.get_or_create` (`polar/event_type/repository.py`) and `organization_service.add_user`, which already handled the duplicate-membership half of this same race:

```python
nested = await session.begin_nested()
try:
    user = await self.create_by_email(session, email, signup_attribution=signup_attribution)
except IntegrityError:
    await nested.rollback()
    user = await repository.get_by_email(email)
    if user is None:
        raise
    return (user, False)
return (user, True)
```

Notes on the approach:

- **Savepoint, not `session.rollback()`.** A full rollback would discard the whole request transaction, not just the failed insert. `begin_nested()` scopes the rollback to the `INSERT` and leaves the caller's transaction usable — this is the established convention in this repo.
- **No stale background job.** `create_by_email` enqueues `user.on_after_signup` *after* the flush, so the losing request never enqueues a job for a user it didn't create.
- **The recovery lookup does find the row.** Under READ COMMITTED, the loser's `INSERT` blocks until the winner commits, so the unique violation implies the row is committed and visible to the subsequent statement's snapshot.
- **Re-raise preserves current behaviour** for the case where the conflicting row exists but `get_by_email` filters it out (blocked users are excluded from that query). Returning a blocked user here would change auth behaviour, which is out of scope for this fix.

## Checklist

- [ ] Linting and type checking pass (`uv run task lint && uv run task lint_types`) — `ruff format` and `ruff check` pass on the changed files; `lint_types` and `pytest` could not be run in this environment (no Python 3.14 toolchain, no Docker for the test database), so CI is the first full check

---
_Generated by [Claude Code](https://claude.ai/code/session_01WJrhWohQsPQfvRBwiroRzf)_

---
<a href="https://app.blacksmith.sh/polarsource/codesmith/polar/pr/13476"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788021014&installation_model_id=13063&pr_number=13476&repository=polarsource%2Fpolar&return_to=https%3A%2F%2Fgithub.com%2Fpolarsource%2Fpolar%2Fpull%2F13476&signature=0a2f1b7cdda7a9091f685874ebe966cc98d0f06cd9b5cbdb90957f092e14f99e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

<a href="https://cubic.dev/pr/polarsource/polar/pull/13476?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/user/service.py
+++ b/server/polar/user/service.py
@@ -5,6 +5,7 @@ from uuid import UUID
 import stripe as stripe_lib
 import structlog
 from sqlalchemy import func, update
+from sqlalchemy.exc import IntegrityError
 
 from polar.exceptions import PolarError
 from polar.integrations.polar.service import polar_self as polar_self_service
@@ -76,14 +77,23 @@ class UserService:
     ) -> tuple[User, bool]:
         repository = UserRepository.from_session(session)
         user = await repository.get_by_email(email)
-        created = False
-        if user is None:
+        if user is not None:
+            return (user, False)
+
+        nested = await session.begin_nested()
+        try:
             user = await self.create_by_email(
                 session, email, signup_attribution=signup_attribution
             )
-            created = True
+        except IntegrityError:
+            await nested.rollback()
+            user = await repository.get_by_email(email)
+            if user is None:
+                raise
+            log.info("user.get_by_email_or_create.concurrent_creation", user_id=user.id)
+            return (user, False)
 
-        return (user, created)
+        return (user, True)
 
     async def create_by_email(
         self,
--- a/server/tests/user/test_service.py
+++ b/server/tests/user/test_service.py
@@ -2,6 +2,7 @@ import pytest
 import stripe as stripe_lib
 from pytest_mock import MockerFixture
 from sqlalchemy import select
+from sqlalchemy.exc import IntegrityError
 
 from polar.kit.utils import utc_now
 from polar.models import (
@@ -14,6 +15,7 @@ from polar.models import (
 from polar.models.user import IdentityVerificationStatus, OAuthPlatform
 from polar.models.user_organization import OrganizationRole
 from polar.postgres import AsyncSession
+from polar.user.repository import UserRepository
 from polar.user.schemas import UserDeletionBlockedReason, UserUpdate
 from polar.user.service import (
     IdentityAlreadyVerified,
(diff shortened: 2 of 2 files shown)


maximevast (approved):
LGTM!

PR #13478, opened 2026-07-31
Always check for `deleted_at`, drop redundant index

pieterbeulque:
I do believe this also fixes a potential bug where we could have billed for deleted entries.

### DIFF
--- a/server/polar/models/billing_entry.py
+++ b/server/polar/models/billing_entry.py
@@ -38,12 +38,6 @@ class BillingEntry(RecordModel):
     __tablename__ = "billing_entry"
     __table_args__ = (
         Index(
-            "ix_billing_entries_s_oi_pp",
-            "subscription_id",
-            "order_item_id",
-            "product_price_id",
-        ),
-        Index(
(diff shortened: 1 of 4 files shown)


server/polar/models/billing_entry.py
```diff
@@ -37,12 +37,6 @@ class BillingEntryType(StrEnum):
 class BillingEntry(RecordModel):
     __tablename__ = "billing_entry"
     __table_args__ = (
-        Index(
-            "ix_billing_entries_s_oi_pp",
-            "subscription_id",
-            "order_item_id",
-            "product_price_id",
-        ),
```
joebon, line 45: How do you know this is redundant @pieterbeulque ?

pieterbeulque: We already have another index `ix_billing_entries_s_d_oi_pp` that indexes `subscription_id`, `deleted_at`, `order_item_id`, and `product_price_id` (so all three of this redundant index + `deleted_at`). By making sure every query respects the soft-deletion we don't need the index without the `deleted_at` anymore.

joebon (approved):

PR #13494, opened 2026-07-31
feat(orders): improve order export controls

strandhvilliam:
## Summary

Replaces the one-click "Export" button on the sales page with an options dialog, and extends
the `/v1/orders/export` endpoint with status, date range, time zone and column selection.

| Collapsed | Open |
|---|---|
| Screenshot 2026-07-31 at 16 48 40 | Screenshot 2026-07-31 at 16 48 51 |

## What

Backend:
- New `polar/order/export.py` holding the CSV generation logic, moved out of `endpoints.py` layer to it's own file.
- `GET /v1/orders/export` gains `status`, `created_after`, `created_before`, `timezone` and
  `columns` query parameters.
- 16 selectable columns (the previous 7 remain the default, in order): customer name, billing
  name, billing country, subtotal, discount, tax, total, refunded amount and billing reason are
  new. `billing_reason` is mapped from the internal enum to the public one.
- `created_at` is rendered in the requested IANA time zone, and the downloaded filename now
  carries the date bounds (e.g. `polar-orders-2026-01-01-2026-07-31.csv`).

Frontend:
- `ExportOrdersModal` + `ExportOrdersColumns` on the sales page: local/UTC toggle, date range
  picker and a collapsible column picker.
- The dialog inherits the product and status filters currently applied to the table.

## Why

The experience of exporting was lacking compared to other features that users may expect from similar payment platforms with transaction export functionality. This adds a more deliberate way of exporting data for orders so that anyone doing accounting or reconciliation don't have to post-process the file by hand as much. This can be the starting point to improving export functionality on other pages as well.

## How

CSV rows are built as a `column -> value` mapping and then projected in the requested column
order, so the header and the row stay in sync and adding a column is a single entry in three
places. The time zone is a validated `str` (`AfterValidator` checking `ZoneInfo`) rather than an
enum of every IANA zone, the generated TypeScript client inlines such an enum at every use
site, which added ~1200 lines.

## Checklist

---
<a href="https://app.blacksmith.sh/polarsource/codesmith/polar/pr/13494"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788101512&installation_model_id=13063&pr_number=13494&repository=polarsource%2Fpolar&return_to=https%3A%2F%2Fgithub.com%2Fpolarsource%2Fpolar%2Fpull%2F13494&signature=9f30262ff2fb7c95eafd711067a3358d9d9262177a77c2560d95fe445e4eeada"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


server/polar/order/export.py
```diff
@@ -0,0 +1,163 @@
...
+    invoice_number = "invoice_number"
+    customer_name = "customer_name"
+    billing_name = "billing_name"
+    billing_country = "billing_country"
+    subtotal_amount = "subtotal_amount"
+    discount_amount = "discount_amount"
+    tax_amount = "tax_amount"
+    total_amount = "total_amount"
+    refunded_amount = "refunded_amount"
+    billing_reason = "billing_reason"
```
strandhvilliam, line 37: Please provide feedback if any of these shouldn't be exposed, or doesn't make sense. The Billing Reason is filtered with `_public_billing_reason` helper to match the `@field_serializer` in `order/schemas.py`

server/polar/order/export.py
```diff
@@ -0,0 +1,163 @@
...
+
+    (results, _) = await order_service.list(
+        session,
+        auth_subject,
+        organization_id=organization_id,
+        product_id=product_id,
+        status=status,
+        created_after=created_after,
+        created_before=created_before,
+        pagination=PaginationParams(limit=1000000, page=1),
```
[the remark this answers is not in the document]
strandhvilliam: It's preexisting, already had the identical PaginationParams(limit=1000000, page=1) before this PR.

A streaming solution could be a follow up fix and not a drive by fix here.

server/polar/order/export.py
```diff
@@ -0,0 +1,163 @@
+from collections.abc import AsyncGenerator, Sequence
```
strandhvilliam, line 1: Open for opinions and pushback on the new file. But the preexisting export logic was contained in the `endpoints.py` file which seems to be the wrong layer to have it, even if it was quite small.

Moving it to the `service.py` seems a bit off to, as this hasn't anything to do with actual order business logic. So I decided to put it in an `export.py` file.

Maybe a export module would be even more fitting to centralize exporting logic for multiple entities

Yopi: I think it sounds like a good idea to move it out. It was probably there just because it was so small before.

I think it's fine to keep it here for now and then we can push all export stuff into a centralized place when we get there

server/polar/order/export.py
Yopi: I believe all of these are exposed via public APIs already, so this should be fine.

Only question is on order amount. Does that actually exist anymore in a non-prefixed way?

Yopi: I mean should we use `net_amount` instead for amount, since that's the name mapping, and it's already a computed field

server/polar/order/export.py
```diff
@@ -0,0 +1,163 @@
...
+        raise ValueError(f"{value!r} is not a valid IANA time zone") from None
+    return value
+
+
+# Kept as a plain string rather than an enum of every IANA zone: the generated
+# TypeScript client inlines such an enum at every use site, adding ~1200 lines.
+OrderExportTimezone = Annotated[str, AfterValidator(_validate_timezone)]
+
+
+def _public_billing_reason(value: OrderBillingReasonInternal) -> OrderBillingReason:
```
Yopi: Should/could this reuse `serialize_billing_reason` to avoid the duplication?

Yopi (approved):
I like it! Nice work 😄

server/polar/order/export.py
strandhvilliam: Yeah I think that makes sense, that would also reduce confusion of difference with `total_amount`. Thanks!

server/polar/order/export.py
strandhvilliam: Alright, perfect!

server/polar/order/export.py
strandhvilliam: Good point, that would probably be best to avoid drift. Only way I could think of was putting a method on the `OrderReasonBillingInternal` class and reusing it in both places. Does this makes sense to do?

```
class OrderBillingReasonInternal(StrEnum):
    """
    Internal billing reasons with additional granularity.
    """

    purchase = "purchase"
    subscription_create = "subscription_create"
    subscription_cycle = "subscription_cycle"
    subscription_cycle_after_trial = "subscription_cycle_after_trial"
    subscription_cancel = "subscription_cancel"
    subscription_update = "subscription_update"

    def to_public(self) -> OrderBillingReason:
        if self in (
            OrderBillingReasonInternal.subscription_cycle_after_trial,
            OrderBillingReasonInternal.subscription_cancel,
        ):
            return OrderBillingReason.subscription_cycle
        return OrderBillingReason(self)
```

strandhvilliam: Lol, that broke it. Had to make it a classmethod so it matches similar `RefundReason.to_stripe`

server/polar/order/export.py
```diff
@@ -0,0 +1,150 @@
...
+from polar.product.schemas import ProductID
+
+from .service import order as order_service
+
+
+class OrderExportColumn(StrEnum):
+    email = "email"
+    created_at = "created_at"
+    product = "product"
+    net_amount = "net_amount"
```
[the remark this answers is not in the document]
strandhvilliam: fixed in [dcb0787](https://github.com/polarsource/polar/pull/13494/commits/dcb0787a84a4962e7f42b672f1785cc1817669db)

PR #13506, opened 2026-08-01
perf(billing_entry): partial index for pending static entries

pieterbeulque:
## Summary

Cycle order creation was timing out (`asyncpg TimeoutError`) for high-volume metered subscriptions while streaming static line items in `BillingEntryRepository.get_static_pending_by_subscription`.

> Rebased on latest `main`. Since this PR was opened, `main` independently landed the query-level `type != 'metered'` filter in `get_static_pending_by_subscription` (and dropped several unused `billing_entry` indexes in #—`71ec0612f44f`). This PR is now rebased onto that and reduced to the piece that isn't in `main`: the **partial index** that filter enables.

## What

- Add a partial index `ix_billing_entry_pending_static` on `billing_entry (subscription_id)` covering only pending non-metered entries (`WHERE deleted_at IS NULL AND order_item_id IS NULL AND type != 'metered'`), in the model and via a concurrent migration.
- Refresh the comment in `get_static_pending_by_subscription` to point at the index the existing `type != 'metered'` filter enables.
- Add a regression test for a subscription that has both static and metered pending entries.

## Why

`get_static_pending_by_subscription` already filters pending entries down to the static (non-metered) ones, but the driving scan still walked the **whole** pending set via `ix_billing_entries_s_d_oi_pp` and heap-checked `type` on every row. On high-volume metered subscriptions that pending set is millions of rows — so even with the filter, the scan blew past the statement timeout:

```
File "polar/order/service.py", line 1308, in create_subscription_order
File "polar/billing_entry/service.py", line 131, in compute_pending_subscription_line_items
File "polar/billing_entry/repository.py", line 81, in get_static_pending_by_subscription
...
asyncpg ... TimeoutError
```

The query-only filter (now in `main`) narrows the rows returned but doesn't stop the scan from visiting every pending metered row. The partial index does.

## How

Metered entries are the only ones created with `type=metered` (via `BillingEntry.from_metered_event`); every static entry (`cycle`/`proration`/`subscription_seats_*`) carries a static price. So "static" ≡ "non-metered" by construction, and a partial index keyed on that predicate contains only the handful of pending static entries per subscription.

- The existing `type != 'metered'` filter already **implies** the index predicate, so the planner drives the static lookup off this tiny index instead of the full pending scan — no query change required to benefit.
- Metered inserts fail the predicate, so they never touch the index: no write overhead on the ingestion hot path. This also complements #`71ec0612f44f`'s index cleanup rather than re-bloating writes.
- The migration creates the index `CONCURRENTLY` inside an `autocommit_block()`, matching the repo's existing index-migration convention. It is chained after `main`'s current head (`aaa111ca733d`), single-head.

## Checklist

> **Note on verification:** the DB-backed test suite runs in CI. Locally I verified everything that doesn't need a database — the compiled query's `WHERE` clause implies the partial-index predicate exactly, the model index DDL and the rendered migration SQL are identical, the migration chain is single-head onto `main`, and ruff/format/mypy are clean on the changed files. The two DB-dependent checks left to CI are the E2E tests (including the new mixed-scenario regression test) and `alembic check` (model ↔ migration parity for the partial index).

https://claude.ai/code/session_01ReQt23G1Raf8bbbLfAnKVQ

### DIFF
--- a/server/polar/billing_entry/repository.py
+++ b/server/polar/billing_entry/repository.py
@@ -71,6 +71,13 @@ class BillingEntryRepository(
     ) -> AsyncGenerator[BillingEntry]:
         statement = (
             self.get_pending_by_subscription_statement(subscription_id, cutoff=cutoff)
+            # Metered entries are the only ones created with `type=metered`, so
+            # the static ones are exactly the non-metered ones. Filtering on
+            # `type` lets the planner use the partial `ix_billing_entry_pending_static`
+            # index and skip the pending metered entries entirely, instead of
+            # scanning them all and joining `product_prices` just to discard them
+            # — which times out on high-volume subscriptions.
+            .where(BillingEntry.type != BillingEntryType.metered)
             .join(BillingEntry.product_price)
             .where(ProductPrice.is_static.is_(True))
             .options(
--- a/server/polar/billing_entry/repository.py
+++ b/server/polar/billing_entry/repository.py
@@ -75,9 +75,11 @@ class BillingEntryRepository(
             .join(BillingEntry.product_price)
             # Metered entries always carry a metered price: `from_metered_event`
             # is the only writer of this type, and it takes the price from a
-            # meter-scoped lookup. So this predicate selects the same rows as
-            # `is_static` alone, but lets the `billing_entry` scan drop them
-            # before the join to `events` rather than after it.
+            # meter-scoped lookup. So filtering on `type` selects the same rows as
+            # `is_static` alone, but lets the planner use the partial
+            # `ix_billing_entry_pending_static` index and drop the pending metered
+            # entries during the `billing_entry` scan instead of walking the whole
+            # pending set — which was timing out on high-volume subscriptions.
             .where(
                 BillingEntry.type != BillingEntryType.metered,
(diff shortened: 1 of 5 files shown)


server/polar/billing_entry/repository.py
```diff
@@ -71,6 +71,13 @@ async def get_static_pending_by_subscription(
     ) -> AsyncGenerator[BillingEntry]:
         statement = (
             self.get_pending_by_subscription_statement(subscription_id, cutoff=cutoff)
+            # Metered entries are the only ones created with `type=metered`, so
+            # the static ones are exactly the non-metered ones. Filtering on
+            # `type` lets the planner use the partial `ix_billing_entry_pending_static`
+            # index and skip the pending metered entries entirely, instead of
+            # scanning them all and joining `product_prices` just to discard them
+            # — which times out on high-volume subscriptions.
```
joebon: Let's drop this comment, it adds no value or context

joebon (approved):
Looks reasonable. I'd drop the silly AI comment but that's a non-blocker

maximevast (approved):

PR #13522, opened 2026-08-04
webhooks: Generate secrets spec compliantly

Yopi:
Now that we switched our secret prefix from polar_whs_ to whsec_ the secrets we generate doesn't seem to be fully [https://www.standardwebhooks.com/](spec compliant).

Namely: Secret serialization base64 encoded, prefixed with whsec_ for easy identification.

Standard Webhooks libraries expect the secret to be base64 encoded and then prefixed with whsec_. We expect you to base64 encode the full secret including the prefix.

This changes that, but signs the secrets in both ways to ensure backward compatibility.

<a href="https://cubic.dev/pr/polarsource/polar/pull/13522?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/webhook/tasks.py
+++ b/server/polar/webhook/tasks.py
@@ -1,4 +1,5 @@
 import base64
+import binascii
 from collections.abc import Mapping
 from ssl import SSLError
 from uuid import UUID
@@ -26,9 +27,19 @@ from polar.worker import (
     enqueue_job,
 )
 
+from .constants import WEBHOOK_SECRET_PREFIX
 from .service import webhook as webhook_service
 
 log: Logger = structlog.get_logger()
+
+
+def standard_webhooks_key(secret: str) -> bytes | None:
+    if not secret.startswith(WEBHOOK_SECRET_PREFIX):
+        return None
+    try:
+        return base64.b64decode(secret[len(WEBHOOK_SECRET_PREFIX) :], validate=True)
+    except binascii.Error:
+        return None
 
 
 # Safety-guard max_retries: enough for all ordering retries within the age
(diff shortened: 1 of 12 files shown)


server/polar/webhook/tasks.py
```diff
@@ -107,13 +118,16 @@ async def _webhook_event_send(
...
+    # Dual-signed: legacy raw-string-key verifiers must keep working
+    secret = event.webhook_endpoint.secret
+    b64secret = base64.b64encode(secret.encode("utf-8")).decode("utf-8")
+    signature = StandardWebhook(b64secret).sign(str(event.id), ts, event.payload)
+    standard_key = standard_webhooks_key(secret)
+    if standard_key is not None:
+        standard_signature = StandardWebhook(standard_key).sign(
+            str(event.id), ts, event.payload
+        )
+        signature = f"{standard_signature} {signature}"
```
frankie567, line 130: I feels like a dangerous breaking change. I'm quite sure there are naive implementations out there that will just read the full `webhook-signature` header without splitting on space first.

Yopi: Yeah I had a similar feeling. You are supposed to support multiple signatures to be able to do a roll out of new keys, but you never know what people do.

Not sure what would be a better way, basing it on when the secret was created maybe and only doing one signature?

frankie567: Yes, we probably need to store `secret_generated_at` in DB, and have a cutoff.

PR #13549, opened 2026-08-05
Hide Balance Events from timeline

emilwidlund:
- **fix customer selector**
- **hide balance events from timeline**

<a href="https://cubic.dev/pr/polarsource/polar/pull/13549?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/apps/web/src/components/Timeline/Timeline.tsx
+++ b/clients/apps/web/src/components/Timeline/Timeline.tsx
@@ -35,7 +35,11 @@ const buildSegments = (events: schemas['Event'][]): TimelineSegment[] => {
   }
 
   for (const event of events) {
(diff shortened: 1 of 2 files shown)


strandhvilliam (approved):
Usually not the biggest fan of having different logic on `null` vs `undefined`, as can be confusing.

Although in this case it kinda makes sense since undefined is a missing entry/fallback while null is to not show it. 🤷‍♂️

PR #13568, opened 2026-08-06
feat(compass): add compass thread persistence layer

strandhvilliam:
PR 1 of 3 in the _threads to compass_ POC

For more context, see the draft https://github.com/polarsource/polar/pull/13550

Adds the data layer for server-side Compass conversations: the compass_threads / compass_thread_messages tables, their models, and the repository/service that read and write them.

Nothing calls this yet, the API and frontend land in follow-ups. 

Follow up PR: https://github.com/polarsource/polar/pull/13574

<a href="https://cubic.dev/pr/polarsource/polar/pull/13568?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/polar/compass/threads/service.py
```diff
@@ -0,0 +1,309 @@
...
+    and the freshness rule in the system prompt has nothing to act on. The
+    date is the turn's date in the merchant's timezone, so it compares
+    directly against the "today" the run context provides.
+    """
+    for message in messages:
+        if not isinstance(message, ModelRequest):
+            continue
+        for part in message.parts:
+            if isinstance(part, ToolReturnPart) and isinstance(part.content, str):
+                part.content = f"[fetched {fetched_on.isoformat()}] {part.content}"
```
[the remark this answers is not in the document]
strandhvilliam: This is an artifact of having split the PR. The actual logic is changed in the next PR

server/polar/compass/threads/service.py
```diff
@@ -0,0 +1,309 @@
...
+
+
+def _title_from_prompt(prompt: str) -> str:
+    title = " ".join(prompt.split())
```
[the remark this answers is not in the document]
strandhvilliam: We sanitize this in next pr, result of the PR split

server/polar/compass/threads/service.py
```diff
@@ -0,0 +1,309 @@
...
+        )
+        thread = await repository.create(thread, flush=True)
+
+        overflow_statement = repository.apply_recency_order(
+            repository.get_readable_statement(auth_subject).where(
+                CompassThread.organization_id == organization_id
+            )
+        ).offset(THREADS_CAP)
+
+        for overflow in await repository.get_all(overflow_statement):
```
psincraian: I don't see the point in soft deleting those in the backend. I would simply ignore them in the frontend

strandhvilliam: Yeah I can see your point! I've been thinking about how and if we even should 'limit' the total amount of threads in available. But probably better as you say, just send in a pagination cursor and filter from there. The soft_deletion after thread cap pattern is a bit weird

server/polar/compass/threads/service.py
```diff
@@ -0,0 +1,309 @@
...
+                model_messages=model_messages,
+            ),
+            flush=True,
+        )
+        # Keep list order by last activity. Inserting a message does not
+        # update the thread row on its own.
+        await thread_repository.update(thread, update_dict={"modified_at": utc_now()})
+        return message
+
+    async def build_message_history(
```
psincraian, line 250: I see different problems:

1. At message 13 the system prompt dissapears, and this can be a problem if we don't inject it again.
2. Today's date is hardcoded in the main agent and not here, so going back will not provide the current date. But probably this is giving confusion to agents, as powerful models already have a tool for the current date time

server/polar/compass/threads/service.py
```diff
@@ -0,0 +1,309 @@
...
+                    session,
+                    thread_id,
+                    prompt=prompt,
+                    parts=parts,
+                    model_messages=model_messages,
+                )
+
+        return record
+
+    async def record_turn(
```
psincraian, line 222: I think here we should widen the required_scopes. For example, if in the first turn I ask about the metrics and in the second turn I ask about listing customers the scopes are not updated

server/polar/compass/assistant/agent.py
```diff
@@ -18,6 +18,9 @@
     *CUSTOMER_TOOLS_WITH_SCOPES,
 ]
 
+ASSISTANT_SCOPES: set[Scope] = {scope for _, scope in _ALL_TOOLS_WITH_SCOPES}
```
psincraian: Maybe for this compasss, I would use 5.6-luna as it has been my mainly driver in an agents with tools and it's really powerful and cheap!

server/polar/compass/threads/service.py
strandhvilliam: 1. Yes this is valid feedback as of this PR, but it's actually addressed in the next PR. (I think it was a bad idea to split the PR's so much 🤦‍♂️), basically using the instructions instead of system prompt its rebuild for every turn

Image

strandhvilliam: 2. Our idea was that tool calls made in the past should have a timestamp on them, so that the model is aware of this to get context of when something happend and compare to today. So that if we continue a thread on a later date and make same call it's less confusion.

At least that is the thinking. Maybe it's clearer in the next PR.

But yes, should probably rethink it a bit and verify that the model actually doesn't get confused

server/polar/compass/assistant/agent.py
strandhvilliam: Nice! Let's do it!

server/polar/compass/threads/service.py
strandhvilliam: Huh you are probably right, I just don't see it.

My thinking was that the thread on creation will get the user scopes (and assistant base scopes) which will be used on every turn onwards. 

But maybe I have misunderstood how scopes are provided on the auth_subject?

server/polar/compass/threads/service.py
strandhvilliam: Dropped the THREADS_CAP and soft deleting in favor of pagination instead.

server/polar/compass/threads/service.py
psincraian: Right, my example was wrong. Seems that scopes are the same on every turn.

But now I wonder if we really need this column at all. A dashboard login has almost all the scopes, and those depends on the role. So this always stores all the scopes and the check pass. Do we really need it?

strandhvilliam: Think you might be right, and I might have been overengineering this a bit so I really appreciate all the input!

I guess the goal was to keep the old hashing thinking from the version where we didn't persist anything. So like a polar access token couldn't see a thread it isn't supposed to, and if you as a user get your role downgraded, you shouldn't be able to read an old thread from when you had more permissions.

But I feel this is hard to reason about, so you're probably right to just drop the column and only care about the dashboard login scopes. 😅

psincraian: That makes sense. Do you plan to be able to use Compass from an API with access tokens?

If not, then I will keep things more simple

strandhvilliam: Dropped the column and simplified it a bit since a thread now belongs only to a specific user anyway.

Don't think that an org access token should be able to access it, as Compass is a dashboard thing. Kept the user tokens working with similar shape as the payouts module tough. Think it's simpler than my initial way.

psincraian (approved):
Nice job!

PR #13585, opened 2026-08-07
tf/ecs: Set up ECS cluster and service module

Yopi:
This spins up an ECS cluster in each environment, and adds a module to spin up ECS/Fargate services inside those clusters.

This is an alternative place to run Docker services.

<a href="https://cubic.dev/pr/polarsource/polar/pull/13585?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/terraform/modules/ecs_cluster/main.tf
+++ b/terraform/modules/ecs_cluster/main.tf
@@ -1,0 +1,18 @@
+resource "aws_ecs_cluster" "this" {
+  name = "polar-${var.environment}"
+
+  setting {
+    name  = "containerInsights"
+    value = "enabled"
+  }
+}
+
+resource "aws_ecs_cluster_capacity_providers" "this" {
+  cluster_name       = aws_ecs_cluster.this.name
+  capacity_providers = ["FARGATE", "FARGATE_SPOT"]
+
+  default_capacity_provider_strategy {
+    capacity_provider = "FARGATE"
+    weight            = 1
+  }
+}
--- a/terraform/modules/ecs_cluster/main.tf
+++ b/terraform/modules/ecs_cluster/main.tf
@@ -1,5 +1,6 @@
 resource "aws_ecs_cluster" "this" {
   name = "polar-${var.environment}"
+  tags = var.tags
 
   setting {
     name  = "containerInsights"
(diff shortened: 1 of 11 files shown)


terraform/modules/ecs_cluster/main.tf
```diff
@@ -0,0 +1,18 @@
...
+
+  setting {
+    name  = "containerInsights"
+    value = "enabled"
+  }
+}
+
+resource "aws_ecs_cluster_capacity_providers" "this" {
+  cluster_name       = aws_ecs_cluster.this.name
+  capacity_providers = ["FARGATE", "FARGATE_SPOT"]
```
psincraian, line 13: Just to take into account that Fargate is more limited than ECS. It don't allow to run Daemon on the same instance. 

I think it's fine to start as it's way easier, and migration should be smooth

terraform/modules/ecs_service/variables.tf
```diff
@@ -0,0 +1,87 @@
...
+}
+
+variable "command" {
+  description = "Container command override."
+  type        = list(string)
+  default     = null
+}
+
+variable "cpu" {
+  description = "Task CPU units."
```
psincraian: Maybe just to make things easier, I would try to abstract this a bit so we have some default instances.  Something like instanceType that can be small, mid, large

psincraian (approved):

frankie567 (commented):
For which cases will we need to run arbitrary Docker containers in AWS?

Yopi:
> For which cases will we need to run arbitrary Docker containers in AWS?

If we are replacing Render: It would things like the API service, metabase, etc

terraform/modules/ecs_service/variables.tf
Yopi: That could be nice, yeah

frankie567:
Crazy idea: couldn't we use a Lambda for the API backend? :D

terraform/modules/ecs_cluster/main.tf
Yopi: Hm, that's true. You can get most of it via sidecar containers that you would want a daemon for, but there might be something that we'd want

Yopi:
> Crazy idea: couldn't we use a Lambda for the API backend? :D

Haha, interesting! Let's discuss it 😄

PR #13594, opened 2026-08-10
org: Make benefit-grants readable without a write token

Yopi:
Without this it is not possible to do e.g. /benefit-grants as a user with only read only permissions (impersonation)

---
## Summary by cubic
Allow org admins with read-only sessions to view benefit grants, billing details, and payment methods. Adds a user-only auth guard so GET works without a write token; write actions remain blocked.

- New Features
  - Added `AuthorizeOrgManageUserRead` (user-only; allows `organizations:read` or `organizations:write`).
  - Switched GET to use it: GET /organizations/{id}/benefit-grants, /billing, /payment-methods; updated client schema to list `organizations:read`.
  - Tests (benefit-grants): 200 for read-only admin; 401 anonymous; 403 for non-admin member and for PATCH updates.

<sup>Written for commit be90cb8a9f7e8a82152cff944dda179d37f8f8d1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/13594?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/authz/dependencies.py
+++ b/server/polar/authz/dependencies.py
@@ -188,6 +188,20 @@ AuthorizeOrgManageUser = Annotated[
         )
     ),
 ]
+# User-only counterpart to `AuthorizeOrgManageRead`.
+AuthorizeOrgManageUserRead = Annotated[
+    AuthzContext[User],
+    Depends(
+        OrgPolicyGuard(
+            org_policy.can_manage,
+            allowed_subjects={User},
+            required_scopes={
+                Scope.organizations_read,
+                Scope.organizations_write,
+            },
+        )
+    ),
+]
 AuthorizeOrgAccess = Annotated[
     AuthzContext[User | Organization], Depends(OrgPolicyGuard())
 ]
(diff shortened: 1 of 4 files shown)


server/polar/authz/dependencies.py
```diff
@@ -188,6 +188,20 @@ async def dependency(
         )
     ),
 ]
+# User-only counterpart to `AuthorizeOrgManageRead`.
```
joebon, line 191: Why can we use `AuthOrgManageRead` directly then?

Yopi: I don't think we want to allow org tokens here since we are passing the user id of the caller `external_member_id=str(authz.auth_subject.subject.id)`

frankie567 (approved):

server/polar/authz/dependencies.py
joebon: Oh, these are polar self endpoints! Ok, thanks

joebon (approved):

PR #13599, opened 2026-08-10
chore: update to typescript v7 in web

simonxabris:
## Summary

Upgrades typescript to use v7 in web for typechecking

## Why

v7 is *much* faster than previous version because of the go rewrite. Numbers on my machine:

   | Compiler | Runs | Median | Average |
   | --- | ---: | ---: | ---: |
   | TypeScript 5.9.3 | 16.557–17.550s | **17.353s** | 17.159s |
   | TypeScript 7.0.2 | 2.962–3.406s | **3.122s** | 3.161s |

## How

Since typescript v7 does not offer a programmatic API that next and other tools use, I used the recommended method from the announcement blog post (https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/) to use v7 for typechecking but let programmatic consumers use v6.

## Checklist

### DIFF
--- a/clients/apps/web/package.json
+++ b/clients/apps/web/package.json
@@ -113,6 +113,7 @@
     "@types/react": "19.2.14",
     "@types/react-dom": "19.2.3",
     "@types/web": "^0.0.345",
+    "@typescript/native": "npm:typescript@7.0.2",
     "@vitejs/plugin-react": "^5.2.0",
     "@vitest/ui": "^4.1.4",
     "autoprefixer": "^10.5.0",
@@ -121,7 +122,7 @@
     "remark": "^15.0.1",
     "remark-mdx": "^3.1.1",
     "tailwindcss": "^4.2.2",
-    "typescript": "5.9.3",
+    "typescript": "npm:@typescript/typescript6@6.0.2",
     "vite-tsconfig-paths": "^5.1.4",
     "vitest": "^4.1.4"
   }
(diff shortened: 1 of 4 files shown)


strandhvilliam (approved):
LGTM! Looking forward to turbo fast type checking 😎

clients/apps/web/package.json
```diff
@@ -113,6 +113,7 @@
     "@types/react": "19.2.14",
     "@types/react-dom": "19.2.3",
     "@types/web": "^0.0.345",
+    "@typescript/native": "npm:typescript@7.0.2",
```
pieterbeulque, line 116: I'm not sure how it works, but I trust you!

pieterbeulque (approved):

strandhvilliam:
Screenshot 2026-08-10 at 16 23 00

I think the `knip` command stopped working though, for checking unused files. Don't know if this is still even used. You know @pieterbeulque ?

pieterbeulque:
@strandhvilliam Not really used, but not entirely useless either. Why does it not work anymore? I think it's a reasonably maintained package?

simonxabris:
@pieterbeulque @strandhvilliam I'll check whats wrong with knip and fix it.

simonxabris:
Knip is working again, turns out it has a peer dependency on ts which was resolved to the version installed in `web`, but since that changed to an incompatible version that broke it. Explicitly installing a compatible version in root where knip is fixes it.

strandhvilliam:
That's perfect! 😀

PR #13605, opened 2026-08-10
fix: fix `useDebounceCallback` race condition

simonxabris:
## Summary

**Related Issue**: polarsource/feedback#350

PR fixes the `useDebounceCallback` hook so there are no race condition when user does a second action within the 500ms debounce time.

## Why

Users could accidentally overwrite their filtering/pagination if they set their preferences really fast.

## How

Hook now stores the latest callback in a ref and that ref is updated when sorting or pagination changes. This means that when the debounce timer expires in always has access to the latest change.

## Checklist

### DIFF
--- a/clients/apps/web/src/hooks/utils.ts
+++ b/clients/apps/web/src/hooks/utils.ts
@@ -1,4 +1,10 @@
-import { useCallback, useEffect, useRef, useState } from 'react'
+import {
+  useCallback,
+  useEffect,
+  useLayoutEffect,
+  useRef,
+  useState,
+} from 'react'
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const useDebouncedCallback = <T extends (...args: any[]) => any>(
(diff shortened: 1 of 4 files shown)


maximevast (approved):
LGTM!

clients/apps/web/src/hooks/utils.ts
```diff
@@ -1,25 +1,45 @@
...
   callback: T,
   delay: number,
-  dependencies?: unknown[],
 ) => {
-  const timeout = React.useRef<NodeJS.Timeout>(undefined)
+  const timeout = useRef<ReturnType<typeof setTimeout> | undefined>(undefined)
+  const callbackRef = useRef(callback)
 
-  return React.useCallback(
+  useLayoutEffect(() => {
```
pieterbeulque, line 17: Unexpected, but I think it's valid here

pieterbeulque (approved):

clients/apps/web/src/hooks/utils.ts
simonxabris: It's used here, because layout effects are applied synchronously, while simple effects are not so theoretically there can be a race condition between rendering and the effect being applied and we could end up reading a stale value.

PR #13617, opened 2026-08-11
fix(events): render subscription event cards from event metadata

pieterbeulque:
## Summary

`SubscriptionEventCard` rendered the **live** subscription's product name and current price, which doesn't reflect what each event was actually about. This sources the card's content from `event.metadata` instead — the data recorded at the time of the event.

## What

- Split `SubscriptionEventCard` into two sub-cards dispatched on `event.name`, dropping the `useSubscription` fetch.
- **`subscription.cycled` / `subscription.revoked`**: show the event's own product (`metadata.product_id`) and the `amount` / `currency` / `recurring_interval` / `recurring_interval_count` **as recorded on the event** (via `AmountLabel`), instead of the subscription's current price.
- **`subscription.product_updated`**: show the **old → new** product transition (`old_product_id` → `new_product_id`) instead of the current product.
- The subscription-detail link is preserved via `metadata.subscription_id` (always present).

## Why

For an event log, the current subscription state is the wrong source of truth — it describes the subscription *as it is now*, not the event. This was especially misleading for `subscription.product_updated`: the event exists because the product *changed*, yet the card showed only the current product, hiding the transition entirely. `cycled` / `revoked` similarly showed a re-derived current price rather than the amount actually billed/revoked.

## How

Each event type already carries the relevant snapshot in its metadata, so the card reads from there. Optional metadata fields (`product_id`, `amount`, `currency`, …) are rendered only when present, so a `cycled`/`revoked` event without a product id renders immediately (the disabled `useProduct(undefined)` query never enters a loading state) rather than showing a perpetual "Loading…".

## Checklist

- [ ] New functionality is covered by tests — this is a presentational fix to an existing component; the `EventCard` family has no unit tests, and none were added.
- [x] Linting and type checking pass — frontend change: `pnpm typecheck` (tsc --noEmit), `eslint --max-warnings 0`, and `oxfmt --check` all pass.

https://claude.ai/code/session_015X8ZNhNTaS1LYQtheCwEik

---
_Generated by [Claude Code](https://claude.ai/code/session_015X8ZNhNTaS1LYQtheCwEik)_

<a href="https://cubic.dev/pr/polarsource/polar/pull/13617?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/apps/web/src/components/Events/EventCard/SubscriptionEventCard.tsx
+++ b/clients/apps/web/src/components/Events/EventCard/SubscriptionEventCard.tsx
@@ -1,11 +1,12 @@
-import ProductPriceLabel from '@/components/Products/ProductPriceLabel'
-import { useSubscription } from '@/hooks/queries'
+import AmountLabel from '@/components/Shared/AmountLabel'
+import { useProduct } from '@/hooks/queries'
 import { OrganizationContext } from '@/providers/maintainerOrganization'
 import AllInclusiveOutlined from '@mui/icons-material/AllInclusiveOutlined'
+import ArrowForwardOutlined from '@mui/icons-material/ArrowForwardOutlined'
 import { schemas } from '@polar-sh/client'
 import { Status, type StatusColor } from '@polar-sh/orbit'
 import Link from 'next/link'
-import { useContext, useMemo } from 'react'
+import { useContext } from 'react'
 import { EventCardBase } from './EventCardBase'
 
 export interface SubscriptionEventCardProps {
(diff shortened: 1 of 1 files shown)


joebon (approved):
lgtm

PR #13621, opened 2026-08-11
fix(license-key): drive the benefit grant when key status changes

maximevast:
Superseeds https://github.com/polarsource/polar/pull/12659

## Problem

Changing a license key's status from the dashboard wrote only the key. The parent benefit grant kept its old state.

Two symptoms:

- Key re-enabled, grant still revoked — the key validates, but the customer portal hides it.
- Key revoked, grant still granted — the portal advertises a benefit that fails validation.

Just over 400 grants in production disagree with their key, across more than 120 organizations. The oldest dates to December 2024 and new ones are still accruing.

## Fix

A status change now enqueues `benefit.grant` or `benefit.revoke`, so the grant follows the key through the normal lifecycle.

A grant is revoked exactly when its key is revoked. `granted` and `disabled` both keep the entitlement — the portal already renders a disabled key.

## Backfill

`scripts.backfill_license_key_grant_state` realigns the existing rows: 307 to revoke, 96 to grant. Dry-run by default.

It writes the timestamps directly and emits no webhooks. Some of these rows diverged over a year ago; replaying them as fresh grant activity would misinform every subscriber.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/13621?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/license_key/service.py
+++ b/server/polar/license_key/service.py
@@ -183,13 +183,7 @@ class LicenseKeyService:
         if already_applied:
             return
 
-        enqueue_job(
-            "benefit.revoke" if revoke else "benefit.grant",
-            customer_id=grant.customer_id,
-            benefit_id=grant.benefit_id,
-            member_id=grant.member_id,
-            **grant.scope,
-        )
+        enqueue_job("license_key.sync_benefit_grant", license_key_id=license_key.id)
 
     async def validate(
         self,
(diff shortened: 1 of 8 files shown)


Yopi (approved):
LGTM.

I think the bot might have a point in regards of having stale state encoded in the tasks, but not sure if there would be a better way to solve it without adding a new task that reads the current status

PR #13625, opened 2026-08-11
feat(merchant-migration): link the moved cards to imported customers

psincraian:
## Summary

Read the cards that have landed on Polar's Stripe account for an imported customer, store them, and return the one their subscription was already being charged on.

## Why

PAN copy preserves the source `cus_…` id but creates new `pm_…` ids, so a card can be on Polar's Stripe account and still be invisible to Polar: the renewal path charges `subscription.payment_method_id or customer.default_payment_method_id`, and neither exists until we do this.

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_011PzjzyWCvUEeXWuG4UGy8R)_

<a href="https://cubic.dev/pr/polarsource/polar/pull/13625?utm_source=github" rel="nofollow noreferrer noopener" target="_blank">``&lt;img alt="Review in cubic" src="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"&gt;``</a>

### DIFF
--- a/server/polar/merchant_migration/cards.py
+++ b/server/polar/merchant_migration/cards.py
@@ -19,13 +19,22 @@ from polar.models import Customer, PaymentMethod
 from polar.payment_method.service import payment_method as payment_method_service
 from polar.postgres import AsyncSession
 
+from .canonical import CanonicalPaymentMethod
+
 CARD_TYPE = "card"
 
 
 async def link_payment_method(
-    session: AsyncSession, customer: Customer
+    session: AsyncSession,
+    customer: Customer,
+    *,
+    source_method: CanonicalPaymentMethod | None = None,
 ) -> PaymentMethod | None:
     """Mirror the cards on Polar's Stripe account onto the Polar customer.
+
+    ``source_method`` is the method the source subscription was charging; the
+    copy of it wins, so a customer with several stored cards keeps being charged
+    on the one they were already being charged on.
 
     Returns the payment method to charge, or None while nothing has landed for
     this customer yet. Idempotent: re-running picks up cards that arrived since,
@@ -56,7 +65,9 @@ async def link_payment_method(
     if not payment_methods:
         return None
 
-    preferred = _preferred(payment_methods, customer.default_payment_method_id)
+    preferred = _preferred(
+        payment_methods, customer.default_payment_method_id, source_method
+    )
     # The renewal falls back to the customer default when a subscription has no
     # method of its own, so an imported customer should have one.
     if customer.default_payment_method_id is None:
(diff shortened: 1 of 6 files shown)


server/polar/merchant_migration/cards.py
```diff
@@ -0,0 +1,103 @@
...
+def _is_copy_of(
+    payment_method: PaymentMethod, source_method: CanonicalPaymentMethod
+) -> bool:
+    """A copy keeps the card but not its id, so match on what survives."""
+    details = {
+        "last4": source_method.last4,
+        "brand": source_method.brand,
+        "exp_month": source_method.exp_month,
+        "exp_year": source_method.exp_year,
+    }
```
frankie567, line 116: I'm no expert but I guess it's possible to have several cards with the same last 4 digits and same expiration date. Could we leverage fingerprint instead?

psincraian: I did a small test in two Stripe accounts that I have and the fingerprint changes 🥲

psincraian: Maybe I can raise an error when this happens and we take a look manually to see if we find something

server/polar/merchant_migration/cards.py
```diff
@@ -0,0 +1,103 @@
...
+from polar.models import Customer, PaymentMethod
+from polar.payment_method.service import payment_method as payment_method_service
+from polar.postgres import AsyncSession
+
+from .canonical import CanonicalPaymentMethod
+
+CARD_TYPE = "card"
+
+
+async def link_payment_method(
```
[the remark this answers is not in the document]
psincraian: Those bots... This will be done in a follow up PR.

server/polar/merchant_migration/cards.py
psincraian: Done — it raises now instead of picking one.

`_preferred` collects **all** the copies that match instead of returning the first, and two or more is `AmbiguousCopiedCard`. No status code on it deliberately, so it lands in Sentry as a hard error rather than being rendered as a normal business outcome: charging the wrong card is worse than not charging, and there's nothing sensible to guess between two identical-looking cards.

Zero matches is unchanged — that's "the card didn't land", not ambiguity, and it still falls through to the existing ladder.

On the fingerprint: matches what the API reference implies. The uniform-fingerprint guarantee is scoped to a Connect platform and *its connected accounts*, and the merchant's account isn't connected to ours — so there was never a version of this that worked across a copy. Noted in the code so nobody re-tries it:

```python
"""A copy keeps the card but not its id, and not its fingerprint either —
that is per-account. Brand, last4 and expiry are what is left."""
```

One caveat worth flagging: the callers don't handle this yet. In the cutover it would fail the whole chain rather than just that subscription, so I'd catch it there and record the subscription as `failed` with the reason — that keeps the rest of the migration moving and still surfaces it for a manual look. Happy to do that in #13626 if you agree it's the right shape.

There's also an exact answer available if we want to stop guessing entirely: Stripe uploads a **CSV mapping old → new ids** to the recipient's Dashboard Documents after a copy, and we're the recipient. It needs an Ops step to fetch, so it fits the PAN checklist rather than this function — but it would make this whole heuristic unnecessary.

---

frankie567 (approved):

PR #13613, opened 2026-08-11
feat(email): dedup reminders on deduplication_key instead of email_props

joebon:
Follow up to #13456
Waiting for #13633

Backfill has been applied already, in sandbox and production

---
## Summary by cubic
Switch reminder email deduplication to `EmailLog.deduplication_key` and add a partial unique index to prevent duplicates per recipient. Stable SQL-generated keys keep repo queries aligned with the sender and avoid dupes when props change.

- **Refactors**
  - Use `deduplication_key` comparisons in `payment_method` and `subscription` repositories via new SQL helpers.
  - Added `payment_method_expiration_reminder_key_sql`, `subscription_renewal_reminder_key_sql`, and `subscription_trial_conversion_reminder_key_sql`.
  - Updated tests to match sender-written keys and verify suppression for already-reminded periods, including renewal/trial windows and prior-period behavior.

- **Migration**
  - Add partial unique index `ix_email_logs_deduplication_key` on (`deduplication_key`, `to_email_addr`) where the key is not null, allowing multiple nulls but enforcing one reminder per key+recipient.

<sup>Written for commit f1914146c0a572e31e70ee27421d2715d30367b0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/13613?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #13633, opened 2026-08-11
feat(email): add index on email_logs deduplication_key

joebon:
Preparing for #13613

---
## Summary by cubic
Add a concurrent, partial (WHERE deduplication_key IS NOT NULL) non-unique index on email_logs(deduplication_key) to speed up the reminder scanner’s dedup lookup. The migration drops any invalid leftover before creating it, and the same partial index is defined in the model; dedup stays in application logic via NOT EXISTS.

<sup>Written for commit 44be3f7e16bd77ebd092db6f9d7f2b84a8e546aa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/13633?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/migrations/versions/2026-08-11-1505_add_index_on_email_logs_deduplication_key.py
+++ b/server/migrations/versions/2026-08-11-1505_add_index_on_email_logs_deduplication_key.py
@@ -1,0 +1,47 @@
+"""add index on email_logs deduplication_key
+
+Revision ID: 89e805c471b0
+Revises: 9a1c47f0b3d2
+Create Date: 2026-08-11 15:05:09.141659
+
+"""
+
+from alembic import op
+
+# Polar Custom Imports
+
+# revision identifiers, used by Alembic.
+revision = "89e805c471b0"
+down_revision = "9a1c47f0b3d2"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+INDEX = "ix_email_logs_deduplication_key"
+
+
+def upgrade() -> None:
+    with op.get_context().autocommit_block():
+        # Drop any INVALID leftover from an interrupted concurrent build first.
+        op.drop_index(
+            INDEX,
+            table_name="email_logs",
+            if_exists=True,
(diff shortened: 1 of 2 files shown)


server/migrations/versions/2026-08-11-1505_add_index_on_email_logs_deduplication_key.py
```diff
@@ -0,0 +1,48 @@
...
+
+def upgrade() -> None:
+    with op.get_context().autocommit_block():
+        # Drop any INVALID leftover from an interrupted concurrent build first.
+        op.drop_index(
+            INDEX,
+            table_name="email_logs",
+            if_exists=True,
+            postgresql_concurrently=True,
+        )
```
Yopi, line 30: What is this?

Yopi (approved):

server/migrations/versions/2026-08-11-1505_add_index_on_email_logs_deduplication_key.py
joebon: @Yopi when creating indexes concurrently and it fails there will be a broken index existing, so it needs to be dropped before trying to create one (again). That's how we handle failures and retries of such indices afaik

Yopi: Ah, so we always pre-empt it now?

joebon: Yes exactly, this construction deletes such an index first if it exists, then proceeds to create it

PR #13650, opened 2026-08-12
Clear customer_tax_id when disabling business purchase on checkout

### LINKED ISSUE #13649: Checkout: disabling "purchase as a business" doesn't clear customer_tax_id, leaving stale reverse-charge tax on the order
Toggling off "purchase as a business" after entering a VAT ID does not clear `Checkout.customer_tax_id`, so the checkout can complete as a personal purchase while still carrying a business VAT ID and 0% reverse-charge tax treatment on the resulting order/invoice.

**Repro flow:**

1. Enable "purchase as a business" and enter a valid VAT ID → `PATCH {customer_tax_id: "..."}` is stored, tax recalculates to reverse-charge (0% VAT).
2. Disable "purchase as a business" before completing checkout → frontend only sends `PATCH {is_business_customer: false}`; `customer_tax_id` is never resent as `null` (the tax-ID field unmounts via `shouldUnregister: true` in `CheckoutFormProvider`, so it's dropped from the form payload without an explicit clear).
3. Backend `_update_checkout` in `server/polar/checkout/service.py` only nulls `customer_tax_id` when the field is explicitly present with `null` in the request (`"customer_tax_id" in checkout_update.model_fields_set`). Since it's absent, the previously-stored VAT ID is left untouched (and even re-validated).
4. Tax calculation (`_update_checkout_tax` → Stripe Tax) keys purely on whether `customer_tax_id` is set, independent of `is_business_customer`, so reverse-charge/0% VAT persists through confirm and lands on the final order/invoice and Stripe customer tax ID, even though `is_business_customer = false`.

**Expected:** turning off "purchase as a business" should clear `customer_tax_id` (and business billing name) and revert the tax calculation to standard/non-reverse-charged treatment.

**Fix candidates:**

* Frontend: send `customer_tax_id: null` (and clear business billing name) alongside `is_business_customer: false` when unchecking.
* Backend (more robust, also covers non-UI API updates): in `_update_checkout`, enforce that `is_business_customer = false` implies `customer_tax_id = null`, similar to the existing reset pattern in `_reset_payment_form_fields_if_needed`.

Related: #8889 (resets `is_business_customer` on payment-form-not-required), PR #13572 (separate fix for clearing `tax_id` on the Customer/Customer Portal API, same class of bug on a different field).

<sub>Sent by @joebon from [Plain thread investigation](https://app.stilla.ai/m/memo_01kzta843tf9386ra1dpjp8tp1).</sub>

joebon:
## Summary by cubic
Clears business tax details when a user turns off “Purchase as a business” in checkout, so VAT is calculated correctly. Prevents stale VAT IDs from causing reverse-charge (0% VAT) on orders and invoices. Fixes #13649.

- **Bug Fixes**
  - When `is_business_customer` is set to false, clear `customer_tax_id`, `customer_tax_id_number`, and `customer_billing_name`.
  - Added a test to confirm these fields are cleared when disabling business purchase.

<sup>Written for commit 4ad65ec20691ee7e03ca1dd06bdb616f5c2921eb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/13650?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/checkout/service.py
+++ b/server/polar/checkout/service.py
@@ -2221,7 +2221,17 @@ class CheckoutService:
         if checkout_update.customer_billing_address:
             checkout.customer_billing_address = checkout_update.customer_billing_address
 
-        if (
+        # Turning off business purchase must clear the business billing fields.
+        # Otherwise a stale tax ID keeps reverse-charge (0% VAT) treatment on the
+        # order and invoice even though the checkout is no longer a business purchase.
+        disabling_business_customer = (
+            checkout_update.is_business_customer is False
+            and "is_business_customer" in checkout_update.model_fields_set
+        )
+        if disabling_business_customer:
+            checkout.customer_billing_name = None
+
+        if disabling_business_customer or (
             checkout_update.customer_tax_id is None
             and "customer_tax_id" in checkout_update.model_fields_set
         ):
--- a/server/tests/checkout/test_service.py
+++ b/server/tests/checkout/test_service.py
@@ -3511,6 +3511,31 @@ class TestUpdate:
         assert checkout.customer_billing_address is not None
         assert checkout.customer_billing_address.country == "US"
 
+    async def test_disabling_business_customer_clears_tax_id(
+        self,
+        session: AsyncSession,
+        save_fixture: SaveFixture,
+        checkout_one_time_custom: Checkout,
+    ) -> None:
+        checkout_one_time_custom.is_business_customer = True
+        checkout_one_time_custom.customer_billing_name = "ACME Inc."
+        checkout_one_time_custom.customer_tax_id = (
+            "FR61954506077",
+            TaxIDFormat.eu_vat,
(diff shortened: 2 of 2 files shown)


pieterbeulque (approved):
I thought this was going to be a front-end fix, but this feels solid.

joebon:
> I thought this was going to be a front-end fix, but this feels solid.

Yeah, we'll need the same coupled clearing for non-frontend triggered API requests also so I think this is better @pieterbeulque

PR #13653, opened 2026-08-12
Fix repeating discounts in interval-change charge previews

frankie567:
## Summary

- preserve the next renewal timestamp before applying a pending subscription update in charge previews
- evaluate repeating discount expiration at the start of the upcoming billing period
- cover monthly-to-yearly pending updates with a regression test

## Root cause

Applying an interval-changing pending update mutates `current_period_end` to the end of the new interval before the preview checks discount repetition. The preview therefore evaluated expiration as much as a year too late instead of at the renewal boundary.

## Impact

Charge previews now match the actual subscription cycle behavior, so customers see repeating discounts that will still apply to their next charge.

## Testing

- `uv run pytest tests/subscription/test_service_charge_preview.py -q` (14 passed)
- `uv run task lint_check`
- `uv run task lint_types`

Fixes polarsource/feedback#303

<a href="https://cubic.dev/pr/polarsource/polar/pull/13653?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/tests/subscription/test_service_charge_preview.py
+++ b/server/tests/subscription/test_service_charge_preview.py
@@ -360,6 +360,56 @@ class TestCalculateChargePreview:
 
         assert preview.base_amount == 1000
 
+    async def test_pending_interval_update_keeps_repeating_discount(
+        self,
+        session: AsyncSession,
+        save_fixture: SaveFixture,
+        organization: Organization,
+        product: Product,
+        customer: Customer,
+    ) -> None:
+        discount = await create_discount(
+            save_fixture,
+            type=DiscountType.percentage,
+            basis_points=2500,
+            duration=DiscountDuration.repeating,
+            duration_in_months=2,
+            organization=organization,
+        )
+        annual_product = await create_product(
+            save_fixture,
+            organization=organization,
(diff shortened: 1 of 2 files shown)


server/tests/subscription/test_service_charge_preview.py
```diff
@@ -360,6 +360,56 @@ async def test_pending_update_archived_prices_falls_back_to_current_product(
...
+        subscription = await create_active_subscription(
+            save_fixture,
+            product=product,
+            customer=customer,
+            discount=discount,
+            current_period_start=datetime(2024, 1, 1, tzinfo=UTC),
+            current_period_end=datetime(2024, 2, 1, tzinfo=UTC),
+        )
+
+        from polar.subscription.update import generate_subscription_update
```
frankie567: Add this import at the top.

PR #13656, opened 2026-08-12
refactor: migrate data table search parameters to nuqs

strandhvilliam:
## Summary

Refactors discounts, custom fields and webhook/endpoint page to use nuqs for managing query filters instead of the custom made solution which caused some unnecessary roundtrips. Improves DX and reduces complexity.

Matches previous refactors accomplishing the same for orders table, sales table etc.

<a href="https://cubic.dev/pr/polarsource/polar/pull/13656?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/custom-fields/CustomFieldsPage.tsx
```diff
@@ -132,7 +77,7 @@ const ClientPage: React.FC<ClientPageProps> = ({
 
   const customFieldsHook = useCustomFields(organization.id, {
     ...getAPIParams(pagination, sorting),
-    type,
+    type: type ?? undefined,
```
Comment (pieterbeulque) [inline clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/custom-fields/CustomFieldsPage.tsx:80]: Is this to map `null` to `undefined`?

pieterbeulque (approved):
Blind approval but I like the clean-up! Since we don't fetch any data I assume there's no harm in moving it from SSR to client-side?

clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/custom-fields/CustomFieldsPage.tsx
Comment (strandhvilliam) [reply clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/custom-fields/CustomFieldsPage.tsx:80]: Yeah, nuqs returns `null` for missing parameter  but the api itself expects `undefined`

strandhvilliam:
Yeah, we fetch that data client side anyway like you say. And by updating the query parameters, previously parsing of the query parameters would have needed to make a roundtrip to the server component instead of just directly using the updated value in client.

Not like it will be a noticeable improvement but we had no advantage of SSR here anyway. It cleans it up and nuqs has good DX. If we need the parameters server side, nuqs has easy ways to get them there too.

clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/custom-fields/CustomFieldsPage.tsx
```diff
@@ -29,75 +25,24 @@ import {
...
-      | DataTableSortingState
-      | ((old: DataTableSortingState) => DataTableSortingState),
-  ) => {
-    const updatedSorting =
-      typeof updaterOrValue === 'function'
-        ? updaterOrValue(sorting)
-        : updaterOrValue
+const ClientPage: React.FC<ClientPageProps> = ({ organization }) => {
+  const { pagination, setPagination, sorting, setSorting } =
+    useDataTableQueryState({
```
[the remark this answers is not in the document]
Comment (strandhvilliam) [reply clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/settings/custom-fields/CustomFieldsPage.tsx:41]: This is along the other patterns we've used for nuqs

PR #13662, opened 2026-08-12
fix(customer): top customers avatar falls back to owner email

pieterbeulque:
## Summary

**Related Issue**: Closes polarsource/feedback#354

Team customers created without a direct email got `avatar_url: null` from `GET /v1/customers/top`, while the customer **detail** and **create** endpoints correctly return an avatar derived from the owner member's email. This fixes that inconsistency so the same customer no longer shows an avatar everywhere except the top-customers widget.

## What

- `OrderRepository.get_revenue_by_customer` now ranks on and returns the `Customer` entity — `(Customer, order_count, net_revenue)` — instead of scalar `(id, email, name, …)` columns.
- The `/v1/customers/top` endpoint builds `TopCustomer` from the `Customer` object, so `avatar_url` comes straight from the `Customer.avatar_url` property.
- The compass assistant's `top_customers_by_revenue` tool consumes the same repository method and now also derives its avatar from `Customer.avatar_url` (same latent bug, fixed consistently).
- Added a regression test: a team customer with `email=None` plus an owner member returns the owner-email avatar from `/top`.

## Why

`Customer.avatar_url` already implements the fallback:

```python
email = self.email or (self.owner.email if self.owner else None)
return _avatar_url_for_email(email) if email else None
```

The `/top` endpoint replicated only the `email`-based half (`_avatar_url_for_email(email) if email else None`), so team customers without an email showed no avatar despite having one on their detail page and at creation. Deriving from the model property gives a single source of truth.

## How

Selecting the full `Customer` entity (grouped by its primary key) lets the `owner` relationship — already `lazy="selectin"` — load automatically, so `Customer.avatar_url`'s owner fallback resolves with no extra query wiring. Both callers of the shared repository method were updated to the new tuple shape.

## Checklist

- [ ] Linting and type checking pass — `ruff check` passes; `uv run task lint_types` (mypy) and the pytest suite could **not** be run in this ephemeral environment (the pinned Python 3.14.6 toolchain and Postgres aren't provisioned here). The SQLAlchemy query change was validated in an isolated SQLAlchemy repro (entity select + `select_from`/`join` + group-by-on-PK + `selectin` owner load).

https://claude.ai/code/session_011gv3a3vnBRbyBdiayBV5oL

---
_Generated by [Claude Code](https://claude.ai/code/session_011gv3a3vnBRbyBdiayBV5oL)_

<a href="https://cubic.dev/pr/polarsource/polar/pull/13662?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/order/repository.py
+++ b/server/polar/order/repository.py
@@ -65,10 +65,6 @@ class OrderRepository(
         """Customers ranked by paid net revenue: (customer, order count, net
         revenue in cents), descending.
 
-        Returns the `Customer` entities so callers can derive display fields
-        (e.g. `avatar_url`, which falls back to the owner member's email for
-        team customers) from a single source of truth.
-
         Partially refunded orders count with the refunded portion subtracted,
         so the ranking reflects money actually kept.
 
--- a/server/polar/order/repository.py
+++ b/server/polar/order/repository.py
@@ -61,9 +61,13 @@ class OrderRepository(
         start: datetime | None = None,
         end: datetime | None = None,
         limit: int = 10,
-    ) -> Sequence[tuple[UUID, str | None, str | None, int, int]]:
-        """Customers ranked by paid net revenue: (customer_id, email, name,
-        order count, net revenue in cents), descending.
+    ) -> Sequence[tuple[Customer, int, int]]:
+        """Customers ranked by paid net revenue: (customer, order count, net
+        revenue in cents), descending.
+
+        Returns the `Customer` entities so callers can derive display fields
+        (e.g. `avatar_url`, which falls back to the owner member's email for
+        team customers) from a single source of truth.
 
         Partially refunded orders count with the refunded portion subtracted,
         so the ranking reflects money actually kept.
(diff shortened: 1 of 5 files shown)


server/polar/order/repository.py
```diff
@@ -77,19 +81,18 @@ async def get_revenue_by_customer(
         )
         statement = (
             select(
-                Customer.id,
-                Customer.email,
-                Customer.name,
+                Customer,
```
pieterbeulque, line 80: Assuming the column selection wasn't on purpose for slow query reasons

server/polar/order/repository.py
```diff
@@ -61,9 +61,13 @@ async def get_revenue_by_customer(
...
-    ) -> Sequence[tuple[UUID, str | None, str | None, int, int]]:
-        """Customers ranked by paid net revenue: (customer_id, email, name,
-        order count, net revenue in cents), descending.
+    ) -> Sequence[tuple[Customer, int, int]]:
+        """Customers ranked by paid net revenue: (customer, order count, net
+        revenue in cents), descending.
+
+        Returns the `Customer` entities so callers can derive display fields
+        (e.g. `avatar_url`, which falls back to the owner member's email for
+        team customers) from a single source of truth.
```
pieterbeulque:
```suggestion
```

frankie567 (approved):
The fix is good.

However, that `/top` endpoint is a bit misguided at first glance. We could have achieved a similar thing with the `list` endpoint by adding sorting criterions and filters.

strandhvilliam:
Think the /top etc endpoints are from https://github.com/polarsource/polar/pull/13609.

Maybe a pattern like the metrics endpoint? Like with the different metric slugs but for customers?

strandhvilliam (approved):

PR #13673, opened 2026-08-12
config: Don't mask the DB password in DB config

Yopi:
Without this we connect with a masked password which fails connection. This is likely the root cause why we couldn't connect to pgbouncer when we rolled it out initially.

---
## Summary by cubic
Fixes DB connections by emitting an unmasked Postgres DSN from config, ensuring the real password is used (including with pgbouncer). Fallback host/port behavior remains intact for both drivers.

- **Bug Fixes**
  - Build DSN with URL.render_as_string(hide_password=False) so the password isn’t masked.
  - Added tests for `asyncpg` and `psycopg2` covering no-fallback and fallback cases, password passthrough, and defaulting fallback port to the primary port.

<sup>Written for commit 0bf299855aa56c9bc8d3c8e7cf3ee072defba665. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/13673?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/config.py
+++ b/server/polar/config.py
@@ -612,20 +612,18 @@ class Settings(BaseSettings):
                     path=database,
                 )
             )
-        return str(
-            URL.create(
-                f"postgresql+{driver}",
-                username=username,
-                password=password,
-                database=database,
-                query={
-                    "host": [
-                        f"{host}:{port}",
-                        f"{fallback_host}:{fallback_port or port}",
-                    ]
-                },
-            )
-        )
+        return URL.create(
+            f"postgresql+{driver}",
+            username=username,
+            password=password,
+            database=database,
+            query={
+                "host": [
+                    f"{host}:{port}",
+                    f"{fallback_host}:{fallback_port or port}",
+                ]
+            },
+        ).render_as_string(hide_password=False)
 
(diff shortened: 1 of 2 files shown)


frankie567 (approved):
Makes sense, but that's still surprising given we were using this code to connect with SQLAlchemy without any issue 🤔

Yopi:
> Makes sense, but that's still surprising given we were using this code to connect with SQLAlchemy without any issue 🤔

This path is only used when we have multiple hosts specified. Which we did when we were trialing pgbouncer before the summer and it crashed, so I think this is the reason why we got incorrect password back then 😄

frankie567:
> This path is only used when we have multiple hosts specified. Which we did when we were trialing pgbouncer before the summer and it crashed, so I think this is the reason why we got incorrect password back then 😄

Oh I see, you use `URL` directly instead of `PostgresDsn` 🤔 Could be worth to consolidate both branches, because it's quite confusing like this

PR #13674, opened 2026-08-12
Fix Sentry reporting from Lambda workers

### LINKED ISSUE #13366: Investigate why AWS Lambda errors aren't showing up in Sentry
Following [INC-7](https://polar-sh.slack.com/archives/C0BKEMZ9QE9/p1784878975164389) (Lambda workers not running in sandbox), we noticed errors from the AWS Lambda workers aren't appearing in Sentry. Investigate why Lambda error reporting isn't reaching Sentry and fix the integration.

<sub>Sent by @frankie567 from [AWS Lambda Sentry error tracking investigation](https://app.stilla.ai/m/memo_01ky9h7m4ffg2tv2e73dmdp2kd).</sub>

frankie567:
## Summary

- enable Sentry's AWS Lambda integration for the Lambda worker entrypoint
- explicitly capture task and retry-backoff exceptions that are intentionally handled by the SQS partial-batch flow

## Root cause

The worker caught task exceptions so it could apply retry backoff and return an SQS partial-batch response. Sentry's logging integration is configured for breadcrumbs only, so those handled exceptions never became Sentry events. Default Sentry integrations are disabled globally, which also meant the AWS Lambda integration was absent and the Lambda lifecycle did not drain the Sentry transport.

Fixes #13366

## Validation

- `uv run task lint`
- `uv run task lint_types`
- ADR compliance check — no violations

### DIFF
--- a/server/polar/sentry.py
+++ b/server/polar/sentry.py
@@ -6,6 +6,7 @@ import sentry_sdk
 from dramatiq import get_broker
 from sentry_sdk.integrations.argv import ArgvIntegration
 from sentry_sdk.integrations.atexit import AtexitIntegration
+from sentry_sdk.integrations.aws_lambda import AwsLambdaIntegration
 from sentry_sdk.integrations.dedupe import DedupeIntegration
 from sentry_sdk.integrations.dramatiq import DramatiqIntegration as _DramatiqIntegration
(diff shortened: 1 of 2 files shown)


Yopi (approved):
🙌

PR #13687, opened 2026-08-12
server/order: stop the stale payment lock sweep from scanning every order

maximevast:
## Summary

The hourly stale payment lock sweep has been timing out since Jul 29 — 137 failures.
A timeout enqueues nothing, so stale locks go unreleased.
https://logfire-us.pydantic.dev/polar/polar/?q=span_name+like+%27%25stale_payment_lock%25%27&since=2026-07-29T00%3A00%3A00Z

`Order.is_payment_lock_stale.is_(True)` renders as `(payment_lock_acquired_at <= now() - interval '1 hour') IS TRUE`. Postgres won't match that against `ix_orders_payment_lock_acquired_at` and falls back to a parallel sequential scan of all 5.3M orders.

## Checklist

- [ ] This PR addresses a single concern (one bug fix, one feature, one refactor)
- [ ] The diff is reasonably sized and easy to review
- [ ] New functionality is covered by tests
- [ ] Linting and type checking pass (`uv run task lint && uv run task lint_types`)
- [ ] No unrelated changes or drive-by fixes are included

---
## Summary by cubic
Stops the hourly stale payment lock sweep from scanning every order by making the predicate indexable. Previously the query rendered with IS TRUE, Postgres skipped the partial index, and the sweep often timed out and released no locks. Now the sweep uses the bare predicate, performs an index scan, and logs each release.

- The repository uses `Order.is_payment_lock_stale` directly to match the partial index and order by `payment_lock_acquired_at`.
- The task logs a warning when releasing a stale lock with `order_id` and `payment_lock_acquired_at`.
- A test pins the SQL shape to keep the predicate indexable and prevent regressions.

<sup>Written for commit bdc9f2b9d5df118343b7618914b801846be950e4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/13687?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #13694, opened 2026-08-12
server: keep SQLAlchemy boolean predicates indexable

### LINKED ISSUE #13690: Audit `IS TRUE` / `IS FALSE` wrappers on SQLAlchemy hybrid predicates
## Context

Follow-up to #13687, which fixes the stale payment-lock sweep by changing:

```python
.where(Order.is_payment_lock_stale.is_(True))
```

to:

```python
.where(Order.is_payment_lock_stale)
```

The original form renders the hybrid comparison as:

```sql
(payment_lock_acquired_at <= now() - interval '1 hour') IS TRUE
```

That prevents PostgreSQL from matching the predicate to `ix_orders_payment_lock_acquired_at` (`WHERE payment_lock_acquired_at IS NOT NULL`), causing a sequential scan of the orders table.

Original PR and discussion: https://github.com/polarsource/polar/pull/13687

## Investigation

SQLAlchemy 2.0.51 generates these three shapes for a boolean SQL expression `p`:

| SQLAlchemy | SQL |
| --- | --- |
| `.where(p)` | `WHERE p` |
| `.where(p == True)` | `WHERE p = true` |
| `.where(p.is_(True))` | `WHERE p IS TRUE` |

A local PostgreSQL 15.1 experiment with 100,000 rows and a partial B-tree index reproduced the production behavior:

| Predicate | Plan |
| --- | --- |
| bare timestamp inequality | index-only scan |
| timestamp inequality `= true` | index-only scan |
| timestamp inequality `IS TRUE` | sequential scan + sort |

PostgreSQL deliberately simplifies `p = true` to `p` and `p = false` to `NOT p`, partly so equivalent expressions can be recognized during partial-index matching. It does not perform the same rewrite for `IS TRUE` or `IS FALSE`.

The same behavior was reproduced for ordinary boolean partial indexes:

| Partial-index predicate | Query predicate | Plan |
| --- | --- | --- |
| `WHERE flag` | `WHERE flag` | index scan |
| `WHERE flag` | `WHERE flag IS TRUE` | sequential scan |
| `WHERE NOT flag` | `WHERE NOT flag` | index scan |
| `WHERE NOT flag` | `WHERE flag IS FALSE` | sequential scan |

This also affects timestamp-presence hybrids such as `is_deleted`, whose SQL expression is `deleted_at IS NOT NULL`:

```sql
-- Current is_deleted.is_(False)
WHERE (deleted_at IS NOT NULL) IS FALSE

-- Negated hybrid
WHERE NOT (deleted_at IS NOT NULL)
-- PostgreSQL normalizes this to deleted_at IS NULL
```

With selective test data, the wrapped form performed a sequential scan while the negated form used the `deleted_at` index. Polar has several partial indexes with `WHERE deleted_at IS NULL`, so this may prevent more index matches than the one found in #13687.

A repository search currently finds approximately 248 `.is_(True)` / `.is_(False)` usages. They are not all equivalent migration candidates: the results mix hybrid properties, real boolean columns, JSON expressions, hybrid definitions, and scalar/value contexts.

References:

- PostgreSQL partial-index predicate matching: https://www.postgresql.org/docs/current/indexes-partial.html
- PostgreSQL boolean equality simplification: https://doxygen.postgresql.org/clauses_8c.html

## Proposed convention

For boolean hybrid properties and boolean expressions used as predicates:

```python
.where(Model.condition)   # true
.where(~Model.condition)  # strictly false
```

`sqlalchemy.not_(Model.condition)` can be used where it is clearer.

Do **not** use Python `not`: SQLAlchemy hybrid attributes can be evaluated immediately by Python. For example, `not Order.is_deleted` evaluates to the ordinary value `False`, which can turn a query into `WHERE false`.

Keep `.is_(True)` / `.is_(False)` where a non-null scalar boolean is required, or where three-valued SQL semantics have been explicitly reviewed.

## Semantic caveats

In predicate positions such as `WHERE`, `JOIN ... ON`, and `CASE WHEN`:

- bare `p` and `p IS TRUE` select the same rows;
- `NOT p` and `p IS FALSE` select the same rows;
- both alternatives exclude `NULL` / unknown.

They are not interchangeable in every SQL context:

- In a projection, `p IS FALSE` always returns a non-null boolean, while `NOT p` can return `NULL`.
- If the intended meaning is “not true, including NULL,” use `p.is_not(True)`; neither `~p` nor `p.is_(False)` includes NULL.
- Inner identity tests can be part of a hybrid's definition and semantics. For example, JSON capability expressions may intentionally use `...as_boolean().is_(True)`. Removing a redundant outer wrapper is different from changing the hybrid definition.
- Constraints and other contexts where SQL `UNKNOWN` has distinct behavior need separate review.

## Suggested scope

- [ ] Inventory `.is_(True)` / `.is_(False)` usages and classify predicate vs scalar contexts.
- [ ] Prioritize hot queries and expressions backed by partial/selective indexes, especially soft-deletion, archival, timestamp-presence, and status hybrids.
- [ ] Replace predicate-context wrappers with bare expressions or `~expression` where semantics are preserved.
- [ ] Verify representative queries with compiled PostgreSQL SQL and `EXPLAIN`.
- [ ] Add SQL-shape regression tests for performance-critical predicates.
- [ ] Document or lint the convention to prevent new outer `IS TRUE` / `IS FALSE` wrappers on hybrid predicates.

frankie567:
## Summary

- use bare SQLAlchemy boolean expressions for true predicates and explicit negation for false predicates
- avoid outer `IS TRUE` / `IS FALSE` wrappers across backend queries, hybrid expressions, and maintenance scripts
- preserve `IS NOT TRUE` where `NULL` intentionally represents a disabled setting
- add ADR-0009 to document the convention and its three-valued-logic exception

## Why

This generalizes the fix discovered in #13687.

SQLAlchemy's `.is_(True)` and `.is_(False)` wrap an existing boolean expression in `IS TRUE` or `IS FALSE`. PostgreSQL does not simplify those identity tests when proving that a query predicate implies a partial-index predicate. That can hide an otherwise indexable comparison—such as a timestamp inequality or null check—and cause a sequential scan.

Using the boolean expression directly, or negating it explicitly, keeps the underlying predicate visible to the planner.

The migration preserves identity tests where they express different SQL semantics. In particular, `expression.is_not(True)` remains appropriate when both explicit `false` and `NULL` should match.

## Impact

Query results remain equivalent for strict true and false predicate filtering, while PostgreSQL can make use of matching ordinary and partial indexes. ADR-0009 records the convention for future backend queries.

## Validation

- lint passes
- tests pass
- branch diff checked against all Accepted ADRs: no violations

Closes #13690

<a href="https://cubic.dev/pr/polarsource/polar/pull/13694?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/scripts/prepare_organizations_members.py
+++ b/server/scripts/prepare_organizations_members.py
@@ -85,14 +85,12 @@ async def prepare(
             .where(
                 Organization.deleted_at.is_(None),
                 Organization.status != OrganizationStatus.BLOCKED,
-                Organization.feature_settings["seat_based_pricing_enabled"]
-                .as_boolean()
-                .is_(True),
+                Organization.feature_settings[
+                    "seat_based_pricing_enabled"
+                ].as_boolean(),
                 or_(
                     Organization.feature_settings["member_model_enabled"].is_(None),
-                    Organization.feature_settings["member_model_enabled"]
-                    .as_boolean()
-                    .is_(False),
+                    ~Organization.feature_settings["member_model_enabled"].as_boolean(),
                 ),
             )
             .order_by(Organization.slug.asc())
--- a/handbook/docs.json
+++ b/handbook/docs.json
@@ -105,7 +105,8 @@
               "engineering/decisions/0005-auth-subject-and-scopes",
               "engineering/decisions/0006-migration-and-backfill-safety",
               "engineering/decisions/0007-no-default-in-output-schemas",
-              "engineering/decisions/0008-encrypt-secrets-at-rest"
+              "engineering/decisions/0008-encrypt-secrets-at-rest",
+              "engineering/decisions/0009-use-bare-boolean-expressions-in-sqlalchemy-predicates"
             ]
           },
           "engineering/tech-notes",
--- a/handbook/engineering/decisions/0009-use-bare-boolean-expressions-in-sqlalchemy-predicates.mdx
+++ b/handbook/engineering/decisions/0009-use-bare-boolean-expressions-in-sqlalchemy-predicates.mdx
@@ -1,0 +1,52 @@
+---
+title: "ADR-0009: Use bare boolean expressions in SQLAlchemy predicates"
+description: "Avoid wrapping SQLAlchemy query predicates in IS TRUE or IS FALSE so PostgreSQL can match indexes."
+---
+
+<Info>
+**Status**: Accepted
+
+**Area**: Backend
+
+**Date**: 2026-08-12
+</Info>
+
+## Context
+
+SQLAlchemy hybrid properties often expand to comparisons such as
+`payment_lock_acquired_at <= cutoff` or `deleted_at IS NOT NULL`. Wrapping such an expression
+in `.is_(True)` or `.is_(False)` produces an outer `IS TRUE` or `IS FALSE` node; PostgreSQL's
+limited predicate implication logic may then fail to see the underlying indexable condition,
+particularly when matching a partial index, and choose a sequential scan instead.
+
+## Decision
+
+In SQL predicate positions, write a boolean expression directly for truth and negate it with
+`~expression` or `sqlalchemy.not_(expression)` for falsehood. Do not use `.is_(True)` or
+`.is_(False)` unless SQL identity semantics are intentionally required.
+
+```python
+statement.where(Order.is_payment_lock_stale)
+statement.where(~Order.is_deleted)
+```
+
+Identity tests remain valid when their three-valued behavior is the requirement. For example,
+`expression.is_not(True)` means "false or unknown" and therefore includes `NULL`, while
+`~expression` means strict falsehood and excludes `NULL`. The rule does not apply to
+`.is_(None)` / `.is_not(None)`, which express SQL null checks.
+
+## Consequences
+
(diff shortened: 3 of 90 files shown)


server/scripts/prepare_organizations_members.py
```diff
@@ -85,14 +85,12 @@ async def prepare(
             .where(
                 Organization.deleted_at.is_(None),
                 Organization.status != OrganizationStatus.BLOCKED,
-                Organization.feature_settings["seat_based_pricing_enabled"]
-                .as_boolean()
-                .is_(True),
+                Organization.feature_settings[
+                    "seat_based_pricing_enabled"
```
joebon, line 89: my eyes are bleeding

joebon (approved):
Nice job and improvement!

pieterbeulque (approved):

PR #13705, opened 2026-08-13
feat: add download statistics to `/file` and display in dashboard

simonxabris:
## Summary

**Related Issue**: polarsource/feedback#338

Changes the `/v1/files` endpoint to return download statistics and displays them on the downloadable benefit page.

## What

Each file now returns a `downloaders` and `downloads` fields which represent how many people downloaded the file and how many times.

API also has a `benefit_id` param and when provided, only statistics for those benefits are displayed. When not provided, statistics across all benefits are returned.

## Why

Users need visibility into how people interact with their downloadable benefits.

## Screenshot

Screenshot 2026-08-13 at 9 31 15

## Checklist

### DIFF
--- a/server/polar/file/endpoints.py
+++ b/server/polar/file/endpoints.py
@@ -26,7 +26,6 @@ from polar.routing import APIRouter
 
 from . import auth
 from .schemas import (
-    DownloadableFileRead,
     FileCreate,
     FileDownload,
     FilePatch,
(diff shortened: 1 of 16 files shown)


server/polar/file/endpoints.py
```diff
@@ -62,21 +63,35 @@ async def list(
...
         None, title="FileID Filter", description="Filter by file ID."
     ),
+    benefit_id: UUID4 | None = Query(
+        None,
+        title="Benefit ID",
+        description=(
+            "Scope download statistics to this benefit. If omitted, statistics "
+            "are aggregated across all benefits."
+        ),
+    ),
```
frankie567: In terms of API this is not a very good pattern: the query parameter influences the output of the API, and it also doesn't act as a pure filter.

I think we need a dedicated API under benefits: `GET /v1/benefits/{id}/files`

* For benefit that are not `downloadables`, return a 404
* Returns a paginated list of downloadable files, including their download stats

Under the hood, it's almost like a pure REST endpoint on the `Downloadable` entity, but we massage that to look like a file.

WDYT?

simonxabris: Makes perfect sense, I'll get back to it, thanks!

emilwidlund:
@simonxabris Neat! Great work! 

Minor suggestion - I wonder if we should make use of the `<DataTable />` component here. To keep it consistent with similar tables of data. Should be fairly straightforward to do. What do you think?

simonxabris:
@emilwidlund Yes that occurred to me, but then the "Dont make everything a data table" comment echoed in my head as well so I was like well just 2 more pieces of information, maybe the data table is an overkill. But yeah, if you think so let's go with the datatable!

simonxabris:
@emilwidlund What do you think of adding the download button in a separate "actions" column at the end? Or do we follow other data tables where there's a dropdown with actions? I feel like for one main action we shouldn't hide it behind a dropdown.

Screenshot 2026-08-13 at 11 04 19

emilwidlund:
> @emilwidlund Yes that occurred to me, but then the "Dont make everything a data table" comment echoed in my head as well so I was like well just 2 more pieces of information, maybe the data table is an overkill. But yeah, if you think so let's go with the datatable!

Haha, love the thinking. I think for this purpose it's the right call. It's mainly that I want to stay away from having the big entity pages as big table views only.

emilwidlund:
> @emilwidlund What do you think of adding the download button in a separate "actions" column at the end? Or do we follow other data tables where there's a dropdown with actions? I feel like for one main action we shouldn't hide it behind a dropdown.
> 
> Screenshot 2026-08-13 at 11 04 19

This is great! Looks good to me! :)

frankie567 (approved):
Looks great 👏

PR #13707, opened 2026-08-13
ci: Migrate preview envs to exe.dev instead of DO

Yopi:
This spins up a separate VM per PR instead of all preview envs sharing the same VM.

This hosts the frontend as well as the backend in the PR VM

<a href="https://cubic.dev/pr/polarsource/polar/pull/13707?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


frankie567 (approved):
Didn't go through everything but I like how it looks much more simpler and straightforward 👏

Yopi:
> Didn't go through everything but I like how it looks much more simpler and straightforward 👏

It shouldn't be worse!

PR #13708, opened 2026-08-13
fix(customer-portal): show a subscription's locked price, not the live catalog price

pieterbeulque:
## Summary

**Related Issue**: N/A — reported by a merchant via support (lowered a subscription price from $24.99 → $19.99/yr; the customer portal then read as a price *increase* and the "Next Charge" total didn't match the headline).

Existing subscribers stay locked into the price they subscribed with, even after the merchant changes the product's catalog price. The customer portal was pulling the product's **current catalog price** in two places, so lowering a price made the portal contradict itself.

## What

- **`CustomerSubscriptionHeader`** (Subscriptions list): only strike through the catalog price when the customer genuinely pays **less** than it (grandfathered or discounted). Previously any difference triggered the strikethrough, so after a price drop the new lower catalog price was struck through and the higher locked amount shown — `~~$19.99~~ $24.99`, reading as an increase.
- **`CurrentPeriodOverview`** ("Next Charge" card): render the subscription's locked base amount (the charge preview's `base_amount`, the same figure the Subtotal is built from) instead of the product's current catalog price, so the headline and the Subtotal agree. Applies to fixed, custom and seat-based prices; pending plan changes still preview the upcoming product's price and pure-metered products keep their existing rendering.

## Why

When a merchant edits a subscription price, Polar leaves existing subscriptions untouched — the old price row is archived and a new one created, and subscribers keep their locked `subscription.amount`. But the portal derived the *displayed* price from `subscription.product.prices` (the live catalog), which now points at the merchant's new price. Result for the reported case:

- Subscriptions list: `~~$19.99~~ $24.99 /yr` — looks like the price went up.
- Next Charge card: headline `$19.99/yr` while the Subtotal charged `$24.99` — impossible to reconcile.

## How

The backend already exposes both numbers — the locked amount (`subscription.amount` / charge-preview `base_amount`) and the product's live catalog price (`subscription.product.prices`). This was purely a matter of the UI picking the locked value where it had been using the live catalog value. No backend or schema changes.

## Checklist

- [ ] New functionality is covered by tests — a component regression test was added and then removed at the maintainer's request; this is a display-only fix
- [x] Linting and type checking pass — frontend: `pnpm lint` and `pnpm typecheck` (apps/web) pass

https://claude.ai/code/session_01MdxRMsgEsr1ESNKGzzMXgC

---
_Generated by [Claude Code](https://claude.ai/code/session_01MdxRMsgEsr1ESNKGzzMXgC)_

<a href="https://cubic.dev/pr/polarsource/polar/pull/13708?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


clients/apps/web/src/components/CustomerPortal/CurrentPeriodOverview.tsx
```diff
@@ -76,6 +77,19 @@ export const CurrentPeriodOverview = ({
       price.price_currency === subscription.currency && isSeatBasedPrice(price),
   )
 
+  // The subscription is locked into the price it was created with, which can
```
strandhvilliam: I get why a comment may be needed here, but maybe we can make it more concise?

strandhvilliam (approved):

PR #13733, opened 2026-08-13
Add Resend tags (category + tenant_id) to outgoing emails

pieterbeulque:
## Summary

**Related Issue**: N/A

Attach two Resend tags to every email we send, so emails can be filtered and attributed in the Resend dashboard and webhooks:

- `category` — the email template name (e.g. `order_confirmation`, `login_code`)
- `tenant_id` — the organization id, for organization-scoped emails

## What

- `polar/email/sender.py`: added an optional `tags: dict[str, str]` argument to `EmailSender.send()` (abstract + `LoggingEmailSender` + `ResendEmailSender`). `ResendEmailSender` serializes it into Resend's payload shape: `"tags": [{"name": ..., "value": ...}, ...]`.
- `polar/email/tasks.py`: the `email.send` task derives the tags and passes them to the sender — `category` from `template`, `tenant_id` from the organization id in the props.
- `polar/email/repository.py`: made the existing `_extract_organization_id` helper public (`extract_organization_id`) and reused it for the `tenant_id` tag.
- `tests/email/test_tasks.py`: added tests asserting the tags passed to the sender for an org-scoped email and a non-org email.

## Why

We want per-tenant and per-template observability/filtering of transactional email in Resend (deliverability, volume, debugging) without threading extra context through every call site.

## How

All outgoing mail funnels through the `email.send` Dramatiq task → `email_sender.send()`. Deriving the tags there means:

- **No call-site changes** — the task already receives `template` and `props_json`, which is everything needed to build both tags.
- `tenant_id` reuses `extract_organization_id`, the same derivation that already populates `EmailLog.organization_id`, so the tag and the log row stay consistent.
- Tag values are UUIDs / snake_case template names, which already satisfy Resend's tag charset (ASCII letters, numbers, `_`, `-`).

## Follow-up (separate PR)

`tenant_id` is only emitted for templates whose props embed the full `Organization` object (orders, subscriptions, payment-method reminders, webhook-disabled, seat invitations, customer sessions, etc.). A handful of organization-related templates currently carry only `organization_name` (a string, no id) and therefore get `category` but **not** `tenant_id`:

- `organization_invite`
- `organization_offboarded`
- `support_case_organization_new_message`
- the maintainer `notification_*` emails

A follow-up PR should plumb the organization id into those props (and their call sites) so they also carry `tenant_id`.

## Checklist

- [ ] Linting and type checking pass (`uv run task lint && uv run task lint_types`) — could not be run in the ephemeral web session (no Python 3.14.6 available); files compile and logic is verified, please confirm in CI

https://claude.ai/code/session_011jJqvAiamwfGwNnBWHF2xr

---
_Generated by [Claude Code](https://claude.ai/code/session_011jJqvAiamwfGwNnBWHF2xr)_

<a href="https://cubic.dev/pr/polarsource/polar/pull/13733?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/email/repository.py
+++ b/server/polar/email/repository.py
@@ -14,7 +14,7 @@ from polar.models.email_log import (
 log: Logger = structlog.get_logger()
 
 
-def _extract_organization_id(
+def extract_organization_id(
     email_props: dict[str, Any],
 ) -> UUID | None:
     org = email_props.get("organization")
@@ -76,6 +76,6 @@ class EmailLogRepository(RepositoryBase[EmailLog], RepositoryIDMixin[EmailLog, U
                 processor_id=processor_id,
                 error=error,
                 deduplication_key=deduplication_key,
-                organization_id=_extract_organization_id(props),
+                organization_id=extract_organization_id(props),
             )
         )
--- a/server/polar/email/sender.py
+++ b/server/polar/email/sender.py
@@ -66,6 +66,7 @@ class EmailSender(ABC):
         reply_to_name: str | None = DEFAULT_REPLY_TO_NAME,
         reply_to_email_addr: str | None = DEFAULT_REPLY_TO_EMAIL_ADDRESS,
         attachments: Iterable[Attachment] | None = None,
+        tags: dict[str, str] | None = None,
     ) -> str | None:
         pass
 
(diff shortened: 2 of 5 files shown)


Yopi (approved):
Good idea

PR #13740, opened 2026-08-13
Run web and api on Vercel as services

nsidnev:
### Frontend + API

Runs the Polar frontend and API as two Vercel services on one deployment, declared in `vercel.toml`. Dramatiq actors listen for messages from Vercel Queues rather than Redis in this setup. APScheduler cron jobs are handled automatically by directly exposing the scheduler so that the Vercel infra can recognize it and automatically configure Vercel Runtime Cache as a durable store.

Vercel Services enable:

- Synced preview environments
- Atomic deployments with skew protection (client and api are never out of sync)
- Private communication from server-side Next.js to FastAPI

This PR wires the FastAPI server at both the subdomain path (`api.polar.sh`) and a subpath route (`/api/v1`, `/api/backoffice`) to enable both staging & preview environments.

**Prereqs for deployment:**

- Removing "edge" runtime frontend functions: edge runtime is deprecated on Vercel
- `POLAR_JWKS` accepts the JWKS document itself, so keys come from an environment variable instead of the gitignored `.jwks.json`.
- `POLAR_REDIS_URL` and `POLAR_POSTGRES_URL_NON_POOLING` (as injected by managed Postgres integrations, e.g. Neon) to simplify wiring preview environments with Redis and Postgres from the marketplace
- The Turnstile sitekey is overridable via `NEXT_PUBLIC_TURNSTILE_SITE_KEY` (the secret half of the pair was already configurable), for domains outside the production sitekey's allowlist.
- Second-factor login pages hard-navigate when the API is same-origin: `/auth` chains into the one-time `GET /v1/auth/complete`, which the client router would fetch twice, burning the token.
- Public URLs (frontend/API base, checkout URL, allowed hosts) and cookie scope self-configure from the deployment's own URL — deployment URLs change every deploy, so they can't be static configuration. Explicitly configured values have the priority (`model_fields_set`).
- `NEXT_IMAGE_UNOPTIMIZED=1`: Vercel Services don't wire `/_next/image` yet. (soon to be fixed)

---

### Dramatiq workers

Move the background workers from `dramatiq` processes reading Redis to Functions that listen on Vercel Queues.

Summary:

Each `dramatiq` queue becomes a queue topic. The worker Function and its topics are derived from the broker config in `pyproject.toml`:

```toml
[[tool.vercel.subscribers]]
entrypoint = "polar.worker.run:broker"
```

Semantic differences on Vercel Queues:

- Delivery is push-based
- Concurrency comes from serverless function scaling, not worker processes/threads

---

### Scheduled Jobs

Runs the cron jobs without a resident scheduler process.

Current way to set up cron jobs: `polar.worker.scheduler` has the `start()` function called directly.

Vercel adaptation: expose the configured scheduler directly for the Vercel infra to recognize it and auto-configure. The scheduler becomes a queue subscriber driven by queue messages, scoped to its own `__aps_*` wake topics, with durable state in the Vercel Runtime Cache.


frankie567:
Superseded by #13874

PR #13745, opened 2026-08-14
backoffice/subscriptions: Add ability to set a past_due subscription to active

Yopi:
Adds a somewhat generic status transition on subscriptions in the backoffice, but only allow past_due to active for now

<a href="https://cubic.dev/pr/polarsource/polar/pull/13745?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/backoffice/subscriptions/endpoints.py
+++ b/server/polar/backoffice/subscriptions/endpoints.py
@@ -21,6 +21,7 @@ from polar.models import (
 )
 from polar.models.subscription import SubscriptionStatus
 from polar.order.repository import OrderRepository
+from polar.order.service import order as order_service
 from polar.postgres import AsyncSession, get_db_read_session, get_db_session
 from polar.subscription import sorting
 from polar.subscription.repository import SubscriptionRepository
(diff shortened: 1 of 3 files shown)


server/polar/backoffice/subscriptions/endpoints.py
```diff
@@ -658,3 +673,81 @@ async def update_billing_period_end(
...
+                )
+                return
+            if (subscription.status, form.status) == (
+                SubscriptionStatus.past_due,
+                SubscriptionStatus.active,
+            ):
+                await subscription_service.mark_active(session, subscription)
+                if form.void_pending_orders:
+                    await order_service.void_pending_orders_for_subscription(
+                        session, subscription
```
frankie567, line 725: I don't get why we would do this in this situation?

Yopi: I was thinking if a subscription is in past_due, we can set it to active and void the pending orders, or just set it to active again. If the orders are not voided they would continue dunning and might be able to charge at some point further down the line.

In the specific case that we were discussing as a lead up to this, the order is already voided, so it's moot.

frankie567 (approved):

PR #13746, opened 2026-08-14
Add metadata update support to subscription PATCH endpoint

pieterbeulque:
## Summary

Adds support for updating subscription metadata through the PATCH `/v1/subscriptions/{id}` endpoint.

## What

- Added `MetadataInputMixin` to `SubscriptionUpdateBase` schema to enable metadata field in update requests
- Implemented `update_metadata()` method in `SubscriptionService` to handle metadata updates
- Added logic to call `update_metadata()` when metadata is provided in the update request
- Added `has_metadata` property to detect when metadata is explicitly set in the request
- Added comprehensive tests covering metadata replacement and preservation when not provided

## Why

Users need the ability to update subscription metadata (e.g., custom reference IDs) without modifying other subscription properties. This is a common use case for integrations that need to track external identifiers.

## How

The implementation follows the existing pattern used for other subscription update operations:
1. Extended the schema with metadata support via `MetadataInputMixin`
2. Added a dedicated `update_metadata()` method in the service layer
3. Integrated the metadata update into the main `update()` flow with a conditional check
4. Metadata replacement is complete (not merged) with existing metadata, consistent with API design

The tests verify:
- Metadata is correctly replaced when provided
- Metadata is preserved when not included in the update request
- HTTP 200 response is returned on success

## Checklist

- [x] This PR addresses a single concern (metadata update feature)
- [x] New functionality is covered by tests (`TestSubscriptionUpdateMetadata`)

https://claude.ai/code/session_01BfbGArfSwn7Dfh3KmPVYPX

<a href="https://cubic.dev/pr/polarsource/polar/pull/13746?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/subscription/service.py
+++ b/server/polar/subscription/service.py
@@ -1236,6 +1236,11 @@ class SubscriptionService:
                     session, ctx, subscription, trial_end=update.trial_end
                 )
 
+            if update.has_metadata:
+                subscription = await self.update_metadata(
+                    session, ctx, subscription, metadata=update.metadata
+                )
+
         if isinstance(update, SubscriptionUpdateSeats):
             subscription = await self.update_seats(
                 session,
@@ -1807,6 +1812,19 @@ class SubscriptionService:
         )
 
         return subscription
+
+    async def update_metadata(
+        self,
+        session: AsyncSession,
+        ctx: SubscriptionUpdateContext,
+        subscription: Subscription,
+        *,
+        metadata: dict[str, Any],
+    ) -> Subscription:
+        repository = SubscriptionRepository.from_session(session)
+        return await repository.update(
+            subscription, update_dict={"user_metadata": metadata}, flush=True
+        )
 
     async def update_seats(
         self,
--- a/server/polar/subscription/service.py
+++ b/server/polar/subscription/service.py
@@ -1237,9 +1237,7 @@ class SubscriptionService:
                 )
 
             if update.has_metadata:
-                subscription = await self.update_metadata(
-                    session, ctx, subscription, metadata=update.metadata
-                )
+                subscription.user_metadata = update.metadata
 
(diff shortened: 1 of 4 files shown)


server/polar/subscription/service.py
```diff
@@ -1236,6 +1236,11 @@ async def update(
                     session, ctx, subscription, trial_end=update.trial_end
                 )
 
+            if update.has_metadata:
+                subscription = await self.update_metadata(
```
frankie567: It feels a bit too much to have a dedicated method + a flush in this situation. IMO, we could simply do:

```py
if update.has_metadata:
    subscription.user_metadata = update.metadata
```

frankie567 (changes requested):

server/polar/subscription/service.py
frankie567: Yeah no, actually... Let me take that back.

server/polar/subscription/service.py
```diff
@@ -1808,6 +1813,19 @@ async def update_trial(
...
+        self,
+        session: AsyncSession,
+        ctx: SubscriptionUpdateContext,
+        subscription: Subscription,
+        *,
+        metadata: dict[str, Any],
+    ) -> Subscription:
+        repository = SubscriptionRepository.from_session(session)
+        return await repository.update(
+            subscription, update_dict={"user_metadata": metadata}, flush=True
```
frankie567: I think my main concern here is the unnecessary `flush`. We can avoid it.

frankie567 (approved):

PR #13751, opened 2026-08-14
server/refund: assert payment transaction exists before issuing a refund

frankie567:
In the case where a merchant issues a refund very quickly (<2 seconds) after a successful payment, the payment transaction may not yet be available in the database. Since the refund transaction service assumes this transaction exists, we hit an assertion error after having issued the refund on Stripe; which led to an inconsistent state where we try to reconcile a refund from a Stripe webhook we didn't save in our database.

This fix ensures the payment transaction exists before proceeding with the refund, and raise an error if it doesn't.

Fixes https://github.com/polarsource/feedback/issues/383

<a href="https://cubic.dev/pr/polarsource/polar/pull/13751?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/refund/service.py
+++ b/server/polar/refund/service.py
@@ -39,6 +39,7 @@ from polar.order.service import order as order_service
 from polar.payment.repository import PaymentRepository
 from polar.tax.calculation import tax_calculation as tax_calculation_service
 from polar.tax.calculation.base import AlreadyRevertedError
+from polar.transaction.repository import PaymentTransactionRepository
 from polar.transaction.service.refund import (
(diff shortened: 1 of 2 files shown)


pieterbeulque (approved):
I guess Cubic has a point

PR #13760, opened 2026-08-14
refactor(product_price): make pricing tiers reusable across price types

strandhvilliam:
## Summary

Only seat prices have tiered pricing today. The storage format, the calculation, and the validation all live on the seat price type. A new price type with tiers would have to copy all of it.

This PR adds one shared system to store, calculate, and validate tiers. Seat prices are the first price type on it. Future tiered price types (metered tiers, next in the stack) use the shared system and must not add their own format.

**Invariant: seat prices bill the same amounts before, during, and after this change.** Reads still come from `seat_tiers`. This PR changes how the code computes the amount, not the data it computes from.

Stacked on #13792, which adds the `tiers` column this PR writes to.

## The shared format

```json
{
  "tier_type": "graduated",
  "minimum_units": 1,
  "tiers": [
    {"up_to": 10, "price_per_unit": "500"},
    {"up_to": null, "price_per_unit": "300"}
  ]
}
```

Two things differ from the seat format:

- **Tiers are gates, not ranges.** A tier covers everything up to and including `up_to`, starting where the previous tier ended. The last tier's `up_to` is `null` if it's open-ended. Gaps and overlaps aren't expressible, so they can't exist in stored data.
- **`price_per_unit` is a string.** Python reads JSON decimals as floats. Some pricing rates could go below a cent (metered for example have `unit_amount` which is `Numeric(17, 12)`), so strings parse to `Decimal` without loss. `up_to` and `minimum_units` are whole units and stay plain integers.

**`minimum_units` is the smallest purchasable quantity, not a tier bound.** The seat format overloads the first tier's `min_seats` for this. The engine always prices from zero, and the purchase layer enforces the floor. Nothing in this PR reads `minimum_units` for billing.

### Calculation
A `TieredPrice` mixin owns the `tiers` column and gives a price `get_tiered_amount(quantity) -> Decimal`, plus `get_minimum_units()` and `get_maximum_units()`. Module-level `_calculate_volume` / `_calculate_graduated` do the arithmetic. Same results as the seat calculation, with three additions:

- A missing or unknown `tier_type` raises. The old code read a missing type as volume, only the legacy translation keeps that default.
- Quantity 0 costs 0, and a negative quantity raises. Seats can't be 0, but metered usage can.
- A quantity no tier covers still raises for volume, and still bills only the covered units for graduated. Both errors are `ValueError` subclasses, so callers see no change.

### Validation
`validate_tiers_data` holds every rule about the shared format: at least one tier, non-negative rates, `up_to > 0`, unique `up_to` values, only the last tier unbounded, and `minimum_units` between 0 and the last tier's `up_to`.

### Legacy translation
`seat_tiers_to_tiers_data` is the only function that knows the seat format. `max_seats` becomes `up_to`, the first tier's `min_seats` becomes `minimum_units`, numbers become strings, a missing type means volume. It rejects gaps and overlaps, since the shared format can't represent them, that's the one place the contiguity check still lives.

### Seat API
ProductPriceSeatTiers` translates incoming tiers and runs the shared validator, then converts the errors back to seat wording (`max_seats`, "Gap or overlap between tiers"). Request and response shapes are unchanged.

### Dual write
A SQLAlchemy `set` listener on `seat_tiers` translates each write into `tiers`. New and updated rows fill both columns.

### Backfill
`scripts/backfill_product_price_tiers.py` fills `tiers` from existing seat_tiers` rows using the same translation function as the dual-write, so backfilled and freshly written rows are identical. Dry run by default, `--execute` to apply, keyset pagination by id, one commit per batch. Each translated row goes through the validator first, so a corrupt legacy row stops the script instead of landing in the shared column.

## Migration path

- [x] 1. Add the nullable `tiers` column (#13792).
- [x] 2. Calculate seat amounts through the shared engine.
- [x] 3. Write to both `tiers` and `seat_tiers`.
- [x] 4. Add the backfill script. Run it after deploy.
- [ ] 5. Verify both columns produce the same amounts.
- [ ] 6. Read from `tiers`, stop writing `seat_tiers`, drop the column.

Steps 5 and 6 are follow-up PRs. Step 6 is mostly deleting the `get_tiers_data()` override on
`ProductPriceSeatUnit` — it's the only thing still reading `seat_tiers` for billing.

## Testing

- Existing seat pricing tests pass unchanged, including the T-28449 case where the first tier starts above 1.
- Engine tests: volume and graduated across tier boundaries, unsorted tiers, fractional rates, quantity 0, negative quantity, missing and unknown `tier_type`, past the last bounded tier for both types, and `minimum_units` not shifting what gets billed.
- Validator tests for each rule, including `minimum_units` above the last tier's `up_to`.
- Translation tests: bounds to `up_to`, first `min_seats` to `minimum_units`, missing
  `max_seats`, gaps, overlaps, unbounded tier not last.
- Dual-write tests (constructor and update) and backfill tests (dry run, apply, idempotency).

## Out of scope

- Tiers on metered prices, that's #13793, next in the stack.
- Frontend changes (#13794).
- Any change to what seat validation accepts or rejects.
 

<a href="https://cubic.dev/pr/polarsource/polar/pull/13760?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/models/product_price.py
+++ b/server/polar/models/product_price.py
@@ -343,7 +343,7 @@ class TieredPrice:
 
     __abstract__ = True
 
-    tiers: Mapped[Tiers | None] = mapped_column(
+    tiers: Mapped[Tiers] = mapped_column(
         TiersType(none_as_null=True),
         use_existing_column=True,
         nullable=True,
@@ -362,14 +362,8 @@ class TieredPrice:
         default=None,
     )
 
-    def get_tiers(self) -> Tiers:
-        """The tiers this price bills on."""
-        if self.tiers is None:
-            raise ValueError("Price has no tiers")
-        return self.tiers
-
     def get_tiered_amount(self, quantity: int) -> Decimal:
-        return self.get_tiers().calculate(quantity)
+        return self.tiers.calculate(quantity)
 
     def get_minimum_units(self) -> int:
         """The smallest purchasable quantity (inclusive), 0 when unset.
@@ -381,7 +375,7 @@ class TieredPrice:
         open-ended. Defaults to the last tier's bound when unset."""
         if self.maximum_units is not None:
             return self.maximum_units
-        return self.get_tiers().last_bound
+        return self.tiers.last_bound
 
 
 class ProductPriceSeatUnit(TieredPrice, NewProductPrice, ProductPrice):
@@ -413,7 +407,7 @@ class ProductPriceSeatUnit(TieredPrice, NewProductPrice, ProductPrice):
 
     @property
     def is_free(self) -> bool:
-        return all(tier.unit_amount == 0 for tier in self.get_tiers().tiers)
+        return all(tier.unit_amount == 0 for tier in self.tiers.tiers)
 
     __mapper_args__ = {
         "polymorphic_identity": ProductPriceAmountType.seat_based,
@@ -431,7 +425,7 @@ def _write_tiers_from_seat_tiers(
     """Dual-write to the shared `tiers`, `minimum_units` and `maximum_units`
     columns. Delete when `seat_tiers` is dropped."""
     if value is None:
-        target.tiers = None
+        target.tiers = None  # type: ignore[assignment]
         target.minimum_units = None
         target.maximum_units = None
     else:
--- a/server/polar/models/product_price.py
+++ b/server/polar/models/product_price.py
@@ -33,6 +33,7 @@ from polar.enums import (
     SubscriptionRecurringInterval,
     TaxBehaviorOption,
 )
+from polar.exceptions import PolarError
 from polar.kit.currency import format_currency
 from polar.kit.db.models import RecordModel
 from polar.kit.extensions.sqlalchemy.types import StringEnum
@@ -68,16 +69,16 @@ class TierType(StrEnum):
 
 
 class Tier(TypedDict):
-    """A per-unit rate up to and including `up_to`.
+    """A per-unit rate up to and including `bound`.
 
     Each tier starts where the previous one ended. The first starts at
-    zero. `up_to` is None on the last tier if it's unbounded. Rates are
+    zero. `bound` is None on the last tier if it's unbounded. Rates are
     in cents and may be fractional, so they're stored as strings and
     parsed to Decimal before any math.
     """
 
-    up_to: int | None
-    price_per_unit: str
+    bound: int | None
+    unit_amount: str
 
 
 class TiersData(TypedDict):
@@ -86,12 +87,17 @@ class TiersData(TypedDict):
     `maximum_units` columns, not here.
     """
 
-    tier_type: TierType
+    type: TierType
     tiers: list[Tier]
 
 
-class InvalidTiersError(ValueError):
-    pass
+class TiersError(PolarError):
+    """Base error for invalid or unusable tiered pricing data."""
+
+
+class InvalidTiersError(TiersError):
+    def __init__(self, message: str) -> None:
+        super().__init__(message, status_code=400)
 
 
 class UnboundedTierNotLastError(InvalidTiersError):
@@ -111,12 +117,13 @@ class NonContiguousTiersError(InvalidTiersError):
         self.next_min_seats = next_min_seats
 
 
-class InvalidQuantityError(ValueError):
-    pass
+class InvalidQuantityError(TiersError):
+    def __init__(self, message: str) -> None:
+        super().__init__(message, status_code=400)
 
 
 def _parse_tier_type(data: TiersData) -> TierType:
-    tier_type = data.get("tier_type")
+    tier_type = data.get("type")
     try:
         return TierType(tier_type)
     except ValueError as e:
@@ -124,45 +131,45 @@ def _parse_tier_type(data: TiersData) -> TierType:
 
 
 class _ParsedTier(TypedDict):
-    up_to: int | None
-    price_per_unit: Decimal
+    bound: int | None
+    unit_amount: Decimal
 
 
 def _parse_tiers(data: TiersData) -> list[_ParsedTier]:
     parsed: list[_ParsedTier] = [
         {
-            "up_to": tier["up_to"],
-            "price_per_unit": Decimal(tier["price_per_unit"]),
+            "bound": tier["bound"],
+            "unit_amount": Decimal(tier["unit_amount"]),
         }
         for tier in data.get("tiers", [])
     ]
-    return sorted(parsed, key=lambda t: (t["up_to"] is None, t["up_to"] or 0))
+    return sorted(parsed, key=lambda t: (t["bound"] is None, t["bound"] or 0))
 
 
 def _calculate_volume(quantity: int, tiers: list[_ParsedTier]) -> Decimal:
     for tier in tiers:
-        up_to = tier["up_to"]
-        if up_to is None or quantity <= up_to:
-            return tier["price_per_unit"] * quantity
+        bound = tier["bound"]
+        if bound is None or quantity <= bound:
+            return tier["unit_amount"] * quantity
     raise InvalidQuantityError(f"No tier covers quantity {quantity}")
 
 
 def _calculate_graduated(quantity: int, tiers: list[_ParsedTier]) -> Decimal:
     total = Decimal(0)
     remaining = quantity
-    previous_up_to = 0
+    previous_bound = 0
     for tier in tiers:
         if remaining <= 0:
             break
-        up_to = tier["up_to"]
-        tier_capacity = (up_to - previous_up_to) if up_to is not None else None
+        bound = tier["bound"]
+        tier_capacity = (bound - previous_bound) if bound is not None else None
         units_in_tier = (
             remaining if tier_capacity is None else min(remaining, tier_capacity)
         )
-        total += units_in_tier * tier["price_per_unit"]
+        total += units_in_tier * tier["unit_amount"]
         remaining -= units_in_tier
-        if up_to is not None:
-            previous_up_to = up_to
+        if bound is not None:
+            previous_bound = bound
     return total
 
 
@@ -179,31 +186,31 @@ def validate_tiers_data(
         raise InvalidTiersError("At least one tier is required")
 
     for tier in tiers:
-        if tier["price_per_unit"] < 0:
+        if tier["unit_amount"] < 0:
             raise InvalidTiersError(
-                f"Tier price_per_unit must be >= 0, got {tier['price_per_unit']}"
+                f"Tier unit_amount must be >= 0, got {tier['unit_amount']}"
             )
-        up_to = tier["up_to"]
-        if up_to is not None and up_to <= 0:
-            raise InvalidTiersError(f"Tier up_to must be > 0, got {up_to}")
+        bound = tier["bound"]
+        if bound is not None and bound <= 0:
+            raise InvalidTiersError(f"Tier bound must be > 0, got {bound}")
 
     for current, next_tier in pairwise(tiers):
         # Tiers are sorted with None last, so a duplicate None lands here.
-        if current["up_to"] is None:
+        if current["bound"] is None:
             raise UnboundedTierNotLastError()
-        if next_tier["up_to"] == current["up_to"]:
+        if next_tier["bound"] == current["bound"]:
             raise InvalidTiersError(
-                f"Tier up_to values must be unique, got {current['up_to']} twice"
+                f"Tier bound values must be unique, got {current['bound']} twice"
             )
 
-    last_up_to = tiers[-1]["up_to"]
+    last_bound = tiers[-1]["bound"]
     if minimum_units is not None:
         if minimum_units < 0:
             raise InvalidTiersError(f"minimum_units must be >= 0, got {minimum_units}")
-        if last_up_to is not None and minimum_units > last_up_to:
+        if last_bound is not None and minimum_units > last_bound:
             raise InvalidTiersError(
-                f"minimum_units must not exceed the last tier's up_to, "
-                f"got {minimum_units} > {last_up_to}"
+                f"minimum_units must not exceed the last tier's bound, "
+                f"got {minimum_units} > {last_bound}"
             )
     if maximum_units is not None:
         if maximum_units <= 0:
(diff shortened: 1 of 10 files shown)


server/polar/models/product_price.py
```diff
@@ -60,6 +62,136 @@ class ProductPriceSource(StrEnum):
     ad_hoc = "ad_hoc"
 
 
+class TierType(StrEnum):
```
strandhvilliam: Idk if `models/product_price.py` is the correct file for this. Feels like it's home to many different concerns. Would it be more reasonable to move this into for example `products/tiers.py`?

server/polar/models/product_price.py
```diff
@@ -355,85 +577,41 @@ class ProductPriceSeatUnit(NewProductPrice, ProductPrice):
...
+        max_units = self.get_maximum_units()
+        return int(max_units) if max_units is not None else None
 
     __mapper_args__ = {
         "polymorphic_identity": ProductPriceAmountType.seat_based,
         "polymorphic_load": "inline",
     }
 
 
+@event.listens_for(ProductPriceSeatUnit.seat_tiers, "set")
```
strandhvilliam, line 423: Thinking is dual-write during the migration to the shared column. 

Not sure if this is the best way to do this in Python.

server/polar/models/product_price.py
frankie567: Agreed!

server/polar/models/product_price.py
```diff
@@ -60,6 +62,164 @@ class ProductPriceSource(StrEnum):
...
+    """A per-unit rate up to and including `up_to`.
+
+    Each tier starts where the previous one ended. The first starts at
+    zero. `up_to` is None on the last tier if it's unbounded. Rates are
+    in cents and may be fractional, so they're stored as strings and
+    parsed to Decimal before any math.
+    """
+
+    up_to: int | None
+    price_per_unit: str
```
frankie567: For consistency, this field should be named `unit_amount`

server/polar/models/product_price.py
```diff
@@ -60,6 +62,164 @@ class ProductPriceSource(StrEnum):
...
+    """The structure of the shared tiers JSONB column, used by every tiered
+    price type. Purchasable quantity bounds live in the `minimum_units` and
+    `maximum_units` columns, not here.
+    """
+
+    tier_type: TierType
+    tiers: list[Tier]
+
+
+class InvalidTiersError(ValueError):
```
frankie567: We should have a proper error hierarchy that inherits from `PolarError`: they are handled out-of-the-box if they're raised as part of an API call.

server/polar/models/product_price.py
```diff
@@ -346,109 +544,141 @@ def get_amount_and_label(self, units: float) -> tuple[int, str]:
...
+class TieredPrice:
+    """Mixin for prices billed from a shared list of tiers.
+
+    `get_tiered_amount` returns cents for a whole-unit quantity.
+    Subclasses interpret those units (seats, metered usage, …).
+    """
+
+    __abstract__ = True
+
+    tiers: Mapped[TiersData | None] = mapped_column(
```
frankie567: Since you're in a context of a TieredPrice, you can "lie" to the type checker and make it non nullable. So you don't have to do check `is None` everywhere where you're handling a `TieredPrice`.

server/polar/models/product_price.py
```diff
@@ -60,6 +62,164 @@ class ProductPriceSource(StrEnum):
...
+class Tier(TypedDict):
+    """A per-unit rate up to and including `up_to`.
+
+    Each tier starts where the previous one ended. The first starts at
+    zero. `up_to` is None on the last tier if it's unbounded. Rates are
+    in cents and may be fractional, so they're stored as strings and
+    parsed to Decimal before any math.
+    """
+
+    up_to: int | None
```
frankie567: Not really happy with `up_to`. Maybe "bound" or "limit"?

server/polar/models/product_price.py
```diff
@@ -60,6 +62,164 @@ class ProductPriceSource(StrEnum):
...
+    price_per_unit: str
+
+
+class TiersData(TypedDict):
+    """The structure of the shared tiers JSONB column, used by every tiered
+    price type. Purchasable quantity bounds live in the `minimum_units` and
+    `maximum_units` columns, not here.
+    """
+
+    tier_type: TierType
```
frankie567: `type` should be enough

server/polar/models/product_price.py
```diff
@@ -60,6 +62,164 @@ class ProductPriceSource(StrEnum):
...
+    except ValueError as e:
+        raise InvalidTiersError(f"Missing or unknown tier_type: {tier_type!r}") from e
+
+
+class _ParsedTier(TypedDict):
+    up_to: int | None
+    price_per_unit: Decimal
+
+
+def _parse_tiers(data: TiersData) -> list[_ParsedTier]:
```
frankie567: I'm wondering if this we wouldn't be better served by Pydantic models in this context. We would get out-of-the-box:

* All the validation we need
* Serialization/deserialization to/from JSONB
* Methods, so we can tidy up everything under a class
* Usable out-of-the-box in the API

Roughly:

```py
class Tier(BaseModel):
    up_to: int | None = Field(le=0)
    price_per_unit: Decimal

class Tiers(BaseModel):
    tier_type: TierType
    tiers: list[Tiers]

    @model_validator(mode="after")
    def validate_up_to(self) -> typing.Self:
        ...

    def calculate_volume(self) -> Decimal:
        ...
```

frankie567: To work with SQLAlchemy, we can implement a custom type that makes the conversion automatically when reading/writing to the ORM model. Example: https://github.com/polarsource/polar/blob/309f14ebdf1fdd2b59ae31300175faaff0404bba/server/polar/kit/address.py#L236-L249

server/polar/models/product_price.py
frankie567: Looks fine to me!

server/polar/models/product_price.py
```diff
@@ -346,109 +544,141 @@ def get_amount_and_label(self, units: float) -> tuple[int, str]:
...
+    amount_type: Mapped[Literal[ProductPriceAmountType.seat_based]] = mapped_column(
+        use_existing_column=True, default=ProductPriceAmountType.seat_based
+    )
+    seat_tiers: Mapped[SeatTiersData] = mapped_column(
+        postgresql.JSONB,
+        nullable=True,
+    )
+
+    def get_tiers_data(self) -> TiersData:
+        return seat_tiers_to_tiers_data(self.seat_tiers)
```
frankie567: So this will be removed when the backfill has run, right?

server/tests/models/test_product_price.py
```diff
@@ -231,3 +260,509 @@ def test_first_tier_min_seats_above_one(self) -> None:
         assert price.calculate_amount(10) == 10 * 20000
         # 10 at first tier + 5 at second tier.
         assert price.calculate_amount(15) == 10 * 20000 + 5 * 6000
+
```
frankie567, line 288: That'll be good to move that to a dedicated module :D

server/scripts/backfill_product_price_tiers.py
```diff
@@ -0,0 +1,109 @@
...
+        ).scalar_one()
+        typer.echo(
+            f"[dry-run] {count} seat prices would be backfilled. "
+            "Re-run with --execute to apply."
+        )
+        return count
+
+    total = 0
+    last_id = None
+    while True:
```
frankie567: There are ~4K seat prices in prod right now, so I think we can avoid batching and use the `stream` method of ProductPrice repository to iterate on them. That would also make the dry run mode a bit more useful (we'll be able to detect potential invalid tiers data)

frankie567 (changes requested):
That's significant work Villiam 👏  I think we need a bit more of refactoring and polish so it's really clean and solid for the foreseeable future :)

server/polar/models/product_price.py
strandhvilliam: Aah yes! Was initially thinking this would be bad later when adding metered tiers, since it already has a `unit_amount` field for non tiered cost. But when I think about it, that makes naming this `unit_amount` make even more sense :D

server/polar/models/product_price.py
strandhvilliam: Good point. Done!

server/polar/models/product_price.py
strandhvilliam: Sure! Went with `up_to` after internal discussion to match the Stripe spec, but we are not _bound_ to it.

Went with `bound`.

server/polar/models/product_price.py
strandhvilliam: Makes sense 🫡

server/polar/models/product_price.py
strandhvilliam: This is super cool! The custom type to convert automatically is so nice and get cleaner distinction between the pydantic schema model and the orm one. Had a fun evening geeking out about it! :D

server/polar/models/product_price.py
strandhvilliam: Yes, this and a lot of the weird seat-to-newtierformat stuff I've added should be able to be removed and cleaned up.

server/tests/models/test_product_price.py
strandhvilliam: 🫡

server/scripts/backfill_product_price_tiers.py
strandhvilliam: Makes sense, batching is overkill and the `stream` method seems super useful. Thanks!

server/polar/models/product_price.py
strandhvilliam: Aah cool! Didn't know this. Thanks!

server/polar/models/product_price.py
```diff
@@ -60,6 +62,164 @@ class ProductPriceSource(StrEnum):
...
+    maximum_units: int | None = None,
+) -> None:
+    _parse_tier_type(data)
+    tiers = _parse_tiers(data)
+
+    if not tiers:
+        raise InvalidTiersError("At least one tier is required")
+
+    for tier in tiers:
+        if tier["price_per_unit"] < 0:
```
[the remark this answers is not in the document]
strandhvilliam: The price_per_unit < 0 check is gone from a newer commit and this is not relevant anymore

server/polar/models/product_price.py
```diff
@@ -346,109 +334,105 @@ def get_amount_and_label(self, units: float) -> tuple[int, str]:
...
+    `get_tiered_amount` returns cents for a whole-unit quantity.
+    Subclasses interpret those units (seats, metered usage, …).
+    """
+
+    __abstract__ = True
+
+    tiers: Mapped[Tiers] = mapped_column(
+        TiersType(none_as_null=True),
+        use_existing_column=True,
         nullable=True,
```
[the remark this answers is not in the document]
strandhvilliam: This is intentional. The column is nullable because it's shared across price types

server/scripts/backfill_product_price_tiers.py
```diff
@@ -0,0 +1,98 @@
...
+
+async def _run(session: AsyncSession, *, dry_run: bool) -> int:
+    repository = ProductPriceRepository.from_session(session)
+    statement = repository.get_base_statement(include_deleted=True).where(
+        ProductPriceSeatUnit.seat_tiers.isnot(None)
+    )
+    if not dry_run:
+        # Lock each row so a concurrent seat_tiers write cannot commit new
+        # canonical values that this snapshot would then overwrite.
+        statement = statement.with_for_update(of=ProductPrice)
```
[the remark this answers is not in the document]
strandhvilliam: Don't want to reopen the race lock or use batching. It's ~4k rows, so it's fairly cheap

server/tests/product/test_tiers.py
```diff
@@ -0,0 +1,209 @@
+from typing import Any
```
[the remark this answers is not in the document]
strandhvilliam: Still covered. The unbounded max-units case and a missing max-seats key are the same paths we already test. The remaining sort is shared with the translator, which is still covered.

server/polar/models/product_price.py
```diff
@@ -346,109 +334,105 @@ def get_amount_and_label(self, units: float) -> tuple[int, str]:
...
-        raise ValueError(f"No tier found for {seats} seats")
+    def get_tiered_amount(self, quantity: int) -> Decimal:
+        return self.tiers.calculate(quantity)
+
+    def get_minimum_units(self) -> int:
+        """The smallest purchasable quantity (inclusive), 0 when unset.
+        Enforced by the purchase layer, not the pricing engine."""
+        return self.minimum_units or 0
+
+    def get_maximum_units(self) -> int | None:
```
psincraian, line 373: Now, seeing this I think there is a gap 🤔 How can I define a tier like:


Seats:
- Tier 1 : from 1 to 10 seats: $10/seat
- Tier 2: 11+ seats: $9/seat


We don't have an "unbound option", right?

frankie567 (approved):
Now it looks :chief-kiss:! Great work :)

psincraian (commented):
Love the abstraction, I only have a comment

server/polar/models/product_price.py
strandhvilliam: Maybe I'm misunderstanding you, but your example should be supported by setting the last `bound` to `None`.

So:
- Tier 1: { bound: 10, unit_amount: 10.0 } -> 1 - 10
- Tier 2: { bound: null, unit_amount: 9.0 } -> 11+

So the second tier would continue forever, unless one set a hard limit with `maximum_units` (which shouldn't be that relevant for seats but mainly for future metered prices`

psincraian: Right! Missed that part

psincraian (approved):
Really nice job Villiam!

strandhvilliam:
Appreciate the reviews @frankie567 @psincraian! Just added a fix that will keep the current seat charges reading from the old `seat_tiers` so if a charge is done before I run the backfill, it won't crash. Merging and running the backfill in a bit

PR #13765, opened 2026-08-14
Run web and API on Vercel as services

nsidnev:
Runs the Polar frontend and API as two Vercel services on one deployment, declared in `vercel.toml`. Dramatiq actors listen for messages from Vercel Queues rather than Redis in this setup. APScheduler cron jobs are handled automatically by directly exposing the scheduler so that the Vercel infra can recognize it and automatically configure Vercel Runtime Cache as a durable store.

Vercel Services enable:

- Synced preview environments
- Atomic deployments with skew protection (client and api are never out of sync)
- Private communication from server-side Next.js to FastAPI

This PR wires the FastAPI server at both the subdomain path (`api.polar.sh`) and a subpath route (`/api/v1`, `/api/backoffice`) to enable both staging & preview environments.

**Prereqs for deployment:**

- Removing "edge" runtime frontend functions: edge runtime is deprecated on Vercel
- `POLAR_JWKS` accepts the JWKS document itself, so keys come from an environment variable instead of the gitignored `.jwks.json`.
- `POLAR_REDIS_URL` and `POLAR_POSTGRES_URL_NON_POOLING` (as injected by managed Postgres integrations, e.g. Neon) to simplify wiring preview environments with Redis and Postgres from the marketplace
- The Turnstile sitekey is overridable via `NEXT_PUBLIC_TURNSTILE_SITE_KEY` (the secret half of the pair was already configurable), for domains outside the production sitekey's allowlist.
- Public URLs (frontend/API base, checkout URL, allowed hosts) and cookie scope self-configure from the deployment's own URL — deployment URLs change every deploy, so they can't be static configuration. Explicitly configured values have the priority (`model_fields_set`).
- `NEXT_IMAGE_UNOPTIMIZED=1`: Vercel Services don't wire `/_next/image` yet. (soon to be fixed)

---

Supersedes  #13740 with only frontend and API related changes


server/polar/app.py
```diff
@@ -221,7 +223,16 @@ def create_app() -> FastAPI:
         app.add_middleware(FlushEnqueuedWorkerJobsMiddleware)
         app.add_middleware(AsyncSessionMiddleware)
         app.add_middleware(rate_limit.get_middleware, redis=rate_limit_redis)
+    if settings.is_vercel():
+        # The service-binding proxy strips trailing slashes, which would turn
```
frankie567, line 227: Could this be fixed at the proxy level or at least configurable? The trailing slashes redirection is the default and common in FastAPI apps. Having to add a dedicated middleware is not great.

vercel.toml
```diff
@@ -0,0 +1,105 @@
...
+service = "api"
+format = "url"
+env = "POLAR_API_URL"
+
+[services.api]
+root = "server"
+entrypoint = "pyproject.toml"
+# Build the email renderer binary and backoffice assets so they ship in the
+# Function bundle; run migrations last, closest to the deploy.
+buildCommand = "uv run task emails && uv run task backoffice && uv run task db_migrate"
```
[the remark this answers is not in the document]
frankie567: @nsidnev I know it's not really an issue since we'll have dedicated database per preview deployment, but I think having a "pre-deploy" script could be interesting for that + seeding database. Is it something supported by Vercel?

frankie567 (approved):

server/polar/app.py
nsidnev: Agreed, let us take a deep look and fix this behavior so that the middleware could be dropped.

frankie567:
@nsidnev I'll merge the PR as-is so we can have a try at a deployment; let's ship improvements in follow-up PR if needed.

I'll probably clean up the config a bit myself.

PR #13784, opened 2026-08-17
perf: add partial index for linking pending billing entries

psincraian:
## Summary

Adds a partial index on `billing_entry` to support keyset-batched linking of pending entries. Migration-only PR; the code that uses it lands in #13785, as ADR-0006 requires.

**Merge order: #13796 first, then this, then #13785.** The first merge-queue run of this PR failed: the index changes the plan of the pending scan, which exposed a latent plan-dependent row order and made three tests flaky. #13796 makes that ordering deterministic.

## What

- New partial index `ix_billing_entry_pending_link` on `(subscription_id, product_price_id, id)` where `deleted_at IS NULL AND order_item_id IS NULL`.
- Matching `Index` on the `BillingEntry` model.

## Why

Subscription renewals for high-volume metered subscriptions time out. The fix walks pending entries in `id` order, one price at a time, in batches that resume after the last linked id. That scan needs `id` in an index next to the subscription and price columns. Only unlinked rows live in the index, so it stays small and linking updates never write to it.

## How

`CREATE INDEX CONCURRENTLY` inside an autocommit block, with the usual drop of any INVALID leftover first. Same pattern as `ix_billing_entry_pending_static`.

## Checklist

### DIFF
--- a/server/migrations/versions/2026-08-17-1200_add_pending_link_billing_entry_index.py
+++ b/server/migrations/versions/2026-08-17-1200_add_pending_link_billing_entry_index.py
@@ -1,0 +1,50 @@
+"""add partial index for linking pending billing entries
+
+Revision ID: f3d81c25ab90
+Revises: d54b19ea954c
+Create Date: 2026-08-17 12:00:00.000000
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# Polar Custom Imports
+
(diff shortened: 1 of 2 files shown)


joebon (approved):
Looks good! Not convinced it'll fix everything since I think the row updates themselves will take too long even if we make the lookups blazing fast. But worth the try and an improvement nonetheless!

PR #13785, opened 2026-08-17
perf: keyset-paginate pending billing entry linking

psincraian:
## Summary

Fixes subscription renewals that time out on high-volume metered subscriptions.

**Merge order: #13796, then #13784, then this.**

## What

- `_link_pending` now walks pending entries in `id` order and resumes each batch after the last linked id, instead of restarting a bare `LIMIT` scan.
- Each batch returns only a count and the last id, computed server-side in a CTE.
- Meter linking runs one price at a time so the scan stays on the new index.
- `order.create_subscription_order` and `subscription.cycle` get a 600s time limit. Both run this loop on whole-cycle backlogs.
- The locked-subscription invariant leeway goes from 5 to 15 minutes so healthy long runs don't page.

## Why

The old loop re-issued the same `UPDATE ... LIMIT` statement until done. Rows updated earlier in the transaction stay visible to the scan until commit, so each batch re-scanned everything already linked. With hundreds of thousands of pending entries the task could not finish inside the 60s worker limit. The rollback threw away all progress, so every retry started from zero and failed the same way. Production traces show 16 attempts in a row killed at exactly 60s with no progress.

## How

Keyset pagination on `id`, served by `ix_billing_entry_pending_link` from #13784. Each batch does bounded work, so batch time stays flat instead of growing.

Measured, not estimated: on a local database seeded with 600,000 pending entries for one subscription (schema built from the models, so all production indexes are present), the real repository code links the full set in about 24 seconds. EXPLAIN confirms the new index serves the scan with the cursor as an index condition. Production adds load and cold caches, but the 600s budget leaves wide headroom. Not measured end to end in production yet; the first real cycle after deploy will confirm.

The linking stays in one transaction, so billing semantics do not change: amounts are computed from the pending set before linking, and partial progress is never committed.

## Checklist

### DIFF
--- a/server/polar/billing_entry/repository.py
+++ b/server/polar/billing_entry/repository.py
@@ -1,10 +1,9 @@
 from collections.abc import AsyncGenerator, Sequence
 from datetime import datetime
 from itertools import batched
-from typing import cast
 from uuid import UUID
 
-from sqlalchemy import CursorResult, Select, and_, func, or_, select, update
+from sqlalchemy import Select, and_, func, or_, select, update
 from sqlalchemy.orm.strategy_options import contains_eager, joinedload
 
 from polar.config import settings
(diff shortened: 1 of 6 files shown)


joebon (approved):
NICE! Let's give it a spin

PR #13787, opened 2026-08-17
Fix read performance of meter quantities endpoint

pieterbeulque:
Bug fix: customer and external customer filters could match events across organizations sharing the same external ID.

Performance improvement: resolve the passed in customer ID's first to a list of tuples (customer_id, external_customer_id), then find events for those ID's. This path avoids sequential scans throughout.

### DIFF
--- a/server/polar/event/repository.py
+++ b/server/polar/event/repository.py
@@ -21,7 +21,7 @@ from sqlalchemy import (
     update,
 )
 from sqlalchemy.dialects.postgresql import aggregate_order_by, insert
-from sqlalchemy.orm import joinedload
+from sqlalchemy.orm import aliased, joinedload
 
 from polar.authz.types import AccessibleOrganizationID
 from polar.kit.repository import RepositoryBase, RepositoryIDMixin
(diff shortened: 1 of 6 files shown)


server/polar/event/repository.py
```diff
@@ -289,26 +289,49 @@ async def get_first_user_event_timestamp(
         return result.scalar_one_or_none()
 
     def get_customer_id_filter_clause(
-        self, customer_id: Sequence[UUID]
+        self, organization_id: UUID, customer_id: Sequence[UUID]
```
pieterbeulque: We need the filter on organization ID to filter correctly by external ID.

server/polar/event/repository.py
```diff
@@ -289,26 +289,57 @@ async def get_first_user_event_timestamp(
...
+            )
+            .union_all(
+                select(event_by_external_customer_id.id).where(
+                    event_by_external_customer_id.organization_id == organization_id,
+                    event_by_external_customer_id.external_customer_id.in_(
+                        external_customer_ids
+                    ),
+                )
+            )
+        )
```
Yopi: I wonder if this would look clearer if you did something like:

```suggestion
        event_ids_by_customer_id = select(Event.id).where(
            Event.organization_id == organization_id,
            Event.customer_id.in_(customer_ids),
        )
        event_ids_by_external_customer_id = select(Event.id).where(
            Event.organization_id == organization_id,
            Event.external_customer_id.in_(external_customer_ids),
        )
        matching_event_ids = event_ids_by_customer_id.union_all(
            event_ids_by_external_customer_id
        )
```

Yopi (approved):
Good idea!

Yopi (approved):

server/polar/meter/service.py
```diff
@@ -327,14 +327,30 @@ async def get_quantities(
             Event.organization_id == meter.organization_id,
         ]
         event_repository = EventRepository.from_session(session)
-        if customer_id is not None:
-            event_clauses.append(
-                event_repository.get_customer_id_filter_clause(customer_id)
```
frankie567, line 332: Maybe naive but: what if we added an optional `organization_id` to `get_customer_id_filter_clause` so we could do something like this:

```py
    def get_customer_id_filter_clause(
        self, customer_id: Sequence[UUID], organization_id: UUID | None = None
    ) -> ColumnElement[bool]:
        external_customer_statement = select(Customer.external_id).where(
            Customer.id.in_(customer_id)
        )
        if organization_id is not None:
            external_customer_statement = external_customer_statement.where(
                Customer.organization_id == organization_id
            )
        statement = or_(
            Event.customer_id.in_(customer_id),
            Event.external_customer_id.in_(
                external_customer_statement
            ),
        )
        if organization_id is not None:
            statement = and_(Event.organization_id == organization_id, statement)
        return statement
```

Wouldn't that achieve the same optimization?

pieterbeulque: That would fix the bug, but on its own is not enough to avoid the sequential scan. I think the Postgres planner can't figure it out when joining events with customers, and splitting into two queries is the most straight-forward way to keep things on a fast path.

frankie567 (approved):

PR #13790, opened 2026-08-18
server/auth: issue TOTP codes using SHA-1 instead of SHA-256

frankie567:
Lot of shitty authenticator apps don't support SHA-256, and don't even issue a warning when the `algorithm` parameter is set to SHA-256. So they silently generates SHA-1 codes, which are invalid, and the user is left with a broken 2FA setup, not understanding why.

Thus, switch new enrollments to SHA-1 to maximize compatibility with existing authenticator apps. Since we store the algorithm in our enrollment database, existing enrollments will continue to work with SHA-256.

Remove the related warning that's no longer relevant in the TOTP setup modal.

Fix https://github.com/polarsource/feedback/issues/390

<a href="https://cubic.dev/pr/polarsource/polar/pull/13790?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/apps/web/src/components/Settings/TOTPSetupModal.tsx
+++ b/clients/apps/web/src/components/Settings/TOTPSetupModal.tsx
@@ -24,7 +24,6 @@ export interface TOTPSetupModalProps {
 const TOTPSetupContent = ({ onEnabled }: { onEnabled: () => void }) => {
   const [code, setCode] = useState('')
   const [error, setError] = useState<string | null>(null)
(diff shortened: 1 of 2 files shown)


pieterbeulque (approved):
Approved with disgust 😁

PR #13797, opened 2026-08-18
Redirect authenticated users away from auth page

pieterbeulque:
- Pass on `?refresh` from the session refresh modal to opt out of redirect behavior
 - Redirect to existing `?return_to`, if any, only fall back to `/start`

### DIFF
--- a/clients/apps/web/src/app/(main)/auth/page.tsx
+++ b/clients/apps/web/src/app/(main)/auth/page.tsx
@@ -20,16 +20,25 @@ export default async function Page(props: {
     error?: string
     return_to?: string
     from?: string
+    refresh?: string
   }>
 }) {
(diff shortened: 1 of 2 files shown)


frankie567:
Yep, it's on purpose: we have a mechanism to refresh a session to perform sensitive operations. So this is important.

pieterbeulque:
@frankie567 yeah, sorry about that. I covered that use case now in an update.

frankie567:
Sorry but is it such an issue that we need to add that complexity to the code base?

pieterbeulque:
@frankie567 - Fair, I bumped into it from a user report. I've asked for details on how they land on this page, because I don't think we link to it from anywhere. Let's pause for now. Thanks!

PR #13799, opened 2026-08-18
products: let recurring products set a meter cycle

joebon:
# Summary

Last step in meter cycling: Bring the frontend parts to allow merchants to create recurring products with meter cycling different from the normal subscription cycling.

### Toggle not visible because product neither has metered price nor meter credit benefit
1-hidden-until-something-is-metered

### Toggle off
2-revealed-by-metered-price

### Toggle on
3-monthly-meters-on-yearly-billing

### Invalid cadence configured
4-invalid-combination

### Disabled when editing
6-locked-on-an-existing-product

---
## Summary by cubic
Let recurring products set a separate meter cycle so resets, meter‑credit grants, and overage settlement can run on a different cadence. Previously the meter cycle always matched the billing cycle; now you can opt into a cycle that must evenly divide the billing cycle. This is accepted on create only and gated by an organization feature flag.

- UI/behavior:
  - Toggle appears under Price Configuration next to the free trial when the product is recurring, the org’s meter cycling feature is enabled, and the product has a metered price or a meter‑credit benefit; existing products always show a saved cycle read‑only even if the flag later turns off.
  - Enabling defaults to monthly for month/year billing and daily for day/week; turning the toggle off, switching to one‑time, removing the last metered trigger, or otherwise losing the gate clears `meter_interval` and `meter_interval_count`. Field is disabled on update.
  - Client‑side rule mirrors the backend divisibility check and updates inline on changes; the count input is digits‑only, stores null on empty, and enforces required/min/max with clear messages.

- Dev notes:
  - Exports `recurringIntervalValues` from `@polar-sh/client`; adds a `ProductMeterCycleMixin` to carry `meter_interval` fields in `ProductForm` types and passes `hasMeterCreditBenefit` through `ProductForm`.
  - Adds `meterIntervalDividesBillingInterval`, `defaultMeterInterval`, and `shouldShowMeterCycle` utilities and frontend tests for divisibility, defaults, gating (including saved cycles), and validation.

<sup>Written for commit bdfe684fca83131ba7de995efa5215caf21c8cf6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/13799?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/apps/web/src/components/Products/CreateProductPage.tsx
+++ b/clients/apps/web/src/components/Products/CreateProductPage.tsx
@@ -62,6 +62,11 @@ export const CreateProductPage = ({
     [enabledBenefits],
   )
 
+  const hasMeterCreditBenefit = useMemo(
+    () => enabledBenefits.some((b) => b.type === 'meter_credit'),
+    [enabledBenefits],
+  )
+
   const getDefaultValues = () => {
     if (sourceProduct) {
       return productToCreateForm(sourceProduct)
@@ -228,6 +233,7 @@ export const CreateProductPage = ({
             <ProductForm
               organization={organization}
               update={false}
+              hasMeterCreditBenefit={hasMeterCreditBenefit}
               benefitsSlot={
                 <Benefits
                   organization={organization}
--- a/clients/apps/web/src/components/Products/EditProductPage.tsx
+++ b/clients/apps/web/src/components/Products/EditProductPage.tsx
@@ -59,6 +59,11 @@ export const EditProductPage = ({
     [enabledBenefits],
   )
 
+  const hasMeterCreditBenefit = useMemo(
+    () => enabledBenefits.some((b) => b.type === 'meter_credit'),
+    [enabledBenefits],
+  )
+
   const form = useForm<ProductEditOrCreateForm>({
     defaultValues: {
       ...product,
@@ -244,6 +249,7 @@ export const EditProductPage = ({
             <ProductForm
               organization={organization}
               update={true}
+              hasMeterCreditBenefit={hasMeterCreditBenefit}
               benefitsSlot={
                 <Benefits
                   organization={organization}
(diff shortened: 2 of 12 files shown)


pieterbeulque:
I think you make the right decisions, but it feels pretty complex.

Can we maybe show this option only when the meter is relevant, i.e. meter credits are granted or a metered price is added? The UI feels very technical for what could be a simple concept.

cc @strandhvilliam @emilwidlund for inspiration

strandhvilliam:
Nice, this is so cool!

Think I agree with Pieter on maybe showing it when relevant. I haven't looked into it too much but in my head one would go in the order of:
1. Select recurring sub (this shows 'Add metered price' button)
2. Add metered price and configure it with the proper meter etc
3. Select if meter should have separate cycle than the subscription

Maybe something like:
Screenshot 2026-08-18 at 12 19 48

joebon:
> Nice, this is so cool!
> 
> Think I agree with Pieter on maybe showing it when relevant. I haven't looked into it too much but in my head one would go in the order of:
> 
> 1. Select recurring sub (this shows 'Add metered price' button)
> 2. Add metered price and configure it with the proper meter etc
> 3. Select if meter should have separate cycle than the subscription
> 
> Maybe something like: Screenshot 2026-08-18 at 12 19 48

Thanks!
I did something along the lines of this. The one UX gotcha is that a metered price is not the only thing that I think should guard this. To @pieterbeulque's point if the product has a meter credit benefit then meter cycling is _also_ applicable. In the latter case the UI positioning is less obvious to me. WDYT?

pieterbeulque (approved):
We can test run this! I'm sure it'll need tweaks once somebody with no context bumps into this, but only one way to find out 😁 great work.

PR #13803, opened 2026-08-18
docs(engineering): add metered billing periods design document

joebon:


handbook/engineering/design-documents/metered-billing-periods.mdx
```diff
@@ -0,0 +1,369 @@
...
+fan-out runs at ~5 events/s against a ceiling in the hundreds — so *creating* entries is not
+what breaks. *Claiming* them is. Both of those hold at today's volume and are revisited under
+[Deferred: rollups](#deferred-rollups); neither is a reason to think the entry table is free
+forever.
+
+## Goals
+
+- **Settlement cost stops scaling with the number of billing entries.** It is the one cost that
+  grows without bound as usage grows.
+- An invoice's scope is immutable, so invoicing cannot change what a later invoice sees.
```
frankie567, line 68: That's one way of seeing it. It'sproposed here to determine the immutability of the invoice from its period, while currently it's immutable by looking at the billing entries it settles.

The good thing with the current design is that we can trigger an invoice at any time, and it'll invoice what's pending. It's not tied to any time period.

handbook/engineering/design-documents/metered-billing-periods.mdx
```diff
@@ -0,0 +1,399 @@
...
+    currency: Mapped[str]
+
+    __table_args__ = (UniqueConstraint("subscription_id", "meter_id", "starts_at"),)
+```
+
+### Three operations
+
+**Open** — at subscription start, cycle, or a meter-interval boundary, insert one period per
+metered price on the subscription. The price is snapshotted onto the row. Idempotent: a price
+that already has an accruing period is left alone.
```
frankie567, line 114: How does that handle metered price change mid-period?

handbook/engineering/design-documents/metered-billing-periods.mdx
```diff
@@ -0,0 +1,399 @@
...
+| `_link_pending`, `link_pending_by_subscription_and_price` / `_and_meter` | — |
+| `PendingByPrice` / `PendingByMeter`, `processed_meters`, the active-price scan | the meter is a column on the period |
+| `_get_metered_line_item` and `_get_metered_line_item_by_meter` | one function |
+| nowhere to record what a period has been billed | `billed_amount`, a high-water mark |
+
+Entries themselves are untouched: still one per event, still written at ingestion, still
+carrying the customer-to-subscription resolution decided at that moment. That resolution —
+direct subscription over seat holder, the trial boundary, the price in force — is subtle,
+already correct, and inherited for free by aggregating over entries rather than re-deriving it
+from events at settlement.
```
frankie567, line 197: I don't understand that part

frankie567 (commented):
Overall, I'm not sure I understand what the proposal here is. LLM verbiage makes it quite hard to follow.

Though, I would say my main concern is that we lose one of the core principle we set since the beginning: billing entries are the source of truth for billing. For every order item, we are able to backtrack to each singular event by following billing entries. If I understand this proposal properly, we lose that link by offloading this to `MeterPeriod`.

In this context, it feels like we're simply making Billing Entries completely useless. Maybe it's the right choice, but it feels like a change in semantics: we're invoicing by looking at time periods rather than looking at granular events.

handbook/engineering/design-documents/metered-billing-periods.mdx
joebon: Well, it doesn't - or any price changes would ignored mid-meter cycle. Is that behavior desirable? I.e. that a merchant should be able to change metered prices and that would take immediate effect on already running subscriptions?

handbook/engineering/design-documents/metered-billing-periods.mdx
joebon: I.e. we keep using billing entries instead of sourcing from events directly.

But that's written in a shitty way, and reasonably don't even need to be mentioned.

handbook/engineering/design-documents/metered-billing-periods.mdx
joebon: Yeah, and I really like the current design! The problematic aspect of it is the cost of assigning `order_item_id` for all billing entry rows though. This is one proposed way of solving that.

joebon:
> Overall, I'm not sure I understand what the proposal here is. LLM verbiage makes it quite hard to follow.
> 
> Though, I would say my main concern is that we lose one of the core principle we set since the beginning: billing entries are the source of truth for billing. For every order item, we are able to backtrack to each singular event by following billing entries. If I understand this proposal properly, we lose that link by offloading this to `MeterPeriod`.
> 
> In this context, it feels like we're simply making Billing Entries completely useless. Maybe it's the right choice, but it feels like a change in semantics: we're invoicing by looking at time periods rather than looking at granular events.

Sorry for the LLM verbosity @frankie567 !

I opted for keeping a lot of investigative slop around, but that was a shitty call. Dropping

handbook/engineering/design-documents/metered-billing-periods.mdx
frankie567: I was referring to a customer upgrading/downgrading in a middle of a cycle; where the unit price is different. In the current state, if a customer switches on Jan 15 from a 0.05$/unit price to a 0.01$/unit, we're able to:

* Bill usage from Jan 1 to Jan 15 at 0.05$
* Bill usage from Jan 15 to Jan 31 at 0.01$

Only exception being the max/min/avg aggregations which would be billed at 0.01$

psincraian: That is my main concern also 🤔 Product changes has been one of the hard problems at Polar

handbook/engineering/design-documents/metered-billing-periods.mdx
```diff
@@ -0,0 +1,399 @@
...
+  candidates; the measurements below say neither is the binding constraint.
+- Cash balances. See [wallets](/engineering/design-documents/wallets).
+- Building pre-aggregation infrastructure now. See [Deferred: rollups](#deferred-rollups).
+
+## Design
+
+### One table
+
+```python
+class MeterPeriod(RecordModel):
```
psincraian, line 89: What would be the plan to see what events have we billed and why? I think for debugging we may need to store that information somehow.

psincraian, line 89: How does this work with the Meter cycle that you implemented now? Would love to see some data examples

handbook/engineering/design-documents/metered-billing-periods.mdx
```diff
@@ -0,0 +1,399 @@
...
+    __tablename__ = "meter_periods"
+
+    subscription_id: Mapped[UUID]
+    meter_id: Mapped[UUID]
+    product_price_id: Mapped[UUID]      # the price this period bills at
+
+    starts_at: Mapped[datetime]         # half-open window [starts_at, ends_at)
+    ends_at: Mapped[datetime]
+    status: Mapped[MeterPeriodStatus]   # accruing | settled
+    billable: Mapped[bool]              # False during trial
```
psincraian, line 100: And what is the window on that case? Do we have a window for trial?

psincraian (commented):
I like the window idea, some questions tough

PR #13815, opened 2026-08-18
feat(organization): support partial updates of settings groups

pieterbeulque:
## Summary

**Related Issue**: N/A

`PATCH /v1/organizations/{id}` couldn't partially update the `customer_email_settings`, `subscription_settings`, or `customer_portal_settings` groups: callers had to send each group in full. This makes them behave like `feature_settings` and `dispute_settings`, which already support partial merges.

## What

- Added dedicated partial-update schemas — `OrganizationCustomerEmailSettingsUpdate`, `OrganizationSubscriptionSettingsUpdate`, and `OrganizationCustomerPortalSettingsUpdate` (with nested `CustomerPortalUsageSettingsUpdate` / `CustomerPortalSubscriptionSettingsUpdate` / `CustomerPortalCustomerSettingsUpdate`) — where every field is optional.
- `OrganizationUpdate` now references these `*Update` variants instead of the full model TypedDicts.
- The service merges each group into the organization's current settings instead of replacing it wholesale. `customer_portal_settings` is deep-merged so a nested group (e.g. `subscription`) can be changed without resending its siblings.
- Regenerated the TypeScript client (`clients/packages/client/src/v1.ts`).

## Why

Previously these three groups were typed as their full TypedDicts and written by wholesale assignment. Because those TypedDicts require every key, omitting any nested field returned a `422`, and there was no way to toggle a single flag (e.g. turning off one `customer_email` notification) without re-sending the whole object. `feature_settings` and `dispute_settings` already merge partial input; this brings the remaining settings groups in line.

## How

- Each `*Update` schema mirrors its settings model but with all fields `| None = None`, so a subset validates. `subscription_settings.proration_behavior` keeps its public JSON-schema restriction (`invoice` / `prorate` / `next_period`), so the generated client surface is unchanged.
- In `OrganizationService.update`, each group is merged with `model_dump(mode="json", exclude_unset=True, exclude_none=True)` on top of the stored value, and the two newly-handled groups are excluded from the generic `update_dict` so they aren't re-applied. `customer_email_settings` is merged onto `resolve_default_customer_email_settings(...)` so the stored value stays complete.
- Full-object updates continue to work (a complete payload merges to the same result as before), so this is backwards compatible.

## Testing

- Added endpoint tests for partial `customer_email_settings`, `subscription_settings`, and deep-merged `customer_portal_settings` updates, plus a full-object `customer_email_settings` update.
- Ran `TestUpdateOrganization` (20 passed) and `tests/organization/test_service.py` + `test_schemas.py` (350 passed) against a local Postgres/Redis.
- `ruff check`, `ruff format --check`, and `mypy` pass on the changed files.

> Note: the DB-backed suite was run against a locally-installed Postgres because the sandbox couldn't pull the Docker images; the standard CI infra will re-run the full suite.

## Checklist

https://claude.ai/code/session_01Jfigrw4Jbz1MHSEMxgLBVY

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jfigrw4Jbz1MHSEMxgLBVY)_

<a href="https://cubic.dev/pr/polarsource/polar/pull/13815?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/organization/schemas.py
+++ b/server/polar/organization/schemas.py
@@ -13,7 +13,7 @@ from pydantic import (
     computed_field,
     model_validator,
 )
-from pydantic.json_schema import SkipJsonSchema
+from pydantic.json_schema import SkipJsonSchema, WithJsonSchema
 from pydantic.networks import HttpUrl
 
 from polar.auth.permission import ROLE_PERMISSIONS, OrganizationPermission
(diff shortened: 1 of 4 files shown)


server/polar/organization/schemas.py
```diff
@@ -399,6 +399,81 @@ def _merge_customer_email_settings_defaults(value: Any) -> Any:
 ]
 
 
+class OrganizationCustomerEmailSettingsUpdate(Schema):
+    """Partial update for customer email settings.
```
frankie567, line 403: Those doctoring will make their way to the OpenAPI, SDK and docs, so they should be a bit less technical or dropped entirely.

server/polar/organization/service.py
```diff
@@ -566,13 +569,48 @@ async def update(
...
+        if update_schema.customer_email_settings is not None:
+            merged_email_settings = {
+                **organization.customer_email_settings,
+                **update_schema.customer_email_settings.model_dump(
+                    mode="json", exclude_unset=True, exclude_none=True
+                ),
+            }
+            # Resolve defaults after merging so a legacy organization that lacks
+            # `payment_method_expiration_reminder` derives its fallback from the
+            # `subscription_cycled` value set in this same request, not the old one.
```
frankie567, line 595: That's why when we add a new setting, we shouldn't be lazy and make a proper migration. Otherwise, we need this kind of workarounds.

@cubic-dev Please remember that for the next time you see a new or removed organization setting.

pieterbeulque: Yeah, maybe I'll clean up the data first so that we don't need to do this...

PR #13827, opened 2026-08-18
fix: stop the event loop watchdog from freezing the worker

psincraian:
## Summary

A worker stopped processing its queue and never came back. This PR fixes the cause, and makes the worker restart itself if it happens again.

## Why

We found two problems.

**1. The watchdog froze the worker.**

We have a watchdog that logs thread stacks when the event loop stops responding. It used `faulthandler.dump_traceback(all_threads=True)`.

That call can freeze the whole process when a thread is stuck inside a blocking C call. That is exactly when the watchdog runs. So the watchdog made things worse. 

We reproduced it locally. One blocking job, and the worker never returned. The log showed one warning and then silence. That matches what we saw in production.

**2. A blocking S3 upload ran on the event loop.**

`create_order_invoice` is `async`, but `s3.upload` is normal blocking boto3 code. Dramatiq runs all async jobs of a process on one event loop. So one slow upload stops every other job in that process.

`receipt/service.py` already handled this correctly. The invoice path did not.

## What

- The watchdog no longer uses `faulthandler`. It builds the stack text in plain Python.
- The watchdog finds the event loop thread by id. It used to look for a thread name that dramatiq never sets, so that log field was always empty.
- The watchdog exits the process after 3 misses in a row (about 60 seconds).
- The three blocking `s3.upload` calls now run in a thread with `asyncio.to_thread`.
- The S3 client now has timeouts. Before this, one call could take about 5 minutes.

## How

A stuck event loop cannot shut down cleanly, so the watchdog calls `os._exit`. Dramatiq sees the process die, stops the rest, and the platform starts a new one.

`WORKER_EVENT_LOOP_WATCHDOG_MAX_MISSES` controls this. Set it to `0` to only log and never exit.

## How we tested it

We reproduced the freeze locally with a temporary job that blocks on purpose, using the same setup as production (2 processes, 8 threads each). That job was only a local test and is not part of this PR (check the job code here https://github.com/polarsource/polar/pull/13827/changes/4369e3235b0cd03b261e20c9a44fc5a8ed26659c#diff-ccd625cb8c09c1f129a73ae8e068ea94cccfdb66339b7419bdbcabb5249b2e0e) 

| Case | Before | After |
| --- | --- | --- |
| One process stuck | Queue slow, the other process kept working | Same |
| Both processes stuck | Queue dead, nothing moved | Both exit in about 75s, service restarts |

One note: this was tested on macOS. The freeze itself may behave differently on Linux. The fix is safe either way.

## Follow up (not in this PR)

`send_confirmation_email` still builds the invoice PDF inline. That puts heavy PDF and S3 work on the low priority queue, even though a separate queue already exists for it.

## Checklist

### DIFF
--- a/server/polar/config.py
+++ b/server/polar/config.py
@@ -67,8 +67,8 @@ class Settings(BaseSettings):
     WORKER_HEALTH_CHECK_INTERVAL: timedelta = timedelta(seconds=30)
     WORKER_MAX_RETRIES: int = 20
     WORKER_MIN_BACKOFF_MILLISECONDS: int = 2_000
-    # Consecutive watchdog misses (~20s each) before the worker exits to be
-    # restarted. 0 disables the exit and only logs.
+    # Misses in a row before the worker exits and gets restarted. Each miss
+    # takes about 20s. Set to 0 to only log.
     WORKER_EVENT_LOOP_WATCHDOG_MAX_MISSES: int = 3
     WORKER_PROMETHEUS_DIR: Path = Path(tempfile.gettempdir()) / "prometheus_multiproc"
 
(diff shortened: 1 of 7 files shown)


Yopi (approved):
Nice catch!

PR #13836, opened 2026-08-18
aws/lambda: Map up env variables and secrets

Yopi:
* Since we don't want to encode secrets in env vars in the lambda, we need
to have something that reads them out from secrets manager via the AWS API
and then puts them as env vars in the runtime.

* Map up the rest of the env variables: We are reading these from the Render module since they are handled and mapped in the Render module already. To avoid us having to do the same mapping twice, and to avoid having to do the refactor now, we simply let the Render service module export the adjusted variables. These are then inserted into the lambda via env variables and via secrets manager to be read out
by the lambda runtime on startup

<a href="https://cubic.dev/pr/polarsource/polar/pull/13836?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


terraform/modules/render_service/outputs.tf
```diff
@@ -22,3 +22,156 @@ output "cron_job_ids" {
   description = "Map of cron job names to their IDs"
   value       = { for name, cron in render_cron_job.cron : name => cron.id }
 }
+
+output "worker_env_vars" {
```
Yopi: I don't love this. And maybe I should just stop being lazy and not do this, I think that's acceptable feedback.

terraform/sandbox/aws.tf
```diff
@@ -81,42 +81,32 @@ locals {
...
-  }
+  lambda_worker_environment = merge(
+    module.sandbox.worker_env_vars,
+    {
+      POLAR_JWKS                    = "/tmp/jwks.json"
+      POLAR_POSTGRES_DATABASE       = "polar_sandbox"
+      POLAR_POSTGRES_HOST           = module.pgbouncer_aws.host
+      POLAR_POSTGRES_PORT           = module.pgbouncer_aws.port
+      POLAR_POSTGRES_USER           = local.db_user
+      POLAR_POSTGRES_SSL            = "false"
```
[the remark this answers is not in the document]
Yopi: This is the same we have discussed for the pgbouncer PRs already

terraform/production/aws.tf
```diff
@@ -60,44 +60,32 @@ locals {
...
-    POLAR_LOGFIRE_TOKEN   = var.logfire_token
-    POLAR_POSTGRES_PWD    = local.db_password
-    POLAR_RESEND_API_KEY  = var.backend_resend_api_key_production
-    POLAR_SECRET          = var.backend_secret_production
-    POLAR_SENTRY_DSN      = var.backend_sentry_dsn_production
-    POLAR_SLACK_BOT_TOKEN = var.slo_report_slack_bot_token
-    TAILSCALE_AUTHKEY     = var.lambda_worker_tailscale_token
-  }
+  lambda_worker_environment = merge(
+    module.production.worker_env_vars,
```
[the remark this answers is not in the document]
Yopi: Fixed

frankie567:
> Since we don't want to encode secrets in env vars in the lambda

We don't? :D

terraform/modules/render_service/outputs.tf
frankie567: So we export Render variables to use them as input of Lambdas? That's indeed a bit unexpected :p 

I guess the "right" solution is to build up an environment variables module that reads from Terraform / Secrets Manager and feed it to Render and Lambdas modules?

Yopi:
> > Since we don't want to encode secrets in env vars in the lambda
> 
> We don't? :D

Lambda has a problem where, unlike ECS, you can't read secret environment variables from the secrets manager. So they all end up as "regular" environment variables which are readable in all APIs / interfaces. 

So to read environment variables from the secrets manager we need to do the hack around in the Dockerfile. I don't think it's too bad here, but it's a bit annoying

terraform/modules/render_service/outputs.tf
Yopi: Exactly. I refactored this to do that instead. Was getting super annoyed with it 😄

frankie567 (approved):
Great!

PR #13852, opened 2026-08-19
server: upgrade Ruff and introduce new rules

frankie567:
## Summary by cubic
Upgrades `ruff` and applies new rules across the server to standardize typing, literals, and exception handling. Behavior should be unchanged; silent exceptions now log via `structlog`, and static assets are versioned by an 8‑char SHA‑256 content hash.

- Typing/style: prefer `Sequence[...]` with tuple defaults; normalize `AsyncGenerator[T]` annotations; convert `dict(...)`/`list(...)` to literals; simplify boolean/ternary expressions; remove `from __future__ import annotations`; fix Pydantic validator signatures/returns; drop no‑op `pass`/`return None`.
- Exceptions/logging: replace bare/silent `except` blocks with `logger.exception` in load tests and backoffice (orders, reviews) and services; use `log.exception` in Discord/GitHub/Tinybird paths; consolidate multi‑exception handlers.
- Datetime: use `datetime.fromisoformat(value)` without `"Z"` normalization in Stripe/Tinybird and account risk; make `MIN_DATETIME` and Tinybird stats buckets UTC‑aware. Sanity‑check any callers that may send `"Z"` timestamps.
- Validation: checkout seat constraints remain enforced end‑to‑end (require `min_seats <= max_seats`; `seats` must lie within bounds).
- I/O and S3: `VersionedStaticFiles` delegates to a module‑level, cached `get_file_version` returning an 8‑char SHA‑256; `ip_geolocation` streams writes via `writelines`; S3 presign/public URL params now use dict literals; CSV/streaming and repository `AsyncGenerator` signatures normalized.
- Touchpoints to verify: auth factors/cookies; checkout amount parsing and confirm validation; customer portal charge preview/downloadables; S3 multipart/download presign and public URLs; Stripe Connect refresh/account risk parsing; Discord/GitHub role/invite flows; events stats/timeseries and Tinybird client; file download responses and CSV streaming.
- Rollout: update local tooling and `pre-commit` to the new `ruff` and run `ruff --fix`. No runtime config changes or database migrations.

<sup>Written for commit 4e4c70f0935fef93761994b7e3ca502c87b27df1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/13852?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/load_tests/scenarios/checkout.py
```diff
@@ -73,7 +77,7 @@ def update_customer_details(self):
                 else:
                     response.failure(f"Failed to update: {response.text}")
         except Exception:
-            pass
+            logger.exception("Failed to update checkout customer details")
```
psincraian, line 80: Love this, would it be possible to force to pass the exception to the logging to see what was the error?

server/migrations/env.py
```diff
@@ -11,11 +11,11 @@
 
 def include_object(object, name, type_, reflected, compare_to):
     # Exclude tables/indexes marked with skip_autogenerate
-    if type_ in ("table", "index") and hasattr(object, "info"):
-        if object.info.get("skip_autogenerate"):
-            return False
-
-    return True
+    return not (
```
psincraian, line 14: ❤️

psincraian (approved):
Love this

Image

server/scripts/set_organization_embed_hosts.py
```diff
@@ -230,7 +230,7 @@ async def _rdap(domain: str, client: httpx.AsyncClient) -> dict[str, Any] | None
         )
         response.raise_for_status()
         payload = response.json()
-    except (httpx.HTTPError, ValueError):
+    except httpx.HTTPError, ValueError:
```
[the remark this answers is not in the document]
frankie567: @cubic-dev It's now valid in Python 3.14 after the introduction of [PEP758](https://peps.python.org/pep-0758/). Please remember that.

server/polar/startup_program/service.py
```diff
@@ -53,7 +53,7 @@ class StartupProgramService:
     invited.
     """
 
-    async def mark_invited(self, organization: Organization) -> "Discount":
+    async def mark_invited(self, organization: Organization) -> Discount:
```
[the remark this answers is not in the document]
frankie567: @cubic-dev

This finding assumes Python 3.13’s eager annotation evaluation. The project requires Python 3.14, where annotations are deferred by default under PEP 649/749. Consequently, forward references and names imported only under `TYPE_CHECKING` are not evaluated when the function or class is defined, so they do not cause the claimed import-time `NameError`.

This was also verified by importing the affected modules with the project’s Python 3.14 interpreter.

References:

- [[Python 3.14 annotation semantics](https://docs.python.org/3.14/library/annotationlib.html#annotation-semantics)](https://docs.python.org/3.14/library/annotationlib.html#annotation-semantics)
- [[What’s New in Python 3.14: deferred annotations](https://docs.python.org/3.14/whatsnew/3.14.html#pep-649-and-749-deferred-evaluation-of-annotations)](https://docs.python.org/3.14/whatsnew/3.14.html#pep-649-and-749-deferred-evaluation-of-annotations)
- [[PEP 649](https://peps.python.org/pep-0649/)](https://peps.python.org/pep-0649/)
- [[PEP 749](https://peps.python.org/pep-0749/)](https://peps.python.org/pep-0749/)

Please remember that.

server/load_tests/scenarios/checkout.py
frankie567: I think that's what `logger.exception` does automatically (and probably why this linter rule exists). So I think it already does what you want out of the box :)

PR #13854, opened 2026-08-19
product: seat prices from the shared tiers columns

strandhvilliam:
## Summary

Backfill has been run successfully. This moves seat based reads to the `tiers` column instead of the `seat_tiers` one. Does not drop the old seat_tiers column however for now, but is unused.

- Stacked on merged #13760 (`c94fbd3c1e`). Merge **after** the `tiers` backfill has run.
- Seat billing, persistence, and min/max bounds read `tiers` / `minimum_units` / `maximum_units`. The `seat_tiers` DB column is unused (still mapped so Alembic does not drop it).
- No breaking API change: HTTP still sends and returns `seat_tiers`; that payload is now a derived view of the shared columns.

## Stack
```
#13792 tiers column (merged)
  └── #13760 reusable pricing tiers (merged)
        └── **this PR** (seat cutover, after backfill)
              └── #13793 metered tiers backend
                    └── #13794 metered tiers frontend
```


server/polar/models/product_price.py
```diff
@@ -379,65 +380,64 @@ def get_maximum_units(self) -> int | None:
...
+            self.tiers = None  # type: ignore[assignment]
+            self.minimum_units = None
+            self.maximum_units = None
+            return
+        self.tiers = seat_tiers_to_tiers(value)
+        self.minimum_units, self.maximum_units = seat_tiers_unit_bounds(value)
+
     def calculate_amount(self, seats: int) -> int:
-        amount = seat_tiers_to_tiers(self.seat_tiers).calculate(seats)
+        amount = self.get_tiered_amount(seats)
```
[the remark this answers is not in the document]
strandhvilliam: Seat prices are always created with tiers before anyone can buy them, so billing never runs with missing tiers, a crash there would mean bad data

server/polar/models/product_price.py
```diff
@@ -379,65 +380,64 @@ def get_maximum_units(self) -> int | None:
...
+    @property
+    def seat_tiers(self) -> SeatTiersData:
+        if self.tiers is None:
+            return {"seat_tier_type": SeatTierType.volume, "tiers": []}
+        return tiers_to_seat_tiers(self.tiers, self.minimum_units, self.maximum_units)
+
+    @seat_tiers.setter
+    def seat_tiers(self, value: SeatTiersData | None) -> None:
+        if value is None:
+            self.tiers = None  # type: ignore[assignment]
```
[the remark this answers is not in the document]
strandhvilliam: Will remove the column in a bit. So this is not needed. Backfill wont run again

server/polar/product/tiers.py
```diff
@@ -170,6 +170,12 @@ def __init__(self, message: str) -> None:
         super().__init__(message, status_code=400)
 
 
+def integral_price_per_seat(unit_amount: Decimal) -> int:
+    if unit_amount != unit_amount.to_integral_value():
+        raise ValueError(f"Seat tier rates must be whole cents, got {unit_amount}")
```
pieterbeulque: `whole cents` is not the right term with zero-decimal currencies.

This should be something like `Tier rates must be in smallest currency unit` to be consistent with other places in the codebase (⌘ + F `smallest currency unit`)

server/scripts/backfill_product_price_tiers.py
```diff
@@ -1,12 +1,12 @@
 """Backfill the shared `tiers`, `minimum_units` and `maximum_units` columns
 on seat-based product prices.
 
-Translates each row's legacy `seat_tiers` with the same functions the
-dual-write hook uses, so backfilled and newly written rows match.
-Re-running the script is safe: the translation is deterministic.
+Translates each row's unused `seat_tiers` column with the same functions
```
pieterbeulque: Is this needed?

strandhvilliam: We can technically delete the backfill, just kept it in case needing to do it again, with some adjustments as per the changes. But we should not run the backfill again.

server/polar/product/tiers.py
strandhvilliam: My bad, I wrote that intentionally. Will change!

server/scripts/backfill_product_price_tiers.py
strandhvilliam: Deleted 🫡

pieterbeulque (approved):

PR #13855, opened 2026-08-19
aws/lambda: Increase max retries for webhook lambda

Yopi:
Since we are treating the webhook retries a bit differently from other tasks, we need to give it more retries.

<a href="https://cubic.dev/pr/polarsource/polar/pull/13855?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/terraform/production/aws.tf
+++ b/terraform/production/aws.tf
@@ -111,7 +111,7 @@ locals {
     "high-priority"         = { timeout_seconds = 120 }
     "medium-priority"       = { timeout_seconds = 120 }
     "low-priority"          = { timeout_seconds = 660 }
-    "webhooks"              = { timeout_seconds = 120 }
+    "webhooks"              = { timeout_seconds = 120, max_retries = 250 } # Must stay above webhook_event.send's max_retries.
     "tinybird"              = { timeout_seconds = 120 }
     "invoices-and-receipts" = { timeout_seconds = 240 }
   }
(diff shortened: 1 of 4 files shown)


terraform/production/aws.tf
```diff
@@ -111,7 +111,7 @@ locals {
     "high-priority"         = { timeout_seconds = 120 }
     "medium-priority"       = { timeout_seconds = 120 }
     "low-priority"          = { timeout_seconds = 660 }
-    "webhooks"              = { timeout_seconds = 120 }
+    "webhooks"              = { timeout_seconds = 120, max_retries = 250 } # Must stay above webhook_event.send's max_retries.
```
psincraian, line 114: So each webhook can be up to 250 requests to the original server? That seems a lot

Yopi: Not precisely:

https://github.com/polarsource/polar/blob/main/server/polar/webhook/tasks.py#L34-L41

We allow only 10 real retries for requests, but since we have the ambition to send them in order we allow 200 (`60*1000/300` = 200) ordering retries. In which if the task runs, and it realises it's out of order it sleeps for a minute.

psincraian (approved):

PR #13856, opened 2026-08-19
scripts: Add SQS DLQ redrive script

Yopi:
When tasks end up in the DLQ for whatever reason, we need a way to put them back in the original queue if we can fix the issue that placed them there in the first place.

This script allows us to inspect and redrive messages from any queue.

<a href="https://cubic.dev/pr/polarsource/polar/pull/13856?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -104,6 +104,7 @@ dev = [
   # * https://futuresearch.ai/blog/litellm-pypi-supply-chain-attack/
   # * https://pypi.org/project/litellm/
   # "litellm==1.82.3",
+  "botocore[crt]>=1.43.38",
 ]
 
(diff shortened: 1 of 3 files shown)


psincraian (approved):
LGTM! 

But the AWS already provides this in the UI, and in my case I prefer to use the UI when possible 🤔

Yopi:
Fair! Let's work with the UI for now

PR #13858, opened 2026-08-19
Add timeout configuration to generated SDKs

### LINKED ISSUE #13823: Allow configuring HTTP timeout in the new SDKs
The new 1.0 Python/TypeScript SDKs use HTTPX/fetch default timeouts with no way to override them. Add options to configure the client's HTTP timeout (e.g. connect/read timeout) so integrators can tune it for slower endpoints like offline order finalization.

Raised by a customer in Slack: they needed a longer timeout for the credit-card-on-file order routes than the current \~5s default.

<sub>Sent by @frankie567 from [HTTPX defaults issue creation](https://app.stilla.ai/m/memo_01m0arpcfmfknsfmk1spmzt38h).</sub>

frankie567:
## Summary

- add client-level timeout configuration, expressed in seconds, to the generated Python and TypeScript SDKs
- add per-call overrides through `request_timeout` in Python and a final `RequestOptions` argument in TypeScript
- forward timeout overrides through paginated requests
- use HTTPX timeout configuration in Python and the modern `AbortSignal.timeout()` API in TypeScript
- document the API and add generator-template tests

Generated SDK outputs are intentionally excluded; this PR changes only the generator templates.

Closes #13823

## Validation

- `just lint` in `sdk/generator`
- `just test` in `sdk/generator` (83 passed)

<a href="https://cubic.dev/pr/polarsource/polar/pull/13858?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/sdk/generator/python/template/polar/base.py
+++ b/sdk/generator/python/template/polar/base.py
@@ -9,6 +9,7 @@ import typing_extensions
 
 _EnvironmentT = typing.TypeVar("_EnvironmentT", bound=str)
 _ModelT = typing.TypeVar("_ModelT")
+RequestTimeout: typing.TypeAlias = float | httpx.Timeout
 
 
 class PolarError(Exception):
@@ -72,6 +73,7 @@ class BuildRequestMixin:
         path_params: dict[str, typing.Any] | None = None,
         query_params: dict[str, typing.Any] | None = None,
         body: typing.Any | None = None,
+        request_timeout: RequestTimeout | None = None,
     ) -> httpx.Request:
         url = url.format(**(path_params or {}))
 
@@ -85,13 +87,23 @@ class BuildRequestMixin:
             else:
                 params[k] = v
 
-        return self._client.build_request(method, url, params=params, json=body)
+        timeout = self._client.timeout if request_timeout is None else request_timeout
+        return self._client.build_request(
+            method, url, params=params, json=body, timeout=timeout
+        )
 
 
 class SyncClientBase(BuildRequestMixin):
-    def __init__(self, base_url: str, version: str, access_token: str) -> None:
+    def __init__(
+        self,
+        base_url: str,
+        version: str,
+        access_token: str,
+        timeout: RequestTimeout | None = 5.0,
+    ) -> None:
         self._client = httpx.Client(
             base_url=base_url,
+            timeout=timeout,
             headers={
                 "Polar-Version": version,
                 "Authorization": f"Bearer {access_token}",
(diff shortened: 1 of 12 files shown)


sdk/generator/python/template/polar/version/services/service.py
```diff
@@ -80,6 +81,7 @@ def {{ method.name | operation_name }}(
         {{ param.name }}: {{ param.type | type_annotation }} = ...,
         {% endif %}
         {% endfor %}
+        request_timeout: RequestTimeout | None = None,
```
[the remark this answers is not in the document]
frankie567: We deliberately omit that problem for now as a pragmatic measure.

sdk/generator/typescript/template/src/base.ts
```diff
@@ -160,10 +169,34 @@ export class ClientBase {
...
+    request: [string, RequestInit],
+    requestOptions?: RequestOptions,
+  ): Promise<Response> {
     const [fullUrl, requestInit] = request;
+    const timeout = requestOptions?.timeout ?? this.options.timeout;
+    let signal = requestInit.signal;
+    if (timeout !== undefined) {
+      const timeoutMilliseconds = Math.ceil(timeout * 1000);
+      if (
+        !Number.isFinite(timeout) ||
```
psincraian, line 185: I think this validation don't exist on the python version

sdk/generator/python/template/polar/base.py
```diff
@@ -85,13 +87,23 @@ def build_request(
             else:
                 params[k] = v
 
-        return self._client.build_request(method, url, params=params, json=body)
+        timeout = self._client.timeout if request_timeout is None else request_timeout
```
psincraian, line 90: This default don't exist in TS

frankie567: Good catch! Added in d62d920e18547eb2c92b58749e76dfb8e591800f

sdk/generator/typescript/template/src/base.ts
frankie567: Yes, I think that's fine since HTTPX will throw when issuing the request. The reason we did it for TS was it would throw a confusing PolarNetworkError if we didn't enforce it before issuing the fetch request.

psincraian (approved):

PR #13859, opened 2026-08-19
lambda: Set pool_pre_ping

Yopi:
This is to ensure that a connection is still active. Since lambdas reuse the same execution, but are paused inbetween executing, a db connection might have timed out between executions.

This means that we need to ensure that it's still active before we try to use it.

This only enables the pre pinging on lambda workers, nowhere else.

<a href="https://cubic.dev/pr/polarsource/polar/pull/13859?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


frankie567:
> Since lambdas reuse the same execution, but are paused inbetween executing, a db connection might have timed out between executions.

Could you elaborate on that? Does the lambda keep the connection pool open between executions? Because we already have the `pool_recycle` setting set to 10 minutes that should already handle this kind of situations.

Yopi:
> > Since lambdas reuse the same execution, but are paused inbetween executing, a db connection might have timed out between executions.
> 
> Could you elaborate on that? Does the lambda keep the connection pool open between executions? Because we already have the `pool_recycle` setting set to 10 minutes that should already handle this kind of situations.

I _believe_ the issue is that the lambda execution is fully paused between runs, so it doesn't realise that the socket has closed and it also doesn't run the pool recycle. The socket closes on it while it's hibernating. 

But I might be misdiagnosing this, it could be that there is another network component that has a shorter connection limit than our pool recycle. E.g. NAT gateways have 350s idle timeout, which is less than our 600s pool recycle. We should not be going to pgbouncer via the NAT gateway so that shouldn't be it, but there is potentially something else in play here.

frankie567:
> I believe the issue is that the lambda execution is fully paused between runs, so it doesn't realise that the socket has closed and it also doesn't run the pool recycle. The socket closes on it while it's hibernating.

Isn't there a hook or something for that so that we can properly close the pool?

> But I might be misdiagnosing this, it could be that there is another network component that has a shorter connection limit than our pool recycle. E.g. NAT gateways have 350s idle timeout, which is less than our 600s pool recycle. We should not be going to pgbouncer via the NAT gateway so that shouldn't be it, but there is potentially something else in play here.

Yeah, that's definitely an option

Yopi:
> > I believe the issue is that the lambda execution is fully paused between runs, so it doesn't realise that the socket has closed and it also doesn't run the pool recycle. The socket closes on it while it's hibernating.
> 
> Isn't there a hook or something for that so that we can properly close the pool?

There is, and we can close it, but then we would pay the reconnect penalty every time. This seems to happen every once in a while, so the penalty of a ping is a lot lower than forcefully disconnecting and reconnecting.

Mind you that this PR only enables the pre ping on the lambda.

I'll investigate the other alternative in lowering the recycle timer, but I don't think this is the most horrible choice.

frankie567 (approved):

PR #13868, opened 2026-08-19
feat: add merchant license key rotation (API + dashboard)

psincraian:
## Summary

Adds merchant-facing license key rotation via API and dashboard. A merchant asked about this on Plain. I think it's small enough to one shot this in a LLM and provide this feature. 

- `POST /v1/license-keys/{id}/rotate
- Dashboard: Rotate action + confirm modal on the license key detail modal
- Docs updated for the new endpoint

https://github.com/user-attachments/assets/df7b7a6b-3006-4de8-9694-574b86e319d7

## Test plan

- [x] `tests/license_key/` including rotate + cross-org 404
- [x] Manual E2E: rotate in dashboard, old key validate → 404, new key → 200, same `id`


pieterbeulque:
Can you revert all the i18n? That's not applicable to the dashboard. I've noticed agents are eager with adding translations.

pieterbeulque:
Bumped into the thread in Plain, do you think it would also make sense to expose this rotation functionality in the Customer Portal (as requested) or can that be abused?

psincraian:
@pieterbeulque  In theory it cannot be abused as we should keep track the usage, I can do that in another PR as it requires a new endpoint and a new view

frankie567 (approved):

PR #13869, opened 2026-08-19
product: add unit-based pricing model and migration

strandhvilliam:
Part 1 of the unit-based pricing stack. Schema only, nothing reads these
columns yet, so this can land on its own.

- Adds the `unit_based` amount type and the `ProductPriceUnitBased` model,
  built on the shared tiers columns.
- Adds the nullable `units` columns to checkouts, checkout links,
  subscriptions, subscription updates and orders.


server/migrations/versions/2026-08-18-1943_add_units_columns_for_unit_based_pricing.py
```diff
@@ -0,0 +1,53 @@
...
+    op.add_column("checkouts", sa.Column("units", sa.Integer(), nullable=True))
+    op.add_column("checkouts", sa.Column("min_units", sa.Integer(), nullable=True))
+    op.add_column("checkouts", sa.Column("max_units", sa.Integer(), nullable=True))
+    op.add_column("checkout_links", sa.Column("units", sa.Integer(), nullable=True))
+    op.add_column("subscriptions", sa.Column("units", sa.Integer(), nullable=True))
+    op.add_column(
+        "subscription_updates", sa.Column("units", sa.Integer(), nullable=True)
+    )
+    op.add_column("orders", sa.Column("units", sa.Integer(), nullable=True))
+    op.add_column(
```
strandhvilliam: Added the singular and plural label as columns. But don't wanna bog down the table. Would a shared jsonb `unit_properties` be better? May be too ambiguous though compared to the other price_types who uses columns in a polymorphic way 🤔

server/polar/models/product_price.py
```diff
@@ -439,6 +442,52 @@ def is_free(self) -> bool:
     }
 
 
+class ProductPriceUnitBased(TieredPrice, NewProductPrice, ProductPrice):
```
strandhvilliam: Not sure on naming of this class.

We have `ProductPriceSeatUnit` and `ProductPriceMeteredUnit` so `ProductPriceUnitBased` falls of that pattern. But I think `ProductPriceUnit` sounds almost like a parent to the other ones?

pieterbeulque: Yeah, but `ProductPriceUnit` is more consistent.

server/migrations/versions/2026-08-18-1943_add_units_columns_for_unit_based_pricing.py
pieterbeulque: We could for now also store only a unit label, and make sure we show it always as `/ ${label}` or `per ${label}`, and push this to the future. This will work for the pilot case ("device")

server/migrations/versions/2026-08-18-1943_add_units_columns_for_unit_based_pricing.py
```diff
@@ -0,0 +1,53 @@
...
+revision = "a83cf131398d"
+down_revision = "c7f2a91e4d38"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
+    # Ensures we don't break app by applying a deadlock-inducing migration
+    op.execute("SET LOCAL lock_timeout = '5s'")
+    op.add_column("checkouts", sa.Column("units", sa.Integer(), nullable=True))
```
psincraian, line 25: checkout is a big table, even so this should be safe, I recommend waiting for tomorrow to merge this

psincraian (approved):

server/migrations/versions/2026-08-18-1943_add_units_columns_for_unit_based_pricing.py
strandhvilliam: Good point!

PR #13880, opened 2026-08-19
server: Drop TrailingSlashRestoreMiddleware

nsidnev:
Addresses the https://github.com/polarsource/polar/pull/13765#discussion_r3788999364, the platform has fixed this behavior and the middleware is no longer required.


frankie567 (approved):
Amazing 🎉

PR #13871, opened 2026-08-19
product: add unit-based subscriptions backend

strandhvilliam:
Part 3 of the unit-based pricing stack, the feature end to end.

- Checkout picks a unit count, bounded by the price's min/max.
- Checkout links can lock units so the buyer can't change them.
- Subscriptions are created from the checkout's unit count.
- Units update endpoint, sharing the quantity proration path with seats.
- `subscription.units_updated` system events, unit billing entries and
  order item labels.

A lot of the code is just duplicating `seats` logic. This is intentional repetition for now. Later I will go through and remove duplication and merge seat based and unit based logic where it makes sense.

Everything is gated behind a feature flag that can only be set internally

---
## Summary by cubic
Add backend support for unit-based subscriptions with min/max bounds, proration like seats, and full coverage across checkout, checkout links, subscriptions, billing, and events. Previously only seat-based quantities were supported; now units persist, validate per price/currency, and emit dedicated events and billing entries.

- Introduces ProductPriceUnit and exposes units/min_units/max_units in the API (including generated `v1` client and email OpenAPI); defaults units to get_minimum_purchasable_units; order item labels use ICU-pluralized unit nouns; web timeline renders subscription.units_updated.
- Checkout validates unit bounds and supports checkout-link unit locks; locks require all linked products be unit-based and valid across every currency-specific price, include the failing price currency in errors, and are ignored when incompatible.
- Subscriptions persist units, allow unit↔non-unit plan switches, validate units against the target plan and any pending product change, preview/apply unit changes with shared quantity proration, and emit subscription.units_updated; billing adds subscription_units_increase and subscription_units_decrease. Removes arbitrary unit input caps.

**Rollout**
- Enable unit-based pricing per organization in backoffice.
- Regenerate or update API clients to read/write units/min_units/max_units.
- Update consumers to handle subscription.units_updated and unit increase/decrease billing entries.
- Validate unit locks per currency and keep unit counts within each price’s min/max when changing plans; remove any UI/API caps beyond a minimum of 1 unit.

<sup>Written for commit f9ef71029dc465d1e94198431987def139e0877b. Summary will update on new commits.</sup>

[...]

PR #13890, opened 2026-08-19
checkout: add unit quantity selection

strandhvilliam:
## Summary

Adds unit-based pricing to the checkout UI. Follow-up to the unit-based subscriptions backend (#13871) and the dashboard/portal work (#13889).

We can probably in a follow-up reduce the components and lines by combining some duplicated logic between unit and seat prices!

## What

Users can now pick a unit quantity at checkout, see a per-tier breakdown in the order summary, and see correct amounts for unit-based products in the product switcher and price labels. Handles volume and graduated tiers, min/max unit limits, fixed base fee + per-unit combos, and merchant-defined unit labels.

## Why

The backend and dashboard already support unit-based prices, but checkout had no quantity selector and rendered blank or wrong amounts for those products, so they weren't purchasable.

## How

Mirrors the existing seat-based path: `getUnitPrice` next to `getSeatPrice`, `utils/units.ts` alongside `utils/seats.ts`, `UnitDetailRow` alongside `SeatDetailRow`, and `CheckoutUnitSelector` mounted where `CheckoutSeatSelector` already is. Tier math lives once in `getUnitTierRows` and is reused by the breakdown, the switcher's "From" price, and the minimum-purchase amount.

| Screenshot |
|---|
| Screenshot 2026-08-21 at 17 41 01 |

## Checklist

---
## Summary by cubic
Adds unit quantity selection to checkout and aligns unit-tier pricing and labels across the switcher, price labels, and breakdown. Previously unit products had no quantity control and could show English labels or first-tier rates; now buyers pick units within limits, inputs are validated, totals use net amounts, and fixed+unit products show the base fee separately.

- Review notes:
  - Checkout UI: renders a unit selector when a unit price exists; enforces min/max (fixed count when min=max); rejects non-numeric/unsafe input; auto-corrects below-min once and disables controls during the correction; headline total uses net_amount.
  - Pricing breakdown: adds unit rows for volume and graduated tiers; shows a base fixed row when combined with per-seat or per-unit pricing.
  - Product switcher: unit prices show net_amount when selected; “From” totals minimum_units against the matching tier; fixed+unit displays “base + per‑unit” with localized unit labels.
  - Price labels: `ProductPriceLabel` resolves the per‑unit rate from minimum_units and tier type.
  - Utilities/i18n: adds `getUnitPrice` and unit helpers (`getUnitTierRows`/`getMinimumUnitAmount`/`getBasePricePerUnit`/`getUnitRows`/`getUnitLabels`); adds `checkout.pricing.perUnit` and `checkout.pricing.units.*` keys in `@polar-sh/i18n` and regenerates locales.

- Migration:
  - If you implement a custom checkout update handler, accept the `units` field in `CheckoutUpdatePublic`.

<sup>Written for commit e09a0a48eb4f0bd3be78d147bf795743b76015df. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/polarsource/polar/pull/13890?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


clients/packages/checkout/src/components/CheckoutUnitSelector.tsx
```diff
@@ -0,0 +1,223 @@
...
+  // Auto-correct unit count if it's below the minimum (only attempt once)
+  useEffect(() => {
+    if (
+      isUnitBased &&
+      needsUnitCorrection &&
+      !isFixedUnits &&
+      !isUpdating &&
+      !autoCorrectAttempted.current
+    ) {
+      autoCorrectAttempted.current = true
```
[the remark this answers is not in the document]
strandhvilliam: This is the same for `seats` so lets keep this for now

clients/packages/checkout/src/components/UnitQuantityControl.tsx
```diff
@@ -0,0 +1,170 @@
...
+  const stopEditing = () => {
+    setIsEditing(false)
+    setInputValue('')
+  }
+
+  const handleInputBlur = () => {
+    const newUnits = Number.parseInt(inputValue, 10)
+    if (
+      !Number.isNaN(newUnits) &&
+      newUnits >= minimumUnits &&
```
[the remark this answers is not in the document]
strandhvilliam: This is fine

clients/packages/checkout/src/utils/units.ts
```diff
@@ -0,0 +1,109 @@
...
+  const forms = resolveUnitLabelForms(source?.unit_label, locale)
+  const unitLabel = forms['=1']?.trim() || forms['other']?.trim() || 'unit'
+  const unitLabelPlural = forms['other']?.trim() || 'units'
+  return { unitLabel, unitLabelPlural }
+}
+
+type UnitTier = schemas['Tier-Output']
+
+function sortTiers(tiers: UnitTier[]): UnitTier[] {
+  return tiers.toSorted((a, b) => {
```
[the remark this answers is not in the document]
strandhvilliam: Will refactor to reduce repetition in follow up PR

clients/packages/checkout/src/components/CheckoutUnitSelector.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+      update({
+        units: minimumUnits,
+      }).catch((err) => {
+        setError(getErrorMessage(err))
+      })
```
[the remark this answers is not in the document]
strandhvilliam: This matches the seats behavior for now. Also the react linter breaks. Skipping this for now, will work fine 99.99%

clients/packages/i18n/src/locales/hu.ts
```diff
@@ -94,6 +94,20 @@ export default {
       },
       inclTax: 'ÁFA (tartalmazza)',
       basePrice: 'Alapár',
+      perUnit: '{unitLabel}enként',
```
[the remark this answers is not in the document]
strandhvilliam: This is autogenerated, so keeping it but might be worth thinking of in future

clients/packages/checkout/src/components/CheckoutUnitSelector.tsx
```diff
@@ -0,0 +1,190 @@
...
+  const displayUnits = Math.max(checkout.units || minimumUnits, minimumUnits)
+  const needsUnitCorrection =
+    checkout.units !== null &&
+    checkout.units !== undefined &&
+    checkout.units < minimumUnits
+
+  const netAmount = checkout.net_amount || 0
+  const currency = checkout.currency ?? 'usd'
+  // Auto-correct unit count if it's below the minimum (only attempt once)
+  useEffect(() => {
```
sebastianekstrom, line 69: Why do we need this `useEffect`?

clients/packages/checkout/src/components/CheckoutUnitSelector.tsx
```diff
@@ -0,0 +1,190 @@
...
+import { getUnitPrice, type ProductCheckoutPublic } from '../guards'
+import { ErrorResponse } from '../providers/CheckoutProvider'
+import { capitalize } from '../utils/string'
+import { getUnitLabels } from '../utils/units'
+import MeteredPricesDisplay from './MeteredPricesDisplay'
+import { UnitQuantityControl } from './UnitQuantityControl'
+
+export interface CheckoutUnitSelectorProps {
+  checkout: ProductCheckoutPublic
+  update: (
```
sebastianekstrom, line 20: Nit pick, but reading through the component it's a bit hard to understand what `update` is, could we make the variable name a bit more verbose?

clients/packages/checkout/src/components/CheckoutUnitSelector.tsx
```diff
@@ -0,0 +1,190 @@
...
+    })
+      .catch((error) => {
+        setError(getErrorMessage(error))
+      })
+      .finally(() => {
+        setIsUpdating(false)
+      })
+  }
+
+  if (compact) {
```
sebastianekstrom, line 117: We always seem to be passing `compact` to this component (except for in the tests), do we need the non-compact mode?

clients/packages/checkout/src/components/UnitDetailRow.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <DetailRow
+      title={t('checkout.pricing.units.count', {
+        count: row.units,
+        unitLabel,
+        unitLabelPlural,
+      })}
+      subtitle={
+        '· ' +
```
sebastianekstrom, line 36: I don't think we use this `·` separator elsewhere in checkout 🤔 I think just a space is fine!

clients/packages/checkout/src/components/UnitQuantityControl.tsx
```diff
@@ -0,0 +1,170 @@
...
+  )
+}
+
+export const UnitQuantityControl = ({
+  units,
+  minimumUnits,
+  maximumUnits,
+  isUpdating,
+  onUpdate,
+  compact = false,
```
sebastianekstrom, line 55: Kinda hard to tell from just looking at the code, but what's the difference between compact and non-compact for this component? The prop adds quite a bit of code, so was just wondering if the UI difference between the two is big enough to justify that.

clients/packages/checkout/src/components/UnitQuantityControl.tsx
```diff
@@ -0,0 +1,170 @@
...
+  minimumUnits: number
+  maximumUnits: number | null
+  isUpdating: boolean
+  onUpdate: (units: number) => Promise<void>
+  compact?: boolean
+}
+
+const NUMERIC_INPUT_PATTERN = /^\d+$/
+
+const QuantityIcon = ({
```
sebastianekstrom: Nit: We have a few icons defined as inline SVG's in the checkout which bothers me a bit since we are kinda reinventing the icon-wheel every time we add a new one. I'd like if we could add a `clients/packages/checkout/src/components/icons` or similar where we gather all of these to help make the UI consistent.

But we can add a separate issue for that.

clients/packages/checkout/src/components/UnitQuantityControl.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  const handleInputBlur = () => {
+    const newUnits = Number.parseInt(inputValue, 10)
+    if (
+      Number.isSafeInteger(newUnits) &&
+      newUnits >= minimumUnits &&
+      (!hasMaximumLimit || newUnits <= maximumUnits) &&
+      newUnits !== units
+    ) {
+      void onUpdate(newUnits)
```
sebastianekstrom: Why do we need `void` here?

clients/packages/checkout/src/components/CheckoutUnitSelector.tsx
strandhvilliam: Same behavior as already exist in `CheckoutSeatSelector.tsx` that (if I understand correctly) is a self-heal if a stale checkout handling the edge case where a minimum is while a checkout is open. Can't really see why this would happen but added it since it's in the similar seat_based one!

clients/packages/checkout/src/components/CheckoutUnitSelector.tsx
strandhvilliam: Totally agree, didn't even know what it did myself. Copied the pattern as the seat ones already have. Think a refactor of both might be in order.

clients/packages/checkout/src/components/CheckoutUnitSelector.tsx
strandhvilliam: Yep only the tests are not using it so it feels unnecessary. Same behavior for `seats` so thinking it will go in the refactor of both `seats` and `units`

clients/packages/checkout/src/components/UnitDetailRow.tsx
strandhvilliam: Screenshot 2026-08-28 at 13 51 42

We do actually in the seat UI aready. Maybe we can use a space for the existing seat based UI too.

clients/packages/checkout/src/components/UnitQuantityControl.tsx
strandhvilliam: Copied from `seats` as well (even though `seats` has that inlined instead of separate component), but as you mentioned not used anyway. Can remove all of this compact stuff in separate refactor of both existing `seats` and new `units`

clients/packages/checkout/src/components/UnitQuantityControl.tsx
strandhvilliam: Yeah. Also looking at my code today the conditional splitting of increment/decrement svg paths makes my brain hurt a bit. Putting it in its own icons/ directory sounds like a good idea!

clients/packages/checkout/src/components/UnitQuantityControl.tsx
strandhvilliam: we don't, just sometimes use it to make it clear we don't care about awaiting the promise. No difference in actual functionality if I'm not totally out riding a bike

strandhvilliam:
I think all concerns above are valid and should be addressed. However they are a result of the existing pattern already existing for seat based pricing, which this mimicks. Instead of addressing them for units here, I suggest we make a separate issue for it and refactor for both units and seats, which probably also means we can reuse a lot of code between them.

Making an issue with the above comments to adress.

sebastianekstrom (approved):

PR #13891, opened 2026-08-20
Add Hong Kong Business Registration Number validation

pieterbeulque:
## Summary

Adds validation for Hong Kong Business Registration Numbers (`hk_br`), which previously had no dedicated validator and were always rejected.

## What

- Added `HKBRValidator` and wired it into `_get_validator()` for `TaxIDFormat.hk_br`.
- Validates the **format only**: 8 digits, after stripping common separators (spaces, dots, dashes, and the parentheses sometimes placed around the trailing check digit, e.g. `1234567(8)`).
- Added test cases for valid 8-digit numbers (plain and separator/parenthesis forms) plus one invalid (too short).

## Why

`hk_br` was declared in the enum and country map but had no validator, so it fell through to the `stdnum` fallback — and since `stdnum` has no `hk.br` module, every Hong Kong BR number was rejected.

We deliberately do **not** verify a check digit: Hong Kong has no publicly documented BR check-digit algorithm (available sources conflict and none is authoritative), so a checksum can't be verified and would risk rejecting genuine numbers. Hong Kong also has no VAT, so the numeric validity of the tax ID is not load-bearing. This matches how Stripe validates `hk_br` and the other non-`stdnum` validators (`AE`/`GE`/`MU`), which are format-only.

## Notes

- No API/schema change — `hk_br` was already part of `TaxIDFormat`; OpenAPI diff is empty.

https://claude.ai/code/session_01Foc6R3DM5SV4v3813XwTgt

### DIFF
--- a/server/polar/tax/tax_id.py
+++ b/server/polar/tax/tax_id.py
@@ -525,7 +525,7 @@ def validate_tax_id(number: str, country: str) -> TaxID:
             try:
                 validator = _get_validator(tax_id_type)
                 return validator.validate(number, country), tax_id_type
-            except (UnsupportedTaxIDFormat, InvalidTaxID):
+            except UnsupportedTaxIDFormat, InvalidTaxID:
                 continue
     raise InvalidTaxID(number, country)
 
--- a/server/polar/tax/tax_id.py
+++ b/server/polar/tax/tax_id.py
@@ -436,6 +436,21 @@ class GEVATValidator(ValidatorProtocol):
         return number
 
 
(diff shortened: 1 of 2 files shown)


pieterbeulque:
Both Cubic & this PR are wrong in the checksum validation. There's no published algorithm, so dumbing it down to an 8 digit validator instead. Since Hong Kong doesn't have VAT, the validity of the tax number doesn't really matter.

frankie567 (approved):

PR #13941, opened 2026-08-21
lambda: Implement group completion for SQS/lambda workers

Yopi:
We're using Dramatiq's group completion to ensure that a job runs after all grouped jobs have completed. This is today only used by enqueue_benefits_grants: We run benefit.revoke after they are done we run enqeueue_grants which enqueues a buch of benefit.grant tasks.

To be able to move these tasks over to the new way, we need to have support for this in the lambda workers.

<a href="https://cubic.dev/pr/polarsource/polar/pull/13941?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/worker/_broker.py
+++ b/server/polar/worker/_broker.py
@@ -205,6 +205,9 @@ class RoutingRedisBroker(RedisBroker):
                         CorrelationID.get(),
                         message.message_id,
                         message.options.get("debounce_key"),
+                        message_options=_sqs.extract_group_completion_options(
+                            message.options
+                        ),
                     )
                 ]
             )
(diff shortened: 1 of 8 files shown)


Yopi:
_This_ should be the last one

joebon (approved):

PR #13969, opened 2026-08-24
product: allow unit-based pricing on one-time products

strandhvilliam:
The initial implementation only allowed unit_based price type products to be recurring subscriptions. This PR allows products to configure one-time purchases for unit-based price types, similarly to how seat based price types already works.

Usage of the unit-based price type is gated behind an internal feature flag


frankie567 (approved):
Very good, Villiam :)

PR #13973, opened 2026-08-25
fix(payout): skip paid_at assertion for manual payout invoices

pieterbeulque:
## Problem

Generating an invoice for a **manual** payout crashes the `payout.invoice` background task, which then retries forever. Observed in Logfire: payout `d738a37d-f2df-4bde-9067-4257ba66578f` failed 52× since Aug 24 (still ongoing), each time with:

```
File "polar/invoice/service.py", line 116, in create_payout_invoice
    assert payout.paid_at is not None
AssertionError
```

## Root cause

`Payout.paid_at` isn't a column — it's a property derived from the first **succeeded** `PayoutAttempt` that has a `paid_at` (`models/payout.py`). A manual payout never goes through a processor, so it has no attempt and `paid_at` stays `None`.

Meanwhile `Payout.status` becomes `succeeded` via the DB trigger, so `trigger_invoice_generation`'s gate (`status == succeeded`) passes and the merchant can request an invoice — which then hits the unconditional `assert payout.paid_at is not None` and crashes. Because the assert is uncaught, Dramatiq retries indefinitely.

## Fix

- Only assert `paid_at is not None` when `payout.processor != PayoutAccountType.manual`.
- Omit the "Paid at" heading item from the rendered invoice when `paid_at` is absent (`InvoiceHeadingItem.value` is `str | datetime`, so passing `None` would fail Pydantic validation).

## Test

Added `TestCreatePayoutInvoice::test_manual_payout_without_paid_at`: builds a real manual payout (no attempt → `paid_at is None`) and asserts invoice generation succeeds and the "Paid at" heading is omitted.

<a href="https://cubic.dev/pr/polarsource/polar/pull/13973?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/invoice/service.py
+++ b/server/polar/invoice/service.py
@@ -4,6 +4,7 @@ import json
 from datetime import datetime
 
 from polar.config import settings
+from polar.enums import PayoutAccountType
 from polar.exceptions import PolarError
 from polar.integrations.aws.s3 import S3Service
 from polar.kit.utils import utc_now
@@ -113,7 +114,8 @@ class InvoiceService:
         # Sanity check to make sure the amounts add up correctly
         assert payout.fees_amount == abs(current_payout_fees_amount)
         assert payout.amount == gross_amount + payout_fees_amount + payment_fees_amount
-        assert payout.paid_at is not None
+        if payout.processor != PayoutAccountType.manual:
+            assert payout.paid_at is not None
 
         items = [
             InvoiceItem(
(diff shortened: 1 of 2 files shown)


pieterbeulque:
Replaced by https://github.com/polarsource/polar/pull/13974

PR #13974, opened 2026-08-25
Fix payout invoice generation crash-loop for manual payouts

pieterbeulque:
## Problem

The `payout.invoice` task was crash-looping on a single manual payout (`d738a37d-…`) — 52 retries since Aug 24, every one an `AssertionError` on `assert payout.paid_at is not None`.

**Root cause:** `Payout.paid_at` is a property derived from a succeeded `PayoutAttempt`. Manual payouts have no attempt, so `paid_at` is always `None`. The DB trigger still flips status to `succeeded`, so invoice generation gets triggered — and then asserts and retries forever.

## Fix

In `polar/invoice/service.py`, fall back to `payout.created_at` for manual payouts instead of asserting on `paid_at`:

```python
if payout.processor == PayoutAccountType.manual:
    paid_at = payout.created_at
else:
    assert payout.paid_at is not None
    paid_at = payout.paid_at
```

The assert is kept for Stripe payouts, where `paid_at` genuinely should always be set — dropping that invariant would hide real bugs. The computed `paid_at` is used for the invoice's "Paid at" heading item.

Once shipped, the stuck payout will generate its invoice on the next retry.

<a href="https://cubic.dev/pr/polarsource/polar/pull/13974?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/invoice/service.py
+++ b/server/polar/invoice/service.py
@@ -4,6 +4,7 @@ import json
 from datetime import datetime
 
 from polar.config import settings
+from polar.enums import PayoutAccountType
 from polar.exceptions import PolarError
 from polar.integrations.aws.s3 import S3Service
 from polar.kit.utils import utc_now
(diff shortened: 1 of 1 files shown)


pieterbeulque:
@stilla-bot Can you create an issue to solve this properly by marking manual payouts as paid properly with a payout attempt!

PR #14023, opened 2026-08-28
Speed up backoffice organization list and detail views

pieterbeulque:
## Problem

The backoffice organization pages load in 6–38s. Confirmed in Logfire (production, last 24h):

- **Detail view** (`/organizations/{id}`): up to **38s**, frequently hitting the 30s statement timeout and returning a **500**.
- **List view** (`/organizations/`): **6–18s**, ~14s of it in a single query.

## Detail view

`PaymentAnalyticsService.get_refund_stats` filtered on `customers.organization_id`, which is two joins away from `refunds` (`refunds → orders → customers`). Postgres can't narrow by org up front, so the plan does a **Seq Scan over the entire `refunds` table** and nested-loops each into orders → 30s timeout.

Fix: `orders` has a direct, indexed `organization_id`. Filter `Order.organization_id` and drop the `customers` join. One join away, via an index.

## List view

The default status/priority ordering (`ORDER BY status DESC, status_updated_at ASC NULLS FIRST`) had **no supporting index** — no index on `status` or `status_updated_at` existed. Every load seq-scanned the whole `organizations` table  and sorted all wide rows in memory before applying `LIMIT`.

Fix: partial composite index on `(status DESC, status_updated_at ASC NULLS FIRST) WHERE deleted_at IS NULL`, built `CONCURRENTLY`. Column directions match the `ORDER BY` exactly so the planner can walk the index and stop at the page limit with no sort. The leading `status` column also serves any status-only filter.

## Deploy safety

- Additive index built `CONCURRENTLY` (no write lock), in its own migration (can't run in a transaction). Safe alongside running code.
- Detail-view change is query-only.
- Migration verified locally: upgrade + downgrade/re-upgrade cycle clean.

### DIFF
--- a/server/migrations/versions/2026-08-28-1435_add_partial_index_on_organizations_.py
+++ b/server/migrations/versions/2026-08-28-1435_add_partial_index_on_organizations_.py
@@ -1,0 +1,50 @@
+"""add partial index on organizations status for backoffice list
+
+Revision ID: b58196b41f41
+Revises: a83cf131398d
+Create Date: 2026-08-28 14:35:09.753552
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# Polar Custom Imports
+
+# revision identifiers, used by Alembic.
+revision = "b58196b41f41"
+down_revision = "a83cf131398d"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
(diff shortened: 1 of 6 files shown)


Yopi (approved):
Nice!

frankie567 (approved):

PR #14063, opened 2026-09-01
server/webhook: make api_version non nullable

frankie567:
The migration was applied, and the backfill as well, so we can now make those columns non nullable. There is a backfill query to be nice with local dev environments, but it should be a no-op in production since the backfill has already been applied.

<a href="https://cubic.dev/pr/polarsource/polar/pull/14063?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


maximevast (approved):

server/migrations/versions/2026-09-03-0700_make_webhookendpoint_webhookevent_api_.py
```diff
@@ -0,0 +1,65 @@
...
+        UPDATE webhook_endpoints
+        SET api_version = '2026-04'
+        WHERE api_version IS NULL
+        """
+    )
+    op.execute(
+        """
+        UPDATE webhook_events
+        SET api_version = '2026-04'
+        WHERE api_version IS NULL
```
joebon, line 36: Hrm, this table holds ~42M rows. I sanity checked this against our read replica:
```
SELECT COUNT(*) FROM webhook_events WHERE api_version IS NULL;
```

That query took 2 minutes(!!) to complete and returned 40M rows so I don't think this will hold and also the assumption that all backfills have been applied seems wrong 😬

frankie567: Yes, it's actually *currently* running, I opened the PR a bit early

frankie567: That's also probably why it took 2 minutes, the backfill is holding big batches 🙂

frankie567: "A bit early"

...

Capture d’écran 2026-09-01 à 16 00 27

PR #14067, opened 2026-09-01
server/webhooks: send version-aware payloads

frankie567:
Actual logic that will send version-aware webhook payloads. The frontend part is deliberately omitted, and will be added in a follow-up PR.

Merge after #14063

<a href="https://cubic.dev/pr/polarsource/polar/pull/14067?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/kit/versioning.py
+++ b/server/polar/kit/versioning.py
@@ -14,10 +14,10 @@ from collections.abc import (
 
 from fastapi import APIRouter, FastAPI
 from fastapi.routing import APIRoute, RouteContext, iter_route_contexts
-from pydantic import Field, GetJsonSchemaHandler
+from pydantic import Field, GetCoreSchemaHandler, GetJsonSchemaHandler
 from pydantic.fields import FieldInfo
 from pydantic.json_schema import JsonSchemaValue
-from pydantic_core import CoreSchema, PydanticOmit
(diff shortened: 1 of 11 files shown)


server/polar/kit/versioning.py
```diff
@@ -64,6 +64,27 @@ def from_scope(
         raw_version = headers.get(VERSION_HEADER)
         return cls.parse(raw_version) if raw_version else default
 
+    @classmethod
+    def __get_pydantic_core_schema__(
```
frankie567, line 68: Since APIVersion is now included in Pydantic schemas, we need to hook into Pydantic logic so it parses and serializes it as a string. Otherwise, it automatically picks up the dataclass and expect a sub-object with year and month fields.

strandhvilliam (approved):
The pydantic hook for custom types thing is pretty cool :D

frankie567:
This PR has become a mess, will reopen

PR #14113, opened 2026-09-02
server: revamp API transaction handling

frankie567:
Before, we were relying on a FastAPI yield dependency to handle the SQLAlchemy session lifecycle. This was a bit brittle because the teardown part is only executed by FastAPI _after_ the response is sent. In some edge cases, it means the client could try to request a resource that has not yet been committed, even though they got a success response. Besides, we had to pollute it with Job Queue management to make sure to clear messages if an exception happens.

With this change, we centralize the transaction management into an ASGI middleware that commits right before the HTTP response starts. The main challenge was to coordinate the case where an endpoint returns an handled exception (like a 404). In this case, we want to rollback the transaction, but the exception is masked to the middleware because of the built-in ExceptionMiddleware that handles it. The solution we found was to wrap the FastAPI route handler to catch exception and save a flag in the request state that can be read by our middleware.

<a href="https://cubic.dev/pr/polarsource/polar/pull/14113?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


frankie567:
@cubic-dev-ai Why do you say in your review "Implementation details of the new middleware are not visible in the diff"? It's right there in `server/polar/middlewares.py`!

frankie567:
@cubic-dev-ai The diff _is_ here, trust me. The middleware is called TransactionalMiddleware and there is +41
-6 diff. Common.

joebon (approved):

PR #14117, opened 2026-09-03
Switch to signing webhooks the standard-compliant way

joebon:
Switch our way of signing webhooks to follow the [https://www.standardwebhooks.com](<https://www.standardwebhooks.com>) approach.

To not break existing webhook integrations relying on our previous way of signing, we'll keep signing webhooks registered (or webhook secrets generated) before a cutoff date time.

SDK `1.0.0-alpha.19` and later supports both signature approaches.

Cut-off date set to *2026-09-08*

### Dashboard screenshots showing notices & badges

Review in cubic

### DIFF
--- a/docs/guides/encore.mdx
+++ b/docs/guides/encore.mdx
@@ -204,7 +204,12 @@ export const webhooks = api.raw(
 ```
 
 <Warning>
-  The `standardwebhooks` library expects the secret to be base64-encoded. Polar provides a raw string (starting with `polar_whs_`), so you must base64-encode the entire secret before passing it to `new Webhook()`.
+  Polar webhook signatures use two key derivations depending on when the secret
+  was generated. Polar SDKs (version 1.0.0-alpha.19 and later) try both. If you
+  verify with `standardwebhooks` directly: secrets created **on or after 8
+  September 2026, 00:00 UTC** can be passed to `new Webhook()` as-is (`whsec_…`).
+  Older Polar secrets need the **entire** secret UTF-8 string base64-encoded
+  first.
 </Warning>
 
 ## Local Development
--- a/docs/guides/encore.mdx
+++ b/docs/guides/encore.mdx
@@ -163,20 +163,22 @@ export const webhooks = api.raw(
       headers[key] = Array.isArray(value) ? value[0] : (value || "");
     }
 
-    // Verify the webhook signature
+    // Verify the webhook signature. Try Standard Webhooks first (secret as-is),
+    // then Polar's legacy HMAC (UTF-8 bytes of the full secret, including `whsec_`).
+    const secret = POLAR_WEBHOOK_SECRET().trim();
     let payload: any;
     try {
-      const base64Secret = Buffer.from(
-        POLAR_WEBHOOK_SECRET().trim(),
-        "utf-8"
-      ).toString("base64");
-      const wh = new Webhook(base64Secret);
-      payload = wh.verify(body, headers);
-    } catch (error: any) {
-      console.error("[Polar] Invalid webhook signature:", error?.message);
-      resp.writeHead(403);
-      resp.end(JSON.stringify({ error: "Invalid signature" }));
-      return;
+      payload = new Webhook(secret).verify(body, headers);
+    } catch {
+      try {
+        const legacySecret = Buffer.from(secret, "utf-8").toString("base64");
+        payload = new Webhook(legacySecret).verify(body, headers);
+      } catch (error: any) {
+        console.error("[Polar] Invalid webhook signature:", error?.message);
+        resp.writeHead(403);
+        resp.end(JSON.stringify({ error: "Invalid signature" }));
+        return;
+      }
     }
 
     // Handle the payload
--- a/docs/integrate/webhooks/delivery.mdx
+++ b/docs/integrate/webhooks/delivery.mdx
@@ -102,15 +102,22 @@ easily validate signatures. Or you can follow their
 in case you want to roll your own.
 
 <Info>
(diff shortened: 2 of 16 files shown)


server/polar/webhook/constants.py
```diff
@@ -1 +1,10 @@
+from datetime import UTC, datetime
+
 WEBHOOK_SECRET_PREFIX = "whsec_"
+WEBHOOK_SECRET_KEY_BYTES = 32
+
+# Secrets created before this instant keep Polar's original HMAC (UTF-8 of the
+# full `whsec_…` string). That includes endpoints stamped with
+# `secret_generated_at` after that column shipped but before spec signing.
+# Bump this if deploy slips so those rows do not flip scheme.
+WEBHOOK_STANDARD_SIGNATURE_CUTOFF = datetime(2026, 9, 8, tzinfo=UTC)
```
pieterbeulque, line 10: I thought we already flipped, and that that was causing some of the user issues?

docs/guides/encore.mdx
```diff
@@ -204,7 +204,12 @@ export const webhooks = api.raw(
 ```
 
 <Warning>
-  The `standardwebhooks` library expects the secret to be base64-encoded. Polar provides a raw string (starting with `polar_whs_`), so you must base64-encode the entire secret before passing it to `new Webhook()`.
+  Polar webhook signatures use two key derivations depending on when the secret
```
pieterbeulque, line 209: I wonder if we should bring a notice in-app for a week or three, maybe for `organization.created_at < cutoff date && webhook.secret_created_at > cutoff date` that we changed this, and that it may require an update to their implementation (and noting that the SDK can handle both).

server/polar/webhook/constants.py
pieterbeulque, line 10: Random idea: should we align this with the `2026-10` API version release, so that it feels less arbitrary of a date? Or is something actively broken right now?

docs/integrate/webhooks/delivery.mdx
```diff
@@ -221,6 +228,7 @@ A common cause is hosting providers like Vercel that redirect between `www` and
 
 ### Invalid signature exceptions
 
-Rolling your own webhook validation logic? Make sure to base64 encode the secret
-you configured on Polar in your code before generating the signature to validate
-against.
+Rolling your own webhook validation logic? Match the scheme for that endpoint
```
pieterbeulque: "Match the scheme" reads a bit weird. We can maybe just start this sentence at "Older Polar secrets…"

pieterbeulque (commented):
I added some thoughts on the DX, but I'll leave it to @Yopi or @frankie567 to chime in on the webhook signature signing itself!

server/polar/webhook/constants.py
joebon: Hrm, nope. We still don't treat the `whsec_` prefix correctly - hence this dance

docs/integrate/webhooks/delivery.mdx
joebon: Agree, thanks! Fixing

server/polar/webhook/constants.py
joebon: Well, we're not following the specification. I'd prefer to not linger longer than necessary on this.

server/polar/webhook/service.py
```diff
@@ -149,7 +162,7 @@ async def create_endpoint(
             organization.id,
             OrganizationPermission.organization_manage,
         )
-        secret = generate_token(prefix=WEBHOOK_SECRET_PREFIX)
+        secret = generate_webhook_secret()
```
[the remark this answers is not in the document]
joebon: That edge case mismatch is acceptable to avoid complexity

server/polar/webhook/tasks.py
```diff
@@ -26,11 +27,35 @@
...
+def sign_webhook(
+    secret: str,
+    msg_id: str,
+    timestamp: datetime,
+    payload: str,
+    *,
+    secret_generated_at: datetime | None,
+) -> str:
+    if uses_standard_webhook_signature(secret_generated_at):
+        return StandardWebhook(secret).sign(msg_id, timestamp, payload)
```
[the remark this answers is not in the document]
joebon: Hrm, I think it's fine to skip this to not over-complicate things.

docs/guides/encore.mdx
joebon: Yeah, that makes sense! I'll try to work something like that into the webhook endpoint pages in dashboard to make this clearer

server/polar/webhook/tasks.py
frankie567: In the SDK implementation, we removed the `whsec_` prefix before computing the signature. Shouldn't we do the same here? Or am I lost again? 😅

joebon: Heh, yeah it's not crystal clear just by looking at the code here. `StandardWebhook` constructor accepts the full secret, including the `whsec_` prefix and handles the stripping internally. See [https://github.com/standard-webhooks/standard-webhooks/blob/main/libraries/python/standardwebhooks/webhooks.py#L29](<https://github.com/standard-webhooks/standard-webhooks/blob/main/libraries/python/standardwebhooks/webhooks.py#L29>)

frankie567: Oh dear! Thanks for the pointer 🙂

frankie567 (approved):

docs/guides/encore.mdx
joebon: WDYT @pieterbeulque?

pieterbeulque (approved):

PR #14119, opened 2026-09-03
chore(adapters): migrate adapters to new sdk

simonxabris:
Migrates all adapters to the new Polar SDK and uses standalone function imports so bundles include only the API operations they need. Also adds missing webhook events.

Related issue: [PLR-59](https://linear.app/polarsh/issue/PLR-59/migrate-adapters-to-new-sdk).

**Better Auth breaking change:** construct the plugin's client with `createPolarCore` instead of `createPolar`. Custom `use` plugins now receive `PolarCore` and call standalone SDK functions. The README and example are updated.

### Bundle sizes

Before/after the standalone-import migration, both using SDK `1.0.0-alpha.20`. Minified ESM, SDK included, framework dependencies external. Better Auth includes creation of the caller's SDK client.

| Adapter | Before | After | Reduction |
| --- | ---: | ---: | ---: |
| Next.js | 76.24 KiB | 13.77 KiB | 81.9% |
| Nuxt runtime | 75.56 KiB | 13.12 KiB | 82.6% |
| TanStack Start | 75.86 KiB | 13.43 KiB | 82.3% |
| Better Auth, including client | 111.34 KiB | 65.47 KiB | 41.2% |
| adapter-utils | 4.32 KiB | 4.32 KiB | 0.0% |

[Full report, gzip sizes, and reproduction instructions](https://github.com/polarsource/polar/blob/chore/update-adapter-utils-sdk/clients/adapters/BUNDLE_SIZE_REPORT.md).

Validation: all five adapter builds and type checks, frontend lint, and 270 tests passed.


clients/adapters/adapter-utils/src/entitlement/entitlement.ts
```diff
@@ -1,19 +1,17 @@
...
+import type { models, webhooks } from '@polar-sh/sdk/2026-04'
 
 export type EntitlementProperties = Record<string, string>
 
-export type EntitlementHandler = (
-  payload:
-    | WebhookBenefitGrantCreatedPayload
-    | WebhookBenefitGrantRevokedPayload,
-) => Promise<void>
+type EntitlementPayload =
```
pieterbeulque, line 5: This isn't your call but the Entitlement naming really caught me off guard here, haha

clients/adapters/nextjs/src/checkout/checkout.ts
```diff
@@ -46,32 +46,32 @@ export const Checkout = ({
...
-        customerName: url.searchParams.get('customerName') ?? undefined,
-        customerBillingAddress: url.searchParams.has('customerBillingAddress')
+        customer_email: url.searchParams.get('customerEmail') ?? undefined,
+        customer_name: url.searchParams.get('customerName') ?? undefined,
+        customer_billing_address: url.searchParams.has('customerBillingAddress')
           ? JSON.parse(url.searchParams.get('customerBillingAddress') ?? '{}')
           : undefined,
-        customerTaxId: url.searchParams.get('customerTaxId') ?? undefined,
-        customerIpAddress:
+        customer_tax_id: url.searchParams.get('customerTaxId') ?? undefined,
```
pieterbeulque: should we also update the search params casing?

pieterbeulque (approved):

clients/adapters/nextjs/src/checkout/checkout.ts
simonxabris: Yes that's a good idea to keep things consistent with the API

PR #14122, opened 2026-09-03
Only delete a payout account when no organization uses it

pieterbeulque:
A single Stripe payout account can be linked to more than one organization. Organization deletion right now would try to delete the payout account, so it checks on pending payouts. This works only under the assumption that one payout account is used by one organization.

In this PR, only actually delete a payout account when it is no longer referenced by any organization. When still shared, just unlink the requesting org and keep the account intact. This allows us to also loosen the check on pending payouts.

*Note:* this unblocks support, but it also feels a bit redundant, because from credit card networks, shared payout accounts can be considered transaction laundering, and we should remove the possibility of re-using payout accounts across merchants.

### DIFF
--- a/server/polar/organization/service.py
+++ b/server/polar/organization/service.py
@@ -886,18 +886,7 @@ class OrganizationService:
     async def _delete_payout_account(
         self, session: AsyncSession, organization: Organization
     ) -> None:
-        if organization.payout_account_id is None:
-            return
-
-        payout_account_repository = PayoutAccountRepository.from_session(session)
-        payout_account = await payout_account_repository.get_by_id(
-            organization.payout_account_id
-        )
-
-        if payout_account is None:
-            return
-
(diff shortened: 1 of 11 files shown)


server/polar/organization/service.py
```diff
@@ -886,18 +886,7 @@ async def soft_delete_organization(
...
-        payout_account_repository = PayoutAccountRepository.from_session(session)
-        payout_account = await payout_account_repository.get_by_id(
-            organization.payout_account_id
-        )
-
-        if payout_account is None:
-            return
-
-        await payout_account_service.delete(session, payout_account, unlink=True)
+        await payout_account_service.unlink_and_maybe_delete(session, organization)
```
frankie567, line 889: We can probably avoid the private method `_delete_payout_account` then

frankie567 (changes requested):

frankie567 (approved):

PR #14130, opened 2026-09-03
Centralize shared clients dependencies with pnpm catalog

simonxabris:
## Summary

We have the same dependencies scattered throughout the workspace. The `catalog` feature helps centralize them in one place so bumping a dep is a one line change, makes it harder to forgot to bump a dep everywhere.<br><br>This PR only handles deps that already used the same version everywhere, there a more shared deps with different versions across the workspace like `vitest` , `typescript` etc. I plan to handle those in separate PRs to not make a big messy PR.

Open in Web Open in Cursor


pieterbeulque:
Interesting! I believe some/most people on the team are still on pnpm 10.x (me included), and I wasn't aware of pnpm catalog. We should do this, though.

It might be worthwhile fixing the scattered dependencies (react / vitest / zod / …) first to apply this throughout everything?

simonxabris:
Well my idea was the opposite, to get the easy stuff in and then reconcile the deps that have different versions. I don't want to create a huge PR with random changes to vitest setups, typescript configs etc. I'd rather do those sperately if it makes sense.

simonxabris:
Oh and this is supported from pnpm 9.5 so hopefully works for everyone out of the box

pieterbeulque (approved):

PR #14134, opened 2026-09-03
feat(sdk): allow methods to receive arbitrary metadata keys

simonxabris:
## Summary

The generated SDK types did not allow for arbitrary metadata on events, only `_llm` and `_cost`. Modified the generator to properly handle `additionalProperties` and when present, allow arbitrary keys.

### DIFF
--- a/sdk/generator/python/template/polar/base.py
+++ b/sdk/generator/python/template/polar/base.py
@@ -222,8 +222,16 @@ def _register_extra_items_typed_dict(
 ) -> None:
     global _retort
 
-    field_types = typing.get_type_hints(model)
-    required_keys = model.__required_keys__
+    field_types = typing.get_type_hints(model, include_extras=True)
+    required_keys = set(model.__required_keys__)
+    for key, field_type in field_types.items():
+        field_origin = typing.get_origin(field_type)
+        if field_origin is typing.Required:
+            required_keys.add(key)
+        elif field_origin is typing.NotRequired:
+            required_keys.discard(key)
+        if field_origin in (typing.Required, typing.NotRequired):
+            field_types[key] = typing.get_args(field_type)[0]
 
     def load_extra_items_typed_dict(data: object) -> dict[str, typing.Any]:
         if not isinstance(data, collections.abc.Mapping):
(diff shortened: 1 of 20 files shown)


sdk/generator/python/template/polar/version/inputs.py
```diff
@@ -11,14 +15,14 @@
...
-{% if model.additional_properties %}
+{% if model.additional_properties and not model.fields %}
 {{ model.name }}: typing.TypeAlias = dict[str, {{ model.additional_properties | type_annotation }}]
 {% if model.description %}
 """{{ model.description }}"""
 {% endif %}
 
 {% else %}
-class {{ model.name }}(typing.TypedDict):
+class {{ model.name }}({% if model.additional_properties %}typing_extensions.TypedDict, extra_items={{ model.additional_properties | type_annotation }}{% else %}typing.TypedDict{% endif %}):
```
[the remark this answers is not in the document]
simonxabris: This is not actually an issue since `extra_items` can only ever be a primitive value and not enum, model or anything complex

sdk/generator/python/template/polar/base.py
```diff
@@ -56,6 +57,20 @@ def __init__(self, status_code: typing.Literal[429], retry_after: int | None = N
...
+
+    @abc.abstractmethod
+    def _mark_additional_properties_model(self) -> None: ...
+
+
+class AdditionalPropertiesMixin(_AdditionalPropertiesModel):
+    __slots__ = ()
+
+    def _mark_additional_properties_model(self) -> None:
+        pass
```
frankie567: That approach definitely makes sense, however, I'm wondering if it wouldn't be better to output TypedDict with extra_items for those schemas with `additionalProperties`. The only place we use that is in metadata. So I think it's a bit more idiomatic and sensible to be able to do:

```python
customer.metadata["foo"]
```

Rather than:

```
customer.metadata.additional_properties.foo
```

The only touchy thing is if we want to handle the `_llm` and `_cost` so they're output as TypedDict instead of dataclasses. I think it's *acceptable* in a first step if we have `event.metadata["_llm"].tokens` instead of `event.metadata["_llm"]["tokens"]`

What do you think?

simonxabris: Makes perfect sense, I'm on it!

frankie567 (approved):

PR #14141, opened 2026-09-04
server/webhooks: send version-aware payloads

frankie567:
Actual logic that will send version-aware webhook payloads. The frontend part is deliberately omitted, and will be added in a follow-up PR.

<a href="https://cubic.dev/pr/polarsource/polar/pull/14141?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/webhook/service.py
+++ b/server/polar/webhook/service.py
@@ -55,6 +55,7 @@ from polar.organization.resolver import get_payload_organization
 from polar.user_organization.service import (
     user_organization as user_organization_service,
 )
+from polar.version import CURRENT_API_VERSION
 from polar.webhook.repository import (
     WebhookDeliveryRepository,
     WebhookEndpointRepository,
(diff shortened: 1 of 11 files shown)


server/polar/webhook/service.py
```diff
@@ -851,7 +852,12 @@ async def send(
     ) -> list[WebhookEvent]:
         now = utc_now()
         payload = WebhookPayloadTypeAdapter.validate_python(
-            {"type": event, "timestamp": now, "data": data}
+            {
+                "type": event,
+                "timestamp": now,
+                "api_version": CURRENT_API_VERSION,
```
joebon: This tripped me a bit, can we add an inline comment it's just a placeholder and will be properly populated further down?

joebon (approved):

server/polar/webhook/service.py
frankie567: Sure!

PR #14144, opened 2026-09-04
Checkout: Compact mobile view

sebastianekstrom:
## Summary

In the mobile view of the checkout we throw a lot of numbers at the user, this is most certainly bad for conversion. To mitigate this I've created an experiment that hides all order details behind a `Order summary` accordion.

However there are a few types of products where this doesn't make sense, because user input is needed (PWYW, seats, etc.) so this feature will only be available for products that's either a one-time purchase or a subscription.

Everything is hidden behind and experiment (`checkout_collapsed_order_summary`).

## Screenshots

| Control | Treatment  |
|--------|--------|
| Screenshot 2026-09-04 at 16 49 41 | Screenshot 2026-09-04 at 16 49 34 |

<a href="https://cubic.dev/pr/polarsource/polar/pull/14144?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/apps/web/src/hooks/useIsMobileViewport.ts
+++ b/clients/apps/web/src/hooks/useIsMobileViewport.ts
@@ -1,0 +1,16 @@
+import { useSyncExternalStore } from 'react'
+
+const MD_BREAKPOINT_MEDIA_QUERY = '(min-width: 768px)'
+
+const subscribe = (onChange: () => void) => {
+  const mediaQueryList = window.matchMedia(MD_BREAKPOINT_MEDIA_QUERY)
+  mediaQueryList.addEventListener('change', onChange)
+  return () => mediaQueryList.removeEventListener('change', onChange)
+}
+
+const getSnapshot = () => !window.matchMedia(MD_BREAKPOINT_MEDIA_QUERY).matches
+
+const getServerSnapshot = () => false
+
+export const useIsMobileViewport = (): boolean =>
+  useSyncExternalStore(subscribe, getSnapshot, getServerSnapshot)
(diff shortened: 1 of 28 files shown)


clients/apps/web/src/hooks/useIsMobileViewport.ts
```diff
@@ -0,0 +1,16 @@
+import { useSyncExternalStore } from 'react'
+
+const MD_BREAKPOINT_MEDIA_QUERY = '(min-width: 768px)'
```
pieterbeulque, line 3: I'd include `pointer: coarse` so that we exclude laptops with a small window, maybe? Since we don't interactively opt people in and out of the experiment, and we want to target actual mobile devices

clients/packages/i18n/src/locales/nl.ts
```diff
@@ -263,6 +263,7 @@ export default {
       description:
         'Je hebt al een gratis proefperiode voor dit product gebruikt, dus er wordt vandaag kosten in rekening gebracht. Ga hieronder verder om je aankoop af te ronden.',
     },
+    orderSummary: 'Besteloverzicht',
```
pieterbeulque, line 266: It's not the most natural but I can't think of anything better.

pieterbeulque (approved):

clients/apps/web/src/hooks/useIsMobileViewport.ts
sebastianekstrom: The thing is that the actual toggling of the new UI is done through media queries in the actual UI, so then we would have a mismatch of users who is exposed (min-width 768 + pointer coarse) and people who actually see the new UI (min width 768). I also think it's kinda fine not focusing only mobile devices for this feature, small windows still benefit seeing the compact layout!

pieterbeulque: *That* is true, my friend!

PR #14148, opened 2026-09-04
Require successful charge on manual early trial conversion

pieterbeulque:
Similar to how we attempt the charge before updating a subscription, attempt the charge before converting a trial early.

We do this only for explicit early conversions by setting `trial_end` to `now`, the automatic trial conversion still will flip trialing → active, then a failed payment will mark it as past due. TBD if we also want to change that, it may make sense to change from trialing → canceled instead of trialing → past due, but I think that's a different discussion. Since this focuses only on the manual trial-conversion path, the merchant has control, so we don't have to think about that.

<a href="https://cubic.dev/pr/polarsource/polar/pull/14148?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/features/subscriptions/manage.mdx
+++ b/docs/features/subscriptions/manage.mdx
@@ -75,7 +75,7 @@ curl --request PATCH \
 You can add, extend, or end a [trial](/features/subscriptions/trials) on any subscription:
 
 - **Add or extend** a trial by setting `trial_end` to a future date. If the subscription is currently active, its status switches to `trialing` and the next charge is postponed to the new date.
-- **End a trial immediately** by setting `trial_end` to `"now"`. The subscription becomes `active` and a new billing cycle — and charge — starts on the spot.
+- **End a trial immediately** by setting `trial_end` to `"now"`. Polar charges the first billing cycle synchronously and only changes the subscription to `active` if payment succeeds. If payment fails, the request returns `402` and the subscription remains `trialing` with its original trial end.
 
 ```bash cURL
 curl --request PATCH \
--- a/docs/features/subscriptions/trials.mdx
+++ b/docs/features/subscriptions/trials.mdx
@@ -52,7 +52,7 @@ For existing subscriptions, you can add, extend or cancel a customer's trial per
 
 To add or extend a trial, set a new trial end date in the future. If the subscription was active, its status will be changed to **trialing**, and the billing will be postponed until the end of the trial.
 
-To cancel a trial, click on the **End trial** button. The subscription will become active immediately, and the customer will be charged immediately for a new billing cycle.
+To cancel a trial, click on the **End trial** button. Polar charges the customer immediately for a new billing cycle and only changes the subscription to **active** if payment succeeds. If payment fails, the subscription remains **trialing** with its original trial end.
 
 <img
   className="block dark:hidden"
--- a/server/polar/subscription/service.py
+++ b/server/polar/subscription/service.py
@@ -56,6 +56,7 @@ from polar.event.system import (
 )
 from polar.exceptions import (
     BadRequest,
+    PaymentNotReady,
     PolarError,
     PolarRequestValidationError,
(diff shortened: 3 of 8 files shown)


server/polar/subscription/service.py
```diff
@@ -495,10 +502,14 @@ async def __aexit__(
...
-        if effect == "cycle":
+    def set_billing_effect(
+        self, effect: Literal["invoice", "cycle", "cycle_sync"]
+    ) -> None:
+        if effect == "cycle_sync":
+            self._billing_effect = "cycle_sync"
+        elif effect == "cycle" and self._billing_effect != "cycle_sync":
             self._billing_effect = "cycle"
-        elif effect == "invoice" and self._billing_effect != "cycle":
+        elif effect == "invoice" and self._billing_effect is None:
```
frankie567: Is it necessary to have all this guard logic?

server/polar/subscription/service.py
```diff
@@ -4325,5 +4342,23 @@ async def _create_subscription_update_order(
...
         )
 
+    async def _create_subscription_cycle_order(
+        self,
+        session: AsyncSession,
+        subscription: Subscription,
+        billing_reason: OrderBillingReasonInternal,
+        *,
+        cutoff: datetime,
+    ) -> Order:
```
frankie567, line 4342: Could we factorize with `_create_subscription_update_order`? One inline import is already too much for my delicate eyes 😛

server/polar/subscription/service.py
pieterbeulque: No, that's really odd, sorry.

frankie567 (approved):

docs/features/subscriptions/manage.mdx
```diff
@@ -75,7 +75,7 @@ curl --request PATCH \
 You can add, extend, or end a [trial](/features/subscriptions/trials) on any subscription:
 
 - **Add or extend** a trial by setting `trial_end` to a future date. If the subscription is currently active, its status switches to `trialing` and the next charge is postponed to the new date.
-- **End a trial immediately** by setting `trial_end` to `"now"`. The subscription becomes `active` and a new billing cycle — and charge — starts on the spot.
+- **End a trial immediately** by setting `trial_end` to `"now"`. Polar charges the first billing cycle synchronously and only changes the subscription to `active` if payment succeeds. If payment fails, the request returns `402` and the subscription remains `trialing` with its original trial end.
```
pieterbeulque:
```suggestion
- **End a trial immediately** by setting `trial_end` to `"now"`. Polar will charge the first billing cycle and only change the subscription to `active` if payment succeeds. If payment fails, the subscription remains `trialing` with its original trial end.
```

docs/features/subscriptions/trials.mdx
```diff
@@ -52,7 +52,7 @@ For existing subscriptions, you can add, extend or cancel a customer's trial per
 
 To add or extend a trial, set a new trial end date in the future. If the subscription was active, its status will be changed to **trialing**, and the billing will be postponed until the end of the trial.
 
-To cancel a trial, click on the **End trial** button. The subscription will become active immediately, and the customer will be charged immediately for a new billing cycle.
+To cancel a trial, click on the **End trial** button. Polar charges the customer immediately for a new billing cycle and only changes the subscription to **active** if payment succeeds. If payment fails, the subscription remains **trialing** with its original trial end.
```
pieterbeulque:
```suggestion
To cancel a trial, click on the **End trial** button. Polar will charge the customer immediately and change the subscription to **active** if payment succeeds.```

PR #14154, opened 2026-09-05
clients/web: scope AI assistant MCP token to the current organization

pieterbeulque:
## Bug

The AI product-creation assistant creates products under the **wrong organization**.

`generateOAT()` in `clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/products/new/ai/chat/route.ts` cached the MCP access token in the `polar_mcp_session` cookie and returned it **unconditionally** whenever the cookie existed. The token is minted with `sub_type: 'organization', sub: organizationId` — i.e. bound to whichever org was open the first time the assistant was used — and the cookie name is a fixed constant, not per-org.

Server-side, an organization-scoped token always resolves to its own org (and 422s if a different `organization_id` is passed), so after switching organizations in the dashboard, the assistant kept reusing org A's token and deterministically created products in org A.

## Fix

Minimal, value-scoped cache:

- New colocated helper `mcpSessionCookie.ts` with `serializeMCPSessionCookie` / `parseMCPSessionCookie` — the cookie value now stores `{ organizationId, token }` as JSON.
- `generateOAT()` only reuses the cached token when its `organizationId` matches the organization the user currently has open; on mismatch (or any unparseable value) it mints a fresh org-scoped token and overwrites the cookie.
- The `hasToolAccess` heuristic in `POST` now also checks the cached token's org instead of mere cookie presence, so a stale token from another org no longer counts as tool access.
- Legacy bare-token cookie values fail JSON parsing and are treated as a cache miss, so existing sessions transparently migrate to the scoped format.

## Verification

- New unit tests `mcpSessionCookie.test.ts` (4 tests: same-org hit, cross-org miss, legacy bare-token miss, malformed/incomplete payload miss) — `pnpm vitest run` passes.
- `pnpm typecheck` in `clients/apps/web` — clean.
- `oxlint --type-aware` on the touched files and `oxfmt --check` — clean.

<a href="https://cubic.dev/pr/polarsource/polar/pull/14154?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/products/new/ai/chat/mcpSessionCookie.test.ts
+++ b/clients/apps/web/src/app/(main)/dashboard/[organization]/(header)/products/new/ai/chat/mcpSessionCookie.test.ts
@@ -1,0 +1,43 @@
+import { describe, expect, it } from 'vitest'
+import {
+  parseMCPSessionCookie,
+  serializeMCPSessionCookie,
+} from './mcpSessionCookie'
+
+const ORGANIZATION_ID = 'e2c8a9d0-1234-4f56-8a9b-0c1d2e3f4a5b'
+const OTHER_ORGANIZATION_ID = 'f3d9b0e1-5678-4a9b-8c0d-1e2f3a4b5c6d'
+const TOKEN = 'polar_oat_test'
+
+describe('parseMCPSessionCookie', () => {
+  it('returns the token when the cookie was minted for the same organization', () => {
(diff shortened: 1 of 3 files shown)


pieterbeulque:
Will pick it back up later.

PR #14229, opened 2026-09-07
Implement users syncing with Resend audience

frankie567:
It's time we have an automatically maintained audience so we can send Broadcasts to our active users. We add a `resend_id` column to User so we can store the Resend Contact ID and therefore properly unsubscribe/delete the old email and add the new one. We also make sure to delete the contact when deleting a user.

<a href="https://cubic.dev/pr/polarsource/polar/pull/14229?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/migrations/versions/2026-09-07-1553_add_user_resend_id.py
```diff
@@ -0,0 +1,36 @@
...
+# revision identifiers, used by Alembic.
+revision = "226afe0d783a"
+down_revision = "fec1e6242c3f"
+branch_labels: tuple[str] | None = None
+depends_on: tuple[str] | None = None
+
+
+def upgrade() -> None:
+    # Ensures we don't break app by applying a deadlock-inducing migration
+    op.execute("SET LOCAL lock_timeout = '5s'")
```
maximevast: 5 sec is enough for the user table ?

maximevast (approved):

server/migrations/versions/2026-09-07-1553_add_user_resend_id.py
frankie567: @stilla-bot I think it's fine for a nullable column, right? It should be fast, even on a big table.

frankie567: Good catch @maximevast, I tweaked the migration so the index is built concurrently

PR #14234, opened 2026-09-07
feat(checkout): enforce embed hosts through frame-ancestors

maximevast:
## Summary

Actual enforcement of for [ENG-12](https://linear.app/polarsh/issue/ENG-12/enforce-embed-hosts-through-frame-ancestors-csp). 

~To be merged later when we checked the logs for breaking framed page.~

Behind a feature flag, safe to merge now https://github.com/polarsource/polar/pull/14349

 
## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/14234?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/apps/web/src/proxy.ts
+++ b/clients/apps/web/src/proxy.ts
@@ -8,6 +8,7 @@ import {
   DISTINCT_ID_COOKIE,
   DISTINCT_ID_HEADER,
 } from './experiments/constants'
+import { checkoutCSP } from './csp.mjs'
 import { getServerURL } from './utils/api'
 import { createServerSideAPI } from './utils/client'
 import { CONFIG } from './utils/config'
@@ -95,7 +96,8 @@ const requiresAuthentication = (request: NextRequest): boolean => {
 }
 
 const CHECKOUT_CLIENT_SECRET = /^\/checkout\/([^/]+)/
-const NO_FRAME_ANCESTORS = ["'none'"]
+const NONE = "'none'"
+const NO_FRAME_ANCESTORS = [NONE]
 
 const isFramed = (request: NextRequest): boolean => {
   const destination = request.headers.get('Sec-Fetch-Dest')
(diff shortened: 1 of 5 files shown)


clients/apps/web/src/proxy.ts
```diff
@@ -328,17 +330,26 @@ export async function proxy(request: NextRequest) {
...
   const response = NextResponse.next({
     request: {
       headers: requestHeaders,
     },
   })
 
+  const checkout = request.nextUrl.pathname.match(CHECKOUT_CLIENT_SECRET)
+  if (checkout) {
+    const frameAncestors = isFramed(request)
+      ? await getFrameAncestors(request, checkout[1])
```
frankie567, line 341: Follow-up improvement: I would be okay if we simply add a property `embed_policy` directly in the CheckoutPublic object. Would avoid an extra request.

frankie567 (approved):

clients/apps/web/src/proxy.ts
```diff
@@ -328,17 +330,26 @@ export async function proxy(request: NextRequest) {
...
+      : NO_FRAME_ANCESTORS
+    response.headers.set(
+      'Content-Security-Policy',
+      checkoutCSP(frameAncestors.join(' ')),
+    )
+    if (frameAncestors.length === 1 && frameAncestors[0] === NONE) {
+      response.headers.set('X-Frame-Options', 'DENY')
```
pieterbeulque: Does this cover anything that the CSP wouldn't?

maximevast: My Claude says that it's:

* for old browser (IE11 which we probably don't support anyway)
* to make the pentesters happy.

Basically, it does not add anything to modern browsers, but does not conflict with the CSP implementation (it's ignored by modern browsers).

Happy to remove it if you prefer.

clients/apps/web/src/proxy.ts
maximevast: We would avoid 1 extra endpoint, not the extra request, the proxy needs to know the policy before we renders the page.

frankie567: Of course, bad idea then :p

clients/apps/web/src/proxy.ts
pieterbeulque: I feel like it may clean up the code a bit (e.g. removing that `NONE` constant)

clients/apps/web/src/proxy.ts
pieterbeulque: I wonder what the impact will be on our TTFB for a checkout in production

maximevast: It adds ~20ms server side, but only for framed checkout (~15% of all checkouts since yesterday)

PR #14241, opened 2026-09-08
Allow OAT creation for OAuth sessions

pieterbeulque:
We rolled back OAT creation for sessions authenticated by an OAT, rightfully so. When rolling that back, we limited these endpoints to only web sessions.

To allow the CLI to create OAT, so that agents can implement Polar directly, this PR opens up these endpoints to OAuth sessions too. One of the reasons OAT-creating-OAT wasn't good was the risk of scope escalation, but with OAuth, this doesn't apply, I made sure to test for that.

We will not open this up further to API key sessions, and we also won't add expose these endpoints to the MCP as tools, as that would by definition leak the created tokens.

About `OrganizationAccessTokenAuthorizer`, I'm not fond of the naming but it does clearly state what it does, and I'm having trouble coming up with a better name.

<a href="https://cubic.dev/pr/polarsource/polar/pull/14241?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/packages/client/src/v1.ts
+++ b/clients/packages/client/src/v1.ts
@@ -5890,10 +5890,15 @@ export interface paths {
     /**
      * List
      * @description List organization access tokens.
+     *
+     *     **Scopes**: `organization_access_tokens:read` `organization_access_tokens:write`
      */
     get: operations['organization_access_tokens:list']
     put?: never
-    /** Create */
+    /**
+     * Create
+     * @description **Scopes**: `organization_access_tokens:write`
+     */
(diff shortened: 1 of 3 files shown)


pieterbeulque:
I declare bankruptcy

PR #14261, opened 2026-09-08
Allow changing current period end on a due-to-cancel subscription

pieterbeulque:
If the subscription is set to cancel at period end, we can still allow updates to the current period's billing end. This has no customer impact, and there are cases where the merchant may want to grant some extra days at the end of the final period as a goodbye gesture, maybe.

<a href="https://cubic.dev/pr/polarsource/polar/pull/14261?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


frankie567 (approved):
Why not!

PR #14262, opened 2026-09-08
fix(payout): block cancellation after an active or successful attempt

frankie567:
## Summary

Block payout cancellation when any attempt is pending, in transit, or successful: the money can no longer be retrieved safely.

## What

Use the same eligibility check in the cancellation service and the backoffice Cancel Payout action. Payouts with no attempts or only failed attempts remain cancelable, subject to the existing payout status check.

## Why

A payout can still have a cancelable status while an attempt is pending. Checking payout status alone could allow the merchant balance to be credited after the money has already been sent.

## How

Refresh attempts after acquiring the existing payout lock, then reject cancellation before any reversal. Return a 409 conflict for non-cancelable payouts.

Validation: 66 payout service and ledger tests pass, including regression cases for pending, in-transit, and successful attempts followed by a failed attempt. Ruff, formatting, and mypy pass for all four changed files. No migrations or manual deployment steps are required.

## Checklist

- [x] This PR addresses a single concern
- [x] Linting and type checking pass for the changed files

<a href="https://cubic.dev/pr/polarsource/polar/pull/14262?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/tests/payout/test_service.py
+++ b/server/tests/payout/test_service.py
@@ -1048,6 +1048,46 @@ class TestTransferStripe:
 
 @pytest.mark.asyncio
 class TestCancel:
+    @pytest.mark.parametrize(
+        "attempt_status",
+        [
+            PayoutAttemptStatus.pending,
+            PayoutAttemptStatus.in_transit,
+            PayoutAttemptStatus.succeeded,
+        ],
+    )
+    async def test_non_failed_attempt_blocks_cancellation(
+        self,
+        save_fixture: SaveFixture,
+        session: AsyncSession,
+        organization: Organization,
+        user: User,
+        stripe_service_mock: MagicMock,
+        payout_transaction_service_mock: MagicMock,
+        attempt_status: PayoutAttemptStatus,
+    ) -> None:
+        account = await create_account(save_fixture, user)
+        payout_account = await create_payout_account(
+            save_fixture, organization, user, type=PayoutAccountType.stripe
+        )
+        payout = await create_payout(
+            save_fixture,
+            account=account,
+            payout_account=payout_account,
+            attempts=[attempt_status, PayoutAttemptStatus.failed],
+        )
+        payout.status = PayoutStatus.pending
+        await save_fixture(payout)
+
+        with pytest.raises(PayoutNotCancelable) as exc:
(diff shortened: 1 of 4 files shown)


server/tests/payout/test_service.py
```diff
@@ -1048,6 +1048,47 @@ async def test_zero_decimal_currency_adjustment(
 
 @pytest.mark.asyncio
 class TestCancel:
+    @pytest.mark.parametrize(
+        "attempt_status",
+        [
+            PayoutAttemptStatus.pending,
+            PayoutAttemptStatus.in_transit,
```
psincraian, line 1055: I thought we should be able to cancel pending and in_transit payouts. For example, when we deny an organization we try to cancel existing payouts

psincraian (commented):
I have some doubts

server/tests/payout/test_service.py
frankie567: The problem with in transit payouts is that most of the time, it's way too late to cancel them and recover the money. I hit two cases in the past days. For example: [https://dashboard.stripe.com/acct_1LzIVeDG1jUQrXwC/connect/accounts/acct_1Q1lIzReJCOZLTih/activity](<https://dashboard.stripe.com/acct_1LzIVeDG1jUQrXwC/connect/accounts/acct_1Q1lIzReJCOZLTih/activity>)

We cancel the payout, re-credit the money on Polar's side, but Stripe is not able to get the money back. It triggers an automatic withdrawal payout (for which we don't handle webhooks) that fails as well. Result: negative balance on Stripe account, and positive balance on Polar. The merchant can then trick us and trigger a payout on another Stripe account (that's actually what happened here).

THat's why I think we can't reliably cancel a payout that's already in flight.

psincraian: Makes sense then

psincraian (approved):

PR #14286, opened 2026-09-09
backoffice: Set up a private backoffice service

Yopi:
This spins up a variant of the backend on a separate host which runs as a private Render service
It also runs caddy with automatic certificate generation together with Cloudflare and Let's Encrypt.

It serves the service via Tailscale.

<a href="https://cubic.dev/pr/polarsource/polar/pull/14286?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/Dockerfile
frankie567, line 1: This is becoming quite crowded over there. Wouldn't it be worth to have a dedicated Dockerfile for the backoffice?

server/docker/run_private_backoffice.sh
```diff
@@ -0,0 +1,43 @@
...
+
+for port in 9002 10000; do
+  start curl --fail --silent --show-error --output /dev/null --noproxy '*' \
+    --retry 60 --retry-all-errors --retry-delay 1 --retry-max-time 60 --max-time 2 \
+    "http://127.0.0.1:$port/healthz"
+  wait "${children[-1]}"
+  unset 'children[-1]'
+done
+
+CLOUDFLARE_API_TOKEN="$cloudflare_api_token" \
```
frankie567, line 39: Maybe we could skip the nice subdomain for now so we can just avoid Caddy/Cloudflare in a first intent?

Yopi: The idea was that this was less work, otherwise we'll either have to make the cookies and impersonation work in a different way, or not work at all

server/Dockerfile
Yopi: Yes, there might be a point 😄

server/docker/run_private_backoffice.sh
frankie567: Fair point. I'm quite convinced I'll need to revamp the backoffice authentication; but don't know when I'll be able to do that. So fine 🙂

server/Dockerfile
Yopi: To not have to set up more images upstream / build things separately / point services to different images, I reduced the length of the Dockerfile a bit for now instead

frankie567: Comments felt useful to me 😂

Yopi: Hahah, sorry. Too aggressive 😄

server/docker/download_ipinfo.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+set -eu
+
+mkdir /data
+access_token=$(cat /run/secrets/IPINFO_ACCESS_TOKEN)
+expected_sha256=$(curl -fsSL "https://ipinfo.io/data/free/country_asn.mmdb/checksums?token=${access_token}" | jq -r '.checksums.sha256')
+curl -fsSL "https://ipinfo.io/data/free/country_asn.mmdb?token=${access_token}" -o /data/country_asn.mmdb
```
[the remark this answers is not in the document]
Yopi: I believe this was the case already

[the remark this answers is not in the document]
frankie567: It was indeed :) But good heads-up

frankie567 (approved):

PR #14302, opened 2026-09-09
CLI: Org implied environment

sebastianekstrom:
## Summary

Make the CLI's environment follow the active organization, so only login and logout needs  `--sandbox` or `--production`.

Fixes https://linear.app/polarsh/issue/PLR-112/make-organization-implied

<a href="https://cubic.dev/pr/polarsource/polar/pull/14302?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/clients/packages/cli/src/services/config.ts
+++ b/clients/packages/cli/src/services/config.ts
@@ -1,27 +1,28 @@
 import { homedir } from 'node:os'
 import { Context, Effect, FileSystem, Layer, Path, Schema } from 'effect'
-import { AuthError, type PolarEnvironment } from '@/schemas/Auth'
+import { AuthError, OrganizationSelection } from '@/schemas/Auth'
 
-const EnvironmentConfig = Schema.Struct({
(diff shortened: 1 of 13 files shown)


sebastianekstrom:
@cubic-dev-ai re review

clients/packages/cli/src/services/config.ts
simonxabris, line 1: I dont think we need to preserve the previous way of doing this and do a migration, we can just do a hard cutover to the new mechanism no?

sebastianekstrom: Fixed!

simonxabris (approved):

PR #14308, opened 2026-09-09
redis: Configure application to use TLS redis (rediss)

Yopi:
<a href="https://cubic.dev/pr/polarsource/polar/pull/14308?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/terraform/modules/render_service/main.tf
+++ b/terraform/modules/render_service/main.tf
@@ -157,9 +157,7 @@ resource "render_env_group" "redis" {
   environment_id = var.render_environment_id
   name           = "redis-${var.environment}"
   env_vars = {
-    POLAR_REDIS_HOST = { value = var.redis_config.host }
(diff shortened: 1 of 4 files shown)


Yopi:
The bot is not entirely correct on its assessment. Connecting via PrivateLink won't work with hostname validation afaik.

Worker override is used by the drain workers, which shouldn't use rediss, since it's for the Render Redis.

frankie567 (approved):
Good! I think that'll allow us to cleanup some boilerplate added by Vercel in config.py :D

PR #14332, opened 2026-09-10
webhooks: Dont re-enqueue task if event doesnt exist

Yopi:
Otherwise we continue retrying forever and ever

<a href="https://cubic.dev/pr/polarsource/polar/pull/14332?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/webhook/tasks.py
+++ b/server/polar/webhook/tasks.py
@@ -6,6 +6,7 @@ from uuid import UUID
 
 import httpx
 import idna
+import sentry_sdk
 import structlog
 from apscheduler.triggers.cron import CronTrigger
 from dramatiq import Retry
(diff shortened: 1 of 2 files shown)


Yopi:
I don't think we ever end up trying before the event exists, so I don't think this is a retryable path?

server/polar/webhook/tasks.py
```diff
@@ -79,7 +79,8 @@ async def _webhook_event_send(
         webhook_event_id, options=repository.get_eager_options()
     )
     if event is None:
-        raise Exception(f"webhook event not found id={webhook_event_id}")
+        log.warning("Webhook event not found, skipping", id=webhook_event_id)
```
frankie567: Could we add a manual Sentry exception capture? I get the idea, but arriving here is a clear indication that something is wrong earlier in the chain.

frankie567 (changes requested):

server/polar/webhook/tasks.py
Yopi: Definitely. The reason for this PR is that we have a few of these left retrying in the drain workers

frankie567 (approved):

PR #14336, opened 2026-09-10
fix: close invariant DB sessions before Slack notifications

pieterbeulque:
## Summary

- run invariant queries through `AsyncReadSessionMaker`
- return invariant failures from the check and close the database session before posting to Slack
- add a regression test that enforces notification happens after session exit

## Production impact

While deploying #14331, `CREATE INDEX CONCURRENTLY` remained in `waiting for old snapshots`. The blockers were repeated `production.worker` sessions left `idle in transaction` by the failed-subscription invariant, including one nearly two hours old. The invariant query opened a transaction and the failure path awaited Slack before leaving its session scope.

This change removes that external I/O from the database transaction and routes the read-only invariant query through the read session maker.

## Verification

- `POLAR_ENV=testing uv run python -m pytest tests/observability/invariants -q` — 36 passed
- `uv run task lint_check`
- `uv run task lint_types`

<a href="https://cubic.dev/pr/polarsource/polar/pull/14336?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/polar/observability/invariants/tasks.py
```diff
@@ -51,5 +51,8 @@ async def check_invariant(invariant_cls_name: str) -> None:
     except StopIteration as e:
         raise InvariantDoesNotExistError(invariant_cls_name) from e
 
-    async with AsyncSessionMaker() as session:
-        await invariant_service.check(session, invariant_cls)
+    async with AsyncReadSessionMaker() as session:
```
pieterbeulque: I'm not sure about this, because maybe we don't want invariants to be influenced by replica lag?

server/tests/observability/invariants/test_tasks.py
```diff
@@ -0,0 +1,53 @@
...
+from polar.observability.invariants.rules.subscriptions_canceled_deleted_customer import (
+    SubscriptionsCanceledDeletedCustomerInvariant,
+)
+from polar.observability.invariants.service import invariant as invariant_service
+
+_check_invariant = tasks.check_invariant.__wrapped__  # type: ignore[attr-defined]
+
+
+@pytest.mark.asyncio
+async def test_notifies_after_database_session_closes(
```
pieterbeulque: I think this test is obsolete, would you agree?

server/polar/observability/invariants/tasks.py
frankie567: Yes, that was the idea

server/tests/observability/invariants/test_tasks.py
frankie567: Doesn't really make sense, yes

frankie567 (commented):
Hmm... Are you sure this would make a difference? This seems to assume that the Slack notification is the blocking factor, but I don't think that's it. Even in case of exception, the SQLALchemy session context manager should properly close the transaction.

Looking at a Logfire trace: https://logfire-us.pydantic.dev/polar/polar?q=trace_id%3D%2701a08ab51dde7303826cd0e5c173db75%27+and+span_id%3D%27ad0ec4d0e4b1374e%27&spanId=ad0ec4d0e4b1374e&traceId=01a08ab51dde7303826cd0e5c173db75&env=-clear-&since=2026-09-10T09%3A15%3A50.626167Z&until=2026-09-10T10%3A15%3A50.626167Z

It looks like the query times out (which is another problem), but then the task continues to run until it hits its own timeout. That's fishy.

pieterbeulque:
Fixed both comments. I agree it's not the root cause, Codex jumped to conclusions there, but this still feels worth merging.

pieterbeulque:
@frankie567 Keep or close?

frankie567:
I would close ☺️

PR #14342, opened 2026-09-10
Fix N+1 product load in subscription charge-preview

pieterbeulque:
## The bug

`GET /v1/customer-portal/subscriptions/{id}/charge-preview` (and the dashboard equivalent) spent seconds re-loading the same product hundreds of times.

`compute_pending_subscription_line_items` iterates every static pending billing entry and calls `_get_static_price_line_item`, which did:

```python
product_repository = ProductRepository.from_session(session)
product = await product_repository.get_by_id(price.product_id)
```

once **per entry**. A subscription with N pending proration entries issued N product loads that almost all resolve to the *same* one or two products. The product is only read for `product.name` and `OrderItem.format_price_label(...)`.

## The fix

Memoize the product lookup in a per-computation `dict[UUID, Product]`, created in `compute_pending_subscription_line_items` and threaded into `_get_static_price_line_item`. Each distinct `product_id` is now fetched at most once per computation. The cache is request/computation-scoped (not a module-level global), so there is no cross-request/org correctness risk.

This path is shared with the **real** cycle-order / invoice generation (`order.service.build_subscription_order` also calls `compute_pending_subscription_line_items`), so it speeds those up too. Behavior is identical — same line items, labels and amounts, just fewer queries.

## Test

Added `test_static_entries_load_product_once_per_computation` in `tests/billing_entry/test_service.py`: creates a subscription with several static pending proration entries on the same product, spies on `ProductRepository.get_by_id`, and asserts it is called exactly once while the resulting line items are unchanged.

## Verification

- `uv run task lint` — passed
- `uv run task lint_types` — passed (no issues in 1648 files)
- `POLAR_ENV=testing uv run python -m pytest tests/billing_entry/ tests/subscription/` — 452 subscription + 18 billing_entry tests pass, including the new one and the charge-preview service tests.

<a href="https://cubic.dev/pr/polarsource/polar/pull/14342?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/tests/billing_entry/test_service.py
+++ b/server/tests/billing_entry/test_service.py
@@ -39,7 +39,6 @@ from polar.product.guard import (
     is_metered_price,
     is_seat_price,
 )
-from polar.product.repository import ProductRepository
 from polar.product.tiers import Tiers, TierType
 from tests.fixtures.database import SaveFixture
 from tests.fixtures.random_objects import (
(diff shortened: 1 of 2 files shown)


server/tests/billing_entry/test_service.py
```diff
@@ -1078,6 +1079,53 @@ async def test_static_price(
         await session.refresh(entries[2])
         assert entries[2].order_item_id == order_item_2.id
 
+    async def test_static_entries_load_product_once_per_computation(
```
pieterbeulque: @stilla can you remove this test, it's not necessary

pieterbeulque:
@joebon is this a naive implementation of avoiding N+1 queries?

joebon (approved):
Perhaps there's a fancier way of doing the same but this def looks like it's doing the job

PR #14345, opened 2026-09-10
server/worker: set idle transaction timeout in AWS Lamdba runner

frankie567:
We apparently have a networking issue between AWS Lambda and Render external database endpoint: the client cancellation sent by asyncpg seems to be lost in the void, leading to the task timing out and transactions stuck in `idle in transaction` state in the database. Until we find the root cause, this change will force Postgres to kill idle transactions after 60 seconds.

<a href="https://cubic.dev/pr/polarsource/polar/pull/14345?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/worker/_runner.py
+++ b/server/polar/worker/_runner.py
@@ -288,9 +288,17 @@ async def run_task(
         )
 
 
-def bootstrap(*, pool_pre_ping: bool = False) -> None:
+def bootstrap(
+    *,
+    pool_pre_ping: bool = False,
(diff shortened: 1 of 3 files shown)


Yopi (approved):
Clever

PR #14400, opened 2026-09-11
Fix: ensure merchant provided webhook secrets always sign the legacy way

joebon:
A merchant provided webhook secret can be any string so can't in general be used to sign webhooks the standard way

<a href="https://cubic.dev/pr/polarsource/polar/pull/14400?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pieterbeulque:
Should we allow merchants to provide secrets?

frankie567:
> Should we allow merchants to provide secrets?

No, it's deprecated and should be removed in future versions

joebon:
> Should we allow merchants to provide secrets?

No, but we've done so historically. That request param was marked as deprecated so I removed that param entirely in the next api version so only exists in current. But the param is used currently, both by custom api clients and our sdks

server/tests/webhooks/test_service.py
```diff
@@ -103,6 +128,28 @@ async def test_organization_endpoint_valid(
         )
         assert updated_endpoint.url == "https://example.com/hook-updated"
 
+    @pytest.mark.auth(
```
joebon, line 131: Maybe this one should be dropped though as it's just duplicating the the service test

frankie567: Yeah, not strictly necessary

frankie567 (approved):

server/tests/webhooks/test_service.py
joebon: Nvm, keeping just to verify it can't find it's way in either via create nor update

pieterbeulque:
Okay!

PR #14413, opened 2026-09-11
Explore `x-tool` extensions on OpenAPI

pieterbeulque:
I'm not sure about the `x-tool-*` naming, to be fair.

`x-tool-name` is quite MCP-specific, though it could help match the CLI with a `gsub('_', ' ')`, maybe? Not sure how you're doing that right now, @simonxabris.

The rest can be shared across the MCP & CLI, especially the annotations can be helpful to render & determine whether something needs a `--confirm`, for example.

<a href="https://cubic.dev/pr/polarsource/polar/pull/14413?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/server/polar/benefit/endpoints.py
+++ b/server/polar/benefit/endpoints.py
@@ -49,7 +49,13 @@ BenefitNotFound = {
     "/",
     summary="List Benefits",
     response_model=ListResource[BenefitSchema],
-    openapi_extra={"parameters": [get_metadata_query_openapi_schema()]},
+    openapi_extra={
+        "x-tool-name": "benefits_list",
+        "x-tool-title": "List benefits",
+        "x-tool-description": "List all existing benefits.",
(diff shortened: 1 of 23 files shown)


frankie567:
I think we could mostly automate this:

* `x-tool-name` seems to be `{tag}_{route_name}`
* `x-tool-title` is equal to `summary` 
* `x-tool-description` is equal to `description` without the scope

So I think with a clever route API class, we could preset all those by default, and use explicit overrides only when needed.

PR #14496, opened 2026-09-15
server/oauth2: name the signing key in the id_token header

maximevast:
Polar's id_tokens carried no kid, so a relying party holding more than one published key had no way to pick the right one. The signing key is now named in the JOSE header, and resolved on each issuance rather than frozen at import.

Also moves off authlib's deprecated get_jwt_config: it copies every config key into the claims, so kid would have landed in the payload instead of the header.

Step 1 of 4 on [PLR-142](https://linear.app/polarsh/issue/PLR-142/overhaul-jwks-management). Prerequisite for publishing a second key.
## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/14496?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


frankie567 (approved):
Great, thanks!

PR #14501, opened 2026-09-15
license-key: Add member to the license key response

Yopi:
This makes it possible to use the license key APIs to verify the member, otherwise there is no specific way to identify which member on a license key that has gotten the issuance.

<a href="https://cubic.dev/pr/polarsource/polar/pull/14501?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/tests/license_key/test_endpoints.py
```diff
@@ -944,6 +946,57 @@ async def test_omitted_increment_usage(
...
+        repository = LicenseKeyRepository.from_session(session)
+        lk = await repository.get_by_id(UUID(granted["license_key_id"]))
+        assert lk is not None
+        lk.member_id = member.id
+        await save_fixture(lk)
+
+        response = await client.post(
+            "/v1/customer-portal/license-keys/validate",
+            json={"key": lk.key, "organization_id": str(lk.organization_id)},
+            headers={"Polar-Version": str(NEXT_API_VERSION)},
```
frankie567, line 983: We should have a pytest marker for that, will put Codex on it 😀

frankie567 (approved):

server/tests/license_key/test_endpoints.py
Yopi: Thanks! Should have thought of it as well 😄

psincraian (approved):
LGTM!

PR #14498, opened 2026-09-15
handbook: record how the JWKS is stored and rotated (ADR-0010)

maximevast:
## Summary

Don't know if this is useful to merge but would still want your opinion on it.

## Checklist

- [ ] This PR addresses a single concern (one bug fix, one feature, one refactor)
- [ ] The diff is reasonably sized and easy to review
- [ ] New functionality is covered by tests
- [ ] Linting and type checking pass (`uv run task lint && uv run task lint_types`)
- [ ] No unrelated changes or drive-by fixes are included

<a href="https://cubic.dev/pr/polarsource/polar/pull/14498?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #14506, opened 2026-09-15
terraform: add the JWKS signing key

maximevast:
Adds the asymmetric KMS key that will sign OAuth2 id_tokens and SSO client assertions, with kms:Sign and kms:GetPublicKey on the app's role. Nothing uses it yet; the signer comes next.

No enable_key_rotation: KMS only rotates symmetric encryption keys, so rotating this one means publishing a second key and moving the set over. See ADR-0010 (#14498).

Step 3 of PLR-142.

## Checklist

<a href="https://cubic.dev/pr/polarsource/polar/pull/14506?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


terraform/modules/jwks_signing_key/main.tf
```diff
@@ -0,0 +1,71 @@
...
+  }
+}
+
+# Asymmetric KMS key that signs OAuth2 id_tokens and SSO client assertions. The
+# private key never leaves KMS; the app signs through kms:Sign and publishes the
+# public half from kms:GetPublicKey.
+#
+# No enable_key_rotation here: KMS supports automatic rotation only for
+# symmetric encryption keys. Rotating this one means creating a second key and
+# moving the published set over to it — see ADR-0010.
```
Yopi, line 18: Does ADR-0010 exist?

terraform/modules/jwks_signing_key/main.tf
```diff
@@ -0,0 +1,71 @@
+terraform {
```
Yopi, line 1: I think it would make more sense that this module was a generic kms_key or aws_kms_key module, and then you can allow certain config via variables.

Unless you think this is basically the only one that will look roughly this shape?

Yopi (approved):
I think you can generalize this a little bit to make it more reusable. But otherwise looks good

terraform/modules/jwks_signing_key/main.tf
maximevast: Now it does !

terraform/modules/jwks_signing_key/main.tf
maximevast: I'd rather not optimize too early but please remind me about this conversation if i'm proved wrong later !

Yopi: I don't think it's early optimization in this case. It's a bit like dependency inversion. Instead of this module knowing that it will be used for purely JWKS signing (so you write about it in the description of the key, in the output field, etc). It should only know that it's a KMS key, and then when you instantiate the module you can say this is for JWKS.

Then all of the information regarding what the key is used for is on the caller side `module jwks_signing_key source=kms` . I think it will be easier to read and parse in both ends

Yopi: If you would need to do a lot of work to get there, then I would tend to agree with you

PR #14518, opened 2026-09-15
pgbouncer: Set statement timeout to 30s

Yopi:
To prevent long running queries from erroring out

<a href="https://cubic.dev/pr/polarsource/polar/pull/14518?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


frankie567 (commented):
How does it relate to DATABASE_CONNECT_TIMEOUT_SECONDS?

Yopi:
> How does it relate to DATABASE_CONNECT_TIMEOUT_SECONDS?

DATABASE_CONNECT_TIMEOUT_SECONDS is the timeout for establishing a connection to the server
DATABASE_COMMAND_TIMEOUT_SECONDS is the timeout for the database queries from the application pov.

But given the issue with Render and that it's not firing for pgbouncer queries, this sets the timeout per query. I was hoping to get rid of the timeouterrors.

But I think this is not going to work in the way I was hoping.

PR #14576, opened 2026-09-17
models: add discord_guild_connections

maximevast:
## Summary

`discord_guild_token` is a computed_field and thus prevent us from migrating jwt signing to KWS because it would require a blocking call to AWS everytime we serialize a discord benefit.
This stack aims to store the relation between a discord guild connection and an organization instead of signing it every time.

It will also fix a "bug" where a multi org user could create a benefit for a discord server belonging to orgA on orgB, even if they are readonly on orgA.

Not to be merged as a stack since we have migration, backfill, and a frontend PR that needs to land in their own order. 

## Checklist

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

<a href="https://cubic.dev/pr/polarsource/polar/pull/14576?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


server/migrations/versions/2026-09-16-1851_add_discord_guild_connections.py
```diff
@@ -0,0 +1,118 @@
+"""add discord_guild_connections
+
+Revision ID: d38387d01757
+Revises: 0ca20ea806d4
```
frankie567: You'll need to rebase your migrations

frankie567 (approved):

PR #14627, opened 2026-09-18
feat: srub PII at log ingestion

simonxabris:
## Summary

Fixes [PLR-139](https://linear.app/polarsh/issue/PLR-139/pii-scrubbing-at-log-ingestion)

Adds a `pii.py` module that exports `scrub_event` and `scrub_value` functions that are then integrated into the following places:

 - server/polar/logging.py — structlog and standard logging, before Logfire forwarding and console/JSON output.
 - server/polar/logfire.py — configures Logfire’s scrubbing callback/patterns and installs our span processor.
 - server/polar/observability/pii_span_processor.py — scrubs span names, attributes, events, links, and status descriptions before export.
 - server/polar/observability/s3_span_exporter.py — scrubs span/event/link attributes before S3 upload.
 - server/polar/sentry.py — scrubs backend Sentry events and breadcrumbs.
 - server/polar/worker/_asyncio.py — scrubs watchdog stack dumps written directly to stderr.

Also adds a post deploy check that emits synthetic logs with PII looking values and checks if they are properly redacted when they arrive in the target system.


server/tests/observability/observability/conftest.py
```diff
@@ -6,13 +6,103 @@
...
 
+import logging
 import os
+from functools import partial
+from io import StringIO
+
+# Set up test environment before any polar imports
+os.environ["POLAR_ENV"] = "testing"
+
+from collections.abc import Iterator
```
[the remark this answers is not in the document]
simonxabris: This does not trigger E402: Ruff permits environment-variable setup before imports. Both `uv run ruff check tests/observability/observability/conftest.py` and the explicit `uv run ruff check --select E402 tests/observability/observability/conftest.py` pass, and the reviewed commit passed the Server: Linters CI job. Keeping the setup before application imports; no code change needed.

.github/workflows/deploy-environment.yml
```diff
@@ -102,6 +108,65 @@ jobs:
         env:
           RENDER_API_TOKEN: ${{ secrets.RENDER_API_TOKEN }}
 
+  validate-logs:
+    name: "Backend: Validate stored logs (${{ inputs.environment }})"
+    needs: [deploy-backend, promote-frontend]
+    if: always() && needs.deploy-backend.result == 'success' && (needs.promote-frontend.result == 'success' || needs.promote-frontend.result == 'skipped')
```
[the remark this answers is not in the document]
simonxabris: The skipped-promotion behavior is correct, but we want to retain validation in this case: the backend has successfully deployed and is already live even if the frontend build failed. The backend deployment tag also advances independently, so a subsequent frontend-only retry may skip the backend and its validation. Adding the proposed guard could therefore leave the deployed backend unchecked. Keeping the current condition rather than suppressing this check after a frontend build failure.

server/tests/observability/observability/test_s3_span_exporter.py
```diff
@@ -101,11 +101,36 @@ def scrubbing_exporter() -> S3SpanExporter:
...
+                        "events": [{"attributes": attributes}],
+                        "links": [{"attributes": attributes}],
+                    }
+                )
+            )
+        )
+        assert url not in json.dumps(result)
+        for entry in [result, *result["events"], *result["links"]]:
+            assert entry["attributes"]["http.url"] == REDACTED
+            assert entry["attributes"]["http.route"] == attributes["http.route"]
```
Yopi, line 128: I'm a bit worried about removing the URL, but maybe the route will be enough for our debugging

server/polar/pii_validation/README.md
```diff
@@ -0,0 +1,96 @@
+# Post-deployment PII validation
```
Yopi: There is a trade-off here, and this might be on the right side of it, but it feels like *a lot* to just avoid accidentally logging PII

Yopi (approved):
Nice work. It's quite a beast. It's a bit difficult to tell if this is doing too much or if it's a good idea to go the extra mile 😄

server/polar/observability/pii.py
```diff
@@ -0,0 +1,184 @@
+from __future__ import annotations
+
+from collections.abc import Mapping
+from typing import Any
+
+import re2
```
frankie567, line 6: What's re2? Why do we need it?

server/polar/pii_validation/README.md
frankie567: Agree with Jesper here, having a full module for testing + CI step for this is overkill. I'm fine with just having the focused tests in tests/

server/polar/observability/pii.py
```diff
@@ -0,0 +1,184 @@
...
+from stdnum import iban, luhn
+
+REDACTED = "[Redacted]"
+
+SENSITIVE_KEYS = frozenset(
+    {
+        "http.url",
+        "url",
```
frankie567, line 13: Isn't it a bit too much to scrub URL? That's usually useful debug information, not sure it's a PII

server/polar/sentry.py
```diff
@@ -45,7 +46,29 @@ def before_send(event: Event, hint: Hint) -> Event | None:
...
+    return scrubbed
+
+
+def before_breadcrumb(breadcrumb: Breadcrumb, hint: BreadcrumbHint) -> Breadcrumb:
+    scrubbed = scrub_event(breadcrumb)
+    if data := scrubbed.get("data"):
+        for key in ("url", "from", "to"):
+            if key in data:
+                data[key] = REDACTED
+    return scrubbed
```
frankie567: I would say that's a bit overkill as a first step, I would prefer to not add too much logic on Sentry's side, and evaluate if we need to do some scrubbing.

server/polar/sentry.py
```diff
@@ -59,6 +82,10 @@ def configure_sentry(*, aws_lambda: bool = False) -> None:
         default_integrations=False,
         auto_enabling_integrations=False,
         before_send=before_send,
+        before_breadcrumb=before_breadcrumb,
+        send_default_pii=False,
+        include_local_variables=False,
+        max_request_body_size="never",
```
frankie567: Let's revert that

frankie567 (changes requested):
Looks great, even if it's a lot of stuff for one PR 😅

I think we can slim it down a bit, at least in a first step.

server/polar/pii_validation/README.md
simonxabris: I'm happy to remove it but the task did list a mechanism like this as needed and so I thought It's necessary for SOC2. I too think it's a lot 😄

server/polar/observability/pii.py
simonxabris: Might be premature, but since we're using regexes, a malicious input can basically DDOS our system. RE2 is an alternative regex implementation that is more efficient.

For a long JWT like input that did not contain any dots, normal python regex took 1.6s while re2 took 0.7ms.

server/polar/pii_validation/README.md
simonxabris: If that works for SOC2 I'm fine with that, I'll remove it.

server/polar/observability/pii.py
simonxabris: I included it because there are URLs that can contain PII, namely:

- /v1/checkouts/client/{client_secret}
- Customer portal links
- Email verification links
- Stripe dashboard

But maybe this is too broad and we should only redact the sensitive parts

server/tests/observability/observability/test_s3_span_exporter.py
simonxabris: I included it because there are URLs that can contain PII, but maybe it's an overkill here

server/polar/sentry.py
simonxabris: Reverted this!

server/polar/pii_validation/README.md
simonxabris: Removed

server/polar/observability/pii.py
frankie567: Hmm, right because of the query parameters. Ok, let's keep it, as Jesper says, the path might be enough.

server/polar/observability/pii.py
frankie567: Hmm, okay, let's try it

server/polar/observability/pii.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def _scrub_string(value: str) -> str:
+    # RE2 requires a plain str, not a str-based Enum.
+    value = _EMAIL_RE.sub(REDACTED, str.__str__(value))
+    value = _JWT_RE.sub(REDACTED, value)
+    value = _BEARER_RE.sub(REDACTED, value)
+    value = _STRIPE_SECRET_RE.sub(REDACTED, value)
+    value = _IBAN_RE.sub(_replace_iban, value)
+    value = _PAN_RE.sub(_replace_pan, value)
```
frankie567, line 155: If it looks like an IBAN/PAN, then let's redact them immediately. No need to add more overhead to double check their validity

frankie567 (approved):

PR #14655, opened 2026-09-18
server/transaction: prepare deletion of useless columns

frankie567:
Defer and stop writing the following columns on Transaction:

- payment_user_id and payment_organization_id: this was used in the Pledge era. The same information can still be derived from pledge_id.
- customer_id: the raw Stripe Customer ID. Not useful, and can be derived either from the Customer foreign key or via the Stripe API and the charge_id.

<a href="https://cubic.dev/pr/polarsource/polar/pull/14655?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


maximevast (approved):
Less is more!

PR #14658, opened 2026-09-18
clients/web: pin API version

frankie567:
We should do what we recommend ourselves :D

<a href="https://cubic.dev/pr/polarsource/polar/pull/14658?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


clients/apps/web/src/utils/client/index.ts
```diff
@@ -31,10 +31,13 @@ const sessionFreshnessMiddleware: Middleware = {
   },
 }
 
+const NEXT_API_VERSION = '2026-10'
```
pieterbeulque, line 34: Should we read this from an env var at build time that is shared across everything in the front-end, including here and the `pnpm run generate` script in `packages/client`?

pieterbeulque (approved):

clients/apps/web/src/utils/client/index.ts
frankie567: Hmm yeah, we could, but that requires a bit of extra tooling 🙂 Leaving that as a follow-up 😀

PR #14659, opened 2026-09-18
logs: Set deletion protection and log retention

Yopi:
Governance mode sets deletion protection for 7 years for logs stored in S3

<a href="https://cubic.dev/pr/polarsource/polar/pull/14659?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


terraform/modules/s3_buckets/main.tf
```diff
@@ -369,6 +369,25 @@ resource "aws_s3_bucket" "logs" {
...
+    status = "Enabled"
+  }
+}
+
+resource "aws_s3_bucket_object_lock_configuration" "logs" {
+  bucket = aws_s3_bucket_versioning.logs.id
+
+  rule {
+    default_retention {
+      mode  = "GOVERNANCE"
```
frankie567, line 387: For backups, Freda bot yelled at me that GOVERNANCE was not hard enough (an admin can still delete it) and nudged to use COMPLIANCE mode, which can't be bypassed. I would recommend you to double check to save you another PR 😀

frankie567 (approved):

terraform/modules/s3_buckets/main.tf
Yopi: Haha I felt like compliance mode would introduce a risk if we ever _need_ to delete an old log. I'll scream at the bot

PR #14687, opened 2026-09-21
stop sending local variables to sentry

joebon:
Currently we're sending all existing local variables (and their values) to sentry upon an exception/crash.
Since that can include highly sensitive data and there's no scrubbing applied to that data let's turn that feature off entirely.

If this turns out to be problematic in some specific circumstances we can always revisit this decision and deliberately bring back specific pieces of data, still without potentially shipping sensitive data to sentry.

<a href="https://cubic.dev/pr/polarsource/polar/pull/14687?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


frankie567 (approved):
Good call!

